PR #3254, opened 2018-02-22
Correct docs to config.pluginmanager.get_plugin()

altendky:
`getplugin()` is deprecated in favor of `get_plugin()`.

https://github.com/pytest-dev/pytest/blob/dd97c940359079fd874795a731d54d1abb7d4c8e/_pytest/config.py#L261


nicoddemus:
Thanks, appreciate it!

Issue #3209, opened 2018-02-12
Logging plugin is not referenced in plugins.rst


thisch:
The logging plugin is missing in the plugin reference list in
https://pytest.readthedocs.io/en/3.4.0/plugins.html#pytest-default-plugin-reference

PR #3275, opened 2018-03-03
Add logging plugin to plugins list

maiksensi:
As outlined in #3209 the logging plugin should be referenced in the plugins list. This PR adds the logging plugin to the list.


nicoddemus:
Thanks @maiksensi!

PR #3282, opened 2018-03-05
[blocked] #3224 allow to use file:line number

### LINKED ISSUE #3224: Allow test selection by file:lineno
`pytest tests/test_foo.py:42` should run the (nearest) test at line 42 in `tests/test_foo.py`.

When it also ignores a trailing colon it would allow for easy copy'n'pasting from failures.

Via #5973:

> To get a better overview of how it can work take a look at the [Rspec documentation](https://relishapp.com/rspec/rspec-core/v/2-3/docs/command-line/line-number-appended-to-file-path)

feuillemorte:
Allowed test selection by file:line_number. For example, you can run ``pytest tests/test_foo.py:5`` and Pytest will find a nearest test.
Fixes #3224 

Problems:
Pytest collects all tests in file but run only one test (or several if parametrize):

$ cat tests/test_1.py
```
1  def test_01():
2      assert 1
3
4
5  def test_02():
6      assert 1
7
8
9  def test_03():
10     assert 1
```
pytest tests/test_1.py:6 -v
```
================================================================================= test session starts ==================================================================================
platform linux -- Python 3.6.4, pytest-3.4.2.dev130+gc95f3316, py-1.5.2, pluggy-0.6.0 -- /home/feuillemorte/github/pytest-dev/pytest/pytest_3224/env/bin/python
cachedir: .pytest_cache
rootdir: /home/feuillemorte/github/pytest-dev/pytest/pytest_3224, inifile: tox.ini
collected 3 items                                                                                                                                                                      

tests/test_1.py::test_02 PASSED                                                                                                                                                  [100%]

=============================================================================== 1 passed in 0.01 seconds ===============================================================================
```

### DIFF
--- a/_pytest/main.py
+++ b/_pytest/main.py
@@ -310,9 +310,7 @@ class Session(nodes.FSCollector):
         self.trace = config.trace.root.get("collection")
         self._norecursepatterns = config.getini("norecursedirs")
         self.startdir = py.path.local()
-        self.line = None
-        self.nearest_func_name = None
-
+        self.file_line = None
         self.config.pluginmanager.register(self, name="session")
 
     def _makeid(self):
@@ -392,10 +390,13 @@ class Session(nodes.FSCollector):
         else:
             if rep.passed:
                 for node in rep.result:
-                    if self.line and self.nearest_func_name:
-                        self.items.extend([item for item in self.genitems(node) if self.nearest_func_name in item.name])
-                    else:
-                        self.items.extend(self.genitems(node))
+                    items_list = []
+                    items_list.extend(self.genitems(node))
+                    if self.file_line:
+                        # print('ITEM', items_list[1].location)
+                        # items_list = [item for item in items_list if self.select_by_function_name in item.name]
+                        items_list = self._find_nearest_test_function(items_list)
+                    self.items.extend(items_list)
             return items
 
     def collect(self):
@@ -475,51 +476,112 @@ class Session(nodes.FSCollector):
         if self.config.option.pyargs:
             parts[0] = self._tryconvertpyarg(parts[0])
         relpath = parts[0].replace("/", os.sep)
-        pattern_line = r':(?P<line>[0-9]+)$'
-        match = re.search(pattern_line, relpath)
-        if match:
-            relpath = ''.join(relpath.split(':')[:-1])
-            self.line = match.group("line")
+
+        # if match:
+        #     relpath = ''.join(relpath.split(':')[:-1])
+        #     self.line = match.group("line")
         path = self.config.invocation_dir.join(relpath, abs=True)
         if not path.check():
-            if self.config.option.pyargs:
-                raise UsageError(
-                    "file or package not found: " + arg +
-                    " (missing __init__.py?)")
-            else:
-                raise UsageError("file not found: " + arg)
-        if self.line:
-            self.nearest_func_name = self._get_nearest_func_name(self.line, path)
+            try:
+                path = self._filter_by_line(relpath)
+            except UsageError:
+                if self.config.option.pyargs:
+                    raise UsageError(
+                        "file or package not found: " + arg +
+                        " (missing __init__.py?)")
+                else:
+                    raise UsageError("file not found: " + arg)
         parts[0] = path
         return parts
 
-    def _get_nearest_func_name(self, line, path):
+    def _filter_by_line(self, path):
+        PATTERN_LINE = re.compile(r':(?P<line>[0-9]+)$')
+
         path = str(path)
-        start_line = line = int(line) - 1
-        if not os.path.isfile(path):
-            raise UsageError("{} is not a file".format(path))
-        with open(path) as f:
-            content = f.readlines()
-        at_flag = False
-        pattern_func = r'[\t ]*?def test(?P<func_name>[a-zA-Z0-9_-]+)\(*'
-        pattern_at = r'[\t ]*?@'
-        if line < len(content) and re.match(pattern_at, content[line]):
-            at_flag = True
+        match = re.search(PATTERN_LINE, path)
 
-        while 0 <= line < len(content):
-            match = re.match(pattern_func, content[line])
-            if match:
-                return "test" + match.group("func_name")
-            if at_flag:
-                if re.match(pattern_at, content[line]):
-                    line = line + 1
+        if match:
+            self.file_line = path[match.start() + 1:match.end()]
+            path = path[:match.start()]
+        path = self.config.invocation_dir.join(path, abs=True)
+        if not path.check():
+            raise UsageError()
+
+        return path
+
+    def _find_nearest_test_function(self, items):
+        """
+        Finds the nearest test function given a path in the form:
+
+        path/to/test.py:12
+
+        With ``12`` being a line number, search upwards or downwards in the file
+        until it finds a test function declaration.
+
+        :return: a filtered item list
+        """
+
+        items = sorted(items, key=lambda x: x.location)
+        mid = self._find_nearest_point_to_line(items)
+        return self._find_all_items_nearest_to_point(items, mid)
+
+    def _find_nearest_point_to_line(self, items):
+        """
+        Find nearest index of items list to file line
+        :param items:
+        :return: index of item list nearest to file line
+        """
+        line = int(self.file_line) - 1
+        mid = 0
+        lo = 0
+        hi = len(items)
+        while lo < hi:
+            mid = (lo+hi)//2
+            midval = items[mid].location[1]
+            next_to_midval = items[mid+1].location[1] if len(items) > mid + 1 else items[mid].location[1]
+            print('LO', lo, hi, midval, self.file_line)
+            if line not in list(range(midval, next_to_midval)):
+                if midval < line:
+                    lo = mid+1
+                elif midval > line:
+                    hi = mid
                 else:
-                    line = start_line - 1
-                    at_flag = False
+                    return mid
             else:
-                line = line - 1
+                return mid
+        return mid
 
-        raise UsageError("No test found near line {}".format(self.line))
+    def _find_all_items_nearest_to_point(self, items, point):
+        """
+        Find all items where location is equal as location of item[point]
+        :param items: items list
+        :param point: index of items list
+        :return: filtered items list
+        """
+        if not items:
+            return []
+        items_list = [items[point]]
+        loc = items[point].location[1]
+        start_point = point
+        up = True
+        down = True
+
+        while up or down:
+            if down:
+                point += 1
+                if point < len(items) and items[point].location[1] == loc:
+                    items_list.append(items[point])
+                else:
+                    point = start_point
+                    down = False
+            elif up:
+                point -= 1
+                if point > 0 and items[point].location[1] == loc:
+                    items_list.append(items[point])
+                else:
+                    point = start_point
+                    up = False
+        return items_list
 
     def matchnodes(self, matching, names):
         self.trace("matchnodes", matching, names)
--- a/_pytest/main.py
+++ b/_pytest/main.py
@@ -393,8 +393,6 @@ class Session(nodes.FSCollector):
                     items_list = []
                     items_list.extend(self.genitems(node))
                     if self.file_line:
-                        # print('ITEM', items_list[1].location)
-                        # items_list = [item for item in items_list if self.select_by_function_name in item.name]
                         items_list = self._find_nearest_test_function(items_list)
                     self.items.extend(items_list)
             return items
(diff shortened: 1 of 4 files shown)


feuillemorte:
[#3224 allow to use file:line number](https://trello.com/c/QcR68cKH/6-3224-allow-to-use-fileline-number)

feuillemorte:
Seems, that my regexp doesn't work on windows :) It's strange. My tests:
```
>>> relpath = str('C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_collect_line0\\test_1.py:4')
>>> pattern_line = r':(?P<line>[0-9]+)$'
>>> match = re.search(pattern_line, relpath)
>>> match
<_sre.SRE_Match object; span=(95, 97), match=':4'>
```

So, my code should work.. I'm trying to reproduce it, but I don't have windows at all :(

maiksensi:
Is it just me or isn't even the file found on windows? 

```bash 
---------------------------- Captured stderr call -----------------------------
ERROR: file not found: C:\Users\appveyor\AppData\Local\Temp\1\pytest-of-appveyor\pytest-0\test_collect_line0\test_1.py:4
```

feuillemorte:
It's because this re.search don't work :(
```
        pattern_line = r':(?P<line>[0-9]+)$'
        match = re.search(pattern_line, relpath)
        if match:
            relpath = ''.join(relpath.split(':')[:-1])
            self.line = match.group("line")
```
but it should work and works locally on linux :)
https://regex101.com/r/0kZAJV/1

maiksensi:
I cannot promise anything, but I can give it a shot tomorrow on windows 7 with py27 to see whether I can reproduce this.

_pytest/main.py
```diff
@@ -388,7 +392,10 @@ def _perform_collect(self, args, genitems):
         else:
             if rep.passed:
                 for node in rep.result:
-                    self.items.extend(self.genitems(node))
+                    if self.line and self.nearest_func_name:
```
nicoddemus: `line` and `nearest_func_name` are not very descriptive names in my opinion, I suggest we drop `self.line` altogether (it is not needed at this point) and keep only the name of the function, I suggest `self._filter_by_function_name` instead. Also I suggest to avoid creating the intermediary list and duplicating the `genitems` call (more to make the code neater than anything else):

```python
items = self.genitems(node)
if self._filter_by_function_name:
    items = (x for x in items if self.select_by_function_name in item.name)
self.items.extend(items)
```

_pytest/main.py
```diff
@@ -468,6 +475,11 @@ def _parsearg(self, arg):
         if self.config.option.pyargs:
             parts[0] = self._tryconvertpyarg(parts[0])
         relpath = parts[0].replace("/", os.sep)
+        pattern_line = r':(?P<line>[0-9]+)$'
```
nicoddemus: Let's move all this logic into a separate free-function, this will separate that logic from the rest of this code and allow us to test the function independently:

```python
self._filter_by_function_name, relpath = _find_nearest_test_function(relpath)
```

```python
def _find_nearest_test_function(path_expr):
    """
    Finds the nearest test function given a path in the form:

    path/to/test.py:12

    With ``12`` being a line number, search upwards in the file 
    until it finds a test function declaration.

    :return: a tuple (function_name, real_path). function_name can be None in 
    case a function could not be found, and real_path will return the actual
    path stripping any trailing ``:<linnum>`` found.
    """    
```

nicoddemus: Please compile these regexes and store them into the function as attributes:

```python
def _find_nearest_test_function(path_expr):
    ...

_find_nearest_test_function.PATTERN_LINE = re.compile(r':(?P<line>[0-9]+)$')
```

This will make the calls slightly faster. This advice also applies to the regexes in `_get_nearest_func_name`.

_pytest/main.py
```diff
@@ -476,9 +488,39 @@ def _parsearg(self, arg):
...
+            self.nearest_func_name = self._get_nearest_func_name(self.line, path)
         parts[0] = path
         return parts
 
+    def _get_nearest_func_name(self, line, path):
+        path = str(path)
+        start_line = line = int(line) - 1
+        if not os.path.isfile(path):
+            raise UsageError("{} is not a file".format(path))
+        with open(path) as f:
```
nicoddemus: Hmm tricky, this might blowup depending on the encoding of the file and of the file system... we probably need to use something else to properly handle the special encoding comment that might appear at the top of the file. I wonder if we can use something in `_pytest.code` for that.

_pytest/main.py
```diff
@@ -476,9 +488,39 @@ def _parsearg(self, arg):
...
+            content = f.readlines()
+        at_flag = False
+        pattern_func = r'[\t ]*?def test(?P<func_name>[a-zA-Z0-9_-]+)\(*'
```
nicoddemus: This regex won't catch functions using non-ascii names. I think we can use `str.isidentifier` instead for that.

changelog/3224.feature
```diff
@@ -0,0 +1 @@
+Allowed test selection by file:line_number. For example, you can run ``pytest tests/test_foo.py:5`` and Pytest will find a nearest test.
```
nicoddemus: We also need to add some docs to it, I suggest we add them to https://docs.pytest.org/en/latest/usage.html after *Run tests by node ids*.

_pytest/main.py
```diff
@@ -468,6 +475,11 @@ def _parsearg(self, arg):
         if self.config.option.pyargs:
             parts[0] = self._tryconvertpyarg(parts[0])
         relpath = parts[0].replace("/", os.sep)
+        pattern_line = r':(?P<line>[0-9]+)$'
+        match = re.search(pattern_line, relpath)
+        if match:
+            relpath = ''.join(relpath.split(':')[:-1])
+            self.line = match.group("line")
```
blueyed: You can use the result from `re.search`, e.g `match.start()` and `.end()` instead of the additional `.split()`.

blueyed: Handle `:lineno` should only be done when the path itself does not exist.  `foo:21` is a valid filename on Linux, and there `foo:21:3` would be used to specify the line number - a bit constructed since it would be `test_foo:21.py` then, but still valid.

_pytest/main.py
feuillemorte:
```
In [18]: 'def test_01():'.isidentifier()
Out[18]: False
```
it returns False :(

nicoddemus: Yeah I meant the second case (combining the regex and `str.isidentifier`), and indeed it works for non-ascii names (I did not know that), but it won't work for non-character names like `def testλ():` (which is valid):

```python
>>> import re
>>> def testλ(): pass
...
>>> re.match(r'[\t ]*?def test(?P<func_name>[a-zA-Z0-9_-]+)\(*', 'def testλ():')
>>> re.match(r'[\t ]*?def test(?P<func_name>[a-zA-Z0-9_-]+)\(*', 'def test_не_англ_имя_01():')
<_sre.SRE_Match object; span=(0, 9), match='def test_'>
```

feuillemorte: https://regex101.com/r/0kZAJV/2

It works well with `def testλ(): pass` :)

The-Compiler: Sorry to be "that guy", but it seems to me that re-reading the file and trying to parse Python with regexes is entirely the wrong approach here.

We already have the function objects collected from this file available somewhere, right? If so, why can't we just use something like [inspect.getsourcelines()](https://docs.python.org/3/library/inspect.html#inspect.getsourcelines) on those?

feuillemorte: @The-Compiler 
this function wants a code object
```
local('/home/username/tests/test_1.py') is not a module, class, method, function, traceback, frame, or code object
```
Could you advice how I can get this object from a path? Or where this object generates in pytest?
```
_pytest._code.compile(path)
``` 
is not working with path too.

nicoddemus: @The-Compiler you are not alone, I'm more than uneasy about the current approach.

I just realized something though, we should just store the line number given by the user and *then* filter the items in `pytest_collection_modifyitems` because each `item` already has a `location` attribute which returns a tuple of `(path, linenum, name)`. This will make things *much*
 simpler. 😁

feuillemorte: @nicoddemus I like it :) But you may have decorators. for example:

```
6 @pytest.mark.parametrize('x', [1,2])
7 def test_02(x):
8     assert 1
```
User run command `pytest test_1.py:6`
I filter it and get item.location:
```
test_01 -> 3
test_02 -> 7
```
And I need to run test on 7th string, but if I don't have decorator I need to run test on 3th string because it may look like 
```
3 def test_01():
4     assert 1
5
6
7 def test_02():
8     assert 1
```
How I can understand it in this case?

feuillemorte: @nicoddemus I mean this situation:
```
1 import pytest
2
3
4 def test_01():
5     assert 1
6
7
8 @pytest.mark.parametrize('x', [1,2,)
9 def test_02(x):
10     assert 1
11
```

I forgot `]` in the decorator. Pytest writes my a Syntax Error:
```
    __import__(modname)
E     File "/home/username/pytest_3224/tests/test_1.py", line 8
E       @pytest.mark.parametrize('x', [1,2,)
E                                          ^
E   SyntaxError: invalid syntax
```

User might want to run test with line 8. He will not copy/paste until line, but he can write line in the end.

It's just an example, but If it doesn't matter then ok :)

changelog/3224.feature
feuillemorte: Fixed

_pytest/main.py
feuillemorte: Fixed

_pytest/main.py
feuillemorte: @nicoddemus Fixed, but, could you help with algorithm? May you can advice how to do it more beautiful?

_pytest/main.py
feuillemorte: Fixed

_pytest/main.py
feuillemorte: Fixed

feuillemorte: Fixed

_pytest/main.py
feuillemorte: Fixed

_pytest/main.py
```diff
@@ -308,7 +310,7 @@ def __init__(self, config):
         self.trace = config.trace.root.get("collection")
         self._norecursepatterns = config.getini("norecursedirs")
         self.startdir = py.path.local()
-
+        self.file_line = None
```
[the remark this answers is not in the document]
nicoddemus: Also, please add some unittests for the cases I mention above. 😉

_pytest/main.py
```diff
@@ -470,15 +476,106 @@ def _parsearg(self, arg):
...
+
+        return path
+
+    def _find_nearest_test_function(self, items):
+        """
+        Finds the nearest test function given a path in the form:
+
+        path/to/test.py:12
+
+        With ``12`` being a line number, search upwards or downwards in the file
```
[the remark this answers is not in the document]
feuillemorte: @nicoddemus downwards means that pytest search downwards in one case: if you enter line before the first test. In other cases pytest search upwards

_pytest/main.py
```diff
@@ -388,7 +390,11 @@ def _perform_collect(self, args, genitems):
         else:
             if rep.passed:
                 for node in rep.result:
-                    self.items.extend(self.genitems(node))
+                    items_list = []
```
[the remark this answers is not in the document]
feuillemorte:
```
items = self.genitems(node)
```
returns an error that you can't assign generator object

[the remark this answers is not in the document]
nicoddemus: Oh right, but because my code sample shadows the other `items` in the same scope:

```python
            self.items = items = []  # note "items" here
            ...
            if rep.passed:
                for node in rep.result:
                    items = self.genitems(node)  # "items" shadows the previous one
                    self.items.extend(items)
            return items  # oops!
```

Checking the code now, it doesn't seem to make sense to create that alias to `self.items` anymore (probably when it was first written it did, but overtime it lost its usefulness). I suggest to remove the alias altogether:

```python
            self.items = []  
            ...
            if rep.passed:
                for node in rep.result:
                    items = self.genitems(node)
                    if node.fspath in self._input_lines:
                        ...    
                    self.items.extend(items)
            return self.items
```

feuillemorte:
@blueyed Hi! As I remember we decided to do some refactoring first before continue this task... 

"we should split the logic of parsing/handling arguments out of Session and into its own class."

@nicoddemus am I right?

nicoddemus:
> @nicoddemus am I right?

@feuillemorte definitely. I had a partial work on this, but with the addition of the package-scoped fixtures that branch is no longer usable. I'm afraid this will have to remain in the background for a while. 😞

blueyed:
Has anything happened in this regard (the required refactoring)?

nicoddemus:
Unfortunately no; package-scoped fixtures got introduced first, which complicated the collection code even further, so I was driven to work on other things unfortunately...

nicoddemus:
Hi @feuillemorte,

PR #3321, opened 2018-03-19
3245: Fix tests to not fail when running from base directory

jeffreyrack:
This resolves the issue described in https://github.com/pytest-dev/pytest/issues/3245 where the example modules being included in the path causes some of the tests to fail when running from the base directory.

To resolve this, some of the example files/directories have been renamed so that there's not an overlap with modules that are creating by the testing scripts, and with this change now running `pytest` against the base directory will not have any failing tests.


nicoddemus:
Thanks @jeffreyrack!

PR #3349, opened 2018-03-28
3348: raise error on unknown arguments to raises

jeffreyrack:
Per https://github.com/pytest-dev/pytest/issues/3348, I've gone ahead and updated pytest.raises to raise a TypeError if an invalid keyword argument is passed.

In addition, I fixed an issue of a test case in pytest that was passing an invalid keyword argument to pytest.raises.

### DIFF
--- a/_pytest/python_api.py
+++ b/_pytest/python_api.py
@@ -597,6 +597,10 @@ def raises(expected_exception, *args, **kwargs):
             message = kwargs.pop("message")
         if "match" in kwargs:
             match_expr = kwargs.pop("match")
+        if len(kwargs.keys()) > 0:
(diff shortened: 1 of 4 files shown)


The-Compiler:
Really looking forward to dropping Python 2 support so we don't have to do kwargs handling by hand :wink:

nicoddemus:
Thanks @jeffreyrack!
 
Done @RonnyPfannschmidt's suggestion of using `if kwargs` and updated CHANGELOG slightly (did I ever say I love being able to edit files directly in GH's interface? 😉)

> Really looking forward to dropping Python 2 support so we don't have to do kwargs handling by hand 

Definitely! Perhaps next year with 4.0? 🤞

PR #3355, opened 2018-03-29
change collections.abc import to fix deprecation warnings on python 3.7

irmen:
This fixes the deprecation warnings from Python 3.7 related to importing abcs from the collections module, mentioned in issue #3339 

I couldn't fix the same for using the deprecated ``imp`` module, there's a bit too much wizardry going on with it in _pytest/assertion/rewrite.py ...

### DIFF
--- a/_pytest/assertion/util.py
+++ b/_pytest/assertion/util.py
@@ -5,7 +5,11 @@ import pprint
 import _pytest._code
 import py
 import six
-from collections import Sequence
+import sys
+if sys.version_info >= (3, 4):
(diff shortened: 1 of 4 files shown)


nicoddemus:
@irmen thanks! I checked out your fork to fix the linting issues myself, but then realized that the imports should go into `compat.py` which is our version compatibility layer.

Thanks!

nicoddemus:
And oops, I realized I switched the versions. 😅 

I will push another commit in a minute.

PR #3356, opened 2018-03-29
Remove stale CHANGELOG entries

nicoddemus:
Those were not removed in the last release because of a bug in
towncrier; I was about to fallback to don't use the .rst extension
for entries but just noticed that this was fixed in towncrier 17.8.0
so it should no longer be a problem.


nicoddemus:
For reference this towncrier issue was fixed in hawkowl/towncrier#102 (thanks @njsmith!)

PR #3370, opened 2018-04-05
no-capture in sample code demo'ing plugin

dchudz:
I think the current comment `will show "setting up"` here will be wrong by default (the print output will be captured). Would it be better for this sample code to say no-capture so the print statement actually shows up when users run it?


nicoddemus:
Definitely, thanks for the catch! 👍

PR #3386, opened 2018-04-11
Add package scoped fixtures #2283

jonozzz:
In response to #2283, adding package scoped fixtures, that follow python-nose's implementation.

```
package/
  __init__.py
  test_foo.py

# __init__.py
def setup_module():
  print("Setting up the package")

def teardown_module():
  print("Tearing down the package")
```
In this case teardown_module() will be executed when all tests in test_foo.py are done.


nicoddemus:
Hi @jonozzz, thanks a lot for working on this!

Could you please rebase against the latest `features` though? Your PR has a lot of commits which are not related to your change.

Looking forward to review this! 😁

PR #3436, opened 2018-04-30
--pdb treats KeyboardInterrupt as exception

### LINKED ISSUE #3299: Suggestion: KeybaordInterrupt pdb
When debugging infinite-loop issues, it can sometimes be useful to enable dropping to debugger upon Ctrl+C. Currently `--pdb` only catches other exceptions, `KeyboardInterrupt` is specifically ignored (`runner.py:193`, `CallInfo.__init__`). It would be nice to have a switch that makes it swallow `KeyboardInterrupt`-s as well instead. It would of course make it harder to quit (although a few consecutive Ctrl+C's usually due the trick), but it's a reasonable price to pay for those cases.

brianmaissy:
Fixes #3299

### DIFF
--- a/_pytest/runner.py
+++ b/_pytest/runner.py
@@ -176,7 +176,8 @@ def check_interactive_exception(call, report):
 def call_runtest_hook(item, when, **kwds):
     hookname = "pytest_runtest_" + when
     ihook = getattr(item.ihook, hookname)
-    return CallInfo(lambda: ihook(item=item, **kwds), when=when)
+    return CallInfo(lambda: ihook(item=item, **kwds), when=when,
+                    treat_keyboard_interrupt_as_exception=item.config.getvalue("usepdb"))
 
 
 class CallInfo(object):
@@ -184,7 +185,7 @@ class CallInfo(object):
     #: None or ExceptionInfo object.
     excinfo = None
 
-    def __init__(self, func, when):
+    def __init__(self, func, when, treat_keyboard_interrupt_as_exception=False):
         #: context of invocation: one of "setup", "call",
         #: "teardown", "memocollect"
         self.when = when
@@ -192,8 +193,11 @@ class CallInfo(object):
         try:
             self.result = func()
         except KeyboardInterrupt:
-            self.stop = time()
-            raise
+            if treat_keyboard_interrupt_as_exception:
+                self.excinfo = ExceptionInfo()
+            else:
+                self.stop = time()
+                raise
         except:  # noqa
             self.excinfo = ExceptionInfo()
         self.stop = time()
--- a/changelog/3299.feature.rst
+++ b/changelog/3299.feature.rst
@@ -1,0 +1,2 @@
+The ``--pdb`` option now causes KeyboardInterrupt to enter the debugger, instead of stopping the test session.
+On python 2.7, hitting CTRL+C again exits the debugger. On python 3.2 and higher, use CTRL+D.
--- a/_pytest/debugging.py
+++ b/_pytest/debugging.py
@@ -16,7 +16,7 @@ def pytest_addoption(parser):
     group = parser.getgroup("general")
     group._addoption(
         '--pdb', dest="usepdb", action="store_true",
-        help="start the interactive Python debugger on errors.")
+        help="start the interactive Python debugger on errors or KeyboardInterrupt.")
     group._addoption(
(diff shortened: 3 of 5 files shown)


_pytest/runner.py
```diff
@@ -176,24 +176,27 @@ def check_interactive_exception(call, report):
...
+    return CallInfo(lambda: ihook(item=item, **kwds), when=when, usepdb=item.config.getvalue("usepdb"))
 
 
 class CallInfo(object):
     """ Result/Exception info a function invocation. """
     #: None or ExceptionInfo object.
     excinfo = None
 
-    def __init__(self, func, when):
+    def __init__(self, func, when, usepdb=None):
```
blueyed: What about passing in the whole `item`?
Would make it more flexible for future changes.

brianmaissy: I don't mind changing it.

I figured it was cleaner to pass the minimum information I need to accomplish the task, passing more seemed like premature optimization.

nicoddemus: I personally prefer to pass just what you need as well. I would name the parameter differently as well, because it is more about treating `KeyboardInterrupt` as a special case... how about `stop_on_keyboard_interrupt=True`?

_pytest/runner.py
```diff
@@ -382,7 +385,8 @@ def __init__(self, longrepr, **extra):
 def pytest_make_collect_report(collector):
     call = CallInfo(
         lambda: list(collector.collect()),
-        'collect')
+        'collect',
+        None)
```
nicoddemus: We should treat this option as a `bool`, so this should be `False`; but I think we also want to stop on the debugger during collection, so we should pass the parameter here based on `config.getvalue("usepdb")`.

changelog/3299.feature.rst
```diff
@@ -0,0 +1 @@
+The ``--pdb`` option now causes KeyboardInterrupt to enter the debugger, instead of stopping the test session.
```
nicoddemus: Just to make sure: hitting CTRL+C again after landing on the debugger cases the debugger to exit right?

brianmaissy: Yes. I'll try to add a test that verifies that. Should I mention it in the changelog or docs?

_pytest/runner.py
brianmaissy: Sounds good, but in either case we are going to stop on keyboard interrupt. Either we capture the exception or we re-raise it. How about `treat_keyboard_interrupt_as_exception`?

_pytest/runner.py
brianmaissy: what is someone going to do with a debugger in the middle of runner.py?
I guess they could check the sys.path or something like that, but I'm worried about throwing them into a debugger in the middle of the pytest internals

changelog/3299.feature.rst
nicoddemus: I think just mentioning it in the changelog is sufficient, something like:

> ... instead of stopping the test session. Hitting CTRL+C again exits the debugger.

nicoddemus: I see the test failed... not sure if it is easy to fix it; if not I'm OK with dropping the test, I think it is enough to ensure that we enter pdb when we get a `KeyboardInterrupt`.

brianmaissy: I also expected the test to fail because of pexpect's `kill` method not being cross-platform or something like that, but actually the test failed because pdb indeed behaves differently in python 3.6, and the KeyboardInterrupt does not cause it to exit. Are we ok with that?

brianmaissy: looks like it was an intentional change: https://bugs.python.org/issue7245.
Haha now I understand the discussion on the issue.

brianmaissy: The `nosigint` parameter to the Pdb constructor is supposed to disable that functionality, according to the docs (https://docs.python.org/3/library/pdb.html#pdb.Pdb). For some reason it doesn't seem to be working for me, but I can continue looking into it if we care about it.

On the other hand, probably better to let the pdb act like it usually acts for each version of python.

nicoddemus: > On the other hand, probably better to let the pdb act like it usually acts for each version of python.

Definitely, I agree! Thanks for researching this. 👍

nicoddemus:
Awesome work @brianmaissy, thanks a lot!

PR #3445, opened 2018-05-03
Release 3.6.0

nicoddemus:
@The-Compiler would you like to take this out for a spin?


The-Compiler:
Just opened #3446 which should IMHO block the release.

nicoddemus:
Agreed, let's deal with #3446 first

PR #3497, opened 2018-05-22
Import Mapping and Sequence from compat in python_api::approx

hroncok:
Related to https://github.com/pytest-dev/pytest/issues/3339

Fixes a DeprecationWarning on Python 3.7

Adds Mapping to compat

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -144,6 +144,7 @@ Michael Seifert
 Michal Wajszczuk
 Mihai Capotă
 Mike Lundy
+Miro Hrončok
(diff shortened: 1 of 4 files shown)


nicoddemus:
LGTM, thanks!

Would you mind adding yourself to `AUTHORS` and adding a new `trivial` changelog entry? Thanks again!

hroncok:
On it.

hroncok:
Done. Probably some typos in there, because I'm a bit multitasking, so please double check.

PR #3505, opened 2018-05-25
Fix tox spelling (all lower case).

obestwalter:
This is a mini-mini fix, so no changelog entry or anything else is necessary, I guess.


nicoddemus:
Thanks @obestwalter!

obestwalter:
Thanks for tolerating my mild tox-should-be-all-lower-case obsession :)

nicoddemus:
Not at all!

PR #3545, opened 2018-06-06
Improve display of hint about --fulltrace with KeyboardInterrupt

blueyed:
- display the location first
- display hint about --fulltrace in parenthesis


nicoddemus:
Thanks!

PR #3563, opened 2018-06-10
Print <undetermined location> instead of None for warnings

asottile:


asottile:
I remembered the changelog this time _and_ I named the file correctly -- go me!

blueyed:
Thanks for fixing this.

I wonder however why the location is missing in the first place, and if it could be added?

asottile:
I *think* location usually refers to what _test_ triggered the warning, however infra-level things don't really happen in the context of a test (I could be wrong, I haven't personally used the warnings plugin intentionally yet).

PR #3578, opened 2018-06-12
Fix three spaces instead of four in docs

BartoszCki:
Fixing three spaces instead of four in assert.rst


nicoddemus:
Thanks @BartoszCki!

PR #3584, opened 2018-06-14
Fix encoding error with `print` statements in doctests

### LINKED ISSUE #3583: Printing Unicode in doctest under Python 2.7 doesn't work
As of pytest 3.6.1, doctests that `print`s Unicode output (not just evaluating to Unicode; they have to call `print`) are failing with a `UnicodeEncodeError` under Python 2.7 but succeeding under Python 3.3+, regardless of locale or whether `doctest_encoding = utf-8` is set in `pytest.ini`.  This has been observed on both Mac OS X and Ubuntu Trusty & Xenial.

This bug was first noticed in <https://stackoverflow.com/q/50860112/744178>; see there for slightly more information.  Issue #2728 may also be related.

Minimal example:

`pytest-bug.rst`:

```
Here is a failing example::

    >>> print(u'\xE5\xE9\xEE\xF8\xFC')
    åéîøü
```

Running `pytest pyttest-bug.rst` fails under Python 2.7.10 and succeeds under Python 3.6.5.

jwodder:
This fixes #3583.  This fix was suggested by Stack Overflow user phd in
<https://stackoverflow.com/a/50863820/744178>.

### DIFF
--- a/changelog/3583.bugfix.rst
+++ b/changelog/3583.bugfix.rst
@@ -1,0 +1,1 @@
+Fix encoding error with `print` statements in doctests
--- a/src/_pytest/doctest.py
+++ b/src/_pytest/doctest.py
@@ -505,7 +505,7 @@ def _fix_spoof_python2(runner, encoding):
 
         def getvalue(self):
             result = _SpoofOut.getvalue(self)
-            if encoding:
+            if encoding and isinstance(result, bytes):
                 result = result.decode(encoding)
             return result
 
--- a/testing/test_doctest.py
+++ b/testing/test_doctest.py
@@ -655,6 +655,22 @@ class TestDoctests(object):
         result = testdir.runpytest(p, "--doctest-modules")
         result.stdout.fnmatch_lines(["* 1 passed *"])
 
+    def test_print_unicode_value(self, testdir):
+        """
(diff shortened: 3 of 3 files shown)


jwodder:
I would have added the minimal example from #3583 as a test case, but I wasn't sure how exactly to do that.  Someone who knows what they're doing should probably do that.

changelog/3583.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix encoding error with `print` statements in doctests
```
nicoddemus: This should be `rst`:

```
Fix encoding error with ``print`` statements in doctests.
```

jwodder: It is RST; reStructuredText uses single backticks for individual identifiers, while double backticks are for arbitrary inline code.

nicoddemus: Oh you are right, sorry; 

We use inline code in the changelog though because there won't be any references to it in the final document (but TBH I don't remember if they render differently).

jwodder:
@nicoddemus I added a failing test case, but now the fix doesn't fix it; it was working when I tried modifying a copy of pytest in a venv a few hours ago....

PR #3608, opened 2018-06-21
add reamde for .pytest_cache

### LINKED ISSUE #3519: Add a README to .pytest_cache directory

avirlrma:
fixes #3519


avirlrma:
@nicoddemus I have added the tests,the readme content and everything else as well.
Please review the tests

src/_pytest/cacheprovider.py
```diff
@@ -103,6 +103,22 @@ def set(self, key, value):
             with f:
                 self.trace("cache-write %s: %r" % (key, value))
                 json.dump(value, f, indent=2, sort_keys=True)
+                self._ensure_readme()
+
+    def _ensure_readme(self):
+        content_readme = """# pytest cache directory #
```
nicoddemus: Please use `textwrap.dedent` so you don't need to have the lines spill over the right side:

```python

    def _ensure_readme(self):
        content_readme = textwrap.dedent("""\
        # pytest cache directory #
 
        This directory contains data from the pytest's cache plugin, \
        which provides the `--lf` and `--ff` options, as well as the `cache` fixture.
         
        **Do not** commit this to version control.
         
        See [the docs](https://docs.pytest.org/en/latest/cache.html) for more information.
        """)
        if self._cachedir.check(dir=True):
            readme_path = self._cachedir.join("README.md")
            if not readme_path.check(file=True):
                readme_path.write(content_readme)
```

testing/test_cacheREADME.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import absolute_import, division, print_function
+import _pytest
+
+pytest_plugins = ("pytester",)
+
+
+class Helper(object):
```
nicoddemus: I don't think we need this class at all, it only adds to (albeit small) increase in code complexity without any benefit.

I suggest to just move the `check_readme` function to `TestHelper` and delete this class.

testing/test_cacheREADME.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def check_readme(self, testdir):
+        config = testdir.parseconfigure()
+        readme = config.cache._cachedir.join("README.md")
+        return readme.isfile()
+
+
+class TestReadme(Helper):
```
nicoddemus: Let's move this class to `test_cache.py`, I don't think we need to keep this in a separate file just because of a single test.

testing/test_cacheREADME.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def test_readme_passed(self, testdir):
+        testdir.tmpdir.join("test_a.py").write(
```
nicoddemus: It's easier to use `testdir.makepyfile`:

```python
testdir.makepyfile("""
    def test_always_passes():
        assert 1
""")
```

nicoddemus:
> solves #3519

A small detail, `solves` is not one of the keywords recognized by GH to automatically close issues when PRs get merged, see https://help.github.com/articles/closing-issues-using-keywords.

src/_pytest/cacheprovider.py
avirlrma: I knew this was wrong,but even after searching couldn't come up with a solution. Also shouldn't flake point this out?

testing/test_cacheREADME.py
avirlrma: will do.

testing/test_cacheREADME.py
avirlrma: got it

src/_pytest/cacheprovider.py
nicoddemus: > Also shouldn't flake point this out?

I don't think so, the code is not *wrong*, but using dedent makes it more pleasing to the eyes. :grin

avirlrma:
Thanks, @nicoddemus for helping me out all the time and bearing with me.

nicoddemus:
Not at all, your contribution is very much appreciated! 👍

avirlrma:
@nicoddemus I looked into issue [1120](https://github.com/pytest-dev/pytest/issues/1120) and wanted to work on it,Can you have a look and comment there please?

PR #3611, opened 2018-06-22
Doc Fix: skipif's example description

AdamEr8:
Fixed a small mistake in skipif's example description, which read the opposite of what the code describes


doc/en/skipping.rst
```diff
@@ -80,7 +80,7 @@ during import time.
 
 If you wish to skip something conditionally then you can use ``skipif`` instead.
 Here is an example of marking a test function to be skipped
-when run on a Python3.6 interpreter::
+when run on an interpreter earlier than Python3.6 ::
 
     import sys
     @pytest.mark.skipif(sys.version_info < (3,6),
```
AdamEr8: Maybe the reason "requires python3.6" should also change to "requires python3.6 or higher" to fully match the condition

nicoddemus: Agreed!

PR #3654, opened 2018-07-03
Switch to new API

### LINKED ISSUE #3653: Avoid deprecated attrs feature `convert`
https://github.com/pytest-dev/pytest/commit/37793d4cdbf71f597a3c69c827e0c010528d53a5 introduced a use of the `convert` keyword argument to `attr.ib`.  This is deprecated in attrs:

> DeprecationWarning: The `convert` argument is deprecated in favor of `converter`.  It will be removed after 2019/01.                                 

The replacement should be used instead.

exarkun:
Fixes #3653

### DIFF
--- a/changelog/3653.trivial.rst
+++ b/changelog/3653.trivial.rst
@@ -1,0 +1,1 @@
+Fix usage of ``attr.ib`` deprecated ``convert`` parameter.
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks @exarkun!

nicoddemus:
Btw, we don't catch that warning because unfortunately we don't yet capture deprecation warnings at import time.

exarkun:
Thanks for the quick turn-around on this.  My CI was failing but the fixed release came out so quickly I didn't even have time to implement a work-around on my side.  Much appreciated. :)

nicoddemus:
No worries, thanks again for the PR! 👍

PR #3660, opened 2018-07-04
Correct code blocks in docs

asottile:
I know I make this mistake all the time, let's see if pre-commit can't help prevent it :)

if this pygrep hook works well for pytest, I'm eventually planning to collect up a bunch of them and put them in a more centralized place.  I have a few that I've copied from project-to-project myself and feel it would better to share!

### DIFF
--- a/doc/en/mark.rst
+++ b/doc/en/mark.rst
@@ -52,8 +52,6 @@ should add ``--strict`` to ``addopts``:
         serial
 
 
-.. `marker-iteration`
-
 Marker revamp and iteration
 ---------------------------
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,8 @@ repos:
         language: python
         additional_dependencies: [pygments, restructuredtext_lint]
         python_version: python3.6
+    -   id: rst-backticks
+        name: rst ``code`` is two backticks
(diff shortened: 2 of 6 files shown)


doc/en/mark.rst
```diff
@@ -52,8 +52,6 @@ should add ``--strict`` to ``addopts``:
         serial
 
 
-.. `marker-iteration`
```
nicoddemus: Hmm interesting! This should be 

```rst
.. _`marker-iteration`:
```

Nice that the hook catches this.

nicoddemus: In this case removing it is fine, but in general:

```rst
.. something
``` 

this [is a comment in RST](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#comments), so I'm not sure it is correct to remove

```rst
.. `something`
``` 

is something that should be done automatically.

asottile: yeah none of this was automatic unfortunately :(

nicoddemus: Oh I see now, `pygrep` only complains about the matches but doesn't fix them, is that right?

asottile: correct, it's a cheap way to make a linter

nicoddemus:
Seems useful, let's see how it fares! 👍

PR #3663, opened 2018-07-06
Add reference docs for pytest.mark.usefixtures

nicoddemus:


doc/en/reference.rst
```diff
@@ -161,6 +161,20 @@ Skip a test function if a condition is ``True``.
...
 
 
+.. _`pytest.mark.usefixtures ref`:
+
+pytest.mark.usefixtures
+~~~~~~~~~~~~~~~~~~~~~~~
+
+**Tutorial**: :ref:`usefixtures`.
+
+Mark a test function as using the given fixture names.
```
obestwalter: I wonder if it could be a good idea to add an explicit warning/reminder here that you can only mark test functions like this and not other fixtures to address https://github.com/pytest-dev/pytest/issues/1014

nicoddemus: Ahh good idea, will do

obestwalter: Something like:

"Be aware that although you can use fixtures from oher fixtures by adding them to the function parameters, you cannot do with by marking the fixture function"

Maybe even with an example?

This works:

```
@pytest.fixture
def my_fixture(my_other_fixture):
    [...]
```

This will **not** yield the wanted result and currently not result in an error or a warning either.

```
@pytest.mark.usefixtures("my_other_fixture")
@pytest.fixture
def my_fixture_that_sadly_wont_use_my_other_fixture():
    [...]
```

I would even suggest to put this in a warning box, because this can be surprising even for experienced pytest users, especially due to the lack of an error atm.

nicoddemus: Thanks @obestwalter!

I added the warning but the example I added to the tutorial/normal documentation. IMHO we should keep the reference as lean as possible.

doc/en/reference.rst
```diff
@@ -170,9 +170,14 @@ pytest.mark.usefixtures
 
 Mark a test function as using the given fixture names.
 
+.. warning::
+
+    This mark can be used with *test functions* only, having no affect when applied
+    to a **fixture** function.
+
```
obestwalter: perfect :)

PR #3665, opened 2018-07-07
Small tweaks to the changelog entries

nicoddemus:
Some small tweaks on the text I noticed while reviewing the changelog for the upcoming 3.7 release.

Also pushed a tweak to our changelog rendering: now it writes the issue link first:

```
- `#3623 <https://github.com/pytest-dev/pytest/issues/3623>`_: Introduce ``pytester.copy_example`` as helper to do acceptance tests against examples from the project.
```

I want to be able to use a new feature from towncrier 18.6 which no longer wraps lines automatically, which allows us to use multi-line entries, including coding samples which can be very handy for new features. The issue link at the end was a problem because it did appear right after the line in a multi-line string.

Here's a local draft I did of dummy issues using this feature:

```rst
- `#3665 <https://github.com/pytest-dev/pytest/issues/3665>`_: Dummy entry just to demonstrate multi-paragraph changelog entry:

      * Some change
      * Another change.

  And now an example:

      .. code-block:: python

          def somecode():
              pass


- `#3666 <https://github.com/pytest-dev/pytest/issues/3666>`_: Some other entry.

      .. code-block:: python

          def somecode():
              pass
```

I generated the docs locally and it looks good.

### DIFF
--- a/changelog/README.rst
+++ b/changelog/README.rst
@@ -26,7 +26,7 @@ changelog using that instead.
 
 If you are not sure what issue type to use, don't hesitate to ask in your PR.
 
-Note that the ``towncrier`` tool will automatically
-reflow your text, so it will work best if you stick to a single paragraph, but multiple sentences and links are OK
-and encouraged. You can install ``towncrier`` and then run ``towncrier --draft``
+``towncrier`` preserves multiple paragraphs and formatting (code blocks, lists, and so on), but for entries
+other than ``features`` it is usually better to stick to a single paragraph to keep it concise. You can install
+``towncrier`` and then run ``towncrier --draft``
 if you want to get a preview of how your change will look in the final release notes.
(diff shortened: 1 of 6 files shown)


changelog/README.rst
```diff
@@ -26,7 +26,7 @@ changelog using that instead.
 
 If you are not sure what issue type to use, don't hesitate to ask in your PR.
 
-Note that the ``towncrier`` tool will automatically
-reflow your text, so it will work best if you stick to a single paragraph, but multiple sentences and links are OK
-and encouraged. You can install ``towncrier`` and then run ``towncrier --draft``
+``towncrier`` preserves multiple paragraphs and formatting (code blocks, lists, and so on), but for entries
```
nicoddemus: cc @njsmith, since I borrowed this from `trio` ;)

nicoddemus:
> If there's never a case for a multi line entry, should we add a quick local hook linter to prevent this in the future?

That's an interesting idea. 👍 

I'm merging this for now, if we want to tackle the hook we should do it in a separate task. 😁

PR #3669, opened 2018-07-09
Remove some extraneous `# noqa` comments

asottile:
This was partially automated with https://github.com/asottile/yesqa

_with a few caveats_:
- it was run under python2 (chosen arbitrarily, when run under python3 other
  things were changed)
- I used `git checkout -p` to revert the removal of `noqa` comments from
  `cmp()` lines.

### DIFF
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -41,8 +41,8 @@ PY36 = sys.version_info[:2] >= (3, 6)
 MODULE_NOT_FOUND_ERROR = "ModuleNotFoundError" if PY36 else "ImportError"
 
 if _PY3:
-    from collections.abc import MutableMapping as MappingMixin  # noqa
-    from collections.abc import Mapping, Sequence  # noqa
+    from collections.abc import MutableMapping as MappingMixin
+    from collections.abc import Mapping, Sequence
 else:
     # those raise DeprecationWarnings in Python >=3.7
     from collections import MutableMapping as MappingMixin  # noqa
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -71,7 +71,7 @@ def main(args=None, plugins=None):
         return 4
 
 
-class cmdline(object):  # NOQA compatibility namespace
+class cmdline(object):  # compatibility namespace
     main = staticmethod(main)
 
 
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -941,7 +941,7 @@ class TestAssertionRewriteHookDetails(object):
             e = IOError()
             e.errno = 10
             raise e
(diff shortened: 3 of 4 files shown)


testing/code/test_source_multiline_block.py
```diff
@@ -1,4 +1,3 @@
-# flake8: noqa
```
asottile: actually, going to revert this line -- we don't use autopep8 any more so I can merge this back into the other test -- PR incoming for that!

blueyed:
> This was partially automated with https://github.com/asottile/yesqa

How long did it took approximately?

asottile:
> > This was partially automated with https://github.com/asottile/yesqa
>
> How long did it took approximately?

A second or so. yesqa is kind of slow. I basically ran `git ls-files -- '*.py' | xargs yesqa` (though I guess I could have used `pre-commit try-repo` to ease installation or used -P4 for more speed)

blueyed:
I thought it was slower, since it runs flake8 for any removed comment I guess?
A second is good then.. :)
(also depends on the number of "# noqa" comment of course, which shouldn't be that many).

asottile:
@blueyed it's less clever than that unfortunately -- essentially:

1. read the source
2. strip all noqa comments (_new_: if there are none to strip, skip the file)
3. run flake8 on the file
4. put back any comments that caused lint errors to appear

so worst case it runs flake8 once per file -- best case it runs flake8 not at all

asottile:
> So it knows from a single run already which caused problems (makes sense) - which is good I think?!
> How could it be more clever?

first it could batch a lot better, right now it invokes a subprocess per file, flake8 has its own multiprocessing code which would cut out a lot of the overhead (most of flake8's startup time is `pkg_resources` and an absolute _ton_ of regex initialization).  In fact, most of the time invoking flake8 isn't actually linting :(  Even if flake8 were running in `-j1` mode, or serially in-process (no longer possible in flake8 3.x :sob) the overhead would be much much much smaller and it would probably complete on the order of hundreds of ms instead of seconds.

second, there's some flake8 plugins in which this approach is just... well... wrong.  The `# noqa` comment can prevent a lint on a completely different line number -- and well, yesqa is just not smart enough to know that 🤷‍♂️ 

I got quite discouraged after finding some cases like this and don't think it'll ever work 100%, but it _is_ pretty good for _normal_ cases.

yesqa is one of my few projects where I truly believe the `0.0.x` version is appropriate -- it works really well but without a ton of work has some fatal flaws.

There's also cases (not unlike `pytest`) where `# noqa` is used to ignore version-specific errors -- think things like `RecursionError` (py35+) or `cmp(...)` (py2 only) -- both would need `# noqa` comments for their version counterparts -- but `yesqa` only "tests" a single version of python

PR #3680, opened 2018-07-14
Manage GH labels using the new 'labels' tool

nicoddemus:
Using: https://github.com/hackebrot/labels

This will make it easier to manage labels in bulk than using the GitHub's interface. The `labels.toml` file being committed was fetch from the current labels unchanged.

### DIFF
--- a/.github/labels.toml
+++ b/.github/labels.toml
@@ -1,0 +1,149 @@
+["os: cygwin"]
+color = "006b75"
+description = "cygwin platform-specific problem"
+name = "os: cygwin"
+
+["os: linux"]
+color = "1d76db"
+description = "linux platform-specific problem"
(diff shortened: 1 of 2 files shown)


hackebrot:
Yay! 😃

hroncok:
This is awesome. Do you run it as a service to keep it in sync?

nicoddemus:
Not currently, we just run the command-line tool when the file changes. At the moment a service might be overkill given that we don't change labels very often. 👍

webknjaz:
So https://github.com/pytest-dev/pytest/pull/7280/files#diff-a1a2118c193bef589c6c34d938f194516b3ca948057909ce3c9a575ce28de52aL42 removed this from the dev guide but the config still exists...

PR #3694, opened 2018-07-18
Fix mark.rst typos & grammar

drewrisinger:
Fix minor typos and grammar in the documentation.


tadeoos:
@drewrisinger, before we merge this, could you add an entry to the changelog file as per https://github.com/pytest-dev/pytest/tree/master/changelog? (use a PR number instead of issue number, since I don't see any issue associated with this)

Issue #3695, opened 2018-07-18
pytest.approx abs/rel logic seems to be flipped on np.array()


abrammer:
Hopefully, I'm not missing something obvious here. 

Absolute and relative tolerance arguments or logic seem to get flipped when the comparison is against a np.array 

Think I've tracked it down to this line.  https://github.com/pytest-dev/pytest/blob/f93995e15c377d9050eb09bcf9518914067ff402/src/_pytest/python_api.py#L214
`abs` and `rel` arguments should be in the other order according to the `ApproxBase.__init__`
https://github.com/pytest-dev/pytest/blob/f93995e15c377d9050eb09bcf9518914067ff402/src/_pytest/python_api.py#L46

```
pytest.__version__
'3.6.3'
np.__version__
'1.13.1'
```

```
abs_check = pytest.approx(100, abs=0.1, rel=0)
rel_check = pytest.approx(100, abs=0, rel=0.1)
99 == abs_check
# False
np.array(99) == abs_check
# True

99 == rel_check
# True
np.array(99) == rel_check
# False
```

pytestbot:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/pytest-dev/pytest/issues/3593 (pytest.approx raises TypeError for 0-d numpy array), https://github.com/pytest-dev/pytest/issues/3312 (pytest.approx doesn't support comparison between arrays and scalars), https://github.com/pytest-dev/pytest/issues/1994 (Supporting arrays for approx), https://github.com/pytest-dev/pytest/issues/3482 (Extend pytest.approx to handle strings. ), and https://github.com/pytest-dev/pytest/issues/3164 (TypeError with pytest.approx on nested dictionaries).

PR #3696, opened 2018-07-18
bugfix in ApproxNumpy initialisation, use keywords for arguments to fix

abrammer:
With respect to #3695 , flip the arguments in the initialization and also use keywords so it's more obvious and less likely to happen in future.  

Can add tests along the lines of the code in #3695 if desired but it seems like an obvious typo fix

### DIFF
--- a/changelog/3695.bugfix.rst
+++ b/changelog/3695.bugfix.rst
@@ -1,1 +1,1 @@
-Fix ApproxNumpy initialisation argument mixup. abs rel tolerances were previously flipped
+Fix ``ApproxNumpy`` initialisation argument mixup, ``abs`` and ``rel`` tolerances were flipped causing strange comparsion results.
--- a/changelog/3695.bugfix.rst
+++ b/changelog/3695.bugfix.rst
@@ -1,1 +1,2 @@
 Fix ``ApproxNumpy`` initialisation argument mixup, ``abs`` and ``rel`` tolerances were flipped causing strange comparsion results.
+Add tests to check ``abs`` and ``rel`` tolerances for ``np.array`` and test for expecting ``nan`` with ``np.array()``
--- a/changelog/3695.bugfix.rst
+++ b/changelog/3695.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ApproxNumpy initialisation argument mixup. abs rel tolerances were previously flipped
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -211,7 +211,7 @@ class ApproxScalar(ApproxBase):
         the pre-specified tolerance.
         """
         if _is_numpy_array(actual):
-            return all(a == self for a in actual)
+            return all(a == self for a in actual.flat)
 
         # Short-circuit exact equality.
         if actual == self.expected:
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -211,9 +211,7 @@ class ApproxScalar(ApproxBase):
         the pre-specified tolerance.
         """
         if _is_numpy_array(actual):
-            import numpy as np
-
-            return np.all(abs(self.expected - actual) <= self.tolerance)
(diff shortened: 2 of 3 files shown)


asottile:
Let's add a quick test since the current code doesn't actually ensure `ApproxNumpy` works

abrammer:
You think add a test to check tolerances, or just make the current tests such that abs & rel tolerances aren't equivalent.  I can push a commit either way this eve.

asottile:
Anything that fails before your change and passes afterwards. Either approach is fine though the "different params" one is what I'd probably go for.

abrammer:
Something else is broken here.  I typo'd the last rel argument, so it's 10% instead of 1%.  
Fixing the typo means that it fails though, which doesn't make sense as 99 is obviously within 1% of 100.

abrammer:
Justification for f0db64a:

currently pytest will calculate the tolerances against either value if it's a numpy array. 
`np.array(1e7) == pytest.approx(1, abs=1)`

The above will pass a test in 3.6.4.  With the earlier commits 
`np.array(1e7) == pytest.approx(1, rel=1)`
would still pass a test.  

Unless I'm missing something the approx should be 1±1 therefore it should be wrong. 
I have stripped out the repeat ApproxNumpy() initilsiation at 
https://github.com/pytest-dev/pytest/blob/b7419bd9bb2ad5b609532766ceb77e7b732e1dcf/src/_pytest/python_api.py#L214
to handle this.  This may not elegantly handle nans but should just fail.   All current tests pass.

kalekundert:
The extra ``ApproxNumpy`` initialization was there to properly handle corner-cases involving ``nan`` and ``inf``.  In particular, the ``nan_ok`` option will be ignored without it.  That said, looking back at this code now, the extra ``ApproxNumpy`` initialization was also broken because it switched the "actual" and "expected" values, which would cause relative tolerances to be calculated incorrectly.

I think the best way to write this line would be like this:

        if _is_numpy_array(actual):
            return all(a == self for a in actual)

Thanks for the PR and the tests (clearly this code needs more tests)!

nicoddemus:
Thanks @kalekundert for the suggestion! I implemented it myself, I think we can merge this for now, it has been waiting here long enough. I encourage others to open PRs later to improve the tests if needed! 👍

abrammer:
I have changes and tests coming, away from my computer for a couple days so will update with respect to the @kalekundert and tests on Sunday.

nicoddemus:
@abrammer sure think, let's wait for you to get around to this then! Thanks a lot for following up with this. 👍

tadeu:
Just brainstorming, but it would be less errorprone to maintain if all corner cases are tested (it's hard to remember all corner cases when maintaining the code):

* 0-d, 1-d, 2-d
* common values, nan, inf
* check for `rel` only (equality and inequality); cases where `abs` would fail
* check for `abs` only (equality and inequality); cases where `rel` would fail
* no numpy array, numpy array on left-side, numpy array on right-side, numpy array on both sides
* pytest.approx on left-side, pytest.approx on right side
* relative tolerance that would fail if sides are switched

Perhaps an approach with `parametrize` would be helpful.
I could work on this if you think it's useful (would it be another issue?)

abrammer:
@tadeu I think the tests I put in cover most of those cases.   Pulled logic from earlier tests, though maybe a decorator method could be better?

I was going to suggest np.at_least1d(actual) with respect to bf127a6, might be a little more memory efficient but probably doesn't matter for tests.

kalekundert:
Looks good to me.  I think one of ``flat``, ``flatten()``, or ``ravel()`` is needed to handle comparisons with multidimensional arrays, and my understanding is that ``flat`` just makes an iterator and doesn't copy anything.  But I haven't really thought about it that carefully.

abrammer:
Yeah you're right. Flatten copies, flat just iterates. Nice.

tadeu:
@abrammer, awesome improvement on the tests ;)

PR #3729, opened 2018-07-28
Fix "Pytest" to "pytest" in the CHANGELOG

nicoddemus:
Now that we have fixed towncrier to render the proper title, seems fitting
to update the others


nicoddemus:
Merging as this seems to be a no-brainer.

PR #3731, opened 2018-07-29
Use upstream rst-backticks hook

asottile:


asottile:
I decided this was good enough to share :)

nicoddemus:
Failure is unrelated and will be fixed by #3734

PR #3735, opened 2018-07-30
Deprecate pytest namespace

### LINKED ISSUE #2639: deprecate pytest_namespace alltogether
with the merge of  #2636 pytest itself no longer needs the pytest_namespace hook

the need for preinit is gone as well
so to follow up we should remind people that the hook is deprecated

- [ ] print deprecation warnings when hook definitions are found
- [ ] changelog 
- [ ] notes about there not being a replacement
- [x] deprecation roadmap entry

nicoddemus:
This is based on #3732, so it is better to review after that PR gets merged.

Fix #2639

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -42,8 +42,15 @@ def pytest_namespace():
     .. warning::
         This hook has been **deprecated** and will be removed in pytest 4.0.
 
-        Plugins who need this feature are suggested
-        to import ``pytest`` and set attributes explicitly during ``pytest_configure``.
+        Plugins whose users depend on the current namespace functionality should prepare to migrate to a
+        namespace they actually own.
+
+        To support the migration its suggested to trigger ``DeprecationWarnings`` for objects they put into the
+        pytest namespace.
+
+        An stopgap measure to avoid the warning is to monkeypatch the ``pytest`` module, but just as the
+        ``pytest_namespace`` hook this should be seen as a temporary measure to be removed in future versions after
+        an appropriate transition period.
     """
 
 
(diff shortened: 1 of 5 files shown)


src/_pytest/hookspec.py
```diff
@@ -22,17 +23,27 @@ def pytest_addhooks(pluginmanager):
...
 
     This hook is called at plugin registration time.
 
     .. note::
         This hook is incompatible with ``hookwrapper=True``.
+
+    .. warning::
+        This hook has been **deprecated** and will be removed in pytest 4.0.
+
+        Plugins who need this feature are suggested
```
RonnyPfannschmidt: i propose to align this somewhere along
```
Plugins whose users depend on the current name-space functionality should prepare to migrate to a namespace they actually own.

To support the migration its suggested to trigger `DeprecationWarnings` for objects they put into the pytest namespace.

An stopgap measure to avoid the warning is to monkeypatch the `pytest` module.
but just as the `pytest_namespace` hook this should be seen as a temporary measure to be removed.

```
simply to avoid accidentally declaring monkeypatching ´pytest` fine

nicoddemus: Sure will do

RonnyPfannschmidt:
the base dance removed the already merged commits, i'm puzzled as to why its not more easy

nicoddemus:
> the base dance removed the already merged commits, i'm puzzled as to why its not more easy

Yeah it should detect that automatically.

src/_pytest/hookspec.py
nicoddemus: Done

src/_pytest/hookspec.py
```diff
@@ -22,17 +23,34 @@ def pytest_addhooks(pluginmanager):
     """
 
 
-@hookspec(historic=True)
+@hookspec(
+    historic=True,
+    warn_on_impl=RemovedInPytest4Warning(
```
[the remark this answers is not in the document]
nicoddemus: Ooops good catch!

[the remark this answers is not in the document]
nicoddemus: Done 👍

PR #3761, opened 2018-08-01
Fix recursion in pytest.approx() with arrays in numpy<1.13

### LINKED ISSUE #3748: 3.7.0 Maximum recursion depth exceeded
When comparing `numpy.float64` with `float` using `pytest.approx` we reach maximum recursion. It works in pytest `3.5.0` and `3.6.0` thou.
Run on Ubuntu 16.04, python 3.5.2, numpy 1.11.3

It seems that changing
```python
return all(a == self for a in actual.flat)
```
to
```python
return all(self == a for a in actual.flat)
```
in `pytest_api.py` file fixes the issue (see the bottom of traceback below).

```python
import pytest
import numpy as np
val = np.float64(1.0)
val_to_cmp = 1.0
val == pytest.approx(val_to_cmp)

RecursionError                            Traceback (most recent call last)
<ipython-input-8-b44e412779ee> in <module>()
----> 1 val == pytest.approx(val_to_cmp)

~/venvs/dl3/lib/python3.5/site-packages/_pytest/python_api.py in __eq__(self, actual)
    212         """
    213         if _is_numpy_array(actual):
--> 214             return all(a == self for a in actual.flat)
    215 
    216         # Short-circuit exact equality.

~/venvs/dl3/lib/python3.5/site-packages/_pytest/python_api.py in <genexpr>(.0)
    212         """
    213         if _is_numpy_array(actual):
--> 214             return all(a == self for a in actual.flat)
    215 
    216         # Short-circuit exact equality.

... last 2 frames repeated, from the frame below ...

~/venvs/dl3/lib/python3.5/site-packages/_pytest/python_api.py in __eq__(self, actual)
    212         """
    213         if _is_numpy_array(actual):
--> 214             return all(a == self for a in actual.flat)
    215 
    216         # Short-circuit exact equality.

RecursionError: maximum recursion depth exceeded while calling a Python object
```

nicoddemus:
Fix #3748

cc @ironiccode @tigarmo

I tested this patch manually using this example from @tigarmo using `numpy 1.11`:

```python
def testApprox():
    from pytest import approx
    import numpy as np

    a = np.array([1, 2, 3])
    assert a == approx([1, 2, 3])
```

I did not add tests for this because it requires `numpy<1.13`. I thought it overkill to add another CI environment because of this.

### DIFF
--- a/changelog/3748.bugfix.rst
+++ b/changelog/3748.bugfix.rst
@@ -1,0 +1,1 @@
+Fix infinite recursion in ``pytest.approx`` with arrays in ``numpy<1.13``.
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -211,7 +211,9 @@ class ApproxScalar(ApproxBase):
         the pre-specified tolerance.
         """
         if _is_numpy_array(actual):
-            return all(self == a for a in actual.flat)
+            # Call ``__eq__()`` manually to prevent infinite-recursion with
+            # numpy<1.13.  See #3748.
+            return all(self.__eq__(a) for a in actual.flat)
 
         # Short-circuit exact equality.
         if actual == self.expected:
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -211,7 +211,7 @@ class ApproxScalar(ApproxBase):
         the pre-specified tolerance.
         """
         if _is_numpy_array(actual):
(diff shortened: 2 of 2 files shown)


kalekundert:
This was a pretty weird bug, so I thought it would be prudent to put a comment in the code indicating why the comparison has to be done in a certain way.

PR #3764, opened 2018-08-01
Fix `TypeError` when the assertion message is `bytes` in python 3.

### LINKED ISSUE #3763: python 3: `assert ..., bytesobject`: TypeError: string argument without an encoding
Distilled from an actual problem

```python
def test():
    assert False, b'ohai\n'
```

```console
$ pytest --version
This is pytest version 3.7.0, imported from /tmp/dumb-init/venv/lib/python3.6/site-packages/pytest.py
```

```console
$ pytest t.py
============================== test session starts ===============================
platform linux -- Python 3.6.6, pytest-3.7.0, py-1.5.4, pluggy-0.7.1
rootdir: /tmp/dumb-init, inifile: pytest.ini
collected 1 item                                                                 

t.py F                                                                     [100%]

==================================== FAILURES ====================================
______________________________________ test ______________________________________

    def test():
>       assert False, b'foo'

t.py:2: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = b'foo'

    def _format_assertmsg(obj):
        """Format the custom assertion message given.
    
        For strings this simply replaces newlines with '\n~' so that
        util.format_explanation() will preserve them instead of escaping
        newlines.  For other objects py.io.saferepr() is used first.
    
        """
        # reprlib appears to have a bug which means that if a string
        # contains a newline it gets escaped, however if an object has a
        # .__repr__() which contains newlines it does not get escaped.
        # However in either case we want to preserve the newline.
        if isinstance(obj, six.text_type) or isinstance(obj, six.binary_type):
            s = obj
            is_repr = False
        else:
            s = py.io.saferepr(obj)
            is_repr = True
        if isinstance(s, six.text_type):
            t = six.text_type
        else:
            t = six.binary_type
>       s = s.replace(t("\n"), t("\n~")).replace(t("%"), t("%%"))
E       TypeError: string argument without an encoding

venv/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:438: TypeError
============================ 1 failed in 0.06 seconds ============================
```

Not a regression as far as I can tell (back to pytest 3.x at least)

asottile:
Resolves #3763

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -425,20 +425,18 @@ def _format_assertmsg(obj):
     # contains a newline it gets escaped, however if an object has a
     # .__repr__() which contains newlines it does not get escaped.
     # However in either case we want to preserve the newline.
-    if isinstance(obj, six.text_type) or isinstance(obj, six.binary_type):
-        s = obj
-        is_repr = False
-    else:
-        s = py.io.saferepr(obj)
-        is_repr = True
-    if isinstance(s, six.text_type):
-        t = six.text_type
-    else:
-        t = six.binary_type
-    s = s.replace(t("\n"), t("\n~")).replace(t("%"), t("%%"))
-    if is_repr:
-        s = s.replace(t("\\n"), t("\n~"))
-    return s
+    replaces = [(u"\n", u"\n~"), (u"%", u"%%")]
+    if not isinstance(obj, six.string_types):
+        obj = py.io.saferepr(obj)
+        replaces.append((u"\\n", u"\n~"))
+
+    if isinstance(obj, bytes):
+        replaces = [(r1.encode(), r2.encode()) for r1, r2 in replaces]
+
+    for r1, r2 in replaces:
+        obj = obj.replace(r1, r2)
+
+    return obj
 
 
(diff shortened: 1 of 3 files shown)


src/_pytest/assertion/rewrite.py
```diff
@@ -425,20 +425,18 @@ def _format_assertmsg(obj):
...
-    if is_repr:
-        s = s.replace(t("\\n"), t("\n~"))
-    return s
+    replaces = [(u"\n", u"\n~"), (u"%", u"%%")]
+    if not isinstance(obj, six.string_types):
+        obj = py.io.saferepr(obj)
+        replaces.append((u"\\n", u"\n~"))
+
+    if isinstance(obj, bytes):
+        replaces = [(r1.encode(), r2.encode()) for r1, r2 in replaces]
```
nicoddemus: perhaps we should use `errors="replace"` here for safety?

asottile: The only strings it is `.encod()`ing are the literals above.  these encode calls will never fail (barring cosmic rays)

PR #3771, opened 2018-08-02
Fix infinite recursion collection bug with pytest_ignore_collect hook

nicoddemus:
Adding an example for a bug we found at work. This blows up with:

```
.env36\lib\site-packages\pluggy\hooks.py:258: in __call__
.env36\lib\site-packages\pluggy\manager.py:67: in _hookexec
.env36\lib\site-packages\pluggy\manager.py:61: in <lambda>
src\_pytest\python.py:210: in pytest_collect_file
    return ihook.pytest_pycollect_makemodule(path=path, parent=parent)
.env36\lib\site-packages\pluggy\hooks.py:258: in __call__
.env36\lib\site-packages\pluggy\manager.py:67: in _hookexec
.env36\lib\site-packages\pluggy\manager.py:61: in <lambda>
src\_pytest\python.py:215: in pytest_pycollect_makemodule
    return Package(path, parent)
src\_pytest\python.py:552: in __init__
    self, fspath, parent=parent, config=config, session=session, nodeid=nodeid
src\_pytest\nodes.py:351: in __init__
    fspath = py.path.local(fspath)  # xxx only for test_resultlog.py?
.env36\lib\site-packages\py\_path\local.py:158: in __init__
    self.strpath = abspath(path)
w:\Miniconda\envs\py36\lib\ntpath.py:552: in abspath
    return normpath(path)
w:\Miniconda\envs\py36\lib\ntpath.py:494: in normpath
    prefix, path = splitdrive(path)
w:\Miniconda\envs\py36\lib\ntpath.py:142: in splitdrive
    if len(p) >= 2:
E   RecursionError: maximum recursion depth exceeded in comparison
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!
```

The trigger is the `pytest_ignore_collect` hook returning `False`. If it returns `None`, everything works.

This is a regression because returning `False` from the hook worked up until `3.7.0`.

cc @kfasolin

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -561,7 +561,7 @@ class Package(Module):
     def _recurse(self, path):
         ihook = self.gethookproxy(path.dirpath())
         if ihook.pytest_ignore_collect(path=path, config=self.config):
-            return
+            return False
         for pat in self._norecursepatterns:
             if path.check(fnmatch=pat):
                 return False
@@ -594,9 +594,12 @@ class Package(Module):
         return path in self.session._initialpaths
 
     def collect(self):
-        path = self.fspath.dirpath()
+        this_path = self.fspath.dirpath()
         pkg_prefix = None
-        for path in path.visit(fil=lambda x: 1, rec=self._recurse, bf=True, sort=True):
+        for path in this_path.visit(rec=self._recurse, bf=True, sort=True):
+            # we will visit our own __init__.py file, in which case we skip it
+            if path.basename == "__init__.py" and path.dirpath() == this_path:
+                continue
             if pkg_prefix and pkg_prefix in path.parts():
                 continue
             for x in self._collectfile(path):
--- a/changelog/3771.bugfix.rst
+++ b/changelog/3771.bugfix.rst
@@ -1,0 +1,1 @@
+Fix infinite recursion during collection if a ``pytest_ignore_collect`` returns ``False`` instead of ``None``.
--- a/testing/example_scripts/collect/package_infinite_recursion/conftest.py
+++ b/testing/example_scripts/collect/package_infinite_recursion/conftest.py
@@ -1,0 +1,2 @@
+def pytest_ignore_collect(path):
+    return False
--- a/testing/example_scripts/collect/package_infinite_recursion/tests/test_basic.py
+++ b/testing/example_scripts/collect/package_infinite_recursion/tests/test_basic.py
@@ -1,0 +1,2 @@
+def test():
+    pass
--- a/testing/python/collect.py
+++ b/testing/python/collect.py
@@ -1577,3 +1577,9 @@ def test_keep_duplicates(testdir):
     )
     result = testdir.runpytest("--keep-duplicates", a.strpath, a.strpath)
     result.stdout.fnmatch_lines(["*collected 2 item*"])
+
+
+def test_package_collection_infinite_recursion(testdir):
+    testdir.copy_example("collect/package_infinite_recursion")
+    result = testdir.runpytest()
+    result.stdout.fnmatch_lines("*1 passed*")


nicoddemus:
I plan to tackle it soon. If we get to the point where it would be better to merge it, then sure I will mark it as xfail and create a reference issue. 👍

nicoddemus:
I'm studying a fix for this, and I realize that the current package-scoped fixture should be collecting directories, not files. Currently it collects `__init__.py` files as `Package` objects:

https://github.com/pytest-dev/pytest/blob/a76cc8f8c4757452a35e4488a1b3264dc16b3931/src/_pytest/python.py#L213-L216

I believe this is the cause for this bug and the one reported in #3773.

I think we can overhaul `pytest_collect_directory` (#571) to properly support collecting *directories*, but this would require some refactoring in the current code.

For the bug in demonstrated in this PR, I have a fix which seems to work but I believe the long term solution is to allow proper collection of directories.

cc @jonozzz

nicoddemus:
Just pushed the workaround for this PR.

src/_pytest/python.py
```diff
@@ -561,7 +561,7 @@ def __init__(self, fspath, parent=None, config=None, session=None, nodeid=None):
     def _recurse(self, path):
         ihook = self.gethookproxy(path.dirpath())
         if ihook.pytest_ignore_collect(path=path, config=self.config):
-            return
+            return False
```
nicoddemus: This doesn't have anything to do with the fix, just thought I would keep the return value consistent.

src/_pytest/python.py
```diff
@@ -594,9 +594,12 @@ def isinitpath(self, path):
         return path in self.session._initialpaths
 
     def collect(self):
-        path = self.fspath.dirpath()
+        this_path = self.fspath.dirpath()
         pkg_prefix = None
-        for path in path.visit(fil=lambda x: 1, rec=self._recurse, bf=True, sort=True):
+        for path in this_path.visit(rec=self._recurse, bf=True, sort=True):
+            # we will visit our own __init__.py file, in which case we skip it
```
nicoddemus: This is the actual fix. I did not understand what this had to do with `pytest_ignore_collect` though.

nicoddemus:
@RonnyPfannschmidt I know you already approved it, but could you take a look at the new commits and tell what you think?

Would appreciate some input from @jonozzz as well

kfasolin:
Thanks for the fix @nicoddemus  =P

drdavella:
Was this fix released in 3.7.1? This looks very similar to, but not exactly the same as, the issue I reported in https://github.com/astropy/pytest-doctestplus/issues/22.

nicoddemus:
@drdavella not yet, this will be released in `3.7.2`. Can you try the current `master` meanwhile?

jonozzz:
@nicoddemus I didn't even know `pytest_collect_directory` existed! That would've simplified things a lot... Packages are basically directories with a special __init__.py file.

nicoddemus:
@jonozzz if you see the code currently you will notice that `pytest_collect_directory` cannot do much because its contents is discarded; we need to re-purpose it to be part of the collection. But my gut feeling is the same as yours, it would greatly simplify things.

src/_pytest/python.py
jonozzz: `visit()` calls back the `self._recurse(path)` which in turn calls the `pytest_ignore_collect`

nicoddemus:
On second thought, we probably don't want to re-purpose an existing hook, it might be better to deprecate it and add a new hook with the semantics we want to collect a directory.

nicoddemus:
Got curious (although I should be in bed by now) and decided to dig a little, it seems `pytest_collect_directory` *was* used at some point in the past to collect `Directory` nodes: 4a78daf7f3013399ffbf13d3760209b1bbb9b3ec

drdavella:
@nicoddemus yes the fix works for me on `master`, thanks.

PR #3787, opened 2018-08-06
Add option to disable plugin auto-loading

hsoft:
If `PYTEST_DISABLE_PLUGIN_AUTOLOAD` is set, disable auto-loading of
plugins through setuptools entrypoints. Only plugins that have been
explicitly specified are loaded.

ref #3784.

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -701,6 +701,10 @@ class Config(object):
 
         self.pluginmanager.rewrite_hook = hook
 
+        if os.environ.get("PYTEST_DISABLE_PLUGIN_AUTOLOAD"):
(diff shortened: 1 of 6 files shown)


src/_pytest/config/__init__.py
```diff
@@ -701,6 +701,10 @@ def _mark_plugins_for_rewrite(self, hook):
 
         self.pluginmanager.rewrite_hook = hook
 
+        if os.environ.get("PYTEST_DISABLE_PLUGIN_AUTOLOAD", ""):
```
asottile: the default value can be left out here since `get()` returns `None` by default

nicoddemus:
Invited @hsoft to join `pytest-dev`. 🎉

Issue #3789, opened 2018-08-07
Pytest suddenly cannot find module


adamhrosenberg:
Hello. I use pytest on a CI server to test our software. We have been using pytest for years but with the latest update we have an error importing a module during pytest. My logic behind isolating this issue to pytest is: on the CI server, nothing has changed (same python, same code, same OS image) the updated version of pytest is the only thing that has changed.

original versions of what pytest passed with
platform linux2 -- Python 2.7.12, pytest-3.6.3, py-1.5.4, pluggy-0.6.0

current versions where I get failures:
platform linux2 -- Python 2.7.12, pytest-3.7.1, py-1.5.4, pluggy-0.7.1

Let's say the module that cannot import is called X. When within the directory of X, (where the _UnitTest directory is) running pytest allows all of the tests to pass. Changing directory back one layer (or anywhere outside of the X directory), now those tests from module X fail because the import statement failed. 

Also on macOS I had all of the tests passing with pytest 3.6.3. Upgraded pytest, and now the tests related to module X fail. Many other modules import and work however module X does not. I have debugged a lot of it, and all of the paths look good, but it still fails to import. I'm not exactly sure what sets module X apart from the other dozens of module we use successfully. 

I hope this makes sense and I can elaborate more if needed. I am not working in a virtualenv but I can put full pip lists in the comments if more is needed other then the versions I gave above. 

Thank you

Adam

Agi-dev:
Same issue for me, if I downgrade to version <7.1 it's working.

adamhrosenberg:
@Agi-dev thank you for the feedback! glad to know i'm not alone

nicoddemus:
Hi guys, sorry for the delay.

Does `3.7.0` work for you? It might help pin point the problem. Also if any of you can produce a reproducible use case it would be great.

Agi-dev:
Sorry, I've tried to reproduce it with simple example with no success despite the fact I've found a way to make my tests work again.
I've fixed the pb (for me) by deleting a '__init__.py' in the first sub directory of my tests directory. I'm not sure it's the way to go, I'm relatively new in Python programming.

nicoddemus:
We should release `3.7.2` soon (probably next week) with a few fixes in place regarding package collection (#3768, #3526), let's see if they end up solving this issue as well.

adamhrosenberg:
Sounds great

highvelcty:
Here is a way to reproduce the problem:

Make a directory structure such as

```
[...]

Issue #3768, opened 2018-08-02
Pytest 3.7 not collecting tests on mixed Package and Modules


ZaiusDR:
Hi!

I've found an issue with pytest 3.7 while collecting tests on a mixed Python packages and modules folder structure, which worked correctly on previous versions of pytest. For instance if I setup a folder structure like this:

```
.
└── src
    ├── folder_1
    │   └── test
    │       └── test_1.py
    └── folder_2
        └── test
            └── test_2.py
```

Each `test_*.py` file containing a dummy test like:

```
def test_1():
    assert True
```

pytest collects tests correctly:

```
$ py.test --collect-only
=========================================================================================== test session starts ============================================================================================
platform darwin -- Python 2.7.10, pytest-3.7.0, py-1.5.4, pluggy-0.7.1
rootdir: /Users/eduardo/tests, inifile:
collected 2 items
<Module 'src/folder_1/test/test_1.py'>
  <Function 'test_1'>
<Module 'src/folder_2/test/test_2.py'>
  <Function 'test_2'>

======================================================================================= no tests ran in 0.05 seconds =======================================================================================
```

However if I convert one of the folders to a Python package, only the tests inside that folder are being collected:

```
.
└── src
    ├── __init__.py
    ├── folder_1
    │   ├── __init__.py
    │   └── test
    │       └── test_1.py
    └── folder_2
        └── test
            └── test_2.py
```

```
$ py.test --collect-only
=========================================================================================== test session starts ============================================================================================
platform darwin -- Python 2.7.10, pytest-3.7.0, py-1.5.4, pluggy-0.7.1
rootdir: /Users/eduardo/tests, inifile:
collected 1 item
<Package '/Users/eduardo/tests/src'>
  <Package '/Users/eduardo/tests/src/folder_1'>
    <Module 'test_1.py'>
      <Function 'test_1'>

======================================================================================= no tests ran in 0.03 seconds =======================================================================================
```

I'm not pretty sure if this is the intended behaviour or not.

---------------------------------------------------------------------
Debugging info:

`pip list` (fresh venv tested with python2.7 and 3.6 with same results)
```
$ pip list
Package        Version
-------------- -------
atomicwrites   1.1.5
attrs          18.1.0
funcsigs       1.0.2
more-itertools 4.3.0
pathlib2       2.3.2
pip            18.0
pluggy         0.7.1
py             1.5.4
pytest         3.7.0
scandir        1.7
setuptools     40.0.0
six            1.11.0
[...]

PR #3802, opened 2018-08-10
Fix test collection from packages mixed with directories. #3768 and #3789

jonozzz:
This should fix the issue but I still can't find a better fix. See the comment in the code.

What happens is: the files and directories are marked as duplicates when the Session starts collection even though they're not turned into nodes right away (the package nodes will do that later). Ideally is that when Session finds a Package it should stop recursing and move on.

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -216,18 +216,6 @@ def pytest_pycollect_makemodule(path, parent):
     return Module(path, parent)
 
 
-def pytest_ignore_collect(path, config):
-    # Skip duplicate packages.
-    keepduplicates = config.getoption("keepduplicates")
-    if keepduplicates:
-        duplicate_paths = config.pluginmanager._duplicatepaths
-        if path.basename == "__init__.py":
-            if path in duplicate_paths:
-                return True
-            else:
-                duplicate_paths.add(path)
-
-
 @hookimpl(hookwrapper=True)
 def pytest_pycollect_makeitem(collector, name, obj):
     outcome = yield
@@ -554,9 +542,7 @@ class Package(Module):
         self.name = fspath.dirname
         self.trace = session.trace
         self._norecursepatterns = session._norecursepatterns
-        for path in list(session.config.pluginmanager._duplicatepaths):
-            if path.dirname == fspath.dirname and path != fspath:
-                session.config.pluginmanager._duplicatepaths.remove(path)
+        self.fspath = fspath
 
     def _recurse(self, path):
         ihook = self.gethookproxy(path.dirpath())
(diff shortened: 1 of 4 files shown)


src/_pytest/python.py
```diff
@@ -216,18 +216,6 @@ def pytest_pycollect_makemodule(path, parent):
     return Module(path, parent)
 
 
-def pytest_ignore_collect(path, config):
```
nicoddemus: Yeah this does not really seem necessary, `main.pytest_ignore_collect` already seems to do the exactly same thing anyway.

src/_pytest/python.py
```diff
@@ -594,6 +580,15 @@ def isinitpath(self, path):
         return path in self.session._initialpaths
 
     def collect(self):
+        # XXX: HACK!
```
nicoddemus: Strange, I expect `main.pytest_ignore_collect` to handle this properly already:

https://github.com/pytest-dev/pytest/blob/4d8903fd0b89369e0e44f084e93c43d2ce26d736/src/_pytest/main.py#L261-L284

If `pytest_ignore_collect` sees the same path twice, it should ignore the collection when see the path the second time, which would not trigger a new `Package` item to be collected.

src/_pytest/python.py
jonozzz: Initially I copy/pasted the Package class from Session because they both serve very similar functions, and then realized not everything was needed.

src/_pytest/python.py
jonozzz: The problem is it sees that some modules were already "collected" (at Session level) via the recurse() method. Then when the package goes and tries to collect its children and turn them into nodes it can't because they look like duplicates. Wrong! My first reaction was to try to stop marking them as duplicates at Session level, but everything is so tangled up that I can't find a cleaner way.

testing/python/collect.py
```diff
@@ -1583,3 +1583,43 @@ def test_package_collection_infinite_recursion(testdir):
     testdir.copy_example("collect/package_infinite_recursion")
     result = testdir.runpytest()
     result.stdout.fnmatch_lines("*1 passed*")
+
+
+def test_package_with_modules(testdir):
+    """
```
nicoddemus: Just FYI, you could have created the actual files in `testing/example_scripts` and copied the whole directory here like the code above this function does.

The idea is that it is easier to execute the test using pytest as normal during testing/debugging. 😉 

This is fine as is, just thought I would mention this for the future. 👍

nicoddemus:
LGTM! Took the liberty of committing the missing changelog entries. 👍

testing/python/collect.py
```diff
@@ -1583,3 +1583,43 @@ def test_package_collection_infinite_recursion(testdir):
     testdir.copy_example("collect/package_infinite_recursion")
     result = testdir.runpytest()
     result.stdout.fnmatch_lines("*1 passed*")
+
+
+def test_package_with_modules(testdir):
```
blueyed: @jonozzz 
This test unfortunately does not fail for me before the fix, i.e. it does not test the HACK really (which I was looking into to remove/fix properly).
I've copied the test file and then run it against 4d8903f (3.7.2~10).

blueyed: Oh, sorry - confused it with testing/test_collection.py.

src/_pytest/python.py
blueyed: JFI: I've moved it in https://github.com/pytest-dev/pytest/pull/4241.

PR #3807, opened 2018-08-11
Unhide documentation for metafunc.config

### LINKED ISSUE #3746: metafunc.config is not included in API reference
metafunc.config is not included in the [API reference for metafunc](https://docs.pytest.org/en/latest/reference.html#metafunc) - presumably, the documentation comment needs to be directly above the line that sets `self.config`, but it isn't:

```python
class Metafunc(fixtures.FuncargnamesCompatAttr):
    """
    Metafunc objects are passed to the :func:`pytest_generate_tests <_pytest.hookspec.pytest_generate_tests>` hook.
    They help to inspect a test function and to generate tests according to
    test configuration or values specified in the class or module where a
    test function is defined.
    """

    def __init__(self, definition, fixtureinfo, config, cls=None, module=None):
        #: access to the :class:`_pytest.config.Config` object for the test session
        assert (
            isinstance(definition, FunctionDefinition)
            or type(definition).__name__ == "DefinitionMock"
        )
        self.definition = definition
        self.config = config
```

anowlcalledjosh:
Fixes #3746.

### DIFF
--- a/changelog/3746.doc.rst
+++ b/changelog/3746.doc.rst
@@ -1,0 +1,1 @@
+Add documentation for ``metafunc.config`` that had been mistakenly hidden.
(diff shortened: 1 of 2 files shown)


asottile:
Thanks 🎉 !

Issue #3819, opened 2018-08-16
stdout capture (capsys) does not seem to work with real-time cli logging


Sup3rGeo:
- [x] Minimal example if possible

Using the snippet from https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function, just adding a logging call before the second print:

```
import logging
import sys

logger = logging.getLogger(__name__)

def test_myoutput(capsys):  # or use "capfd" for fd-level
    print("hello")
    sys.stderr.write("world\n")
    captured = capsys.readouterr()
    assert captured.out == "hello\n"
    assert captured.err == "world\n"

    logging.info("something")

    print("next")
    captured = capsys.readouterr()
    assert captured.out == "next\n"

```

Running with `pytest --log-cli-level=INFO` makes the test fail as the `captured.out` variable comes empty.

---

- [x] `pip list` of the virtual environment you are using

```
venv-clean-py1) D:\VM\pytest-capsys-logging-bug>pip list
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
atomicwrites (1.1.5)
attrs (18.1.0)
colorama (0.3.9)
more-itertools (4.3.0)
pip (9.0.1)
pluggy (0.7.1)
py (1.5.4)
pytest (3.7.1)
setuptools (28.8.0)
six (1.11.0)
```

- [x] pytest and operating system versions

- pytest (3.7.1)
- windows 10

Sup3rGeo:
Investigating (it took a while, I find it rather complex) it seems that the problem is on this part on the logging module:

https://github.com/pytest-dev/pytest/blob/7d4c4c66d4198d1494eeaed59244d73cd2f4c8aa/src/_pytest/logging.py#L574-L591

So, it seems that logging does `suspend_global_capture()` and then `resume_global_capture()` without taking into account that `stdout` and `stderr` might have the `capsys` streams in place.

When that is the case, logging (in the functions mentioned above) will remove the current `capsys` capture but mistakenly replace it back with the global capture, instead of the `capsys` that was in place.

Deactivating the fixture capture, changing the code to this would work in theory:

```
    def emit(self, record):
        if self.capture_manager is not None:
            self.capture_manager.deactivate_fixture()
            self.capture_manager.suspend_global_capture()
        try:
            (--- etc ---)
        finally:
            if self.capture_manager is not None:
[...]

PR #3822, opened 2018-08-16
Bugfix/capsys with cli logging

Sup3rGeo:
Fixes https://github.com/pytest-dev/pytest/issues/3819

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -129,14 +129,9 @@ class CaptureManager(object):
     @contextlib.contextmanager
     def disabled(self):
         """Context manager to temporarily disables capture."""
-
         # Need to undo local capsys-et-al if exists before disabling global capture
         fixture = getattr(self._current_item, "_capture_fixture", None)
-        if fixture:
-            ctx_manager = fixture.disabled()
-        else:
-            ctx_manager = self._dummy_context_manager()
-
+        ctx_manager = fixture.suspend() if fixture else self._dummy_context_manager()
         with ctx_manager:
             self.suspend_global_capture(item=None, in_=False)
             try:
@@ -340,13 +335,19 @@ class CaptureFixture(object):
             return self._outerr
 
     @contextlib.contextmanager
-    def disabled(self):
+    def suspend(self):
         """Temporarily disables capture while inside the 'with' block."""
         self._capture.suspend_capturing()
         try:
             yield
         finally:
             self._capture.resume_capturing()
+
+    @contextlib.contextmanager
+    def disabled(self):
+        capmanager = self.request.config.pluginmanager.getplugin("capturemanager")
+        with capmanager.disabled():
+            yield
 
 
 def safe_text_dupfile(f, mode, default_encoding="UTF8"):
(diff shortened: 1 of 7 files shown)


src/_pytest/capture.py
```diff
@@ -121,6 +122,25 @@ def suspend_global_capture(self, item=None, in_=False):
                 cap.suspend_capturing(in_=in_)
             return outerr
 
+    @contextlib.contextmanager
+    def disabled(self):
+        """Temporarily disables capture while inside the 'with' block."""
+        if self._current_item is None:
```
asottile: can probably combine with the below and do:

```python
if getattr(self._current_item, '_capture_fixture', None) is None:
    yield
else:
    # use self._current_item._capture_fixture directly 
    ...
```

src/_pytest/logging.py
```diff
@@ -573,8 +573,11 @@ def set_when(self, when):
 
     def emit(self, record):
         if self.capture_manager is not None:
-            self.capture_manager.suspend_global_capture()
-        try:
+            ctx_manager = self.capture_manager.disabled()
+        else:
+            ctx_manager = _dummy_context_manager()
+
+        with ctx_manager:
```
asottile: nice, I love me some context managers!

nicoddemus:
Hi @Sup3rGeo, hold on please, I'm working on this branch with a few minor improvements which I decided to just go ahead with, if you are OK with that. 👍

nicoddemus:
Made some tweaks, let me know what you think:

* Set `self._current_item = None` at the end of the `pytest_runtest_*` calls to avoid leaking the last item in the run.
* Moved the `dummy_context_manager` to `_pytest.compat` so we can reuse it in `capture` and `logging`.
* Renamed `suspend` to `_suspend` in `CaotureFixture` as that's an internal API.
* Renamed `CaptureManager.disabled` to `CaptureManager.global_and_fixture_disabled` to better convey its meaning.

Sup3rGeo:
That's great @nicoddemus, thanks!

Just one little question about the capture - why do we have both a global and a fixture capture?

This is related to your last item - will we ever want to disable global capture without disabling also the fixture captures? As at least, as I understand, the fixture capture is also part of the whole capture plugin (and CaptureManager.disable would just mean to disable the capture plugin altogether)?

But this is really not very important, I agree with merging the way it is now :)

Sup3rGeo:
> I think it is OK to break the abstraction and be very explicit at what it does, so the caller is well aware that we are disabling global and fixture capturing with it

Thanks for the explanation! It makes sense.

So if I understood correctly the purpose of the capsys/etc fixtures is really to steal the capture even from pytest itself. Otherwise we could just have provided tests access to the global capture from a fixture I guess.

nicoddemus:
> So if I understood correctly the purpose of the capsys/etc fixtures is really to steal the capture even from pytest itself. Otherwise we could just have provided tests access to the global capture from a fixture I guess.

One of the reasons is that you can change capture methods: execute with `--capture=sys` and still use the `capfd` fixture.

nicoddemus:
Awesome, thanks @Sup3rGeo!

PR #3830, opened 2018-08-18
Fixtures during teardown can use capsys and capfd to get output from tests

### LINKED ISSUE #3033: capfd not working in 3.3.1
Fixtures that use capfd don't seem to work in version 3.3.1 running on Ubuntu 14.04 LTS with anaconda 3.6.3

```import pytest
import sys

@pytest.fixture()
def fix(capfd):
    yield None
    out, err = capfd.readouterr()
    stdout_filename = "/tmp/stdout"
    stderr_filename = "/tmp/stderr"
    mode = 'w'
    with open(stdout_filename, mode) as stdout:
        stdout.write(out)
    with open(stderr_filename, mode) as stderr:
        stderr.write(err)

def test_a(fix):
    print("lolcatz")
    print("this is a test")
    sys.stderr.write("err")
```

The stdout and stderr files end up empty after running that test. Using this set of pytest dependencies
```attrs (17.3.0)
certifi (2017.11.5)
pip (9.0.1)
pluggy (0.6.0)
py (1.5.2)
pytest (3.3.1)
setuptools (38.2.4)
six (1.11.0)
wheel (0.30.0)
```

Steps of replication are
1. Create the above test
2. Run pytest on it.
3. cat /tmp/std* (they will be empty)
4. Downgrade pytest ( pip install autohil==3.2.5 )
5. Run the test again
6. cat /tmp/std*

nicoddemus:
The problem is that the fixtures didn't live up to the promise to properly accumulate output between setup/call/teardown. This works in `3.2` because we didn't allow to access the fixture during setup/teardown from the first place, so it didn't need to accumulate any buffer.

Follow up to #3066, fixes #3033.

### DIFF
--- a/changelog/3033.bugfix.rst
+++ b/changelog/3033.bugfix.rst
@@ -1,0 +1,1 @@
+Fixtures during teardown can again use ``capsys`` and ``cafd`` to inspect output captured during tests.
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -188,7 +188,6 @@ class CaptureManager(object):
     def pytest_runtest_teardown(self, item):
         self._current_item = item
         self.resume_global_capture()
-        self.activate_fixture(item)
         yield
         self.suspend_capture_item(item, "teardown")
         self._current_item = None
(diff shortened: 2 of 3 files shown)


nicoddemus:
cc @trrichard

PR #3832, opened 2018-08-19
Bugfix/capsys with cli logging (again)

Sup3rGeo:
Solves more problems found related to #3819, improving previous PR #3822 also by refactoring a bit the capture module, mostly by separating the snapping of the global capture from its suspension.

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -62,8 +62,9 @@ def pytest_load_initial_conftests(early_config, parser, args):
     # finally trigger conftest loading but while capturing (issue93)
     capman.start_global_capturing()
     outcome = yield
-    out, err = capman.suspend_global_capture()
+    capman.suspend_global_capture()
     if outcome.excinfo is not None:
+        out, err = capman.snap_global_capture()
         sys.stdout.write(out)
         sys.stderr.write(err)
 
@@ -96,6 +97,8 @@ class CaptureManager(object):
         else:
             raise ValueError("unknown capturing method: %r" % method)
 
+    # Global capturing control
+
     def start_global_capturing(self):
         assert self._global_capturing is None
         self._global_capturing = self._getcapture(self._method)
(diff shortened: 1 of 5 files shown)


nicoddemus:
This is great @Sup3rGeo, thanks a lot.

I see the main problem is that currently we drop the captured output between live log calls because we suspend the capture manager when emitting log records, and this PR fixes that.

Also great work on the refactoring, it reduced the code and improved the readability! 👍

src/_pytest/capture.py
```diff
@@ -62,8 +62,9 @@ def silence_logging_at_shutdown():
     # finally trigger conftest loading but while capturing (issue93)
     capman.start_global_capturing()
     outcome = yield
-    out, err = capman.suspend_global_capture()
+    capman.suspend_global_capture()
     if outcome.excinfo is not None:
+        out, err = capman.snap_global_capture()
```
nicoddemus: Since we are refactoring things, how do you feel about renaming this to `read_global_capture`?

testing/test_capture.py
```diff
@@ -1387,27 +1391,53 @@ def test_pickling_and_unpickling_encoded_file():
     pickle.loads(ef_as_str)
 
 
-def test_capsys_with_cli_logging(testdir):
+def test_capture_with_live_logging(testdir):
```
nicoddemus: Please parametrize this test so it also tests with the `capfd` fixture. 👍

src/_pytest/capture.py
nicoddemus: It is just that `snap_global_capture` does `self._global_capturing.readouterr()`, so I think renaming it here makes sense. Snap is then used just for the "Capture" classes. 😁

Sup3rGeo: Makes sense, will do it

Sup3rGeo:
It seems like there is still a change to be addressed but I don't know how to do it

nicoddemus:
Thanks @Sup3rGeo, top notch work!

> It seems like there is still a change to be addressed but I don't know how to do it

You mean the merge was blocked? That was because we configure the repository to block merges until no reviewers left the PR in the "requested changes" status. 😉 

If nobody else comments until then, I will merge this later. 👍

asottile:
Not sure how important this is, but this appears to have broken `tox`'s test suite (between 3,7,2 and 3.7.3).  To be fair though, tox was using `capfd._start()` directly which is probably a no-no.

https://github.com/tox-dev/tox/pull/949/commits/6c7ea274093807fb93f36a131f63b343782fb903 I believe fixes what I'm seeing

PR #3858, opened 2018-08-22
Use unittest.mock if is only aviable

mimi1vx:
from Python 3.3 is mock part of python standard library in unittest namespace

- [x ] Target the `master` branch for bug fixes, documentation updates and trivial changes.
- [x ] Add yourself to `AUTHORS`   -- already in

### DIFF
--- a/testing/example_scripts/acceptance/fixture_mock_integration.py
+++ b/testing/example_scripts/acceptance/fixture_mock_integration.py
@@ -1,6 +1,9 @@
 """Reproduces issue #3774"""
 
-import mock
+try:
+    import mock
+except ImportError:
+    import unittest.mock as mock
 
 import pytest
 


wimglenn:
Preferably other way around, no?  (try stdlib import first, else fallback to mock from pypi)

nicoddemus:
TBH I'm not sure why this is needed, after all the recommended way to run the tests using `tox`, which install the appropriate dependencies.

asottile:
Here's the context for that comment btw: https://github.com/pytest-dev/pytest/pull/3612#discussion_r197622510

(also I've seen `from unittest import mock` more commonly than `import unittest.mock as mock` -- they do the same thing so nbd)

nicoddemus:
> @nicoddemus because mock package isn't needed on modern python, so some distributions stop providing it for python3.

Ah OK, that makes sense, thanks for the context. 👍

PR #3859, opened 2018-08-23
Some pyupgrade 1.4.x changes

asottile:
I'm doing these slowly because I found one kind of unfortunate thing that I didn't notice in my code (because I only use `from __future__ import unicode_literals`, but `pytest` does not 😭)

The one python2-only quirk is this:

```pycon
>>> '%s' % u'\u2603'
u'\u2603'
>>> '{}'.format(u'\u2603')
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2603' in position 0: ordinal not in range(128)
```

In this PR I've formatted the tests, and one file which was exposed by tests (more to come but I didn't want to get burned out on this branch and just abandon it).


asottile:
Sweet, expect some more gradual PRs similar to this one -- on the plus side this also improves the unicode support in python2 :)

PR #3868, opened 2018-08-24
Use `bytes` directly instead of `binary_type`

asottile:
`bytes` is an alias for `str` in python2.6+

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -402,12 +402,11 @@ def _saferepr(obj):
     JSON reprs.
 
     """
-    repr = py.io.saferepr(obj)
-    if isinstance(repr, six.text_type):
-        t = six.text_type
+    r = py.io.saferepr(obj)
+    if isinstance(r, six.text_type):
+        return r.replace(u"\n", u"\\n")
     else:
-        t = six.binary_type
-    return repr.replace(t("\n"), t("\\n"))
+        return r.replace(b"\n", b"\\n")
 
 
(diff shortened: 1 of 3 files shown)


src/_pytest/assertion/rewrite.py
```diff
@@ -402,12 +402,11 @@ def _saferepr(obj):
...
-    repr = py.io.saferepr(obj)
-    if isinstance(repr, six.text_type):
-        t = six.text_type
+    r = py.io.saferepr(obj)
+    if isinstance(r, six.text_type):
+        return r.replace(u"\n", u"\\n")
     else:
-        t = six.binary_type
-    return repr.replace(t("\n"), t("\\n"))
+        return r.replace(b"\n", b"\\n")
```
asottile: Note that `bytes("string literal")` in python3+ is a `TypeError` so I fixed these as well.

As far as I can tell however this situation is impossible as `__repr__` raises for non `str` types in python3.x

src/_pytest/assertion/rewrite.py
```diff
@@ -446,10 +445,9 @@ def _should_repr_global_name(obj):
 def _format_boolop(explanations, is_or):
     explanation = "(" + (is_or and " or " or " and ").join(explanations) + ")"
     if isinstance(explanation, six.text_type):
-        t = six.text_type
+        return explanation.replace(u"%", u"%%")
     else:
-        t = six.binary_type
-    return explanation.replace(t("%"), t("%%"))
+        return explanation.replace(b"%", b"%%")
```
asottile: this one however seems maybe possible? I'm not sure how this code gets called 🤷‍♂️

Issue #3870, opened 2018-08-24
Arguments not passed from setuptools to pytest (incoming PR)


schmamps:
## Description

The documentation is incorrect.

## System Info

```sh
$ pip list
Package           Version
----------------- -------
astroid           2.0.2
atomicwrites      1.1.5
attrs             18.1.0
beautifulsoup4    4.6.1
colorama          0.3.9
docx2txt          0.8
flake8            3.5.0
flake8-quotes     1.0.0
isort             4.3.4
kinda             0.9.9
lazy-object-proxy 1.3.1
mccabe            0.6.1
more-itertools    4.3.0
mypy              0.620
nltk              3.3
pathlib           1.0.1
pip               18.0
pluggy            0.7.1
py                1.5.4
pycodestyle       2.3.1
pyflakes          1.6.0
pylint            2.1.1
PyPDF2            1.26.0
pytest            3.7.1
python-magic      0.4.15
rope              0.10.7
setuptools        40.0.0
six               1.11.0
typed-ast         1.1.0
typing            3.6.4
wheel             0.31.1
wrapt             1.10.11

$ uname -a
Darwin computer-name 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64
```

## Example

After following the documentation:

```sh
$ python setup.py test -a -vv
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option -a not recognized
```

PR #3871, opened 2018-08-24
Correct cmdclass for doc: Good Integration Practices

schmamps:
Addressing #3870 

If you follow the guide to this point, the correct key for `cmdclass` will be the canonical name 'pytest' rather than the alias 'test' created in setup.cfg.

### DIFF
--- a/doc/en/goodpractices.rst
+++ b/doc/en/goodpractices.rst
@@ -291,7 +291,7 @@ your own setuptools Test command for invoking pytest.
     setup(
         # ...,
         tests_require=["pytest"],
-        cmdclass={"test": PyTest},
+        cmdclass={"pytest": PyTest},
     )
 
 Now if you run::
--- a/changelog/3870.doc.rst
+++ b/changelog/3870.doc.rst
@@ -1,0 +1,1 @@
+correct documentation for setuptools integration


doc/en/goodpractices.rst
```diff
@@ -291,7 +291,7 @@ your own setuptools Test command for invoking pytest.
     setup(
         # ...,
         tests_require=["pytest"],
-        cmdclass={"test": PyTest},
+        cmdclass={"pytest": PyTest},
```
nicoddemus: Isn't this string (`"test"`/`"pytest"`) the name of the command executed in the next section?

```
python setup.py test
```

schmamps: Earlier in the document, `'test'` is created as an alias to the `'pytest'`:
```ini
[aliases]
test=pytest
```

The command doesn't work unless you remove this from `setup.cfg` or replace that alias with the canonical name in `setup.py`.

nicoddemus: Ahh right, thanks for the clarification! 👍

nicoddemus:
Btw thanks on your first contribution! 🎉

schmamps:
Thanks for responding to a PR!

PR #3874, opened 2018-08-25
Travis: use TOXENV=linting for linting stage

blueyed:
This will run it with `--show-diff-on-failure` then, and helps to keep
it in line / in a central place.

See https://travis-ci.org/pytest-dev/pytest/jobs/420595379 for an example failure.

Appveyor uses TOXENV=linting already.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,12 +62,7 @@ jobs:
           repo: pytest-dev/pytest
     - stage: linting
       python: '3.6'
-      env:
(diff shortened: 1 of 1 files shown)


asottile:
oh actually, maybe add `skip_install` / `skip_sdist` / whatever the option I'm forgetting to the `linting` testenv? It seems to still be installing `pytest` which isn't necessary

PR #3875, opened 2018-08-25
Also record coverage for tests

asottile:
As seen in #3873, there's sneaky dead code lurking in `./testing`.  Let's surface that a little bit better.

I started combing through the results but didn't get terribly far.


blueyed:
I've picked up https://github.com/pytest-dev/pytest/pull/2800 again.
I would say to wait with this here until it is in.

I think we should also add branch=1 separately, and I wanted to do it after #2800 to have better reporting for what it changes.

.coveragerc
```diff
@@ -1,4 +1,26 @@
 [run]
+branch = True
+source = .
```
blueyed: Better use `_pytest,testing` here - less to omit then.

.coveragerc
```diff
@@ -1,4 +1,26 @@
...
+    ^\s*raise AssertionError\b
+    ^\s*raise NotImplementedError\b
+    ^\s*return NotImplemented\b
+    ^\s*raise$
+
+    # Don't complain if non-runnable code isn't run:
+    ^if __name__ == ['"]__main__['"]:$
+
+    # Intentionally skipped tests
+    ^\s*@pytest\.mark\.skipif
```
blueyed: `@pytest\.mark\.skipif` should not be excluded: one of the builds should hit it.
(more relevant with/after #2800)

src/_pytest/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 try:
     from ._version import version as __version__
-except ImportError:
+except ImportError:  # pragma: no cover
     # broken installation, we don't even try
     # unknown only works because we do poor mans version compare
     __version__ = "unknown"
```
blueyed: Better leave this code as uncovered.
We're not trying to hit 100%, but is is OK for this to be not covered - also for other things below.

src/_pytest/_argcomplete.py
```diff
@@ -88,7 +88,7 @@ def __call__(self, prefix, **kwargs):
         return completion
 
 
-if os.environ.get("_ARGCOMPLETE"):
+if os.environ.get("_ARGCOMPLETE"):  # pragma: no cover
```
blueyed: This one e.g. would be good to get covered in the long run.

.coveragerc
asottile: hmm but then you have to have a special snowflake `.coveragerc` for each project

.coveragerc
asottile: I don't see a reason to block on #2800 -- until then this improves local runs of isolated environments.

src/_pytest/__init__.py
asottile: > We're not trying to hit 100%

We're not? 😆

asottile: Confirmation that, in fact, _we are_:

https://github.com/pytest-dev/pytest/pull/2800#pullrequestreview-65593151

> Ultimately we should strive to obtain 100% coverage...

.coveragerc
blueyed: ?

I think it is OK for a project to have its own sources defined here - after all you would then have special omits instead.

.coveragerc
blueyed: Yes, it is not blocked by #2800, but with #2800 you would not have added it in the first place, would you? (because it should be covered then, and if not it would mean no test setup is hitting it).

src/_pytest/__init__.py
blueyed: /cc @nicoddemus 

I think the crucial part from there is:

> Even if we "cheat", we gain the benefit of demanding that new code from now on has 100% coverage.

This should be achieved instead by requiring a diff to be covered 100% - something which codecov provides then.

.coveragerc
asottile: 🤷‍♂️ no special `omit`s here -- I copied this from one of my projects

blueyed: Still different than "defaults", no? :)

asottile:
I've rebased and set `source = _pytest,testing`

asottile:
@blueyed -- I'll defer to your review -- let me know if/when this is mergeable!

blueyed:
@asottile 
I've meant to _only_ change `source`, nothing more.
This would then show which tests are not covered.
The next step would then be adding `branch = 1` (maybe also in this PR), but nothing more.

asottile:
what's the reason?

blueyed:
@asottile 
I would rather keep it simpler - but that's just my opinion.

asottile:
I'm sorry, can you highlight what portions of the PR you're in opposition to and your rationale for opposing them? I'm not sure I'm understanding :(

nicoddemus:
I gather that now that #3920 has been merged we can pick this up again?

PR #3877, opened 2018-08-25
tox: coveralls: also report to codecov

blueyed:
This is meant to get base coverage on master for codecov.

Superseded then by #2800, but allows for seeing how #2800 improves things and that nothing regresses, e.g. because of usedevelop not being used there etc.


nicoddemus:
Should we be seeing a report from codecov by now at this point?

nicoddemus:
@blueyed I believe you can login in the `pytestbot` account? It lists you as a collaborator already.

blueyed:
@nicoddemus 
Logged into it explicitly via `/login`, and now only see pytest-dev related projects, but still cannot cancel builds.

nicoddemus:
> There is no "switch account" feature, is there? - so this is a bit of a hassle anyway.

I usually sign out from my account and login as `pytestbot`, and then I can cancel builds...

Issue #3829, opened 2018-08-18
Option to use counter instead of percentage for progress


DanaGoyette:
With pytest (I'm currently using 3.6.1), there's support for showing progress of tests, but when running a collection of 1300 tests, each test contributes very little to the percentage.  I'd prefer to be able to show the bare counter.   Example (with space padding): `test_whatever   [    1/1357]`

In my framework, I'm monkey-patching pytest to get what I want, but it would be more maintainable if pytest supported that style officially.

My current monkey-patch:
```python
    def _get_progress(self):
        """
        Report progress in number of tests, not percentage.
        Since we have thousands of tests, 1% is still several tests.
        """
        collected = self._session.testscollected
        self._PROGRESS_LENGTH = len(' [%d/%d]' % (collected, collected))
        if collected:
            spacing = len(str(collected))
            current = len(self._progress_nodeids_reported)
            fmt = ' [{:%sd}/{}]' % spacing
            return fmt.format(current, collected)
        return ' []'

    TerminalReporter._PROGRESS_LENGTH = len(' [9999/9999]')
    TerminalReporter._get_progress_information_message = _get_progress
```

nicoddemus:
@DanaGoyette I thought about this while developing the feature and imagined using percentages would be more useful, but I of course see the case for a bare counter. It would be great if you would like to contribute a patch with that.

I suspect an ini option would be appropriate: `progress_display_mode=percentage/count`.

PR #3880, opened 2018-08-26
#3829 -- Add the ability to show test progress as number of tests completed instead of a percent.

jeffreyrack:
Implements #3829 by adding a config value that can be used to show the test progress as the number of completed tests instead of a percentage complete.

Wasn't sure where to add the documentation for this, but am willing to go and add it if somebody can suggest a good spot to mention the new config value in.


nicoddemus:
Thanks for tackling this @jeffreyrack!

The counter text is going over the edge when the number of tests make the progress doesn't fit on a single line, at least for me on Windows:

```
...
collected 2131 items

testing\acceptance_test.py ...............s...................[   35 / 2131 ]
............s.x................                       [   66 / 2131 ]
testing\deprecated_test.py ......F.s.....             [   80 / 2131 ]
testing\test_argcomplete.py ss                        [   82 / 2131 ]
```

How does it look on your system?

Also with `-v` the counter text is right next to the status:

```
testing/acceptance_test.py::TestGeneralUsage::test_config_error PASSED[    1 / 2131 ]
```

I think there should be a space between them:

```
testing/acceptance_test.py::TestGeneralUsage::test_config_error PASSED [    1 / 2131 ]
```

Finally, a little of bike shedding: I think we don't need the space between brackets and the numbers: `[  18/2131]` instead of `[   18 / 2131 ]`. It occupies less space and is more pleasant to the eyes (the latter my opinion of course).

nicoddemus:
About where to put the docs: we should put the new option in `reference.rst`, at least.

src/_pytest/terminal.py
```diff
@@ -143,6 +143,12 @@ def pytest_addoption(parser):
         default="progress",
     )
 
+    parser.addini(
+        "progress_display_mode",
```
nicoddemus: Instead of a new option, how about adding a new `count` option to `console_output_style`?

PR #3895, opened 2018-08-28
Avoid possible infinite recursion when writing pyc files in assert rewrite

### LINKED ISSUE #3506: RuntimeError: maximum recursion depth exceeded .... on Windows and Python 2.7
I bumped pytest from 3.5.1 to 3.6.0 ... and the tests fail to run on Windows with a pytest error:

`RuntimeError: maximum recursion depth exceeded`

If I uninstall 3.6.0 and reinstall 3.5.1 things are back to normal.

See https://ci.appveyor.com/project/nexB/license-expression/build/248/job/smx1qy33w1dxrlh5#L531
and https://ci.appveyor.com/project/nexB/license-expression/build/248/job/jpd8c35tgee9mi7y#L532

Things work fine on POSIX and other Python versions on Windows:
https://travis-ci.org/nexB/license-expression/builds/383421178
https://ci.appveyor.com/project/nexB/license-expression/build/248
 
The exact versions and wheel of all the thirdparty packages used are here:
https://github.com/nexB/license-expression/tree/fix-symbol-comparison/thirdparty/dev
https://github.com/nexB/license-expression/tree/fix-symbol-comparison/thirdparty/base

nicoddemus:
What happens is that atomic_write on Python 2.7 on Windows will try
to convert the paths to unicode, but this triggers the import of
the encoding module for the file system codec, which in turn triggers
the rewrite, which in turn again tries to import the module, and so on.

This short-circuits the cases where we try to import another file when
writing a pyc file; I don't expect this to affect anything because
the only modules that could be affected are those imported by
atomic_writes.

The fix has been tested in an actual repository showing the problem: https://github.com/Pylons/pyramid-cookiecutter-starter/pull/56

Fix #3506

### DIFF
--- a/changelog/3506.bugfix.rst
+++ b/changelog/3506.bugfix.rst
@@ -1,0 +1,1 @@
+Fix possible infinite recursion when writing ``.pyc`` files.
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -64,11 +64,16 @@ class AssertionRewritingHook(object):
         self._rewritten_names = set()
         self._register_with_pkg_resources()
         self._must_rewrite = set()
+        # flag to guard against trying to rewrite a pyc file while we are already writing another pyc file,
+        # which might result in infinite recursion (#3506)
+        self._writing_pyc = False
 
     def set_session(self, session):
         self.session = session
 
     def find_module(self, name, path=None):
+        if self._writing_pyc:
+            return None
         state = self.config._assertstate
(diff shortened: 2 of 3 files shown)


stevepiercy:
It looks like [GitHub was unavailable during some of this test run](https://ci.appveyor.com/project/pytestbot/pytest/build/1.0.5001). Would a maintainer please restart the job?

nicoddemus:
@stevepiercy thanks for the ping, restarted it!

nicoddemus:
All good now. 👍

PR #3902, opened 2018-08-28
Fix pytest.org links

stevepiercy:
I corrected various versions of links to `pytest.org` for consistency, to use https, fix 404s, and save redirects.

All URLs should now begin with `https://docs.pytest.org/en/latest/`.

Please note that the project description on GitHub uses the URL `http://pytest.org/`, but I can't fix that one.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,23 +66,23 @@ To execute it::
     ========================== 1 failed in 0.04 seconds ===========================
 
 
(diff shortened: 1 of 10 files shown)


stevepiercy:
Would someone please poke Appveyor to retry the build? It failed due to GitHub not being available.

PR #3903, opened 2018-08-28
Parametrizing scopes with autouse

### LINKED ISSUE #1552: Suggestion: add new 'parametrization' fixture scope
Currently, a fixture can be scoped as function, class, module, or session.

I propose a fixture can also be scoped per *parametrization*.

Consider:

```
import time
@pytest.fixture
def slow_setup():
    time.sleep(1)
    return 'foo'
```

If I have a *group of tests* which I am parametrizing, but that would share this value, it'll do the setup for each one (as expected):

```
# takes 4 seconds
@pytest.mark.parametrize('bar', [1, 2, 3, 4])
def test_parametrized(slow_setup, bar):
    assert slow_setup == 'foo'
```

If I wanted to run the same group of tests but sharing the fixture, I only really have one option (considering that other tests in the same module need the fixture to be re-set up each time, so I can't use module or session scope): set the fixture scope to class, and put the tests inside a class:

```
@pytest.fixture(scope='class')
def slow_setup_cls():
    time.sleep(1)
    return 'foo'

# takes 1 second
class TestParametrized(object):
    @pytest.mark.parametrize('bar', [1, 2, 3, 4])
    def test_parametrized(self, slow_setup_cls, bar):
        assert slow_setup_cls == 'foo'
```

However, the meaning I really intend to convey is "this fixture should be shared across all parametrizations of a function". I propose the following:

```
@pytest.fixture(scope='parametrization')
def slow_setup_pmt():
    time.sleep(1)
    return 'foo'

# takes 1 second
@pytest.mark.parametrize('bar', [1, 2, 3, 4])
def test_parametrized(self, slow_setup_pmt, bar):
    assert slow_setup_pmt == 'foo'
```

This fixture scope conveys, to me: "All parametrizations of this test share the same setup and teardown with regards to this fixture."

SalmonMode:
Closes #1552 

Now, instead of only re-activating fixtures for a given scope if they occur after a parametrized fixture for the same scope that has ``autouse`` as ``True``, all fixtures of that scope will be re-activated.

For example, currently, given the following module:

```python
import pytest


def before_param():
    print("in before_param")


class TestIt():

    @pytest.fixture(scope="class", autouse=True, params=["a", "b"])
    def param_fix(self, before_param, request):
        print("in param_fix")
        return request.param

    def other_fix(self, param_fix):
        print("in other_fix")

    def test_stuff(self, other_fix):
        assert True
```

You used to get the following flow:

```
test_module (module)
└── TestIt (class)
    └── before_param
        ├── param_fix["a"]
        |   └── other_fix
        |       └── test_stuff (function)
        └── param_fix["b"]
            └── other_fix
                └── test_stuff (function)
```

But now, you get this flow:

```
test_module (module)
└── TestIt (class)
    ├── before_param
    |   └── param_fix["a"]
    |       └── other_fix
    |           └── test_stuff (function)
    └── before_param
        └── param_fix["b"]
            └── other_fix
                └── test_stuff (function)
```


SalmonMode:
Woops! Didn't realize I branched from the wrong one. I'll close this and remake it once I can get to my computer

PR #3906, opened 2018-08-29
Preparing release version 3.7.4

asottile:


nicoddemus:
Thanks @asottile!

asottile:
@nicoddemus the [steps](https://github.com/pytest-dev/pytest/blob/master/HOWTORELEASE.rst) don't indicate when I should merge this PR / when I should create the tag (I'll fix them!) -- at what point should those things happen?

nicoddemus:
Ahh right.

After all tests have passed, push the tag to `upstream` with the PR still open. After the Travis job finishes and the deploy is complete, we merge the PR and make the announcements. 

We can update the HOWTORELEASE after `3.7.4` is out to avoid having to wait for the CI all over again.

asottile:
also I think I need to run `git tag 3.7.4` myself as well? I can add that to the steps as well

nicoddemus:
> also I think I need to run git tag 3.7.4 myself as well? I can add that to the steps as well

Yes, that is not very explicit there. Thanks!

asottile:
@nicoddemus lol, do you wait for appveyor?

nicoddemus:
Oh this is running already... just need to wait a bit then.

wimglenn:
appveyor

asottile:
I once had an appveyor build finish before travis-ci -- I restarted it since I assumed something was wrong

blueyed:
Maybe we should trim down the matrix for AV?

asottile:
could also crowdfund, their paid plan has better parallelism iirc

RonnyPfannschmidt:
hmm, neo would disapprove of a low res matrix

nicoddemus:
Just realized that we can configure Travis to skip the stages and go directly to deploy when it is building a tag... what do you guys think about that?

nicoddemus:
Awesome, thanks @asottile!

asottile:
> Your message has been rejected, probably because you are not
> subscribed to the mailing list and the list's policy is to prohibit
> non-members from posting to it.

awkward, I'll subscribe and send that out in a bit!

nicoddemus:
hehehe. Don't forget the tweet as well! 👍

blueyed:
Thanks, @asottile !

> Just realized that we can configure Travis to skip the stages and go directly to deploy when it is building a tag... what do you guys think about that? 

:+1: 

Just did this for pytest-django: https://github.com/pytest-dev/pytest-django/commit/b1f1b26

I think it should run one test job before though.

nicoddemus:
it is easier to skip an entire stage, how about skipping the `test` staging and only run `baseline` and `deploy` when in a tag?

blueyed:
Sounds good!

nicoddemus:
OK, cool: #3908. 👍

PR #3909, opened 2018-08-29
Update release procedure

asottile:


nicoddemus:
>  just tag the release via GitHub (on master), which also triggers a build then already.

This is fine, but I would prefer to tag exactly the same commit that was approved on the PR. If this is possible to do via GitHub's UI, then great!

blueyed:
> This is fine, but I would prefer to tag exactly the same commit that was approved on the PR. If this is possible to do via GitHub's UI, then great!

Sure.. you can specify a commit, and/or browse them via "recent commits" on https://github.com/pytest-dev/pytest/releases/new.

Maybe we should update the document according to this instead?
Or wait for the next iteration/release?

blueyed:
Might be good to have some script to fill in the release notes then also (copy'n'paste via some towncrier output/command).

nicoddemus:
> Might be good to have some script to fill in the release notes then also (copy'n'paste via some towncrier output/command).

The release notes are generated from `tox -e release` 😉

PR #3910, opened 2018-08-29
doc fix: raises accepts tuples instead of lists

### LINKED ISSUE #3907: xfail: mismatch between docs and the actual behaviour
I know this is kinda nitpicky, but for the sake of completeness:

The docs [state](https://github.com/pytest-dev/pytest/blob/master/doc/en/skipping.rst#raises-parameter):

> If you want to be more specific as to why the test is failing, you can specify a single exception, or a **list of exceptions**, in the ``raises`` argument.

However, the actual implementation of the `xfail` marker expects a tuple instead of a list, raising an internal error otherwise. May I suggest to replace ``list`` with ``tuple`` in the above sentence?

hoefling:
Fixes #3907.

### DIFF
--- a/changelog/3907.doc.rst
+++ b/changelog/3907.doc.rst
@@ -1,0 +1,1 @@
+Corrected type of the exceptions collection passed to ``xfail``: ``raises`` argument accepts a ``tuple`` instead of ``list``.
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks!

PR #3911, opened 2018-08-30
improve line width estimate

wimglenn:
as discussed in https://github.com/pytest-dev/py/pull/197

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ def get_environment_marker_support_level():
 def main():
     extras_require = {}
     install_requires = [
-        "py>=1.5.0",
+        "py>=1.5.0",  # if py gets upgrade to >=1.6, remove _width_of_current_line in terminal.py
         "six>=1.10.0",
         "setuptools",
         "attrs>=17.4.0",
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ def get_environment_marker_support_level():
 def main():
     extras_require = {}
     install_requires = [
-        "py>=1.5.0",
+        "py>=1.6.0",
         "six>=1.10.0",
         "setuptools",
         "attrs>=17.4.0",
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ def get_environment_marker_support_level():
 def main():
     extras_require = {}
     install_requires = [
-        "py>=1.6.0",
+        "py>=1.5.0",
         "six>=1.10.0",
         "setuptools",
         "attrs>=17.4.0",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -59,7 +59,7 @@ def get_environment_marker_support_level():
 def main():
     extras_require = {}
     install_requires = [
-        "py>=1.5.0",
+        "py>=1.6.0",
```
nicoddemus: We learned the hard way before to not change dependency requirements in bug-fix releases. 😬 

Any chance of staying compatible with `py-1.5`, but using the new functionality from `py-1.6` when that's installed?

nicoddemus: If not possible/desired, we should target `features` instead.

nicoddemus:
Thanks a lot @wimglenn! Please see my comments and let me know what you think.

wimglenn:
@nicoddemus   Oh, yep, good idea (also has the advantage to simply bump down py version if this change causes teething problems on some platforms).  Addressed in https://github.com/pytest-dev/pytest/pull/3911/commits/c18a5b5179f724e7046742bae60da51dc74469d4

nicoddemus:
Thanks @wimglenn! Took the liberty of moving the code to a function and adding a note to `setup.py` to remove that function once we require `py>=1.6` (probably we should do so in `features` after this gets merged).

wimglenn:
@nicoddemus  is that CI fail due the refactor? or is the codecov diff a new addition?

blueyed:
> or is the codecov diff a new addition?

This.
It seems the diff is not covered fully - but unfortunately the Details view does not load for me currently (https://codecov.io/gh/pytest-dev/pytest/compare/9bd4b0a05ee18e80e4835cd8d42cef1806b9f0f2...19fa01b91dc5289c82cc2c7dfd971eebc39cf28a).

blueyed:
Maybe it is the code in setup.py, and we might want to either cover it, or ignore it if it is not possible easily.

blueyed:
Processing seems to hang on codecov, but the latest commit shows that it is due to not covering all py variants apparently: https://codecov.io/gh/pytest-dev/pytest/commit/c18a5b5179f724e7046742bae60da51dc74469d4

wimglenn:
Ugh.  It will be expensive to test both those branches, because it requires both an old and a new version of `py` installed in test matrix (...or some pretty intrusive mocking)

blueyed:
@wimglenn 
I think we can ignore this - it can be merged still/nonetheless.

nicoddemus:
Thanks again @wimglenn for great work on the PR!

Thanks @blueyed for chipping in as well.

Issue #3853, opened 2018-08-22
Needless message printed with --failed-first and no failed tests


nedbat:
``--failed-first`` always means, run all the tests, but order the recent failure to run first.  But if there were no failures, it always prints a message that it is going to run all the tests. This message is not needed.
```
$ cat test_it.py
def test_it():
    assert 1 == 1

$ pytest
================== test session starts ===================
platform darwin -- Python 2.7.14, pytest-3.7.1, py-1.5.4, pluggy-0.7.1
rootdir: /private/tmp/failedfirst, inifile:
plugins: xdist-1.22.5, forked-0.2
collected 1 item

test_it.py .                                                                                                                                                               [100%]

================ 1 passed in 0.01 seconds ================

$ pytest --failed-first
================== test session starts ===================
platform darwin -- Python 2.7.14, pytest-3.7.1, py-1.5.4, pluggy-0.7.1
rootdir: /private/tmp/failedfirst, inifile:
plugins: xdist-1.22.5, forked-0.2
collected 1 item
run-last-failure: run all (no recorded failures)

test_it.py .                                                                                                                                                               [100%]

================ 1 passed in 0.01 seconds ================

$ pytest -q --failed-first
run-last-failure: run all (no recorded failures)
.                                                                                                                                                                          [100%]
1 passed in 0.01 seconds

```

nicoddemus:
Hi @nedbat,

I think the idea is to make sure the plugin is actually working... I personally like to see the message to make sure there are actually no failures and now all tests are running.

nedbat:
Yes :)

nicoddemus:
#3886 deals with handling `-q` properly, after that gets merged we can them suppress the output when there are no recorded failures on the `features` branch.

dhirensr:
@nedbat @nicoddemus  : I have read the complete issue and also #3886 ! does this ticket say that when --lf ,then the output now is "run-last-failure: run all (no recorded failures)" but instead it should behave exactly as --q and nothing should display?

nicoddemus:
@dhirensr exactly! 😁

dhirensr:
@nicoddemus : where do we change the code like do we send mode as empty string or change the print function ? could you also guide me where is the print ignored in case of --quiet ,I am not able to find that piece of code which ignores the print statement and prints nothing!

dhirensr:
[...]

PR #3912, opened 2018-08-30
Needless message printed with --failed-first and no failed tests #3853

dhirensr:


changelog/3853.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Removed ``"run all (no recorded failures)"`` message printed with ``--failed-first`` and ``--last-failed`` when there are no failed tests.
```
nicoddemus: @dhirensr I took the liberty of updating the CHANGELOG a bit: we don't need to explicitly write about internal details like which tests were updated for example; those are not interesting for final users, which are the public the CHANGELOG is meant for. 😁

PR #3920, opened 2018-08-31
tests/CI: enable branch coverage

blueyed:


nicoddemus:
Not sure why the coverage has dropped...?

Also I noticed that `codecov` reported its findings 3 hours after Travis finished, but `coveralls` was much faster, taking only a few minutes. Is this common or a fluke?

nicoddemus:
> They had DB issues yesterdays, should be faster typically.

Oh OK, thanks.

> Codecov appears to still be swamped.

I did notice in another PR today that the comment from `codecov` was posted right after the build finished, so it seems they are back to normal. 👍 

I get we can merge this as is then?

blueyed:
Yes!

PR #3956, opened 2018-09-07
Move warnings summary above short test summary.

### LINKED ISSUE #3952: Move 'short test summary info' to the end of the output again (regression)
I find the short test summary info displayed by `py.test -ra` very useful, but I have trouble finding it when there are multiple failures with long tracebacks.  I have to page up my terminal scrollback buffer very carefully, so I don't skip over it by accident, especially when I've multiple py.test runs visible in my scrollback.

My life would be markedly improved if the short test summary info got printed _after_ all the failure tracebacks, to be visible right above my shell prompt (and the 'N tests failed' summary line).

asottile:
Resolves #3952

let me know if this should be `s/bugfix/feature/` and target that branch instead

### DIFF
--- a/changelog/3952.bugfix.rst
+++ b/changelog/3952.bugfix.rst
@@ -1,0 +1,1 @@
+Move warnings summary above short test summary.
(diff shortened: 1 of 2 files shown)


asottile:
Looks like the test rightly shows that this will reintroduce #1305

blueyed:
Just for reference: new attempt in #4399.

PR #3967, opened 2018-09-11
tests: use unittest.mock with py34+

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/3965.


blueyed:
Meh, running into https://github.com/tox-dev/tox/issues/706 now.

blueyed:
Squashed, should be good now - although the workaround is a bit unfortunate to have in the first place.

PR #3970, opened 2018-09-12
Raise exception if parametrize collects an empty parameter set

sambarluc:
Currently, `parametrize` skips or xfails a test if no arguments are collected. In both cases, the test is considered successful in most (all?) testing systems. This can potentially hide bugs in the argument collection of parametrize, and thus lead to tests not being performed by mistake. The PR adds an ini option that causes a specific exception to be raised when parametrize collects no arguments.

Implements feature request #3849

### DIFF
--- a/changelog/3849.feature.rst
+++ b/changelog/3849.feature.rst
@@ -1,0 +1,1 @@
+Add ``empty_parameter_set_mark=fail_at_collect`` ini option for raising an exception when parametrize collects an empty set.
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -976,6 +976,7 @@ passed multiple times. The expected format is ``name=value``. For example::
 
     * ``skip`` skips tests with an empty parameterset (default)
     * ``xfail`` marks tests with an empty parameterset as xfail(run=False)
+    * ``fail_at_collect`` raises an exception if parametrize collects an empty parameter set
 
     .. code-block:: ini
 
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -32,11 +32,19 @@ def istestfunc(func):
 
 
 def get_empty_parameterset_mark(config, argnames, func):
+    from ..nodes import Collector
+
     requested_mark = config.getini(EMPTY_PARAMETERSET_OPTION)
     if requested_mark in ("", None, "skip"):
         mark = MARK_GEN.skip
     elif requested_mark == "xfail":
         mark = MARK_GEN.xfail(run=False)
+    elif requested_mark == "fail_at_collect":
+        f_name = func.__name__
+        _, lineno = getfslineno(func)
+        raise Collector.CollectError(
+            "Empty parameter set in '%s' at line %d" % (f_name, lineno)
+        )
     else:
         raise LookupError(requested_mark)
     fs, lineno = getfslineno(func)
--- a/testing/test_mark.py
+++ b/testing/test_mark.py
@@ -13,7 +13,7 @@ from _pytest.mark import (
     transfer_markers,
     EMPTY_PARAMETERSET_OPTION,
 )
-from _pytest.nodes import Node
+from _pytest.nodes import Node, Collector
 
 ignore_markinfo = pytest.mark.filterwarnings(
     "ignore:MarkInfo objects:pytest.RemovedInPytest4Warning"
@@ -1091,7 +1091,14 @@ class TestMarkDecorator(object):
 @pytest.mark.parametrize("mark", [None, "", "skip", "xfail"])
 def test_parameterset_for_parametrize_marks(testdir, mark):
     if mark is not None:
-        testdir.makeini("[pytest]\n{}={}".format(EMPTY_PARAMETERSET_OPTION, mark))
+        testdir.makeini(
+            """
+        [pytest]
+        {}={}
+        """.format(
+                EMPTY_PARAMETERSET_OPTION, mark
+            )
+        )
 
     config = testdir.parseconfig()
     from _pytest.mark import pytest_configure, get_empty_parameterset_mark
@@ -1105,6 +1112,34 @@ def test_parameterset_for_parametrize_marks(testdir, mark):
     assert result_mark.kwargs["reason"].startswith("got empty parameter set ")
     if mark == "xfail":
         assert result_mark.kwargs.get("run") is False
+
+
+def test_parameterset_for_fail_at_collect(testdir):
+    testdir.makeini(
+        """
+    [pytest]
+    {}=fail_at_collect
+    """.format(
+            EMPTY_PARAMETERSET_OPTION
+        )
+    )
+
+    config = testdir.parseconfig()
+    from _pytest.mark import pytest_configure, get_empty_parameterset_mark
+    from _pytest.compat import getfslineno
+
+    pytest_configure(config)
+
+    test_func = all
+    func_name = test_func.__name__
+    _, func_lineno = getfslineno(test_func)
+    expected_errmsg = r"Empty parameter set in '%s' at line %d" % (
+        func_name,
+        func_lineno,
+    )
+
+    with pytest.raises(Collector.CollectError, match=expected_errmsg):
+        get_empty_parameterset_mark(config, ["a"], test_func)
 
 
 def test_parameterset_for_parametrize_bad_markname(testdir):
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@ Alexei Kozlenok
 Anatoly Bubenkoff
 Anders Hovmöller
 Andras Tim
+Andrea Cimatoribus
 Andreas Zeidler
 Andrzej Ostrowski
 Andy Freeland
--- a/src/_pytest/mark/__init__.py
+++ b/src/_pytest/mark/__init__.py
@@ -163,9 +163,9 @@ def pytest_configure(config):
 
     empty_parameterset = config.getini(EMPTY_PARAMETERSET_OPTION)
 
-    if empty_parameterset not in ("skip", "xfail", None, ""):
+    if empty_parameterset not in ("skip", "xfail", "fail_at_collect", None, ""):
         raise UsageError(
-            "{!s} must be one of skip and xfail,"
+            "{!s} must be one of skip, xfail or fail_at_collect"
             " but it is {!r}".format(EMPTY_PARAMETERSET_OPTION, empty_parameterset)
         )
 


src/_pytest/mark/structures.py
```diff
@@ -37,6 +42,8 @@ def get_empty_parameterset_mark(config, argnames, func):
         mark = MARK_GEN.skip
     elif requested_mark == "xfail":
         mark = MARK_GEN.xfail(run=False)
+    elif requested_mark == "fail_at_parametrize":
+        raise ParametrizeError("Empty parameter set")
```
RonnyPfannschmidt: please include the function name in the error just to ease the ux

sambarluc: Do you mean "get_empty_parameterset_mark" in the error message? I was actually thinking that it would be nice to have some introspection providing the name of the test which generated the error, but have no clue if/how that can be done.

RonnyPfannschmidt: no, i meant the name of the function object - so its more clear where the missmark happened

sambarluc: Fixed and added test for it.

sambarluc:
Not exactly sure of what the standard workflow is here. Should I squash the commits? If so, when? Or is that done by the one who decides to merge?

RonnyPfannschmidt:
if you like please squash until the sequence is to your liking, personally i prefer to merge full expressive history lines after doing a bit manual cleanup

sambarluc:
Ok, done. Unless other issues pop up, it's in your hands!

RonnyPfannschmidt:
great job, thanks for the patience and for nicely implementing the proposal

doc/en/reference.rst
```diff
@@ -976,6 +976,7 @@ passed multiple times. The expected format is ``name=value``. For example::
 
     * ``skip`` skips tests with an empty parameterset (default)
     * ``xfail`` marks tests with an empty parameterset as xfail(run=False)
+    * ``fail_at_parametrize`` raises an exception if parametrize collects an empty parameter set
```
nicoddemus: Also how calling the option `fail`? `fail_at_parametrize` seems redundant.

changelog/3849.feature.rst
```diff
@@ -0,0 +1 @@
+Add ini option for raising an exception when parametrize collects an empty set.
```
nicoddemus: Please add the option name:

```rst
Add ``empty_parameter_set_mark=fail`` ini option for raising an exception when parametrize collects an empty set.
```

src/_pytest/mark/structures.py
```diff
@@ -14,6 +14,11 @@
 EMPTY_PARAMETERSET_OPTION = "empty_parameter_set_mark"
 
 
+class ParametrizeError(Exception):
```
nicoddemus: When this is raised this will produce a very long traceback with the pytest internals:

```
============================== ERRORS ===============================
____________ ERROR collecting test-empty-parametrize.py _____________
.env36\lib\site-packages\pluggy\hooks.py:258: in __call__
.env36\lib\site-packages\pluggy\manager.py:67: in _hookexec
.env36\lib\site-packages\pluggy\manager.py:61: in <lambda>
src\_pytest\python.py:255: in pytest_pycollect_makeitem
src\_pytest\python.py:440: in _genfunctions
.env36\lib\site-packages\pluggy\hooks.py:258: in __call__
.env36\lib\site-packages\pluggy\manager.py:67: in _hookexec
.env36\lib\site-packages\pluggy\manager.py:61: in <lambda>
src\_pytest\python.py:164: in pytest_generate_tests
src\_pytest\python.py:975: in parametrize
    function_definition=self.definition,
src\_pytest\mark\structures.py:140: in _for_parametrize
    mark = get_empty_parameterset_mark(config, argnames, func)
src\_pytest\mark\structures.py:47: in get_empty_parameterset_mark
    raise ParametrizeError("Empty parameter set in '%s'" % f_name)
E   _pytest.mark.structures.ParametrizeError: Empty parameter set in 'test'
!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!
====================== 1 error in 0.17 seconds ======================
```

but if we raise `Collector.CollectError` we will much nicer representation:

```
============================== ERRORS ===============================
____________ ERROR collecting test-empty-parametrize.py _____________
Empty parameter set in 'test'
!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!
====================== 1 error in 0.08 seconds ======================
```

src/_pytest/mark/structures.py
```diff
@@ -37,6 +42,9 @@ def get_empty_parameterset_mark(config, argnames, func):
         mark = MARK_GEN.skip
     elif requested_mark == "xfail":
         mark = MARK_GEN.xfail(run=False)
+    elif requested_mark == "fail_at_parametrize":
+        f_name = func.__name__
```
nicoddemus: Please also add the line number and improve the message a bit:

```python
_, lineno = getfslineno(func) 
raise ParametrizeError("Empty parameter set during parametrization of '%s' at line %d" % (f_name, lineno + 1))
```

This produces to me:

```
Empty parameter set during parametrization of 'test' at line 4
```

Which I think reads better. What do you think?

testing/test_mark.py
```diff
@@ -1107,6 +1107,25 @@ def test_parameterset_for_parametrize_marks(testdir, mark):
         assert result_mark.kwargs.get("run") is False
 
 
+def test_parameterset_for_fail_at_parametrize(testdir):
+    testdir.makeini(
+        "[pytest]\n{}={}".format(EMPTY_PARAMETERSET_OPTION, "fail_at_parametrize")
```
nicoddemus: This can be written as:

```python
testdir.makeini("""
    [pytest]
    {}=fail_at_parametrize
""".format(EMPTY_PARAMETERSET_OPTION))
```

Which is a more readable IMHO. 👍

testing/test_mark.py
```diff
@@ -1107,6 +1107,25 @@ def test_parameterset_for_parametrize_marks(testdir, mark):
...
+
+    config = testdir.parseconfig()
+    from _pytest.mark import pytest_configure, get_empty_parameterset_mark
+    from _pytest.mark.structures import ParametrizeError
+
+    pytest_configure(config)
+
+    test_func = all
+
+    with pytest.raises(ParametrizeError) as err:
```
nicoddemus: I suggest to update this to:

```python
with pytest.raises(ParametrizeError, match="Empty parameter set in 'all'"):
    get_empty_parameterset_mark(config, ["a"], test_func)
```

doc/en/reference.rst
RonnyPfannschmidt: i believe the redundancy is needed because its not a mark, it triggers a different behaviour, i believe its important to encode that in the option value

src/_pytest/mark/structures.py
RonnyPfannschmidt: good find, lets use that

doc/en/reference.rst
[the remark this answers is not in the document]
sambarluc: If you ask me, I would definitely pick `fail_at_parametrize` as the default, that is by far the most obvious behaviour we expect in all our tests. However, I don't have a broader overview of how `pytest` is used.

sambarluc: I agree with @RonnyPfannschmidt, `fail` may be interpreted as a marker by a careless reader. A possible alternative would be to have a `fail_at_empty` option in `parametrize` itself. However, I haven't looked at that part of the code, so this is a suggestion purely based on (my) ux. This would make sense (for me), only with a default value of `True`.

src/_pytest/mark/structures.py
sambarluc: Excellent suggestion, of course I had no clue of that, thanks! It's coming...

changelog/3849.feature.rst
sambarluc: Done.

doc/en/reference.rst
nicoddemus: Yeah, I agree, in retrospect seems like a mistake we called the option `empty_parameter_set_mark`, `empty_parameter_set_behavior` or something like that would naturally map to `skip`, `xfail` and `failure`. Calling a mark feels to me like we leaked an implementation detail.

src/_pytest/mark/structures.py
sambarluc: I don't have a strong opinion, as the test name is most often sufficient. However, that does not cost much so I will include the line number.

testing/test_mark.py
sambarluc: Agreed, but that's a copy paste from the test above, so I change both.

doc/en/reference.rst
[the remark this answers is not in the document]
nicoddemus: Yes, I agree, it should error out by default.

Hmmm one way to implement this transition with minimal disturbance to our users: if we find an empty parameter, we issue a warning saying that the default will change in a few versions, and users that like the current default can set it into their `pytest.ini` files. What do you think @RonnyPfannschmidt? 

But sorry, just realized this is not related to this PR and I'm hijacking this thread. 😁

testing/test_mark.py
sambarluc: Nice feature, I did not know it!

doc/en/reference.rst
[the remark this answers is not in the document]
RonnyPfannschmidt: i created a follow-up issue

sambarluc:
Not sure of how badly github deals with changes in history, so I leave the `squash!` and `fixup!` commits until we converge.

nicoddemus:
Thanks @sambarluc for the contribution, much appreciated! 👍

nicoddemus:
Feel free to rebase/squash your commits now. 👍

sambarluc:
Just an idea, before I squash. What about having the option as
'fail_at_collect', instead of 'fail_at_parametrize'? I was thinking that it
may offer less opportunities of confusion.

Le ven. 14 sept. 2018 à 00:17, Bruno Oliveira <notifications@github.com> a
écrit :

RonnyPfannschmidt:
nice idea - in that case i would like to propose the name `fail_instead_at_collect` and i would like to get the input of @nicoddemus as well

sambarluc:
My 5 cents: I would not go for `fail_instead_at_collect`, as it implies that the other options do something else, which is true now but maybe not in the future.

RonnyPfannschmidt:
hmm, that brings me to the point that that particular option wasn't intended to manage it to begin with - lets go with the value you proposed and open a follow-up ticket to think it trough a bit later

nicoddemus:
Hi guys,

I think `fail_instead_at_collect` and `fail_at_collect` mean the same thing actually. Do you see a difference in semantics @RonnyPfannschmidt? What "fail_instead" implies something else is possible to do at collection, what's the "instead of" you are thinking? 

> I'll wait for @nicoddemus before squashing, I guess it's not particularly urgent.

You are right, this will go into `3.9`, which is weeks away. 😁

sambarluc:
Not sure what's wrong with appveyor. How can I retrigger?

RonnyPfannschmidt:
its a codecov failure
@blueyed - what do we do about the regular failures of that service?

PR #3982, opened 2018-09-14
Ignore pytest cache

### LINKED ISSUE #3286: .pytest_cache is showing up in projects git repos
i just noted that this started to happen, as far as i can tell the .cache rename was a breaking change for users

nicoddemus:
Fix #3286

### DIFF
--- a/testing/python/fixture.py
+++ b/testing/python/fixture.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import sys
 import textwrap
 
 import pytest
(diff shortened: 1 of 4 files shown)


testing/python/fixture.py
```diff
@@ -488,6 +489,10 @@ def test_method(self, something):
         assert len(arg2fixturedefs) == 1
         assert arg2fixturedefs["something"][0].argname == "something"
 
+    @pytest.mark.skipif(
+        hasattr(sys, "pypy_version_info"),
+        reason="this method of test doesn't work on pypy",
+    )
     def test_request_garbage(self, testdir):
```
nicoddemus: Change flaky test_request_garbage to provide more debug information

This test fails *very* rarely when running in xdist.


Not a part of the purpose of this PR, but I didn't feel like making a new PR just for that.

PR #3987, opened 2018-09-15
Fix scope determination with indirect parameters

### LINKED ISSUE #3941: class scope fixture is recognized as function scope
[Edit: Bug repro is further simplified. Please see the repro code below]

Hi, I found a case when class scope fixture is recognized as function scope, which fails the test with error "You tried to access the 'function' scoped fixture '...' with a 'class' scoped request object"

Could you please confirm if this is a pytest bug, or if this is not an encouraged pattern and I should try to implement it differently?

platform win32 -- Python 3.6.6, pytest-3.7.3, py-1.5.4, pluggy-0.7.1

```
def class_dep(request):
    print("Fixture setup")

@pytest.fixture(scope="function")
def func_dep():
    pass

@pytest.mark.usefixtures("class_dep")
@pytest.mark.parametrize('class_dep', [10,20], indirect=True)
class TestScenario():
    def test_foo(self, func_dep):
        pass
    
    def test_bar(self):
        pass


> pytest.exe "test_foo1.py::TestScenario" -v -s

test_foo1.py::TestScenario::test_foo[10] Fixture setup
PASSED
test_foo1.py::TestScenario::test_foo[20] Fixture setup
PASSED
test_foo1.py::TestScenario::test_bar[10] Fixture setup
PASSED
test_foo1.py::TestScenario::test_bar[20] Fixture setup
PASSED
```

**"Fixture setup" is printed 4 times.**
If I change def test_foo(self, func_dep): to drop reference to func_dep, then it works as expected ("Fixture setup is printed 2 times").

nicoddemus:
Fix #3941

### DIFF
--- a/changelog/3941.bugfix.rst
+++ b/changelog/3941.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bug where indirect parametrization would consider the scope of all fixtures used by the test function to determine the parametrization scope, and not only the scope of the fixtures being parametrized.
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1141,13 +1141,18 @@ def _find_parametrized_scope(argnames, arg2fixturedefs, indirect):
     """
     from _pytest.fixtures import scopes
 
-    indirect_as_list = isinstance(indirect, (list, tuple))
-    all_arguments_are_fixtures = (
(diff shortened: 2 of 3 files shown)


nicoddemus:
I suspect this fixes other bugs in the issue tracker, will take a look tomorrow.

PR #3990, opened 2018-09-17
coverage: use modules

blueyed:
This should increase coverage for subprocesses, where `source` is used
only from the config file, but not the initial `--source` argument.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,9 @@
 [run]
-source = _pytest,testing
+source = pytest,_pytest,testing/
 parallel = 1
 branch = 1
+
+[paths]
+source = src/
+  .tox/*/lib/python*/site-packages/
+  .tox\*\Lib\site-packages\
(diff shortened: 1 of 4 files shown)


blueyed:
I'm investigating codecov problems with them.
The project compare view (https://codecov.io/gh/pytest-dev/pytest/compare/bb57186dd4ad1bdf57ce05f41b8326a634741f23...2868b39d40945020097e6c08e10bd64b75c8017a) is still timing out on their side unfortunately, but would be required to see how this actually improves things.

.coveragerc
```diff
@@ -1,4 +1,8 @@
 [run]
-source = _pytest,testing
+source = pytest,_pytest,testing/
 parallel = 1
 branch = 1
+
+[paths]
+source = src/
+  .tox/*/lib/python*/site-packages/
```
nicoddemus: This on Windows would be `.tox\*e\Lib\site-packages`. Can we add another entry here and coverage will try one and then the other?

blueyed: Yes, added `.tox\*e\Lib\site-packages\`.

This gets used with `coverage report` - so you could test it after e.g. `tox -e py36-coverage -- testing/deprecated_test.py`.

nicoddemus: Works like a charm! 🤗

PR #3997, opened 2018-09-19
[WIP] Introduce deprecations page

### LINKED ISSUE #3996: Deprecations page
I think we need a page in the docs which lists all features we have deprecated, including rationale, version of deprecation, and how to update code. Currently that information is scattered in the CHANGELOG.

nicoddemus:
Got the page going, wanted to see what others think.

Documenting the deprecations in the CHANGELOG often is done too briefly, while I believe we should write down the rationale and provide examples on how to update the code. I think when users understand the reasoning behind deprecations they are much more understanding. 

Also examples on how to update the code currently are often provided in the docstrings of the features we plan to remove, which will disappear when we effectively remove said feature.

The idea is to keep this page around "forever" as a reference for users which encounter ancient codebases and want to revive them.

fix #3996

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,19 +24,9 @@ pytest 3.8.0 (2018-09-05)
 Deprecations and Removals
 -------------------------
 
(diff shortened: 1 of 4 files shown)


nicoddemus:
Ready for review! 👍

nicoddemus:
cc @okken

PR #3998, opened 2018-09-19
logging: del item.catch_log_handler only in teardown

blueyed:
Without this caplog.record_tuples etc is not available when using
`--pdb`.

Maybe related to https://github.com/pytest-dev/pytest/issues/3099.

### DIFF
--- a/changelog/3998.bugfix.rst
+++ b/changelog/3998.bugfix.rst
@@ -1,0 +1,1 @@
+Fix caplog's catch_log_handler not being available with --pdb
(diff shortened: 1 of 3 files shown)


blueyed:
> Fixes #3099

Will change that to "Ref:" in the commit message, since it might be something different there after all.

nicoddemus:
Can we have a test along this to avoid regressions in the future?

blueyed:
Added a test.

PR #4012, opened 2018-09-21
Fix typo in docstring

maxalbert:
Trivial typo fix in docstring.

Many thanks for everyone's awesome work on `pytest`! I use it every day and could not imagine what my workflow would be without it. Your work is hugely appreciated!


crazymerlyn:
Thanks!

PR #4019, opened 2018-09-22
Show deprecation warnings even if filters are customized

### LINKED ISSUE #4013: supplying filterwarnings overrides default filters
I'm really happy about pytest 3.8 and the work in #2908.

And now that I'm seeing warnings again, I realize how blissful ignorance is.

So I've gone through one project and fixed all of the deprecation warnings for the project under test (there are still many coming from other projects). In order to assert that the project no longer has any warnings, I added this to pytest.ini:

```
filterwarnings = 
    error:::mylib[.*]
```

Now tests pass, but all of the deprecation warnings (from the dependencies) are suppressed. That wasn't what I'd intended by setting the project to error on warnings.

I don't think that's the right behavior. Instead, pytest should configure the filters for DeprecationWarnings and PendingDeprecationWarnings and then allow the user to _override_ those filters but not suppress the defaults with any value.

If the project must suppress the defaults, the docs should at least describe what the defaults are so they can be copied into the configuration wishing to define other filters. I had to traverse to the source to determine what the defaults are:

```
filterwarnings =
	always::DeprecationWarning
	always::PendingDeprecationWarning
```

Is there any reason not to always configure the defaults and allow overrides?

nicoddemus:
Fix #4013

### DIFF
--- a/doc/en/warnings.rst
+++ b/doc/en/warnings.rst
@@ -101,22 +101,28 @@ DeprecationWarning and PendingDeprecationWarning
 ------------------------------------------------
 
 .. versionadded:: 3.8
+.. versionchanged:: 3.9
 
-By default pytest will display ``DeprecationWarning`` and ``PendingDeprecationWarning`` if no other warning filters
-are configured.
+By default pytest will display ``DeprecationWarning`` and ``PendingDeprecationWarning``.
 
-To disable showing ``DeprecationWarning`` and ``PendingDeprecationWarning`` warnings, you might define any warnings
-filter either in the command-line or in the ini file, or you can use:
+Sometimes it is useful to hide some specific deprecation warnings that happen in code that you have no control over
+(such as third-party libraries), in which case you might use the standard warning filters options (ini or marks).
+For example:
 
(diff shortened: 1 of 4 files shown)


doc/en/warnings.rst
```diff
@@ -101,22 +101,28 @@ DeprecationWarning and PendingDeprecationWarning
...
 .. versionadded:: 3.8
+.. versionchanged:: 3.9
 
-By default pytest will display ``DeprecationWarning`` and ``PendingDeprecationWarning`` if no other warning filters
-are configured.
+By default pytest will display ``DeprecationWarning`` and ``PendingDeprecationWarning``.
 
-To disable showing ``DeprecationWarning`` and ``PendingDeprecationWarning`` warnings, you might define any warnings
-filter either in the command-line or in the ini file, or you can use:
+Sometimes it is useful to hide some select deprecation warnings that happen in code that you have no control over
```
blueyed: s/select/specific/ ?

jaraco:
Looks great. Thanks!

PR #4040, opened 2018-09-26
Exclude empty reports for passed tests

ods:
A check is added to avoid empty reports with headline only when `-r P` option is used.

### DIFF
--- a/changelog/4040.trivial.rst
+++ b/changelog/4040.trivial.rst
@@ -1,0 +1,1 @@
+Exclude empty reports for passed tests when ``-rP`` option is used.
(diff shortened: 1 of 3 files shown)


ods:
> this can only happen if logging and output capture get disabled

Sections are not included when nothing was written to them. Without this check I get a lot of empty reports like the following:

```
==================================== PASSES ====================================
_________________________ SomeTestCase.test_something __________________________
______________________ SomeTestCase.test_something_other _______________________
_________________________ SomeTestCase.test_even_more __________________________
__________________________ SomeTestCase.test_and_more __________________________
…
```

nicoddemus:
Great work @ods. Would you mind adding a new CHANGELOG entry? Thanks!

PR #4047, opened 2018-09-27
Add pytest_before_assert hook.

robertschweizer:
This hook will run before every assert, in contrast to the
pytest_assertrepr_compare hook which only runs for failing asserts.

As of now, no parameters are passed to the hook.

We use this hook to print the assertion code of passed assertions to
collect test evidence. This is done using inspect::

    inspect.stack()[7].code_context[0].strip()

Signed-off-by: Schweizer, Robert <rschweizer@definiens.com>

just a guideline):

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -200,6 +200,7 @@ Raphael Pierzina
 Raquel Alegre
 Ravi Chandra
 Roberto Polli
+Robert Schweizer
 Roland Puntaier
 Romain Dorgueil
 Roman Bolshakov
--- a/changelog/4047.feature.rst
+++ b/changelog/4047.feature.rst
@@ -1,0 +1,1 @@
+Add pytest_before_assert hook to run before any assertion.
--- a/doc/en/assert.rst
+++ b/doc/en/assert.rst
@@ -252,6 +252,33 @@ the conftest file:
 .. _assert-details:
 .. _`assert introspection`:
 
+Collecting information about passing assertions
+-----------------------------------------------
+
+The ``pytest_assertrepr_compare`` hook only runs for failing assertions. Information
+about passing assertions can be collected with the ``pytest_before_assert`` hook.
+
+.. autofunction:: _pytest.hookspec.pytest_before_assert
+   :noindex:
+
+For example, to report every encountered assertion, the following hook
+needs to be added to :ref:`conftest.py <conftest.py>`::
+
+    # content of conftest.py
+    def pytest_before_assert():
+        print("Before-assert hook is executed.")
+
+now, given this test module::
+
+    # content of test_sample.py
+    def test_answer():
+        assert 1 == 1
+
+the following stdout is captured, e.g. in an HTML report::
+
+    ----------------------------- Captured stdout call -----------------------------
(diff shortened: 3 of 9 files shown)


nicoddemus:
At first glance it seems there's overlap with #3479, what do you think?

cc @Sup3rGeo

robertschweizer:
You're right, I should have checked #3479 first. 

If the other feature will take much longer, we could consider merging this first as a basic version without parameters being passed to the hook. We'd have to change my before_hook to an assertion pass hook for that.

Otherwise I'm ok with closing this, it's my fault for not checking the other PR.

Sup3rGeo:
Thanks for the ping @nicoddemus.

@robertschweizer, sounds interesting but I think I still don't understand 100% the hook. Can you give a more concrete example of how would you use the hook? Also how to use the inspect command you mentioned to get information? A small example would be ideal.

robertschweizer:
Thanks for still considering this. I'll try and give a full example.

For test documentation, we'd like to collect all the assertions in our test code. In the tests where we use this, most assertions compare against explicit values like string literals. When the assertion passes, it is then completely sufficient to print the original code, without information about the actual operand values.

We started with writing our on assertion functions, probably similar to your first solution of #3457. They just used `inspect.stack()[1]` to access the calling code, and printed one line of it. Now we're looking for a cleaner solution that lets us keep the benfits of nicely formatted Pytest asserts.

Setting this conftest.py:
```python
import inspect

def pytest_before_assert():
    expr = inspect.stack()[7].code_context[0].strip()
    print(expr)
```
This test code:
```python
def test_some_code():
    encountered = get_software_return_value()
    assert encountered == "expected string"
```
prints this output:

>   ----------------------------- Captured stdout call -----------------------------
>   assert encountered == "expected string"

The stack depth of 7 is a bit ugly, but I'm not sure there's a better solution of printing the calling code.

robertschweizer:
Since the other PR has been stuck for a while, what do you think about merging this one?

robertschweizer:
> Reminded me of pytest-pdb which has some methods to find the test etc, and you could use this method to get to the actual `assert` I assume (https://github.com/fschulze/pytest-pdb/blob/b9af6a801fc62ad2a26bbc60b22a2e9af48ab757/pytest_pdb.py#L7-L33).

Thanks for the suggestion! That solution is also pretty dependent on Pytest implementation details, and I can't get it to work properly with the current versions. I'm thinking about relying only on the `pytest_pyfunc_call` function name:

```python
def pytest_before_assert():
    stack = inspect.stack()
    for index, frame_info in enumerate(stack):
        if frame_info.function == "pytest_pyfunc_call":
            print(stack[index - 1].code_context[0])
```

nicoddemus:
Hi @robertschweizer, 

It has been a long time since it has last seen activity, plus we have made sweeping changes on master to drop Python 2.7 and 3.4 support, so this PR has some conflicts which require attention.

Please don't consider this a rejection of your PR, we just want to get this out of sight until you have the time to tackle this again. If you get around to work on this in the future, please don't hesitate in re-opening this!

PR #4069, opened 2018-10-02
Fix source reindenting by using `textwrap.dedent` directly.

### LINKED ISSUE #4066: unexpected indent of test function with docstring
When run test with docstring, FAILURE message indent is unexpected.

For example, 

```python
import unittest
class TestSome(unittest.TestCase):

    def test_bar(self):
        """
        This test fails.
        """
        self.assertEqual(1, 3)
```

run above test with pytest, following message is shown.

```
    def test_foo(self):
        """
            This test fails.
            """
>       self.assertEqual(1, 3)
E       AssertionError: 1 != 3
```

But I expected message like below.

```
    def test_foo(self):
        """
        This test fails.
        """
>       self.assertEqual(1, 3)
E       AssertionError: 1 != 3
```

I found out the cause is probably `tokenize.generate_tokens`. (It is used `src/_pytest/_code/source.deindent` ). However, I do not know how to fix it.

# Version

- macOS 10.14
- Python: 3.6.5
- pytest: 3.8.1

asottile:
Resolves #4066

### DIFF
--- a/src/_pytest/_code/source.py
+++ b/src/_pytest/_code/source.py
@@ -7,6 +7,7 @@ import linecache
 import sys
 import six
 import inspect
+import textwrap
 import tokenize
 import py
 
@@ -23,7 +24,6 @@ class Source(object):
     def __init__(self, *parts, **kwargs):
         self.lines = lines = []
         de = kwargs.get("deindent", True)
-        rstrip = kwargs.get("rstrip", True)
         for part in parts:
             if not part:
                 partlines = []
@@ -33,11 +33,6 @@ class Source(object):
                 partlines = [x.rstrip("\n") for x in part]
             elif isinstance(part, six.string_types):
                 partlines = part.split("\n")
-                if rstrip:
-                    while partlines:
-                        if partlines[-1].strip():
-                            break
-                        partlines.pop()
             else:
                 partlines = getsource(part, deindent=de).lines
             if de:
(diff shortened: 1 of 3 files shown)


src/_pytest/_code/source.py
```diff
@@ -23,7 +24,6 @@ class Source(object):
     def __init__(self, *parts, **kwargs):
         self.lines = lines = []
         de = kwargs.get("deindent", True)
-        rstrip = kwargs.get("rstrip", True)
```
RonnyPfannschmidt: i believe this is a breaking api change - what where our guarantees on those internals

testing/code/test_source.py
```diff
@@ -417,15 +411,15 @@ def f():
 """
 
     lines = deindent(inspect.getsource(f).splitlines())
-    assert lines == ["def f():", '    c = """while True:', "    pass", '"""']
+    assert lines == ["    def f():", '        c = """while True:', "    pass", '"""']
```
RonnyPfannschmidt: this looks pretty much broken now

src/_pytest/_code/source.py
asottile: This isn't an api break, the parameter is ignored (nothing is validated about **kwargs)

testing/code/test_source.py
asottile: This is the same ad the test above, but in list form, I'm pretty sure this is more correct than before (before it introduced an indentation error)

RonnyPfannschmidt: i see what you mean, please make it a indented block with disabled black formatting and a comment that the indent of the last `"""` is the relative ancor

asottile: I feel I could also delete this test since the above one covers the same behavior and is more expressive :man_shrugging:

RonnyPfannschmidt: hmm, true actually

src/_pytest/_code/source.py
```diff
@@ -269,46 +264,7 @@ def getsource(obj, **kwargs):
 
 
 def deindent(lines, offset=None):
```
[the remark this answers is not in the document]
asottile: this is indeed the case, are we ok breaking this api?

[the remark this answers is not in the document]
nicoddemus: `dedent` is in `_pytest._code.source`, so I don't believe it will break any code. @RonnyPfannschmidt's concern was with the `Source` object, which is exposed by `testdir`, but here it is not the case.

@RonnyPfannschmidt please correct me if I'm wrong. 👍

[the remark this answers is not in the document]
asottile: ah ok, I was also going to remove `offset=None` in `Source.deindent` as well, that's slightly closer to a break but still nobody should be calling it :laughing:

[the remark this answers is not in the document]
RonnyPfannschmidt: i all for removing that api

Issue #4073, opened 2018-10-03
Add timeout to testdir.run()


altendky:
I would like to rewrite [this pytest-twisted test](https://github.com/pytest-dev/pytest-twisted/blob/14b16ca114e735d46d52cd36dc2f2306ed0ec636/testing/test_basic.py#L210-L261) but the new way would run indefinitely on failure.  It's already using [`testdir.run()`]() but that doesn't support a timeout.   There's `spawn_pytest()` and `spawn()` but they involve pexpect which doesn't seem right.  What's proper?  I can certainly provide a PR for adding a timeout to `testdir.run()` if that's the answer.

nicoddemus:
Hi @altendky,

Looking at the code, I think it is OK to add this option to `runpytest_subprocess` and `run`. A PR is welcome! 👍

altendky:
https://github.com/pytest-dev/pytest/pull/4078

PR #4078, opened 2018-10-04
Add timeout for Testdir.runpytest_subprocess() and Testdir.run()

altendky:
pytest-dev/pytest#4073

WIP for:
* [x] Deciding between loop/sleep and `signal.alarm()` for py2 timeout
* [x] Ok to throw away all the `kwargs` (other than `timeout`) to `runpytest_subprocess()`?
* [x] Is it really ok to now react to the `timeout` kwarg which was previously ignored?
* [x] Stock checklist below

---

just a guideline):

- [ ] ~~Target the `master` branch for bug fixes, documentation updates and trivial changes.~~


src/_pytest/pytester.py
```diff
@@ -1061,7 +1065,27 @@ def run(self, *cmdargs):
...
+            timeout = kwargs.get('timeout')
+            if timeout is None:
+                ret = popen.wait()
+            elif six.PY3:
+                try:
+                    ret = popen.wait(timeout)
+                except subprocess.TimeoutExpired:
+                    raise self.TimeoutExpired()
+            else:
+                end = monotonic.monotonic() + timeout
```
nicoddemus: I don't think we need to introduce a new dependency to get an accurate timeout here, TBH. I think using `time.time()` on Linux and `time.clock()` on Windows will give us good enough accuracy? To be clear I like having the two implementations separate like this, makes it easier to drop the Python 2-only code later, I'm questioning the external dependency to achieve that.

src/_pytest/pytester.py
```diff
@@ -1061,7 +1065,27 @@ def run(self, *cmdargs):
...
+
+                while True:
+                    ret = popen.poll()
+                    if ret is not None:
+                        break
+
+                    remaining = end - monotonic.monotonic()
+                    if remaining <= 0:
+                        raise self.TimeoutExpired()
```
nicoddemus: I suggest that we add a message here with the command and how many seconds was the timeout.

src/_pytest/pytester.py
```diff
@@ -1119,7 +1143,13 @@ def runpytest_subprocess(self, *args, **kwargs):
         if plugins:
             args = ("-p", plugins[0]) + args
         args = self._getpytestargs() + args
-        return self.run(*args)
+
+        if "timeout" in kwargs:
```
nicoddemus: This is a bit simpler:

```python
return self.run(*args, timeout=kwargs.get("timeout"))
```

testing/test_pytester.py
```diff
@@ -401,3 +401,20 @@ def test_testdir_subprocess(testdir):
...
     result = testdir.runpytest("-k", u"💩")
     assert result.ret == EXIT_NOTESTSCOLLECTED
+
+
+def test_testdir_run_no_timeout(testdir):
+    testfile = testdir.makepyfile("def test_no_timeout(): pass")
+    assert testdir.runpytest_subprocess(testfile).ret == EXIT_OK
+
+
+def test_testdir_run_timeout_expires(testdir):
```
nicoddemus: 👍

src/_pytest/pytester.py
```diff
@@ -1039,7 +1043,7 @@ def popen(self, cmdargs, stdout, stderr, **kw):
 
         return popen
 
-    def run(self, *cmdargs):
+    def run(self, *cmdargs, **kwargs):
```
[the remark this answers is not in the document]
altendky: I'd love to do that, and did do that, but py2.  This re-surprises me each time I find it.

https://repl.it/@altendky/pytest-devpytest4078-001
```python
def f(*args, something=None):
    pass
```

```
Python 2.7.10 (default, Jul 14 2015, 19:46:27)
[GCC 4.8.2] on linux
   
  File "python", line 1
    def f(*args, something=None):
                         ^
SyntaxError: invalid syntax
```

But yes, if we are concerned about extras getting lost I can add some validation.  I was maybe being a bit loose with it given elsewhere in the file there's a random `**kwargs` that is (was?) entirely unused.

altendky:
@nicoddemus and @asottile, thanks for the interest and feedback.  The main thing I dislike about this is rolling my own timeout.  [`subprocess32`](https://pypi.org/project/subprocess32/) was recommended but I can't say I like the comment "It has not been tested by the author on Windows."  Though they do apparently have it [testing on AppVeyor](https://ci.appveyor.com/project/gpshead/python-subprocess32) so perhaps the comment should just be ignored.  How do you feel about that as a dependency as compared with my super simplistic timeout?

setup.py
```diff
@@ -65,6 +65,7 @@ def main():
         "attrs>=17.4.0",
         "more-itertools>=4.0.0",
         "atomicwrites>=1.0",
+        "monotonic",
```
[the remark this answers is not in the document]
altendky: I left it this way because then you can access monotonic consistently via `monotonic.monotonic()`, though sure, at this point we only use it for one thing and only in py2 (and it seems is going away anyways).

nicoddemus:
Thanks @altendky for the quick response!

> How do you feel about that as a dependency as compared with my super simplistic timeout?

I'm OK with the implementation you posted. 👍

src/_pytest/pytester.py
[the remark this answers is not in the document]
asottile: oh, here's keyword-only arguments in python2+3:

```python
def f(**kwargs):
    arg = kwargs.pop('arg', None)
    if kwargs:
        raise TypeError('Unexpected arguments: {}'.format(', '.join(sorted(kwargs))))
```

altendky:
@nicoddemus, also, there are several other functions that could expose and pass through a timeout parameter.  Should they all?  That would be my tendency.  Tests for each or just the one that actually implements the timeout?

nicoddemus:
I only add the timeout parameter to `runpytest_subprocess()` in addition to `run()`, as I believe those will be the ones where the `timeout` parameter will be more useful. We can add to other methods later as need arises.

altendky:
> Thanks @altendky for the quick response!
> 
> > How do you feel about that as a dependency as compared with my super simplistic timeout?
> 
> I'm OK with the implementation you posted.

Too bad it's horrible.  It waits 90% of the timeout even if the process finishes immediately.  First option is just to have a (parameter-settable?) resolution for waiting.  Maybe default to 0.1 seconds.  Second option is maybe something with `signal.alarm`.  Coding them both to see what they look like.

altendky:
https://github.com/pytest-dev/pytest/pull/4078/commits/900cef639710682c67a4ab88bd9e7666b14a9da2

So `signal.alarm()` just feels slick, in a really bad way.  Too much global dependence and effect for my taste.  I vote for the looping.  How about you @nicoddemus and @asottile?

asottile:
yeah probably going to need a loop anyway, there's no signals on windows :)

asottile:
would it be the worst to only support this behaviour in python3+?

altendky:
You think the code sitting there is going to cause enough trouble we don't want to deal with maintaining it?  But in my case it's for pytest-twisted which is supporting 2.7 so I do want it, yes.  :]

asottile:
ah ok I meant more because this is going to be hard to get right / working on python2 (especially python2 windows) 😆

altendky:
Do we really want to be throwing away all the (now other than `timeout`) `kwargs` to `runpytest_subprocess()`?

https://travis-ci.org/pytest-dev/pytest/jobs/437442250#L950

altendky:
Is it just me or is Travis being weird? Seems to be cancelling one build and then starting another.

altendky:
> Do we really want to be throwing away all the (now other than `timeout`) `kwargs` to `runpytest_subprocess()`?
> 
> https://travis-ci.org/pytest-dev/pytest/jobs/437442250#L950

So we've got this sort of weirdly breaking change here.  Previously `Testdir.runpytest_subprocess()` totally ignored all kwargs.  As shown above even the test suite made use of this for some reason (I haven't dug into why).  So, turning a previously ignored kwarg into a timeout feature could break things for people previously passing a timeout kwarg.  How do we categorize this?  Do we want to deprecate the arbitrary throwing away of parameters or such?

altendky:
I'm almost satisfied.  Just the two remaining questions left in the original message for tracking.

* Ok to throw away all the kwargs (other than timeout) to runpytest_subprocess()?
* Is it really ok to now react to the timeout kwarg which was previously ignored?

nicoddemus:
@altendky I think so, the patch looks good as is to me. 👍

src/_pytest/pytester.py
```diff
@@ -1061,7 +1076,39 @@ def run(self, *cmdargs):
...
+
+                resolution = min(0.1, timeout / 10)
+
+                while True:
+                    ret = popen.poll()
+                    if ret is not None:
+                        break
+
+                    remaining = end - time.time()
+                    if remaining <= 0:
```
asottile: I would write `if time.time() > end:` myself, but this is equivalent logic

blueyed:
@altendky 
> Is it just me or is Travis being weird? Seems to be cancelling one build and then starting another.

It cancels previous builds for new commits, assuming that you have pushed fixups.  It is a Travis setting, enabled for pytest.

altendky:
@blueyed thanks but I think it isn't that.  Both builds are for the same commit and start at the same time as far as I can tell.  The GitHub build links start out pointing at the canceled build.

https://travis-ci.org/pytest-dev/pytest/builds/437650065
https://travis-ci.org/pytest-dev/pytest/builds/437650097

src/_pytest/pytester.py
altendky: That's a leftover from when I actually used `remaining` elsewhere.  Thanks for catching it.

blueyed:
Yeah, looks weird (same commit).  Not sure.  Likely related to this feature though I guess.

altendky:
@blueyed there are a lot of duplicates including for other people as well, though some of them actually fully ran both builds instead of canceling.

https://travis-ci.org/pytest-dev/pytest/pull_requests

anyways...

altendky:
I really don't like that we are already getting silly timeout errors.

https://travis-ci.org/pytest-dev/pytest/jobs/437814901#L959

An empty test is taking more than 10 seconds? Should I just massively extend these times so they don't fail for a minute or five or such?

blueyed:
@altendky 
Is this because of Travis being slow sometimes?
Have not investigated into this PR, but shouldn't this failure also hide the internal traceback?  (`__tracebackhide__ = True`)

altendky:
@blueyed, good call, I forgot about that feature.  Thanks.  I'll review that when I get a chance.

altendky:
@blueyed, I assume it was random slowness since only one job failed.  Certainly we want the tests to pass quickly, but we can probably afford to occasionally wait a long time to avoid having flaky tests (so long as test integrity is maintained).

altendky:
I forgot I had permissions to rebuild, and now it passes.  So yeah, intermittent...  :[  I hate time.

altendky:
I'm not used to being able to merge things myself...  Am I supposed to do that now?  Wait for somewhere else to merge it?  It's no matter to me, I can wait.

nicoddemus:
Oh you are free to merge after at least one reviewer has approved and there are no more comments to address. 😁 

I would merge it now, but please do the honors. 🎉

nicoddemus:
Awesome, thanks @altendky!

testing/test_pytester.py
```diff
@@ -413,7 +413,7 @@ def test_testdir_run_with_timeout(testdir):
     testfile = testdir.makepyfile("def test_no_timeout(): pass")
 
     start = time.time()
-    result = testdir.runpytest_subprocess(testfile, timeout=10)
+    result = testdir.runpytest_subprocess(testfile, timeout=120)
```
blueyed: @altendky 
Wouldn't that still fail with `assert duration < 5` below?

blueyed: Unfortunately the build failure is gone, likely because it was restarted?!  (would be good to copy'n'paste those in the future).  (it was for py27-pluggymaster)

altendky: :[ Yes, I should have extended that as well.  Or, really, made one number calculated from the other to avoid this error in the future.

How should I proceed?  Push a new commit to the branch?  Start a new branch?

And :[ again.  I did double check that my link to the broken build was for a specific build.  I didn't realize that Travis was going to reuse that number and effectively break the link.  Lesson learned.  *sigh* Don't restart builds, trigger a new build (empty commit, close/reopen PR, something).

nicoddemus:
> How should I proceed? Push a new commit to the branch? Start a new branch?

As this has already been merged, I believe we need to start a new branch.

blueyed:
I think it is fine for now - just noticed this.
After all it should never take that long in the first place, and if it fails in the future we might be able to debug it from there due to the assertion (5s) failing then.

altendky:
I'd rather fix it now than leave a wrong and confusing thing sitting around.  Unless either of you want to discourage correcting it.

nicoddemus:
> I'd rather fix it now than leave a wrong and confusing thing sitting around. Unless either of you want to discourage correcting it.

If you have the time, I personally prefer this to be fixed now. 😁

Issue #4063, opened 2018-10-02
exclude empty setup/teardown from --durations


azrdev:
Write a test without setup or tear-down:
~~~python
def test_sleep():
    import time
    time.sleep(2)
~~~

Run `pytest --durations 0` on this test. Observe something like 
~~~
======================= slowest test durations =======================
2.00s call     tests/test.py::test_sleep
0.00s teardown tests/test.py::test_sleep
0.00s setup    tests/test.py::test_sleep
~~~

On medium-size testsuites, where one would like to see timings off all tests that run, these useless 0.00s entries clutter the output, so I'd suggest to remove them.

# Versions
python-pytest 3.8.1-1
python 3.7.0-4
from archlinux repos

nicoddemus:
Sounds good. I believe your suggestion should be the default, but we fallback to show everything if the user passes `-vv`. 👍

pecey:
So the current behaviour should happen if the user passes `-vv`, else we should only report values which are non-zero?

azrdev:
@pecey  IMHO that would be misleading. We should only report values that were actually run, i.e. exclude setup/teardown exactly if they were not specified by the user.

crazymerlyn:
Fixed by #4086.

PR #4086, opened 2018-10-08
Exclude durations that are 0.00 seconds long.

jeffreyrack:
Implements #4063 by excluding 0.00 second long durations from the --durations output unless -vv is used.

### DIFF
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -51,6 +51,7 @@ def pytest_terminal_summary(terminalreporter):
 
     for rep in dlist:
         if verbose < 2 and rep.duration < 0.01:
+            tr.write_line("0.00 durations hidden.  Use -vv to show these durations.")
             break
         nodeid = rep.nodeid.replace("::()::", "::")
         tr.write_line("%02.2fs %-8s %s" % (rep.duration, rep.when, nodeid))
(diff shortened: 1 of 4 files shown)


src/_pytest/runner.py
```diff
@@ -49,6 +50,8 @@ def pytest_terminal_summary(terminalreporter):
         dlist = dlist[:durations]
 
     for rep in dlist:
+        if verbose < 2 and rep.duration < 0.01:
```
asottile: should this take into account rounding (`< 0.005` or whatever?)

is `0.009` displayed as 0.01? (I don't know, this could be fine already!)

nicoddemus: It uses `02.2f` for formatting:

```
>>> '%02.2fs' % 0.009
'0.01s'
```

So I think this is fine as is. 👍

asottile: wouldn't that mean it isn't fine since it's not displayed as 0.0s?

nicoddemus: Oh duh you are right, sorry.

asottile: it's certainly an edge case of an edge case though  :laughing:

nicoddemus:
Thanks a lot @jeffreyrack, appreciate it!

PR #3124, opened 2018-01-16
Changes in the logging plugin for 3.4

nicoddemus:
This contains the changes discussed in #3013, except that I did not deprecate any of the configuration variables because I realized it would basically duplicate the number of options which would probably result in more confusion than benefits; we might decide to change/deprecate the options at a later time. 

Summary:

* We no longer modify the level of the root logger unless the user explicitly asks for it; This allows users to configure logging how they better see fit, but one of the effects of this is that by default only `WARNING` messages or above are captured (instead of all messages).
* Live logging is disabled by default;
* Live logging now writes to terminal instead of to `sys.stderr`, which no longer requires users to pass `-s` to see live logging;
* Live logging automatically sets the output to verbose so the logs don't get interleaved with the test progress output;
* Live logging now writes a header separator between the logs captured in each test phase (setup, call, teardown).

PR #4090, opened 2018-10-08
Improve performance of ~3 of the slowest tests

asottile:
cherry picked a few from `pytest testing/{subdir} --durations=5` and wanted to see what I could do to make them faster.

this saves ~11 seconds off of a test run (which isn't much but it's something!)

interesting to note that `runpytest_subprocess` adds about 500ms per invocation. (I guess we should avoid it if possible? @nicoddemus do you know if there's a reason why it was used for #3124?)

### DIFF
--- a/testing/code/test_excinfo.py
+++ b/testing/code/test_excinfo.py
@@ -1371,7 +1371,7 @@ def test_exception_repr_extraction_error_on_recursion():
     def b(x):
         return a(numpy_like())
 
-    with pytest.raises(RecursionError) as excinfo:
+    with pytest.raises(RuntimeError) as excinfo:
         a(numpy_like())
 
     matcher = LineMatcher(str(excinfo.getrepr()).splitlines())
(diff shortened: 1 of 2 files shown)


nicoddemus:
> @nicoddemus do you know if there's a reason why it was used for #3124?

I don't recall why, and looking at the test itself no clear reason jumps to me... if things are passing, I say let's keep your changes. 😁

testing/code/test_excinfo.py
```diff
@@ -1361,40 +1371,30 @@ def a(x):
     def b(x):
         return a(numpy_like())
 
-    try:
+    with pytest.raises(RecursionError) as excinfo:
```
nicoddemus: `RecursionError` is py35+ (or py36+) only; IIRC you need to use `RuntimeError` instead.

asottile: dang, I always forget this 😭

nicoddemus:
Noice! 👍

PR #4105, opened 2018-10-10
Fix trailing whitespace in FixtureDef.__repr__

blueyed:


src/_pytest/fixtures.py
```diff
@@ -899,7 +899,7 @@ def execute(self, request):
         return hook.pytest_fixture_setup(fixturedef=self, request=request)
 
     def __repr__(self):
-        return "<FixtureDef name=%r scope=%r baseid=%r >" % (
+        return "<FixtureDef name=%r scope=%r baseid=%r>" % (
             self.argname,
             self.scope,
             self.baseid,
```
blueyed: Might also have been good to replace "name" with "argname"?

nicoddemus: Hmm yes I think so

PR #4107, opened 2018-10-10
pytester: fix __tracebackhide__ for {re_,fn}match_lines

blueyed:


src/_pytest/pytester.py
```diff
@@ -1352,7 +1354,6 @@ def _match_lines(self, lines2, match_func, match_nickname):
         lines1 = self.lines[:]
         nextline = None
         extralines = []
-        __tracebackhide__ = True
```
nicoddemus: Shouldn't we keep this one?

blueyed: Fixed.

PR #4125, opened 2018-10-12
Run docs, doctesting and linting in the same job to save some CI time

nicoddemus:
Docs also checks the documentation for syntax failures, so it should
part of the baseline to bail out early.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,6 @@ env:
     - TOXENV=py36-numpy
     - TOXENV=py36-pluggymaster PYTEST_NO_COVERAGE=1
     - TOXENV=py27-nobyte
-    - TOXENV=doctesting
-    - TOXENV=docs PYTEST_NO_COVERAGE=1
 
 jobs:
   include:
@@ -61,7 +59,7 @@ jobs:
       env: TOXENV=py27
     - env: TOXENV=py34
     - env: TOXENV=py36
-    - env: TOXENV=linting PYTEST_NO_COVERAGE=1
+    - env: TOXENV=linting,docs,doctesting PYTEST_NO_COVERAGE=1
 
     - stage: deploy
       python: '3.6'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,6 @@
 environment:
   matrix:
-  - TOXENV: "linting"
+  - TOXENV: "linting,docs,doctesting"
     PYTEST_NO_COVERAGE: "1"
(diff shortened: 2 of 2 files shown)


nicoddemus:
The `[skip appveyor]` the commit message [prevents AppVeyor from triggering a build](https://www.appveyor.com/docs/how-to/filtering-commits/#skip-directive-in-commit-message), given that this PR only affects Travis anyway.

blueyed:
@nicoddemus 
With regard to builds: any idea why Travis is queuing two builds per PR?
And are you or somebody else manually canceling the duplicate then?  (which appears to affect the PR on Github then already)

.travis.yml
```diff
@@ -62,6 +61,7 @@ jobs:
     - env: TOXENV=py34
     - env: TOXENV=py36
     - env: TOXENV=linting PYTEST_NO_COVERAGE=1
+    - env: TOXENV=docs PYTEST_NO_COVERAGE=1
```
blueyed: We could also use "TOXENV=linting docs" here (single job).

nicoddemus: Ahh good idea, will do.

nicoddemus: Done 👍

nicoddemus:
> And are you or somebody else manually canceling the duplicate then?

I'm not doing anything, but what do you mean it is queuing two builds per PR? I only see a single build per PR in the dashboard: https://travis-ci.org/pytest-dev/pytest/pull_requests

We have configured Travis to cancel running jobs if new commits are pushed to a branch, but it shouldn't be triggering two builds per PR.

nicoddemus:
Hmmm you mean this?

image

blueyed:
Yes.

.travis.yml
```diff
@@ -61,7 +60,7 @@ jobs:
       env: TOXENV=py27
     - env: TOXENV=py34
     - env: TOXENV=py36
-    - env: TOXENV=linting PYTEST_NO_COVERAGE=1
+    - env: TOXENV=linting docs PYTEST_NO_COVERAGE=1
```
blueyed: I guess this needs quotes, or a list.

nicoddemus: Oops, fixed!

nicoddemus:
No idea what could be causing the duplicated jobs... 🤔

nicoddemus:
Decided to do the same thing for AppVeyor to avoid the cost of spinning a new VM for what is essentially the same job.

PR #4132, opened 2018-10-13
Do not print INTERNALERROR with --pdb

blueyed:
This gets printed by the terminal reporter already, and currently
results in the same error being displayed twice, e.g. when raising an
`Exception` manually from `pytest.debugging.pytest_exception_interact`.

Also fixes print statements in the tests.


blueyed:
Might be nice to have a test for this after all, but do not have time for it now.

changelog/4132.bugfix.rst
```diff
@@ -1 +1 @@
-Fix duplicate printing of internal errors when using `--pdb`.
+Fix duplicate printing of internal errors when using ``--pdb``.
```
blueyed: Thanks, will try to remember this to do right now.. :)

blueyed:
Added a test.

nicoddemus:
Fixed linting. 😉 

@blueyed did you install `pre-commit` so it auto-formats during each commit?

PR #4133, opened 2018-10-13
pdb: handle quitting in post_mortem

blueyed:
`help quit` in pdb says:

> Quit from the debugger. The program being executed is aborted.

But pytest would continue with the next tests, often making it necessary
to kill the pytest process when using `--pdb` and trying to cancel the
tests using `KeyboardInterrupt` / `Ctrl-C`.

TODO:

- [x] changelog
- [x] test?

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -1,10 +1,12 @@
 """ interactive debugging with PDB, the Python Debugger. """
 from __future__ import absolute_import, division, print_function
+
+import os
 import pdb
 import sys
-import os
 from doctest import UnexpectedException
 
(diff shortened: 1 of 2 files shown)


blueyed:
py3 displays it as "_pytest.outcomes.Exit: Quitting debugger", while py2 uses `Exit: Quitting debugger`.
Is this OK?

nicoddemus:
I think so, we can improve this later if need be

blueyed:
I need to touch it anyway again, so might as well add a `__repr__` or somethng for `Exit`.

nicoddemus:
Sounds good, thanks!

PR #4139, opened 2018-10-14
tox.ini: passenv: COVERAGE_*

blueyed:
This is required to pass through COVERAGE_PROCESS_START etc.


blueyed:
Thanks, but would have been very interesting to see the coverage changes via codecov.
Should show up on master now though, so it's OK.

PR #4140, opened 2018-10-14
Faster CI

blueyed:
- CI: run specialized factors in a single job
- AppVeyor: fast_finish

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,19 +12,15 @@ install:
   - pip install --upgrade --pre tox
 env:
   matrix:
-    # note: please use "tox --listenvs" to populate the build matrix below
-    # please remove the linting env in all cases
-    - TOXENV=py27-pexpect
-    - TOXENV=py27-xdist
-    - TOXENV=py27-trial
-    - TOXENV=py27-numpy
-    - TOXENV=py27-pluggymaster PYTEST_NO_COVERAGE=1
-    - TOXENV=py36-pexpect
-    - TOXENV=py36-xdist
-    - TOXENV=py36-trial
-    - TOXENV=py36-numpy
-    - TOXENV=py36-pluggymaster PYTEST_NO_COVERAGE=1
+    # Specialized factors for py27.
+    - TOXENV=py27-pexpect,py27-trial,py27-numpy
     - TOXENV=py27-nobyte
+    - TOXENV=py27-xdist
+    - TOXENV=py27-pluggymaster PYTEST_NO_COVERAGE=1
+    # Specialized factors for py36.
+    - TOXENV=py36-pexpect,py36-trial,py36-numpy
+    - TOXENV=py36-xdist
+    - TOXENV=py36-pluggymaster PYTEST_NO_COVERAGE=1
 
 jobs:
   include:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,12 @@
 environment:
   matrix:
   - TOXENV: "py27"
-  - TOXENV: "py34"
-  - TOXENV: "py35"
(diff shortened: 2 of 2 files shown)


appveyor.yml
```diff
@@ -1,27 +1,29 @@
...
+  - TOXENV: "py36-trial,py36-numpy"
   - TOXENV: "py36-pluggymaster"
     PYTEST_NO_COVERAGE: "1"
-  - TOXENV: "py27-nobyte"
   - TOXENV: "py36-freeze"
     PYTEST_NO_COVERAGE: "1"
+  - TOXENV: "py36-xdist"
+
+matrix:
+  fast_finish: true
```
nicoddemus: I'm not sure this is a good idea. If more than one environment will fail, I would like to get the full picture, otherwise we might get into a situation where you fix a job, then it fails on the next one, then you fix it only to discover it fails yet again...

.travis.yml
```diff
@@ -12,19 +12,15 @@ install:
   - pip install --upgrade --pre tox
 env:
   matrix:
-    # note: please use "tox --listenvs" to populate the build matrix below
-    # please remove the linting env in all cases
-    - TOXENV=py27-pexpect
+    # Specialized factors for py27.
```
nicoddemus: What was the reasoning here for plumping pexpect, trial, and numpy together, while keeping xdist and pluggymaster separate? I'm curious.

appveyor.yml
blueyed: ok.  Will remove the 2nd commit then.

.travis.yml
blueyed: xdist and pluggymaster run the whole test suite, while the others only run parts of it.

Initially I've thought to just install pexpect on CI by default, but then it would slow more jobs down.

Re pexpect I think it could even use a marker or similar to really run only pexpect tests maybe even.

appveyor.yml
blueyed: OTOH you will get the full picture in Travis typically, and AppVeyor is more problematic with regard to build queue length.

After all I would even run less jobs altogether on AppVeyor (e.g. only py27 and py36 or py37, not py34 etc).  Maybe that could be configured for PRs?  (and still run everything on merges to master)

nicoddemus: > OTOH you will get the full picture in Travis typically, and AppVeyor is more problematic with regard to build queue length.

Typically yes, you are right. Well let's leave it like this, and see how it goes. Please just add a comment explaining the rationale for future reference, I think:

```yaml
matrix:
  # stop build if a job fails: usually we get the full picture from
  # travis anyway, and AppVeyor is problematic regarding build queue length
  # we can revert this if this proves problematic
  fast_finish: true
```

What do you think?

> Maybe that could be configured for PRs? (and still run everything on merges to master)

I'm not a big fan on running different sets of jobs in PRs vs master: this only leads to situation of a green PR breaking `master` because the PR did not run all jobs.

.travis.yml
blueyed: > Please add a comment so the rationale is explicit.

"Specialized factors for py27." was meant to be for that.  But maybe put it on the same line then?  Better wording?

appveyor.yml
blueyed: Sounds good, will add the comment.

.travis.yml
nicoddemus: "Specialized tests for py27"? Not sure, the wording might be fine and it was my fault for not understanding it right away hehehe

nicoddemus:
Let's go ahead, no sense in keeping this waiting in account of a single comment. 👍

PR #4141, opened 2018-10-14
tox.ini: fix testpaths

blueyed:
Broken only recently in 307fa7a4 (features).

[ci skip] (since it apparently is not tested/used there)


blueyed:
Maybe it is worth / a good idea to not include "doc" there after all?
It will run first, and takes quite a bit after all.
It would still run on CI, of course.

nicoddemus:
> Maybe it is worth / a good idea to not include "doc" there after all?

Agree, better to rely on CI otherwise it will take up time for all envs.

blueyed:
Cool, amended.

PR #4144, opened 2018-10-14
Fix flaky durations test

nicoddemus:
Unfortunately due to fluctuations in runtime "test_something"
might still appear in the final message.

Example failure:

https://ci.appveyor.com/project/pytestbot/pytest/builds/19494829/job/8lx847u0c78m63wf

### DIFF
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -808,7 +808,6 @@ class TestDurations(object):
         result.stdout.fnmatch_lines_random(
             ["*durations*", "*call*test_3*", "*call*test_2*"]
         )
-        assert "test_something" not in result.stdout.str()
         result.stdout.fnmatch_lines(
             ["(0.00 durations hidden.  Use -vv to show these durations.)"]
         )
--- a/testing/python/fixture.py
+++ b/testing/python/fixture.py
@@ -494,6 +494,12 @@ class TestRequestBasic(object):
         reason="this method of test doesn't work on pypy",
     )
     def test_request_garbage(self, testdir):
+        try:
+            import xdist  # noqa
+        except ImportError:
+            pass
+        else:
+            pytest.xfail("this test is flaky when executed with xdist")
         testdir.makepyfile(
             """
             import sys


nicoddemus:
Here's the output from the failing test:

```
========================== slowest 10 test durations ==========================
0.03s call     test_calls.py::test_3
0.02s call     test_calls.py::test_2
0.02s call     test_calls.py::test_1
0.02s teardown test_calls.py::test_something

(0.00 durations hidden.  Use -vv to show these durations.)
========================== 4 passed in 0.09 seconds ===========================
```

> Why is that? I mean, how does it appear there then? Is it taking a little bit long enough?

Yep, it seems that due to fluctuations on CI time sometimes `test_something` setup, call, or teardown might take more than `0.01`s. I've seen this just yesterday as well during "setup" of `test_something`. So we really cannot rely on that, I'm afraid.

In theory all calls might take more than `0.01`s and the "durations hidden" message won't appear (and the test will fails), but let's see that happen first before also removing that check (or hacking around it).

blueyed:
Looks like there is another flaky one?!
https://travis-ci.org/pytest-dev/pytest/jobs/441305926#L545

```
E       Failed: nomatch: '* 1 passed in *'
E           and: u'============================= test session starts =============================='
E           and: u'platform linux2 -- Python 2.7.6, pytest-3.8.3.dev69+g7ea1efd, py-1.7.0, pluggy-0.7.1'
E           and: u"hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/travis/build/pytest-dev/pytest/.hypothesis/examples')"
E           and: u'rootdir: /tmp/pytest-of-travis/pytest-0/popen-gw37/test_request_garbage0, inifile:'
E           and: u'plugins: xdist-1.23.2, forked-0.2, hypothesis-3.76.0'
E           and: u'collected 1 item'
E           and: u''
E           and: u'test_request_garbage.py .E                                               [100%]'
E           and: u''
E           and: u'==================================== ERRORS ===================================='
E           and: u'________________________ ERROR at teardown of test_func ________________________'
E           and: u''
E           and: u"request = <SubRequest 'something' for <Function 'test_func'>>"
E           and: u''
E           and: u'    @pytest.fixture(autouse=True)'
E           and: u'    def something(request):'
E           and: u'        original = gc.get_debug()'
E           and: u'        gc.set_debug(gc.DEBUG_SAVEALL)'
E           and: u'        gc.collect()'
E           and: u'    '
E           and: u'        yield'
E           and: u'    '
E           and: u'        try:'
E           and: u'            gc.collect()'
E           and: u'            leaked_types = sum(1 for _ in gc.garbage'
E           and: u'                               if isinstance(_, PseudoFixtureDef))'
E           and: u'    '
E           and: u'            # debug leaked types if the test fails'
E           and: u'            print(leaked_types)'
E           and: u'    '
E           and: u'            gc.garbage[:] = []'
E           and: u'    '
E           and: u'>           assert leaked_types == 0'
E           and: u'E           assert 2 == 0'
E           and: u''
E           and: u'test_request_garbage.py:24: AssertionError'
E           and: u'--------------------------- Captured stdout teardown ---------------------------'
E           and: u'2'
E           and: u'====================== 1 passed, 1 error in 4.12 seconds ======================='
E           and: u''
E       remains unmatched: '* 1 passed in *'
…
FAIL testing/python/fixture.py::TestRequestBasic::()::test_request_garbage
```

nicoddemus:
Yeah, that one fails sometimes when running in xdist...

I will skip that test when we are running under xdist then. 👍

Issue #4098, opened 2018-10-09
Add a returncode argument to pytest.exit


nicoddemus:
As discussed in https://github.com/pytest-dev/pytest/issues/3574#issuecomment-396893524, `pytest.exit()` could receive a `returncode` argument to cleanly exit pytest.

BeyondEvil:
If my silly mind will start functioning again after my vacation, I'll take a stab at this...

Now, if I could only set a reminder... 🤔

nicoddemus:
Thanks @BeyondEvil, sounds good! 😁

Zac-HD:
Closed by #4145.

FYI @cacoze [this article](https://help.github.com/articles/closing-issues-using-keywords/) describes the keywords that GitHub recognises to automatically close issues.

nicoddemus:
Awesome work @Zac-HD on triaging all those issues! 👍

BeyondEvil:
Dammit, cacoze beat me to it.

Ah well, you snooze you loose, I suppose. 😊

PR #4145, opened 2018-10-14
Add returncode argument to pytest.exit

cacoze:
Solves #4098

### DIFF
--- a/changelog/4098.feature.rst
+++ b/changelog/4098.feature.rst
@@ -1,1 +1,1 @@
-Add to pytest.exit a returncode argument to cleanly exit pytest.
+Add returncode argument to pytest.exit() to exit pytest with a specific return code.
--- a/changelog/4098.feature.rst
+++ b/changelog/4098.feature.rst
@@ -1,0 +1,1 @@
+Add to pytest.exit a returncode argument to cleanly exit pytest.
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -185,10 +185,12 @@ def wrap_session(config, doit):
             session.exitstatus = EXIT_TESTSFAILED
         except KeyboardInterrupt:
             excinfo = _pytest._code.ExceptionInfo()
-            if initstate < 2 and isinstance(excinfo.value, exit.Exception):
+            exitstatus = EXIT_INTERRUPTED
+            if initstate <= 2 and isinstance(excinfo.value, exit.Exception):
                 sys.stderr.write("{}: {}\n".format(excinfo.typename, excinfo.value.msg))
+                exitstatus = excinfo.value.returncode
             config.hook.pytest_keyboard_interrupt(excinfo=excinfo)
-            session.exitstatus = EXIT_INTERRUPTED
+            session.exitstatus = exitstatus
         except:  # noqa
             excinfo = _pytest._code.ExceptionInfo()
             config.notify_exception(excinfo, config.option)
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -188,7 +188,8 @@ def wrap_session(config, doit):
             exitstatus = EXIT_INTERRUPTED
             if initstate <= 2 and isinstance(excinfo.value, exit.Exception):
                 sys.stderr.write("{}: {}\n".format(excinfo.typename, excinfo.value.msg))
-                exitstatus = excinfo.value.returncode
+                if excinfo.value.returncode is not None:
+                    exitstatus = excinfo.value.returncode
             config.hook.pytest_keyboard_interrupt(excinfo=excinfo)
             session.exitstatus = exitstatus
         except:  # noqa
--- a/src/_pytest/outcomes.py
+++ b/src/_pytest/outcomes.py
@@ -57,10 +57,13 @@ class Exit(KeyboardInterrupt):
 # exposed helper methods
 
 
-def exit(msg):
+def exit(msg, returncode=None):
     """ exit testing process as if KeyboardInterrupt was triggered. """
     __tracebackhide__ = True
-    raise Exit(msg)
+    if returncode:
+        raise SystemExit(returncode)
+    else:
+        raise Exit(msg)
 
 
 exit.Exception = Exit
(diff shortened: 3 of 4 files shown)


src/_pytest/outcomes.py
```diff
@@ -57,10 +57,13 @@ def __init__(self, msg="unknown reason"):
...
 
-def exit(msg):
+def exit(msg, returncode=None):
     """ exit testing process as if KeyboardInterrupt was triggered. """
     __tracebackhide__ = True
-    raise Exit(msg)
+    if returncode:
+        raise SystemExit(returncode)
+    else:
+        raise Exit(msg)
```
nicoddemus: I think we should always raise `Exit`, but pass the argument along to it: `Exit(msg, returncode)`, and treat it accordingly in `main.py`:

https://github.com/pytest-dev/pytest/blob/7e1fac5f9197998e93a836b6ea4393429cb09147/src/_pytest/main.py#L185-L188

Like this:

```python
existstatus = EXIT_INTERRUPTED
if initstate < 2 and isinstance(excinfo.value, exit.Exception):
    sys.stderr.write("{}: {}\n".format(excinfo.typename, excinfo.value.msg))
    exitstatus = excinfo.value.exitstatus
config.hook.pytest_keyboard_interrupt(excinfo=excinfo)
session.exitstatus = exitstatus
```

testing/test_runner.py
```diff
@@ -570,6 +570,15 @@ def pytest_configure(config):
     result.stderr.fnmatch_lines(["Exit: oh noes"])
 
 
+def test_pytest_exit_returncode():
```
nicoddemus: Following my suggestion above, this will need to be changed to a test using `testdir`, something like:

```python
def test_pytest_exit_returncode(testdir):
    testdir.makepyfile("""
        import pytest
        def test_foo():
            pytest.exit("some exit msg", 99)
    """)
    result = testdir.runpytest()
    assert result.ret == 99
```

changelog/4098.feature.rst
```diff
@@ -0,0 +1 @@
+Add to pytest.exit a returncode argument to cleanly exit pytest.
```
nicoddemus: I think:

> Add ``returncode`` argument to ``pytest.exit()`` to exit pytest with a specific return code.

src/_pytest/outcomes.py
nicoddemus: Also, please update the docstring for ``pytest.exit``

src/_pytest/main.py
```diff
@@ -185,10 +185,12 @@ def wrap_session(config, doit):
             session.exitstatus = EXIT_TESTSFAILED
         except KeyboardInterrupt:
             excinfo = _pytest._code.ExceptionInfo()
-            if initstate < 2 and isinstance(excinfo.value, exit.Exception):
+            exitstatus = EXIT_INTERRUPTED
+            if initstate <= 2 and isinstance(excinfo.value, exit.Exception):
```
cacoze: I had to change the the `<` to `<=` to make the statement true. Can you tell me why @nicoddemus ?

src/_pytest/outcomes.py
```diff
@@ -49,18 +49,23 @@ class Failed(OutcomeException):
...
+        self.returncode = returncode
         KeyboardInterrupt.__init__(self, msg)
 
 
 # exposed helper methods
 
 
-def exit(msg):
-    """ exit testing process as if KeyboardInterrupt was triggered. """
+def exit(msg, returncode=None):
```
cacoze: Do you suggest a different default value @nicoddemus ? I thought in using 1 or importing `EXIT_INTERRUPTED` from _pytest/main.py but I wasn't sure.

cacoze:
Thanks for the feedback @nicoddemus . I'm still learning about the project and had no idea that I could use _pytest/main.py to set the return code from the test.

src/_pytest/main.py
nicoddemus: Hmm not really, it should not affect this at all...

src/_pytest/outcomes.py
```diff
@@ -49,18 +49,23 @@ class Failed(OutcomeException):
 class Exit(KeyboardInterrupt):
     """ raised for immediate program exits (no tracebacks/summaries)"""
 
-    def __init__(self, msg="unknown reason"):
+    def __init__(self, returncode=None, msg="unknown reason"):
```
nicoddemus: Please add `returncode` to the end of the parameter list to keep backward compatibility; technically users should not raise this explicitly, but it doesn't hurt to keep compatibility in this case.

src/_pytest/outcomes.py
nicoddemus: Let's keep `None` and handle that in `main`. 👍

src/_pytest/main.py
nicoddemus: I suggest we handle `None` here like this:

```python
            exitstatus = EXIT_INTERRUPTED
            if initstate <= 2 and isinstance(excinfo.value, exit.Exception):
                sys.stderr.write("{}: {}\n".format(excinfo.typename, excinfo.value.msg))
                if excinfo.value.returncode is not None:
                    exitstatus = excinfo.value.returncode
```

src/_pytest/outcomes.py
```diff
@@ -49,18 +49,23 @@ class Failed(OutcomeException):
...
 # exposed helper methods
 
 
-def exit(msg):
-    """ exit testing process as if KeyboardInterrupt was triggered. """
+def exit(msg, returncode=None):
+    """
+    Exit testing process as if KeyboardInterrupt was triggered.
+
+    :param int returncode: return code to be used when exiting pytest..
```
nicoddemus: While we're at it, please add a description for the `msg` parameter as well:

```
:param str msg: message to display upon exit.
```

nicoddemus: Also there's an extra period at the end of this sentence.

blueyed:
Just a btw from reading commits on master: squashing this would have made a lot of sense.

PR #4147, opened 2018-10-14
Stepwise

davidszotten:
add `--stepwise` from `pytest-stepwise` as per discussion in https://github.com/nip3o/pytest-stepwise/issues/14

not sure if this is the best place for the docs (or if they are clear enough), thoughts most welcome

almost all code by @nip3o


src/_pytest/stepwise.py
```diff
@@ -0,0 +1,91 @@
+from _pytest.cacheprovider import Cache
+import pytest
+
+
+def pytest_addoption(parser):
+    group = parser.getgroup('general')
+    group.addoption('--sw', '--stepwise', action='store_true', dest='stepwise',
```
nicoddemus: This will fail the linting check, you can either:

1. install `pre-commit` and execute it:

  ```
  $ pre-commit install
  $ pre-commit run -a
  ```

2. Execute `tox -e linting`.

I recommend 1) because from that point on it will reformat/check every commit.

Commit and push the results. 👍

nicoddemus: I think this will cause an option conflict if `pytest-stepwise` is installed.

We can probably deactivate the plugin and emit a warning recommending for users to uninstall pytest-stepwise and use the builtin functionality instead.

@RonnyPfannschmidt might have another suggestion on how to cope/handle this.

changelog/xxx.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``-sw``, ``--stepwise`` as an alternative to ``--lf -x`` for stopping at the first failure, but starting the next test invocation from that test.  See `the documentation <https://docs.pytest.org/en/latest/cache.html#stepwise>`_ for more info.
```
nicoddemus: Great changelog! 

Make sure to rename it to `4147.feature.rst` now that we have a PR number. 👍

src/_pytest/stepwise.py
davidszotten: actually, my fix for that plugin got merged, but not released so it already doesn't work with current pytest versions (raises a typeerror on launch even if the flags aren't set)

davidszotten:
@nicoddemus think i've addressed all the comments. i'd welcome thoughts on the command line arguments (eg `stepwise-skip` vs `skip`) and the docs

thanks!

src/_pytest/stepwise.py
nicoddemus: Yeah, just realized that:

```
  File "c:\pytest\.env36-sw\lib\site-packages\pytest_stepwise\plugin.py", line 16, in pytest_configure
    config.cache = Cache(config)
TypeError: __init__() missing 1 required positional argument: 'config'
```

So I think we can leave as is if pytest-stepwise doesn't plan any further releases.

src/_pytest/stepwise.py
```diff
@@ -0,0 +1,104 @@
...
+        "--stepwise-skip",
+        action="store_true",
+        dest="stepwise_skip",
+        help="ignore the first failing test but stop on the next failing test",
+    )
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_configure(config):
+    config.cache = Cache.for_config(config)
```
nicoddemus: We need to remove this line because `cacheprovider` already creates the cache for us:

https://github.com/pytest-dev/pytest/blob/f7285b6ab2222e29d80ad1501a301c250db444e0/src/_pytest/cacheprovider.py#L294-L297

We should also remove the `tryfirst=True` option from the hook implementation so `cacheprovider` executes first.

davidszotten:
hm, i can't reproduce the error from travis locally. any advice?

nicoddemus:
> hm, i can't reproduce the error from travis locally. any advice?

Don't worry, that's unrelated and will be fixed by https://github.com/pytest-dev/pytest/pull/4186. 👍 

@RonnyPfannschmidt I would love a second set of eyes before we merge this. 😁

davidszotten:
@nip3o note the comment about _not_ releasing a new version with my fix in `pytest-stepwise` so we don't have to worry about clashes (https://github.com/pytest-dev/pytest/pull/4147#discussion_r225427985)

PR #4173, opened 2018-10-17
Update warnings docs

nicoddemus:
Some points on the document work different in 3.9, plus changed the order
of the sections a bit to make more sense for users reading it for the first time.


asottile:
coveralls failure appears to be https://github.com/urllib3/urllib3/issues/1456

I wonder why it only bit one of the jobs though 🤔

PR #4182, opened 2018-10-17
Fix hook name in deprecations.rst

nicoddemus:
As mentioned in https://github.com/pytest-dev/pytest/issues/4128#issuecomment-430690498


nicoddemus:
Failures are unrelated. :/

PR #4195, opened 2018-10-18
Use safe_str() to format warning message about unicode in Python 2

### LINKED ISSUE #3691: MemoryError using custom warning formatter in python 2.7
- pytest master
- python 2.7
- Windows 10

Running the following test makes pytest hang. Python uses the whole PC memory and eventually it exists with `MemoryError`:

```
# -*- coding: utf8 -*-
from __future__ import unicode_literals
import warnings
import pytest


def _custom_showwarning(message, *a, **b):
    return "WARNING: {}".format(message)

warnings.formatwarning = _custom_showwarning

@pytest.mark.filterwarnings("default")
def test_custom_warning_formatter():
    warnings.warn("¥")
```

Maybe I am doing something wrong here. This problems does not happen with python 3.6.

nicoddemus:
Fix #3691

### DIFF
--- a/changelog/3691.bugfix.rst
+++ b/changelog/3691.bugfix.rst
@@ -1,0 +1,2 @@
+Python 2: safely format warning message about passing unicode strings to ``warnings.warn``, which may cause
+surprising ``MemoryError`` exception when monkey patching ``warnings.warn`` itself.
--- a/src/_pytest/warnings.py
+++ b/src/_pytest/warnings.py
@@ -123,7 +123,7 @@ def warning_record_to_str(warning_message):
     if unicode_warning:
         warnings.warn(
             "Warning is using unicode non convertible to ascii, "
-            "converting to a safe representation:\n  %s" % msg,
+            "converting to a safe representation:\n  {!r}".format(compat.safe_str(msg)),
             UnicodeWarning,
         )
     return msg
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,7 @@ envlist =
 
 [testenv]
 commands =
-    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest --lsof
+    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest --lsof {posargs}
     coverage: coverage combine
     coverage: coverage report
 passenv = USER USERNAME COVERAGE_* TRAVIS
@@ -41,7 +41,7 @@ deps =
     py27: mock
     nose
 commands =
-    pytest -n auto --runpytest=subprocess
+    pytest -n auto --runpytest=subprocess {posargs}
 
 
 [testenv:linting]
@@ -58,7 +58,7 @@ deps =
     hypothesis>=3.56
     {env:_PYTEST_TOX_EXTRA_DEP:}
 commands =
-    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest -n auto
+    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest -n auto {posargs}
 
 [testenv:py36-xdist]
 # NOTE: copied from above due to https://github.com/tox-dev/tox/issues/706.
--- a/testing/test_warnings.py
+++ b/testing/test_warnings.py
@@ -3,6 +3,8 @@ from __future__ import unicode_literals
 
 import sys
 
+import six
+
 import pytest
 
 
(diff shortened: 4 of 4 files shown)


src/_pytest/warnings.py
```diff
@@ -123,7 +123,7 @@ def warning_record_to_str(warning_message):
     if unicode_warning:
         warnings.warn(
             "Warning is using unicode non convertible to ascii, "
-            "converting to a safe representation:\n  %s" % msg,
+            "converting to a safe representation:\n  %s" % compat.safe_str(msg),
```
asottile: I'd probably use `{!r}` :man_shrugging:

asottile: hah -- I suggested this too.  at least I'm consistent: https://github.com/pytest-dev/pytest/issues/3691#issuecomment-405766850

nicoddemus:
What is happening with Travis that it is creating two builds for each PR, cancelling one of them immediately?

image

asottile:
> 
> 
> What is happening with Travis that it is creating two builds for each PR, cancelling one of them immediately?
> 
> image

idk, I've been seeing this only for pytest for the last couple of weeks.  Sometimes it cancels the wrong build too and then you have to manually restart it 🤦‍♂️

nicoddemus:
Fixed the Travis issue, apparently 🎉  (#4197 and #4198)

asottile:
nice! what was the trick? (also fyi kinda neat feature of git: `git commit --allow-empty ...`)

tox.ini
```diff
@@ -18,7 +18,7 @@ envlist =
 
 [testenv]
 commands =
-    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest --lsof
+    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest --lsof {posargs}
```
Zac-HD: Why do we need `{posargs}` here?  Left over from local testing?

changelog/3691.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Python 2: safely format warning message about passing unicode strings to ``warnings.warn``, which may cause
+surprising ``MemoryError`` exception when monkey patching ``warnings.warn`` itself.
```
Zac-HD: *(comment for my own education about pytest's documentation)*

In Hypothesis we have the [`intersphinx_mapping`](https://github.com/HypothesisWorks/hypothesis/blob/192716ef679864e8685a6d759e609d70edfb25b7/hypothesis-python/docs/conf.py#L68-L75) configured so that we can write these as:

    surprising :class:`python:MemoryError` exception when monkey 
    patching :func:`python:warnings.warn` itself.

It allows for somewhat nicer formatting, but more importantly adds a hyperlink - and can check that the link target exists, which caught dozens of typos in our docs.

Has pytest decided not to do this, or just not considered it?  It can be fiddly at times but also pretty useful.

tox.ini
RonnyPfannschmidt: @Zac-HD without posargs its impossible to run specific tests in calls like `tox -- testing/test_some.py` since it wouldnt pass on

changelog/3691.bugfix.rst
RonnyPfannschmidt: i don't recall use taking a deeper look into this so it may be worth trying it - ideally someone with experience setting it up

nicoddemus: @Zac-HD thanks a lot for the pointer, actually or `intersphinx_mapping` is already configured. I tested locally your changes and they actually do work. 😁 

We should remember to use those sphinx references more often. 👍

nicoddemus: Hmm actually I spoke too soon. While it generates the proper docs, our linter complains:

```
ERROR changelog/3691.bugfix.rst:1 Unknown interpreted text role "func".
ERROR changelog/3691.bugfix.rst:1 Unknown interpreted text role "class".
ERROR changelog/3691.bugfix.rst:1 Unknown interpreted text role "func".
ERROR changelog/4192.bugfix.rst:1 Unknown interpreted text role "func".
ERROR changelog/4192.bugfix.rst:1 Unknown interpreted text role "func".
```

src/_pytest/warnings.py
```diff
@@ -123,7 +123,7 @@ def warning_record_to_str(warning_message):
     if unicode_warning:
         warnings.warn(
             "Warning is using unicode non convertible to ascii, "
-            "converting to a safe representation:\n  %s" % msg,
+            "converting to a safe representation:\n  {!r}".format(compat.safe_str(msg)),
```
asottile: oh, when I suggested {!r}, then you don't need `safe_str` at all I think

nicoddemus: Oh. Hmm well not sure if this will be a problem as is.

asottile: No problem, just I was hoping to eventually remove safe_str usages :man_shrugging:

PR #4203, opened 2018-10-19
[WIP] Track visited files and directories when collecting

### LINKED ISSUE #624: recursive symlink significantly slows collection
Originally reported by: **Brian Kearns (BitBucket: [bdkearns](http://bitbucket.org/bdkearns), GitHub: [bdkearns](http://github.com/bdkearns))**

---

placing "ln -s . link" in a collected dir significantly slows down collection. it seems collection doesn't keep track of directories visited to prevent unnecessary recursion?

---
- Bitbucket: https://bitbucket.org/pytest-dev/pytest/issue/624

wjt:
There are two parts to the fix:

* Don't visit directories which have already been visited. This fixes the exponential aspect of the bug.
* ~~Don't visit files which have already been visited. This fixes the more minor problem that, without this additional change, test_noop would be run twice, once as test_noop.py and once as symlink-0/test_noop.py.~~ I decided against this because without more refactoring, it broke `--keep-duplicates`.

Fixes #624.

I also included a fix for tox.ini that makes the instructions in CONTRIBUTING.rst for how to run just a single test work; happy to split that to a separate branch if preferred.

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -392,6 +392,8 @@ class Session(nodes.FSCollector):
         self._initialpaths = frozenset()
         # Keep track of any collected nodes in here, so we don't duplicate fixtures
         self._node_cache = {}
+        # Keep track of visited directories in here, so we don't end up in a symlink-induced loop.
+        self._visited = set()
 
         self.config.pluginmanager.register(self, name="session")
 
@@ -558,7 +560,17 @@ class Session(nodes.FSCollector):
                 return ()
         return ihook.pytest_collect_file(path=path, parent=self)
 
+    def _check_visited(self, path):
+        st = path.stat()
+        key = (st.dev, st.ino)
+        if key in self._visited:
+            return True
+        self._visited.add(key)
+        return False
+
     def _recurse(self, path):
+        if self._check_visited(path):
+            return False
         ihook = self.gethookproxy(path.dirpath())
         if ihook.pytest_ignore_collect(path=path, config=self.config):
             return
--- a/AUTHORS
+++ b/AUTHORS
@@ -223,6 +223,7 @@ Virgil Dupras
 Vitaly Lashmanov
 Vlad Dragos
 Wil Cooley
+Will Thompson
 William Lee
 Wim Glenn
 Wouter van Ackooy
--- a/testing/examples/test_issue624.py
+++ b/testing/examples/test_issue624.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function
+import sys
+
+import six
+
+import py
+import pytest
+
+
+@pytest.mark.skipif(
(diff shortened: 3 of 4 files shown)


wjt:
This may seem a bit far-fetched but I did actually hit this in practice (albeit while writing a test for exactly the same bug in a project tested with pytest https://github.com/gcovr/gcovr/pull/284 but I did hit *that* bug in practice!).

I see that `test_cmdline_python_package_symlink` has some special handling to test that symlinks are supported before running the test – I guess I need the same here, for Windows' benefit. (I don't have a Windows development system at the moment, but I was hoping to see a Windows CI run when I pushed this PR…)

wjt:
I'm told that, on Python 2.7 on Windows, `st_dev` and `st_ino` are always `0` which probably explains why this works so poorly! I guess I'll need to make the cycle-checking conditional somehow.

src/_pytest/main.py
```diff
@@ -558,7 +560,17 @@ def _collectfile(self, path):
                 return ()
         return ihook.pytest_collect_file(path=path, parent=self)
 
+    def _check_visited(self, path):
+        st = path.stat()
```
blueyed: @boxed will not like that.. ;)

You might be interested in https://github.com/pytest-dev/pytest/pull/4237 and the discussion at https://github.com/pytest-dev/pytest/issues/2206.

There is also `_recurse` in the python plugin.

I've also added `seen_dirs` in #4237, and wonder if a combination of `realpath` and this would be better maybe?

wjt: If you think `stat()` (a single syscall) is costly, why would `realpath()` be any better? (Its implementation is, roughly, split the path on the path separator, and call `os.lstat()` on each component.)

wjt: Rather annoyingly, `os.DirEntry` (yielded by `os.scandir()`) includes the inode number but not the device number.

boxed: It was a joke I think. I have been trying to cut down on stat() calls because pytest has made millions of them in some rather simple test scenarios. A single stat() call here and there won't be a big deal. The problem is that there has been very many "just a single" things in pytest from 3.4 to 3.9 and some of those weren't really "single" because they were used in a loop (in a loop!).

So I don't know about this case, but you could try running my test script https://github.com/pytest-dev/pytest/issues/2206#issuecomment-433006348 and see if performance is impacted significantly or not.

nicoddemus:
I think this should target `features` for the same reasons as mentioned in https://github.com/pytest-dev/pytest/pull/4237#issuecomment-433402651. 👍

src/_pytest/main.py
blueyed: Yeah, I was joking.
But I also think that since we're doing `realpath` already for symlink-resolving it might not be necessary to do any `stat` anymore on top.

I think this should be rebased on features, and then maybe only small changes are required to make the test pass.

nicoddemus:
Hi @wjt, 

Issue #4174, opened 2018-10-17
Pytest does not follow symlinks correctly when passed full path and tries to import conftest.py twice


bertjwregeer:
This is an issue with the latest version of pytest, that didn't happen with pytest 3.8.2, likely related to the change at https://github.com/pytest-dev/pytest/pull/4108

Here's the setup:

We set up a symlink for Jenkins that points the workspace to a new folder on different mount:

```/jenkins/workspace -> /scratch0/workspace```

According to Jenkins for a job, `/jenkins/workspace/job-name/` is the `$WORKSPACE`.

We clone a project using pytest into `/jenkins/workspace/job-name/pytest_example/`, it looks like this:

```
.
./conftest.py  # contains: collect_ignore = ['ignore']
./ignore
./ignore/test_stuff.py
./tests
./tests/test_something.py
```

We then run (while in `/jenkins/workspace/job-name/`):

```
pytest $WORKSPACE/pytest_example/ignore/
```

This will blow up stating it is attempting to load a plugin again, here's an example from my local laptop, `temp2` is a symlink to `temp`:

```
(testing) Riley:temp2 xistence$ pytest /Users/xistence/Projects/temp2/pytest_example/ignore/
Test session starts (platform: darwin, Python 3.7.0, pytest 3.9.1, pytest-sugar 0.9.1)
rootdir: /Users/xistence/Projects, inifile:
plugins: sugar-0.9.1, flake8-1.0.2, cov-2.6.0, backports.unittest-mock-1.4

―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ERROR collecting  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
../../.ve/testing/lib/python3.7/site-packages/_pytest/config/__init__.py:388: in _getconftestmodules
    return self._path2confmods[directory]
E   KeyError: local('/Users/xistence/Projects/temp/pytest_example/ignore')

During handling of the above exception, another exception occurred:
../../.ve/testing/lib/python3.7/site-packages/_pytest/config/__init__.py:416: in _importconftest
    return self._conftestpath2mod[conftestpath]
E   KeyError: local('/Users/xistence/Projects/temp/pytest_example/conftest.py')

[...]

PR #4214, opened 2018-10-22
Fix "Plugin already registered" error with symlinks

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4174.


src/_pytest/config/__init__.py
```diff
@@ -380,6 +380,7 @@ def _getconftestmodules(self, path):
         if self._noconftest:
             return []
 
+        path = path.realpath()
```
blueyed: I will move this down to after the cache check.

PR #4217, opened 2018-10-23
Fix MacOS py37 on Travis

nicoddemus:
This has been failing as of 2018-10-23 while installing gcc with
this message:

```
	==> Installing numpy dependency: gcc
	==> Downloading https://homebrew.bintray.com/bottles/gcc-8.2.0.high_sierra.bottl
	######################################################################## 100.0%
	==> Pouring gcc-8.2.0.high_sierra.bottle.1.tar.gz
	Error: The `brew link` step did not complete successfully
	The formula built, but is not symlinked into /usr/local
	Could not symlink include/c++
	Target /usr/local/include/c++
	already exists. You may want to remove it:
	  rm '/usr/local/include/c++'
	To force the link and overwrite all conflicting files:
	  brew link --overwrite gcc
	To list all files that would be deleted:
	  brew link --overwrite --dry-run gcc
	Possible conflicting files are:
	/usr/local/include/c++ -> /usr/local/Caskroom/oclint/0.13.1,17.4.0/oclint-0.13.1/include/c++

```

Example failure: https://travis-ci.org/pytest-dev/pytest/jobs/444989259

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,11 @@ jobs:
       env: TOXENV=py37
       before_install:
         - brew update
+        # remove c++ include files because upgrading python as of 2018-10-23, also
+        # attempts to upgrade gcc, and it fails because the include files already
+        # exist. removing the include files is one of the solutions recommended by brew
+        # this workaround might not be necessary in the future
+        - rm '/usr/local/include/c++'
         - brew upgrade python
         - brew unlink python
         - brew link python


.travis.yml
```diff
@@ -47,15 +47,21 @@ jobs:
       env: TOXENV=py37
       before_install:
         - brew update
+        # update gcc first because it was breaking the py37 update as of 2018-10-23;
+        # might not be necessary in the future
+        - rm '/usr/local/include/c++'
```
nicoddemus: Following what `brew` recommends in its error message:

```
	Could not symlink include/c++
	Target /usr/local/include/c++
	already exists. You may want to remove it:
	  rm '/usr/local/include/c++'
```

@jaraco does this seem correct to you? Any other suggestion?

nicoddemus: Oh sorry, I'm mis-remembering, who contributed the macos builds was not @jaraco. 

@jirikuncar what do you think?

blueyed: Is this symlink in the Travis image already?  You might want to `ls -l` it in the beginning.

blueyed: This happens when also upgrading `numpy` (https://travis-ci.org/pytest-dev/pytest/jobs/444989259#L933), which maybe can be avoided with some option to "brew upgrade python" to not upgrade dependents?  (should be faster then, too)

blueyed:
Pushed a commit that uses `brew upgrade --ignore-dependencies python`, feel free to revert/amend.

nicoddemus:
Thanks for chipping in @blueyed. :+1:

Let's see what @jirikuncar has to say.

nicoddemus:
Well let's get this merged, as having every PR failing because of this is distracting.

nicoddemus:
(We will need at least 1 approval before we can merge this btw)

PR #4220, opened 2018-10-23
Add `empty_parameter_set_mark` ini option documentation to the parametrize doc

anpr:
This is my first try of a pull request, and I tried to address issue #3851.

I didn't find out how to add a link to the reference of `empty_parameter_set_mark` (using `:ref:`), so feel free to edit.


blueyed:
Amended with a reference link.

IMHO it does not really warrant a changelog entry (just adds noise when grepping etc), but have left it.

PR #4224, opened 2018-10-24
_getconftestmodules: avoid isfile()/dirpath()

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/2206#issuecomment-432623646

I cannot see a real difference myself with clock time (1.2s vs 1.15s with some project and `-k doesnotmatch`), but I think it makes sense anyway, and might help with the above comment.

For pytest itself:
new: 1 loop, best of 5: 1.73 sec per loop
old: 1 loop, best of 5: 1.61 sec per loop

However, pytest itself displays "2287 deselected in 1.06 seconds" vs "2287 deselected in 1.22 seconds".

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -387,13 +387,13 @@ class PytestPluginManager(PluginManager):
         if self._noconftest:
             return []
 
-        if path.isfile():
-            directory = path.dirpath()
-        else:
-            directory = path
         try:
-            return self._path2confmods[directory]
+            return self._path2confmods[path]
         except KeyError:
+            if path.isfile():
+                directory = path.dirpath()
+            else:
+                directory = path
             # XXX these days we may rather want to use config.rootdir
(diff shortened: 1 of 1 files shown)


blueyed:
The 2nd commit also uses lru_cache missing in py27.
Needs a solution via https://github.com/pytest-dev/pytest/pull/4227.

blueyed:
I am taking out the 2nd commit (1acb18b).

src/_pytest/config/__init__.py
```diff
@@ -399,7 +399,7 @@ def _getconftestmodules(self, path):
                     mod = self._importconftest(conftestpath)
                     clist.append(mod)
 
-            self._path2confmods[directory] = clist
+            self._path2confmods[path] = clist
```
blueyed: Note that with 1acb18b I've noticed that some other code might expect a dirpath in `self._path2confmods` - haven't checked if I broke this with the previous optimization already though.

PR #4227, opened 2018-10-24
Use functools.lru_cache with _getconftest_pathlist

blueyed:
For pytest's own suite the `cache_info()` looks as follows:

    > session.config._getconftest_pathlist.cache_info()
    CacheInfo(hits=231, misses=19, maxsize=None, currsize=19)

While it does not really make a difference for me this might help with
larger test suites / the case mentioned in
https://github.com/pytest-dev/pytest/issues/2206#issuecomment-432623646.

/cc @boxed

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1,6 +1,7 @@
 """ command line options, ini-file and conftest.py processing. """
 from __future__ import absolute_import, division, print_function
 import argparse
+import functools
 import inspect
 import shlex
 import types
@@ -893,6 +894,7 @@ class Config(object):
             assert type is None
             return value
 
+    @functools.lru_cache(maxsize=None)
     def _getconftest_pathlist(self, name, path):
         try:
             mod, relroots = self.pluginmanager._rget_with_confmod(name, path)
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -894,7 +894,6 @@ class Config(object):
             assert type is None
             return value
 
-    @functools.lru_cache(maxsize=None)
     def _getconftest_pathlist(self, name, path):
         try:
             mod, relroots = self.pluginmanager._rget_with_confmod(name, path)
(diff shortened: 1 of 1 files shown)


boxed:
(Moving the discussion here) Yep. It helps, but there is some other big issue that overshadows this.

nicoddemus:
`AttributeError: 'module' object has no attribute 'lru_cache'` on py27.

Any objections to applying this only for PY3?

blueyed:
> Any objections to applying this only for PY3?

No.
But is not clear to me how to approach this - please feel free to edit / post a diff.

nicoddemus:
Done! 👍

src/_pytest/config/__init__.py
```diff
@@ -907,6 +908,10 @@ def _getconftest_pathlist(self, name, path):
             values.append(relroot)
         return values
 
+    if six.PY3:
+        # once we drop Python 2, please change this to use the normal decorator syntax (#4227)
+        _getconftest_pathlist = functools.lru_cache(maxsize=None)(_getconftest_pathlist)
```
nicoddemus: Old school style! 😎

blueyed: \o/

asottile: btw

```
if six.PY3:
```
is going to [break when python4 comes out](https://github.com/benjaminp/six/blob/e114efceea962fb143c909c904157ca994246fd2/six.py#L37) -- usually prefer `if six.PY2: ... ` / `if not six.PY2: ...`

I've seen other projects do this in a different way:

```python
if six.PY2:
    def lru_cache(*_, **__):
        def dec(fn):
            return fn
        return dec
else:
    from functools import lru_cache
```

neat trick though (saw this from twitter, missed the PR notification)

blueyed: @asottile 
Please consider creating a follow up PR - otherwise I might do it for the follow up via https://github.com/pytest-dev/pytest/pull/4224 (the removed, 2nd commit).
It should go into `pytest.compat` probably, so that it is easier to use everywhere.

nicoddemus: All fair points. I don't mind leaving the code as is, and also don't mind to change to the ones suggested here too. 👍

blueyed: Will move it to compat now.

PR #4231, opened 2018-10-24
Fix flake8: os imported, but unused

blueyed:
Likely due to b6fa4e24.


asottile:
implicit conflict between these comits: 86c7dcff689ecb72a27f6d9dd48abf56bb9d969d (moved the import) 956b3aca97cadf4fd802b60607115d1276f62229 (removed the import)

PR #4247, opened 2018-10-26
_getconftestmodules: use functools.lru_cache

blueyed:
Also renames `_path2confmods` to `_dirpath2confmods` for clarity (it is
expected to be a dirpath in `_importconftest`).

Merges master to avoid conflict with 63691f56 (which should have been done on features probably).

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -214,7 +214,7 @@ class PytestPluginManager(PluginManager):
         self._conftest_plugins = set()
 
         # state related to local conftest plugins
-        self._path2confmods = {}
+        self._dirpath2confmods = {}
         self._conftestpath2mod = {}
         self._confcutdir = None
         self._noconftest = False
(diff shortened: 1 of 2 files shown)


nicoddemus:
There's a simple failure related to renaming the private attribute

nicoddemus:
So the failing test on py27 is testing that we are caching `_getconftestmodules`. This patch makes the test pass:

```diff
diff --git a/src/_pytest/config/__init__.py b/src/_pytest/config/__init__.py
index 20c1e3ec..1124304a 100644
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -395,6 +395,11 @@ class PytestPluginManager(PluginManager):
         else:
             directory = path

+        try:
+            return self._dirpath2confmods[directory]
+        except KeyError:
+            pass
+
```

But it kind of reduces the point of using `lru_cache` a bit, doesn't it?

I guess it is OK to apply the patch above (perhaps inside a `if six.PY2` block to make it clear that is py2-only).

blueyed:
Thanks, added it with `if six.PY2`.

blueyed:
Codecov on AppVeyor failed to upload, otherwise it is green.

PR #4250, opened 2018-10-26
collection: _recurse: skip __pycache__

blueyed:
Taken out of https://github.com/pytest-dev/pytest/pull/4237.
According to https://codecov.io/gh/pytest-dev/pytest/commit/5850de62f11785ade9c980a9243f07d804d1f976 it should not be covered completely: python's `_recurse` was copied from the Session, which might not make sense completely, or would just need some tests for the `pytest_ignore_collect` code therein.


blueyed:
This improves performance from 10.5s to 8.35s with the 1000-files-test-case.

PR #4269, opened 2018-10-30
Upgrade pre-commit hooks

asottile:
@nicoddemus this includes the crlf fix


nicoddemus:
Great, thanks for the quick turn around. 👍

PR #4272, opened 2018-10-30
cacheprovider: display cachedir also in non-verbose mode if customized

blueyed:
TODO:

~~- [ ] test ?~~
- [x] changelog ?


nicoddemus:
I think a trivial changelog is fine, a test would also be nice to avoid regressions.

blueyed:
Added changelog.
Test will come with https://github.com/pytest-dev/pytest/pull/4271.

PR #4277, opened 2018-10-31
pdb: improve msg about output capturing with set_trace

blueyed:
Do not display "IO-capturing turned off/on" when ``-s`` is used to avoid
confusion.

TODO:

- [x] (trivial) changelog entry

### DIFF
--- a/changelog/4277.trivial.rst
+++ b/changelog/4277.trivial.rst
@@ -1,0 +1,4 @@
+pdb: improve message about output capturing with ``set_trace``.
+
+Do not display "IO-capturing turned off/on" when ``-s`` is used to avoid
+confusion.
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -102,6 +102,9 @@ class CaptureManager(object):
 
     # Global capturing control
 
+    def is_globally_capturing(self):
+        return self._method != "no"
+
     def start_global_capturing(self):
         assert self._global_capturing is None
         self._global_capturing = self._getcapture(self._method)
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -80,7 +80,10 @@ class pytestPDB(object):
                 capman.suspend_global_capture(in_=True)
             tw = _pytest.config.create_terminal_writer(cls._config)
             tw.line()
-            tw.sep(">", "PDB set_trace (IO-capturing turned off)")
+            if capman and capman.is_globally_capturing():
+                tw.sep(">", "PDB set_trace (IO-capturing turned off)")
+            else:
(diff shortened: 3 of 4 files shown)


src/_pytest/capture.py
```diff
@@ -102,6 +102,9 @@ def _getcapture(self, method):
 
     # Global capturing control
 
+    def is_globally_capturing(self):
+        return self._method != "no"
+
```
blueyed: I've thought about having a `is_suspended` property instead also.
What do you think?

blueyed: That would be `self._method != "no" and not self._suspended`, which is set internally then - or just `self.suspended`, which never gets set for method "no".

nicoddemus: I think this is fine actually.

nicoddemus:
We just need the change log entry and we are good to go. 👍

blueyed:
Added the changelog.

btw: is there any good mechanism to handle this for PRs better, where you have to create one to get the number?  I've thought about first pushing with `[ci skip]` and then adding it.
Usually it needs a second round due to CI failures, and then you can amend it, but in this case a new CI run just for the changelog would be bad - although linting might fail there, but I have pre-commit installed by now.. ;)

changelog/4277.trivial.rst
```diff
@@ -0,0 +1,4 @@
+pdb: improve msg about output capturing with set_trace
```
blueyed: Punctuation, will add it.

nicoddemus:
> btw: is there any good mechanism to handle this for PRs better, where you have to create one to get the number?

Not sure, but I think we are open to suggestions.

nicoddemus:
The problem with pushing a new commit with `[ci skip]` means that we cancel the currently running builds, and end up with no builds at all.

blueyed:
The first commit to create the PR would have "[ci skip]", but the following one with the changelog (amended) would not have it anymore.

nicoddemus:
Oh right, sorry, I was not clear. 

I got that you suggestion would be to use `[ci skip]` on the first commit (and I agree). I was just commenting that it was not what happened here, which resulted in the cancellation of the running builds for the first commit, and no builds being executed for the last commit either. 😁

blueyed:
I've not tried it here, but in https://github.com/pytest-dev/pytest/pull/4280 (although via PR title).

nicoddemus:
> I've not tried it here, but in #4280 (although via PR title).

65817dd was the last commit here, and it included `[skip ci]` in the commit title, is this one I'm talking about.

But no worries, I'm merging given that e0038b8 executed successfully in Travis and AppVeyor. 👍

PR #4280, opened 2018-10-31
pdb: improve quitting from debugger

blueyed:


changelog/4280.bugfix.rst
```diff
@@ -0,0 +1 @@
+Handle quitting from pdb with ``--trace`` by exiting pytest.
```
nicoddemus: Sorry if this is obvious, but I don't quite get what this means. 😁

blueyed: Using `q` in pdb should exit tests.
Currently it will run all remaining tests afterwards.
This is similar to previous changes from me in this regard.
It needs more work though (adjusting/verifying tests).

blueyed: Let me know if this is still unclear.
I agree that it could be a bit more verbose.
This improves on 86c7dcff.

nicoddemus: Now I understand, thanks!

Perhaps mentioning the old behavior will help to make it clear:

> Pressing `q` when in pdb will now exit pytest, instead of continue to run the remaining tests.

blueyed: Added, thanks.

Please review the tests carefully (we have not coverage for those).

nicoddemus: I have played around with the tests and it seems to me they are behaving well. 👍

blueyed:
TOXENV=py36-pexpect,py36-trial,py36-numpy failed twice on Travis now - debugging it.
Works locally.

blueyed:
Hmm.. not much info via `-s -vv` still.
https://travis-ci.org/pytest-dev/pytest/builds/449369211#L597-L598

Any ideas?

blueyed:
I can reproduce it locally without pdbpp, too.

The problem is that Exit is derived from KeyboardInterrupt, which gets caught by pdb itself then, which displays "--KeyboardInterrupt--" then.
Not sure why this is handled differently in pdbpp, which might actual be a bug/mis-feature there.

See https://github.com/pytest-dev/pytest/pull/4292 for some approach to fix/improve this.
A quickfix/hack/better fix might be to raise a custom exception from pdb, that is derived from outcomes.Exit and bdb.BdbQuit.
And/or fix pytest to handle the BdbQuit (that I assume should be raised) better, although IIRC that would not apply with `--trace`, which uses `pdb.runcall`.

blueyed:
Blocked by https://github.com/pytest-dev/pytest/pull/4292 / https://github.com/pytest-dev/pytest/pull/4299.

changelog/4280.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Handle quitting from pdb with ``--trace`` and parametrized tests by exiting pytest.
+
+Using ``[q]uit`` when in pdb will now exit pytest instead of running
+the remaining tests.
```
blueyed: TODO: reword: this is more generic now.

nicoddemus:
Perhaps we should rebase this on `features`? The diff now shows a ton of commits from `master`, making it difficult to review.

blueyed:
@nicoddemus 
Done.

nicoddemus:
Awesome, thanks!

blueyed:
Changed `Exit` to inherit from `bdb.BdbQuit`.
What do you think?

Should I try to just not inherit `Exit` from `SystemExit` instead?

nicoddemus:
@blueyed can you merge this or do you expect to make further changes?

blueyed:
@nicoddemus 
I'd like to review/revisit it myself at least..  (while coverage is not really helpful for this code, it would still be good to being looking at the codecov report also)

blueyed:
And also test-driving this in general showed another issue the other day (which I've pushed/addressed already), so please try to use it yourself for a bit.
Also reviews would be appreciated in general, of course.. :)

Issue #4046, opened 2018-09-26
Problem with running tests in package's `__init__`


pstiasny:
This might be related to #3749 and/or #3773.

Given the following directory structure:
```
tests/
  __init__.py (1)
  test_foo.py (2)
```

and following setup.cfg:
```
[tool:pytest]
python_files = test_*.py __init__.py
```

Before py.test 3.7:
* `py.test` with no arguments would collect tests in (1) and (2)
* `py.test tests/` would collect tests in (1) and (2)
* `py.test tests/__init__.py` would collect tests in (1)

Until 3.8.1 collecting from `__init__.py` would not work due to #3749.

In 3.8.1 collecting from `__init__.py` works again, but:
* `py.test` with no arguments would behave as before, collecting all tests correctly
* `py.test tests/` would collect tests in (2), omitting (1)
* `py.test tests/__init__.py` would collect tests in (1), but also in (2)

nicoddemus:
Thanks @pstiasny for the report!

blueyed:
Created a backport for master, please test it: https://github.com/pytest-dev/pytest/pull/4286.

PR #3872, opened 2018-08-25
Collect tests from __init__.py files if they match 'python_files'

nicoddemus:
Fix #3773

Followup to #3797

PR #4241, opened 2018-10-25
Move handling of duplicate files

blueyed:
This removes the hack added in
https://github.com/pytest-dev/pytest/pull/3802.

This also improves performance a bit with the test case for #4237:
Before: 7.95s
After: 7.51s

PR #4285, opened 2018-11-01
Fix problems with running tests in package `__init__` files (#4046)

kchmck:
This fixes some behavior changes that were introduced in 3.7.0, detailed by #4046. 

If you have a directory structure like
```
project/
    pytest.ini (has python_files = *.py)
    setup.py
    mypackage/
        __init__.py (contains tests)
        mymodule.py (contains tests)
```
then prior to 3.7.0 you could run `pytest mypackage` from the project directory, and it would pick up all the tests in `__init__.py` and `mymodule.py`. After 3.7.0 the tests in `__init__.py` started being ignored.

There was a partial fix for this in #3872, but it only works if the project directory and not the package directory is passed as the argument (commit https://github.com/pytest-dev/pytest/commit/848508150c7dc148f314ca335ad0d376e87fa3eb adds an example failing test for this.)

This PR seems to fix that and the other issue in #4046. I'm pretty sure the changes here aren't ideal and should probably be cleaned up by someone more familiar with the codebase, but hopefully this can at least serve as a starting point for the fixes.

The issue with `__init__.py` being ignored seems to stem from it being added to the duplicate paths set when it's visited in the root traversal loop, then when it's visited by the directory collection loop it gets ignored as a duplicate even though it was never checked for tests. Removing it from the duplicate set as a fix causes an additional issue of a duplicate nested package being created, which I also added a special-case fix for.

The issue with all package files being collected when only `__init__.py` was specified (test in https://github.com/pytest-dev/pytest/commit/3f07f7e07604b928efdae4409b631e827c7c5e57) appears to stem from `__init__.py` being treated as a `Package` (that gets traversed) when it should be treated as a `Module`.

With these changes, all the existing tests except one pass (commit https://github.com/pytest-dev/pytest/commit/28f4155db9ed1c16cc6aec74663ac913aae1a46f). The test failure seems legitimate to me, though, because it wasn't counting the package created by the `__init__` file. Please let me know if I'm interpreting that wrong.


blueyed:
Created a backport for master: https://github.com/pytest-dev/pytest/pull/4286

kchmck:
Thanks for looking into this! I can confirm that #4241/#4286 fixes the testcase added in https://github.com/pytest-dev/pytest/pull/4285/commits/848508150c7dc148f314ca335ad0d376e87fa3eb but not the one in https://github.com/pytest-dev/pytest/pull/4285/commits/3f07f7e07604b928efdae4409b631e827c7c5e57.

blueyed:
Cool!
Please cherry-pick #4286 here then (without the changelog), and add your other fixes on top.

blueyed:
Cherry-pick https://github.com/pytest-dev/pytest/commit/e04182364338ed9cd2a63945204701c99e297f21, please.

kchmck:
Alright, I've updated on top of https://github.com/pytest-dev/pytest/commit/e04182364338ed9cd2a63945204701c99e297f21

blueyed:
Great work, thanks!
Waiting for somebody else to merge it.

kchmck:
Awesome, thanks

PR #4287, opened 2018-11-01
debugging: make pytest_configure re-entrant

blueyed:
Causes:

    E             def test_1(testdir):
    E                 testdir.runpytest()
    E         >       __import__('pdb').set_trace()
    E
    E         ../../test_trace_after_runpytest.py:3:
    E         …/Vcs/pytest/src/_pytest/debugging.py:81: in set_trace
    E             tw = _pytest.config.create_terminal_writer(cls._config)
    E
    E         config = None, args = (), kwargs = {}, tw = <py._io.terminalwriter.TerminalWriter object at 0x7f1097088160>
    E
    E             def create_terminal_writer(config, *args, **kwargs):
    E                 """Create a TerminalWriter instance configured according to the options
    E                 in the config object. Every code which requires a TerminalWriter object
    E                 and has access to a config object should use this function.
    E                 """
    E                 tw = py.io.TerminalWriter(*args, **kwargs)
    E         >       if config.option.color == "yes":
    E         E       AttributeError: 'NoneType' object has no attribute 'option'

Appears to be related to `pytest_unconfigure` being called then, and this appears to be an inline run.

### DIFF
--- a/changelog/4287.bugfix.rst
+++ b/changelog/4287.bugfix.rst
@@ -1,0 +1,1 @@
+Fix nested usage of debugging plugin (pdb), e.g. with pytester's ``testdir.runpytest``.
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -47,17 +47,24 @@ def pytest_configure(config):
     if config.getvalue("usepdb"):
         config.pluginmanager.register(PdbInvoke(), "pdbinvoke")
 
-    old = (pdb.set_trace, pytestPDB._pluginmanager)
-
-    def fin():
-        pdb.set_trace, pytestPDB._pluginmanager = old
-        pytestPDB._config = None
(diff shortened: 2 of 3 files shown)


blueyed:
Should debugging's pytest_configure be made re-entrant (stacking/popping the stuff on the class), or should it not store it on the class in the first place?
https://github.com/blueyed/pytest/blob/a51fc3b5b2c2c1d7210f94ec8ec793fbc220b181/src/_pytest/debugging.py#L50-L61

blueyed:
I've gone with making it re-entrant.

blueyed:
Thanks for reviewing.

It seems like there is an issue already with the current test suite, it fails with this at the end (https://travis-ci.org/pytest-dev/pytest/jobs/449547322#L1057):

> IndexError: pop from empty list

This appears to indicate that there in on more `configure` than `unconfigure` - maybe even intentional for some test.

PR #4308, opened 2018-11-05
Don't string-compare version numbers

### LINKED ISSUE #4306: BUG: Pytest 3.10 doesn't parse minversion correctly
Pytest 3.10 doesn't correctly parse minversion correctly.  I have 

```
[tool:pytest]
minversion = 3.6
```

in setup.cfg and see `ERROR: setup.cfg:16: requires pytest-3.6, actual pytest-3.10.0'`

Changing 3.6 to 3.06 fixes the problem, but this isn't consistent versioning.

Should correctly pass if if LooseVersion accepts the strings.

```
from distutils.version import LooseVersion
LooseVersion('3.10') > LooseVersion('3.6')

Out[3]: True
```

asottile:
Resolves #4306

### DIFF
--- a/changelog/4306.bugfix.rst
+++ b/changelog/4306.bugfix.rst
@@ -1,0 +1,1 @@
+Parse ``minversion`` as an actual version and not as dot-separated strings.
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -11,6 +11,7 @@ import shlex
 import sys
 import types
 import warnings
+from distutils.version import LooseVersion
 
 import py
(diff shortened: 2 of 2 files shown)


asottile:
the py37 job seems pretty sad for this change :S

it passes locally however:

```
  py37: commands succeeded
  congratulations :)
```

blueyed:
@RonnyPfannschmidt 
> i'd prefer if we used or followed up with or used either a recent setuptools version parse or a packaging version parse

Please create an issue then.

Altogether I would have liked to see a regression test for this also.

Issue #4329, opened 2018-11-07
TypeError: '>' not supported between instances of 'NoneType' and 'float'


blueyed:
```
% pytest 'testing/test_collection.py::TestCollector::()::test_getcustomfile_roundtrip' -s --tb=short
_________________________________________________________________ TestCollector.test_getcustomfile_roundtrip _________________________________________________________________
…/Vcs/pytest/testing/test_collection.py:98: in test_getcustomfile_roundtrip
    node = testdir.getpathnode(hello)
…/Vcs/pytest/src/_pytest/pytester.py:737: in getpathnode
    res = session.perform_collect([x], genitems=False)[0]
…/Vcs/pytest/src/_pytest/main.py:430: in perform_collect
    items = self._perform_collect(args, genitems)
…/Vcs/pytest/src/_pytest/main.py:455: in _perform_collect
    self.ihook.pytest_collectreport(report=rep)
…/Vcs/pluggy/pluggy/hooks.py:284: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
…/Vcs/pluggy/pluggy/manager.py:67: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
…/Vcs/pluggy/pluggy/manager.py:61: in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
…/Vcs/pytest/src/_pytest/terminal.py:491: in pytest_collectreport
    self.report_collect()
…/Vcs/pytest/src/_pytest/terminal.py:500: in report_collect
    if self._collect_report_last_write > t - 0.5:
E   TypeError: '>' not supported between instances of 'NoneType' and 'float'
```

PR #4330, opened 2018-11-07
Fix TypeError in report_collect with _collect_report_last_write

blueyed:
`_collect_report_last_write` might be None, when `pytest_collection` was
not called before.  Not sure if this indicates another problem, but it
can be reproduced with `testing/test_collection.py::TestCollector::()::test_getcustomfile_roundtrip`.

Fixes https://github.com/pytest-dev/pytest/issues/4329

### DIFF
--- a/changelog/4329.bugfix.rst
+++ b/changelog/4329.bugfix.rst
@@ -1,0 +1,1 @@
+Fix TypeError in report_collect with _collect_report_last_write.
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -497,7 +497,10 @@ class TerminalReporter(object):
         if not final:
             # Only write "collecting" report every 0.5s.
             t = time.time()
-            if self._collect_report_last_write > t - 0.5:
+            if (
+                self._collect_report_last_write is not None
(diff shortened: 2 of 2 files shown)


nicoddemus:
Thanks @blueyed, you are on 🔥 

> but it
can be reproduced with testing/test_collection.py::TestCollector::()::test_getcustomfile_roundtrip

What do you mean? This test passes for me, and the output looks OK as well.

blueyed:
> But I worry that this was passing before, because we might introduce a regression by mistake in the future.

When doing this initially I assumed that both hooks were called in sequence always, which apparently is not the case here.
This fix is just was I would have done without that assumption, so I think it is good - but the issue itself of both hooks not being called always might be worth investigating..

nicoddemus:
> This fix is just was I would have done without that assumption, so I think it is good 

Fair enough. 👍

blueyed:
https://github.com/pytest-dev/pytest/pull/4331 contains a test for this, but fails due to the capture plugin.

blueyed:
Included in #4331.

blueyed:
Going to merge this already, since #4331 (which implicitly tests this) might take longer.

Issue #4333, opened 2018-11-07
 pytest version 3.10.1 Fails on python2.7


henrykironde:
python2.7.14/lib/python2.7/site-packages/_pytest/stepwise.py", line 102, in pytest_sessionfinish
    self.config.cache.set("cache/stepwise", [])
AttributeError: 'Config' object has no attribute 'cache'

ref: https://travis-ci.org/weecology/retriever/jobs/451239186

checklist:

- [x] Include a detailed description of the bug or suggestion
- [x] `pip list` of the virtual environment you are using
- [x] pytest and operating system versions
- [x] Minimal example if possible

blueyed:
> AttributeError: 'Config' object has no attribute 'cache'

Appears to be a duplicate of https://github.com/pytest-dev/pytest/issues/4304.

PR #4335, opened 2018-11-07
Refactor class StepwisePlugin.

henrykironde:
Initialize class properties
Check attributes before accessing them, LBYL

checklist

ref: #4333

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -195,6 +195,7 @@ Samuel Dion-Girardeau
 Samuele Pedroni
 Sankt Petersbug
 Segev Finer
+Senyondo Henry
 Serhii Mozghovyi
 Simon Gomizelj
 Skylar Downes
--- a/changelog/4304.bugfix.rst
+++ b/changelog/4304.bugfix.rst
@@ -1,0 +1,1 @@
+Possible bug with the 3.10.0 release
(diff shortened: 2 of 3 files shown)


changelog/4333.bugfix.rst
```diff
@@ -0,0 +1 @@
+pytest version 3.10.1 Fails on python2.7
```
blueyed: 3.10.1 is not there yet.. ;)
Also this should be more specific about the fix.

henrykironde:
@blueyed, this seems to fix the issue. I would love to make a test case but I don't seem to get the organization of the tests. 
However the logic  to create `testing` for the `pytest_sessionfinish` function would be:

- Start a session.

- Initialize **StepwisePlugin** class object and add the `session`. 

If the test are to pass, testing that is covered
The main goal is to have the tests fail, and parse a "--stepwise".

- If active `self.active = config.getvalue("stepwise")` is set, 
we should have some values in `self.lastfailed ` 
from `self.config.cache.set("cache/stepwise", self.lastfailed)`
 else expected values will be [] 
from `self.config.cache.set("cache/stepwise", [])`

nicoddemus:
Thanks @henrykironde for the contribution.

As @blueyed mentioned before, this is a duplicate of #4304.

In pytest many contributed tests use the `testdir` fixture to create small test cases which demonstrate an issue, run pytest, and check the output. There are tons of examples in `testing`, browse around and I'm sure you will get the gist of it.

Closing this now as this issue in particular was already fixed by #4304, but we look forward for your future contributions. 👍

Issue #4321, opened 2018-11-07
Regression: item.nodeid starts with '::' when symlinks are involved


Lekensteyn:
Since upgrading from pytest 3.8.2 to 3.9.3 (and 3.10.0), one of my tests are failing because the nodeid has an additional `::`. It seems related to symlinks. Reproducer with two files:

```py
# tests/conftest.py
def pytest_collection_modifyitems(items):
    for item in items:
        print(item.nodeid)
        assert not item.nodeid.startswith('::')
        #name = item.nodeid.split("::")[0].replace(".py", "").replace("/", ".")
# tests/test_some.py
def test_some():
    pass
```

Working:
```
$ pytest -v tests/
platform linux -- Python 3.7.1, pytest-3.10.0, py-1.7.0, pluggy-0.8.0 -- /usr/bin/python
cachedir: .pytest_cache
rootdir: /tmp/bla, inifile:
plugins: xdist-1.24.0, forked-0.2
collecting ... tests/test_some.py::test_some
collected 1 item
tests/test_some.py::test_some PASSED
```

Not working (I was not able to reproduce it with a symlink at `/tmp/foo` or `/tmp/bla/foo`):
```
$ ln -s $PWD/tests ~/foo
$ pytest -v ~/foo
latform linux -- Python 3.7.1, pytest-3.10.0, py-1.7.0, pluggy-0.8.0 -- /usr/bin/python
cachedir: .pytest_cache
rootdir: /home/peter/foo, inifile:
plugins: xdist-1.24.0, forked-0.2
collecting ... ::test_some.py::test_some
...
INTERNALERROR> AssertionError: assert not True
INTERNALERROR>  +  where True = <built-in method startswith of str object at 0x7fbf2d5a54e0>('::')
INTERNALERROR>  +    where <built-in method startswith of str object at 0x7fbf2d5a54e0> = '::test_some.py::test_some'.startswith
INTERNALERROR>  +      where '::test_some.py::test_some' = <Function 'test_some'>.nodeid
```

Possibly related issues:
#4174 (I also encountered this issue with 3.9.1 and 3.9.2, but not 3.9.3).
#4108

blueyed:
Thanks for the report.
Would be helpful if you could git-bisect this to some commit.

I suggest writing some test in e.g. `testing/test_collection_new.py` for this (look for existing ones in `testing/test_collection.py`).  This would be good then already as a test for a fix, of course.

blueyed:
[...]

PR #4352, opened 2018-11-09
Fix nodes._check_initialpaths_for_relpath for dirs

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/4321#issuecomment-436951894

### DIFF
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -447,7 +447,7 @@ class Collector(Node):
 def _check_initialpaths_for_relpath(session, fspath):
     for initial_path in session._initialpaths:
         if fspath.common(initial_path) == initial_path:
-            return fspath.relto(initial_path.dirname)
+            return fspath.relto(initial_path)
 
 
 class FSCollector(Collector):
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -663,11 +663,11 @@ class TestInvocationVariants(object):
         assert result.ret == 0
         result.stdout.fnmatch_lines(
             [
-                "*test_hello.py::test_hello*PASSED*",
-                "*test_hello.py::test_other*PASSED*",
-                "*test_world.py::test_world*PASSED*",
-                "*test_world.py::test_other*PASSED*",
-                "*4 passed*",
+                "test_hello.py::test_hello*PASSED*",
+                "test_hello.py::test_other*PASSED*",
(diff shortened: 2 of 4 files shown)


testing/acceptance_test.py
```diff
@@ -662,11 +662,11 @@ def test_cmdline_python_namespace_package(self, testdir, monkeypatch):
         assert result.ret == 0
         result.stdout.fnmatch_lines(
             [
-                "*test_hello.py::test_hello*PASSED*",
-                "*test_hello.py::test_other*PASSED*",
-                "*test_world.py::test_world*PASSED*",
-                "*test_world.py::test_other*PASSED*",
-                "*4 passed*",
+                "test_hello.py::test_hello*PASSED*",
+                "test_hello.py::test_other*PASSED*",
```
blueyed: Before this fix there was "hello/" here.

testing/test_session.py
```diff
@@ -323,7 +323,11 @@ def test_one():
 
     result = testdir.runpytest("--rootdir={}".format(path))
     result.stdout.fnmatch_lines(
-        ["*rootdir: {}/root, inifile:*".format(testdir.tmpdir), "*1 passed*"]
+        [
+            "*rootdir: {}/root, inifile:*".format(testdir.tmpdir),
+            "root/test_rootdir_option_arg.py *",
```
blueyed: Before the fix this was `root/test_rootdir_option_argX/test_rootdir_option_arg.py`.

src/_pytest/nodes.py
```diff
@@ -447,6 +447,8 @@ def _prunetraceback(self, excinfo):
 def _check_initialpaths_for_relpath(session, fspath):
     for initial_path in session._initialpaths:
         if fspath.common(initial_path) == initial_path:
+            if initial_path.isdir():
+                return fspath.relto(initial_path)
             return fspath.relto(initial_path.dirname)
```
blueyed: According to codecov only the `isdir` (new) path is covered: https://codecov.io/gh/pytest-dev/pytest/commit/8532756fe549edb6ecb01733f33f714cebdc35b2#D1-450

blueyed: And looking closer it shows that it can only really be a dir here, if it shares a / is the common path with `fspath`, no?

blueyed: Amended.

PR #4358, opened 2018-11-09
Node: do not add "::()" to nodeid

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4127.

### DIFF
--- a/testing/test_collection.py
+++ b/testing/test_collection.py
@@ -510,13 +510,8 @@ class TestSession(object):
                     pass
         """
         )
-        normid = p.basename + "::TestClass::()::test_method"
-        for id in [
-            p.basename,
-            p.basename + "::TestClass",
-            p.basename + "::TestClass::()",
-            normid,
-        ]:
+        normid = p.basename + "::TestClass::test_method"
+        for id in [p.basename, p.basename + "::TestClass", normid]:
             items, hookrec = testdir.inline_genitems(id)
             assert len(items) == 1
             assert items[0].name == "test_method"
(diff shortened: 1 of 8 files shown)


testing/test_collection.py
```diff
@@ -510,13 +510,8 @@ def test_method(self):
...
         )
-        normid = p.basename + "::TestClass::()::test_method"
-        for id in [
-            p.basename,
-            p.basename + "::TestClass",
-            p.basename + "::TestClass::()",
-            normid,
-        ]:
+        normid = p.basename + "::TestClass::test_method"
+        for id in [p.basename, p.basename + "::TestClass", normid]:
```
blueyed: btw: I would really prefer if black would not make a) the diff unnecessarily bigger here by merging this into a single line, and b) keep it for readability.

testing/test_session.py
```diff
@@ -274,16 +274,26 @@ def test_deselect(testdir):
...
+
+            def test_c2(self): pass
     """
     )
     result = testdir.runpytest(
-        "-v", "--deselect=test_a.py::test_a2[1]", "--deselect=test_a.py::test_a2[2]"
+        "-v",
+        "--deselect=test_a.py::test_a2[1]",
+        "--deselect=test_a.py::test_a2[2]",
+        "--deselect=test_a.py::TestClass::test_c1",
```
blueyed: I wonder if we should still support the old syntax, but figured it would not be worth it really.
People might be affected though if they use this in scripts already etc..?!
Anyway, it currently uses `.startswith()`, and therefore a `Node.matches` method should be added/used.

nicoddemus: I agree it is not worth supporting the old syntax. 👍

PR #4337, opened 2018-11-08
Fix collection of symlinked files

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4325.
Fixes https://github.com/pytest-dev/pytest/issues/4321.

TODO:

- [x] https://github.com/pytest-dev/pytest/issues/4321#issuecomment-436951894

Issue #4366, opened 2018-11-10
Tests: test_fail_on_flag (stepwise) is flaky


blueyed:
Seen it before already, and here it is again:

https://ci.appveyor.com/project/pytestbot/pytest/builds/20191315/job/431p82b80ai2je82#L148

Seems to be related to xdist.  TOXENV=py27-trial,py27-numpy,py27-nobyte in this case.

```
================================== FAILURES ===================================
____________________ test_fail_and_continue_with_stepwise _____________________
[gw1] win32 -- Python 2.7.15 c:\projects\pytest\.tox\py27-nobyte\scripts\python.exe
stepwise_testdir = <Testdir local('c:\\users\\appveyor\\appdata\\local\\temp\\1\\pytest-of-appveyor\\pytest-2\\popen-gw1\\test_fail_and_continue_with_stepwise0')>
    def test_fail_and_continue_with_stepwise(stepwise_testdir):
        # Run the tests with a failing second test.
        result = stepwise_testdir.runpytest("-v", "--strict", "--stepwise", "--fail")
        assert not result.stderr.str()
    
        stdout = result.stdout.str()
        # Make sure we stop after first failing test.
        assert "test_success_before_fail PASSED" in stdout
        assert "test_fail_on_flag FAILED" in stdout
        assert "test_success_after_fail" not in stdout
    
        # "Fix" the test that failed in the last run and run it again.
        result = stepwise_testdir.runpytest("-v", "--strict", "--stepwise")
        assert not result.stderr.str()
    
        stdout = result.stdout.str()
        # Make sure the latest failing test runs and then continues.
>       assert "test_success_before_fail" not in stdout
E       AssertionError: assert 'test_success_before_fail' not in '============================= test session starts =============================\nplatform win32 -- Python 2.7.15, pyte...                              [100%]\n\n========================== 6 passed in 0.08 seconds ===========================\n'
C:\projects\pytest\testing\test_stepwise.py:95: AssertionError
---------------------------- Captured stdout call -----------------------------
============================= test session starts =============================
platform win32 -- Python 2.7.15, pytest-3.10.1.dev65+g9d11428a, py-1.7.0, pluggy-0.8.0 -- c:\projects\pytest\.tox\py27-nobyte\scripts\python.exe
cachedir: C:\projects\pytest\.tox\py27-nobyte\.pytest_cache
[...]

PR #4367, opened 2018-11-10
Optimize/revisit determine_setup

blueyed:
Taken out of https://github.com/pytest-dev/pytest/pull/4337.

### DIFF
--- a/src/_pytest/config/findpaths.py
+++ b/src/_pytest/config/findpaths.py
@@ -121,28 +121,34 @@ def determine_setup(inifile, args, rootdir_cmd_arg=None, config=None):
                 break
             except KeyError:
                 inicfg = None
-        rootdir = get_common_ancestor(dirs)
+        if rootdir_cmd_arg is None:
+            rootdir = get_common_ancestor(dirs)
     else:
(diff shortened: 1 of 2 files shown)


src/_pytest/config/findpaths.py
```diff
@@ -121,28 +121,34 @@ def determine_setup(inifile, args, rootdir_cmd_arg=None, config=None):
                 break
             except KeyError:
                 inicfg = None
-        rootdir = get_common_ancestor(dirs)
+        if rootdir_cmd_arg is None:
```
blueyed: Note: only the "is None" branch appears to be covered here: https://codecov.io/gh/pytest-dev/pytest/commit/247bdb2c0f5ba64161de7f24e958919dfe0e02ea#D1-124

blueyed:
Build failure is unrelated (test_fail_and_continue_with_stepwise), #4366.

blueyed:
This was mostly for readability / saneness IIRC.

> the test results indicate it is

Well, it is not 100% covered and we had things already where tests were passing, but they broke things then.

But since this is for features and relatively safe, I'm going to merge this already.

PR #4368, opened 2018-11-11
Travis: use coverage from existing tox env  [skip appveyor]

blueyed:


blueyed:
Coverage decrease is expected due to skipping AppVeyor.

PR #4380, opened 2018-11-13
Travis: move faster builds to baseline

blueyed:
- use py27-pexpect,py27-trial,py27-numpy and py37-xdist in baseline,
  using pexpect there catches errors with pdb tests early, and
  py37-xdist is much faster than py37.
- move py34 and py36 out of baseline.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,7 @@
 environment:
   matrix:
+  - TOXENV: "py37-xdist"
+  - TOXENV: "py27-xdist"
   - TOXENV: "py27"
   - TOXENV: "py37"
   - TOXENV: "linting,docs,doctesting"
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -1,5 +1,7 @@
 environment:
   matrix:
+  - TOXENV: "py37-xdist"
+  - TOXENV: "py27-xdist"
   - TOXENV: "py27"
   - TOXENV: "py37"
```
blueyed: I would argue that it's ok to remove py27 and py37 here then, but left it out for now.
It might be good to run everything with xdist on AppVeyor (but would need new tox factors).

blueyed:
Re AppVeyor (2nd commit): py27-xdist is slower than py27 there.. https://ci.appveyor.com/project/pytestbot/pytest/builds/20250577

Issue #4393, opened 2018-11-14
pytest overwrites .gitignore and README.md when run with --pdb


cdeil:
There's an issue that was introduced in #3982 .

Running `pytest -v test.py --pdb` on a test that raises an exception will create or overwrite the `.gitignore` and `README.md` files in the current working directory.

At least I had one case where this happened, with `Python 2.7.15, pytest-3.10.1, py-1.7.0, pluggy-0.8.0` because for some reason there:
```
cachedir: .
```
See https://gist.github.com/cdeil/a90f0b5727a6187d06ecbf6be68af0ad

This writing of `.gitignore` and `README.md` was introduced in #4390, so cc @nicoddemus .

Can you reproduce the issue?

blueyed:
The README.md should not be overwritten, but .gitignore might be:
https://github.com/blueyed/pytest/blob/ce0ed9cd22f40d9753efeb59e62c9bb2335877cb/src/_pytest/cacheprovider.py#L123-L131

cdeil:
@blueyed - Indeed, `README.md` wasn't overwritten.

In my project there is only a `README.rst`. So pytest created a new `README.md` file.

blueyed:
See https://github.com/pytest-dev/pytest/pull/4394 - however `cache_dir = .` is not really a good setting in  general AFAICT (except for finding corner cases.. ;))

blueyed:
Ah, I've thought you configured it explicitly.
Check what `inifile` is set to (from the header).
Are you using tox by chance?

blueyed:
I get `.pytest_cache`:
```
platform linux2 -- Python 2.7.15, pytest-4.0.0, py-1.7.0, pluggy-0.8.0 -- /tmp/gammapy/venv2/bin/python
cachedir: .pytest_cache
rootdir: /tmp/gammapy, inifile: setup.cfg
collecting ... 
Gammapy test data availability:
gammapy-extra ... no
Gammapy environment variables:
GAMMAPY_EXTRA = not set
collected 0 items / 1 errors                                                                
```

blueyed:
Check the code around here:
https://github.com/pytest-dev/pytest/blob/9dec146edf5baf74a82b91fc85762c8676b604b8/src/_pytest/cacheprovider.py#L279-L282

PR #4394, opened 2018-11-14
cacheprovider: do not write README/.gitignore to existing dir

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4393.

It will not "upgrade" existing dirs anymore - not sure if we want to keep this behavior?!
Could be achieved with more logic, but certainly `cache_dir = .` should be skipped etc.

### DIFF
--- a/testing/test_cacheprovider.py
+++ b/testing/test_cacheprovider.py
@@ -899,5 +899,28 @@ def test_gitignore(testdir):
     config = testdir.parseconfig()
     cache = Cache.for_config(config)
     cache.set("foo", "bar")
-    msg = "# created by pytest automatically, do not change\n*"
-    assert cache._cachedir.joinpath(".gitignore").read_text(encoding="UTF-8") == msg
+    msg = "# Created by pytest automatically.\n*"
+    gitignore_path = cache._cachedir.joinpath(".gitignore")
+    assert gitignore_path.read_text(encoding="UTF-8") == msg
+
+    # Does not overwrite existing/custom one.
+    gitignore_path.write_text("")
+    cache.set("something", "else")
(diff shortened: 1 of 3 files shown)


testing/test_cacheprovider.py
```diff
@@ -899,5 +899,28 @@ def test_gitignore(testdir):
...
     cache = Cache.for_config(config)
     cache.set("foo", "bar")
-    msg = "# created by pytest automatically, do not change\n*"
-    assert cache._cachedir.joinpath(".gitignore").read_text(encoding="UTF-8") == msg
+    msg = "# Created by pytest automatically.\n*"
+    gitignore_path = cache._cachedir.joinpath(".gitignore")
+    assert gitignore_path.read_text(encoding="UTF-8") == msg
+
+    # Does not overwrite existing/custom one.
+    gitignore_path.write_text("")
```
blueyed: Needs to be fixed for py27, otherwise CI is fine.

nicoddemus: Done. 👍

testing/test_cacheprovider.py
```diff
@@ -899,5 +899,29 @@ def test_gitignore(testdir):
...
+        """
+        [pytest]
+        cache_dir = .
+        """
+    )
+    config = testdir.parseconfig()
+    cache = Cache.for_config(config)
+    cache.set("foo", "bar")
+
+    assert os.path.isdir("v")  # cache contents
```
nicoddemus: Added this to ensure `.` is the actual cache dir (just to be safe).

blueyed: Thanks!

nicoddemus:
🙇

PR #4399, opened 2018-11-15
Display "short test summary info" after (main) warnings again

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/3952.

Ref: https://github.com/pytest-dev/pytest/pull/3956.

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -647,9 +647,11 @@ class TerminalReporter(object):
     def pytest_terminal_summary(self):
         self.summary_errors()
         self.summary_failures()
-        yield
         self.summary_warnings()
+        yield
         self.summary_passes()
+        # Display any extra warnings from teardown here (if any).
+        self.summary_warnings()
 
     def pytest_keyboard_interrupt(self, excinfo):
         self._keyboardinterrupt_memo = excinfo.getrepr(funcargs=True)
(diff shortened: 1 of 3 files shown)


src/_pytest/terminal.py
```diff
@@ -647,9 +647,10 @@ def pytest_sessionfinish(self, exitstatus):
     def pytest_terminal_summary(self):
         self.summary_errors()
         self.summary_failures()
-        yield
         self.summary_warnings()
+        yield
```
blueyed: "short test summary info" comes via the skipping plugin, which I found rather confusing.
Should/can this be moved?

It would also might make sense to have "short test summary info" after "summary_passes" even, but I wanted to not touch that behavior (and am not using it - `-ra` does not include `P`).

nicoddemus: Hmmm if we move it to `pytest_terminal_summary`, will it simplify things? It doesn't look so simple to move it, because there's some logic in there to handle summaries for xpass, xfail, etc.

src/_pytest/terminal.py
```diff
@@ -726,11 +727,19 @@ def summary_warnings(self):
             if not all_warnings:
                 return
 
+            final = hasattr(self, "_already_displayed_warnings")
```
nicoddemus: Hmm I'm not sure, I understand the rationale for the change, but I'm not so happy with how this turns out... 🤔

src/_pytest/terminal.py
nicoddemus: Perhaps we should add a comment here explaining why we display warnings twice, what do you think?

blueyed: Makes sense.

blueyed:
Added the comment.

(normally I would have pushed this as a fixup with `[ci skip]`, and then it could have been squashed merged, but yeah.. - full CI run now for this..)

nicoddemus:
I will get this merged now as we know it has passed in the previous commit, and the only change has been a comment. 😁 

Thanks @blueyed!

PR #4411, opened 2018-11-17
Fix collection of testpaths with `--pyargs`

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4405.


hroncok:
For the record, this causes a regression when running the tests with custom PYTHONPATH: https://github.com/pytest-dev/pytest/issues/4913

A fix is ready at https://github.com/pytest-dev/pytest/pull/4914 and I've requested a review from @blueyed.

PR #4456, opened 2018-11-24
Highlight docs with pygments-pytest

asottile:
[`pygments-pytest`](https://github.com/asottile/pygments-pytest) still needs to [get better at -q](https://github.com/asottile/pygments-pytest/issues/7) but I think this is good enough to integrate now.

Here's some examples:


nicoddemus:
Woot!

Just one question, does the `.. code-block:: pytest` block interferes with `regendoc`?

asottile:
> 
> 
> Woot!
> 
> Just one question, does the `.. code-block:: pytest` block interferes with `regendoc`?

yep! made sure `tox -e regen` worked!

one quirk, it added `hypothesis` everywhere so I only committed the actual changes (like my [assertion update](https://github.com/pytest-dev/pytest/pull/4456/files#diff-ea3360880768705b76e7c5bf1b2ba0acR369)).  There's a surprisingly large amount of stuff not handled by `regendoc` right now, and even some that [now makes no sense](https://github.com/pytest-dev/pytest/blob/master/doc/en/example/simple.rst#dynamically-adding-command-line-options)

nicoddemus:
You should tweet about it Anthony

asottile:
Indeed! I'll do that when I'm at a computer :D

asottile:
hmmmm, seems I have to fix rtd 🤦‍♂️ https://readthedocs.org/projects/pytest/builds/8144977/

asottile:
the twits as promised: https://twitter.com/codewithanthony/status/1066524355541139456

PR #4461, opened 2018-11-24
Fix rtd by installing pygments-pytest

asottile:


tox.ini
```diff
@@ -124,11 +124,7 @@ setenv = {[testenv:py27-pluggymaster]setenv}
 skipsdist = True
 usedevelop = True
 changedir = doc/en
-deps =
-    PyYAML
```
asottile: PyYAML seems to only be necessary for regen so I removed it.

asottile:
@blueyed -- maybe try codeclimate in a fork before turning it on here? going to make all the PRs red :(

nicoddemus:
It seems the RTD build has passed, but I don't see the colors in the output... 🤔

asottile:
the css file is 404ing for some reason 🤔 -- the highlight classes are there at least!

blueyed:
@asottile 
Disabled the "Pull request status updates" - hopefully that fixes it.

blueyed:
@nicoddemus 
I would assume that it uses the diff.
My main intention is looking for an alternative for codecov, which fails to report/display diff coverage etc.
But codeclimate cannot merge coverage reports by itself, but need artifacts (which works nicely with CircleCI, but not Tavis).

nicoddemus:
> I would assume that it uses the diff.

Yeah, I would be surprised if that was not the case.

> My main intention is looking for an alternative for codecov, which fails to report/display diff coverage etc.
But codeclimate cannot merge coverage reports by itself, but need artifacts

I see @blueyed, I definitely appreciate you trying out those new tools btw. 👍

asottile:
@nicoddemus is there a way I can get rtd to build a feature branch so I can poke at this? as far as I can tell I'm mostly following how [this suggest](https://docs.readthedocs.io/en/latest/guides/adding-custom-css.html) but I have a few ideas I can try

particularly https://github.com/pytest-dev/pytest/commit/89fd427bdb99cb16500242c2977e72d2519e884a

nicoddemus:
Hmm AFAIK it only builds branches of the repository itself, not from forks. Feel free to push a branch to the main repository and I can enable builds for that branch. Also if you need admin access to RTD let me know. 😁

asottile:
ah I think I can just set up rtfd on my fork and play around there -- can you screenshot the settings page so I can see what settings are being used? (and the advanced settings page too!)

PR #4464, opened 2018-11-25
Display actual test ids in `--collect-only`

### LINKED ISSUE #4458: bytes parametrize values aren't properly escaped
## Problem

Parametrize values of type `byte` don't properly escape non-printable characters leading to inability to select those specific tests with `-k`.

## Minimal example

`test_select.py`:
```python
import pytest


@pytest.mark.parametrize('p', [b'\x01a', b'b'])
def test_a(p):
    pass
```

```shell
~$ pytest --collect-only --quiet
test_select.py::test_a[a]
test_select.py::test_a[b]

no tests ran in 0.04 seconds
```

```shell
~$ pytest -k 'test_a[a]'
======================================== test session starts ========================================
platform darwin -- Python 3.6.6, pytest-4.0.0, py-1.7.0, pluggy-0.8.0
rootdir: /Users/ulo/Envs/tmp-54dcc18e560d86f8/pytest-select, inifile:
collected 2 items / 2 deselected

=================================== 2 deselected in 0.01 seconds ====================================

asottile:
Resolves #4458

### DIFF
--- a/changelog/4458.bugfix.rst
+++ b/changelog/4458.bugfix.rst
@@ -1,0 +1,1 @@
+Display actual test ids in ``--collect-only``.
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -138,7 +138,7 @@ class Node(object):
         return cls
 
     def __repr__(self):
-        return "<%s %r>" % (self.__class__.__name__, getattr(self, "name", None))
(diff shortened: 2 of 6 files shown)


blueyed:
Note: I think it is more useful to have the nodeid in `__repr__`, and use the change from here for `__str__` (https://github.com/blueyed/pytest/pull/22/files).

PR #4485, opened 2018-11-30
RFC Multiline logging

thisch:
This adds support for aligned multi-line log messages:

image

What this change is doing is inserting white-space to all lines except the first of the "message" part of the log msg.

### DIFF
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -3,8 +3,10 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import copy
 import logging
 import re
+import sys
 from contextlib import contextmanager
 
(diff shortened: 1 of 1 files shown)


blueyed:
Interesting.
I wonder though why not `filename:lineno` is used (to make it more condense), but that appears to not have changed here, or has it?

thisch:
I think we should rather make the current integration of the log formatter a bit more extensible. What I have in mind is that the default formatter should be easily replaceable by a custom log formatter. Ideally we could create a collection of predefined log formatters in a new pytest package that can then be selected by name.

s0undt3ch:
Even better then :)

nicoddemus:
Hi @thisch, 

PR #4503, opened 2018-12-03
Upgrade sphinx doc requirements

asottile:
the upstream sphinx issue seems resolved.  I'll watch this carefully as it merges

I also bumped pygments-pytest to get improved `-q` coloring support


asottile:
confirming search works -- not that I had any doubts :)

PR #4521, opened 2018-12-09
[WIP] load_entrypoint_plugins option/marker

blueyed:
Alternative to https://github.com/pytest-dev/pytest/pull/4518.

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -369,9 +369,6 @@ class PytestPluginManager(PluginManager):
             "trylast: mark a hook implementation function such that the "
             "plugin machinery will try to call it last/as late as possible.",
         )
-        config.addinivalue_line(
-            "markers", "load_entrypoint_plugins: only load the given entrypoint plugins"
-        )
         self._configured = True
 
     #
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -252,10 +252,11 @@ class PytestPluginManager(PluginManager):
         ret = super(PytestPluginManager, self).is_blocked(name)
         if not ret:
             try:
-                load_entrypoint_plugins = self.config.getini("load_entrypoint_plugins")
-            except (AttributeError, ValueError):
+                config = self.config
+            except AttributeError:
                 pass
             else:
+                load_entrypoint_plugins = config.getini("load_entrypoint_plugins")
                 if load_entrypoint_plugins is not notset:
                     if name not in load_entrypoint_plugins:
                         return True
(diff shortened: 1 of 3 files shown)


blueyed:
The config hack would not be required if `self.pluginmanager.load_setuptools_entrypoints("pytest11")` would be more modular (https://github.com/pytest-dev/pytest/blob/d1221f14b78872e3b64e5cc98b114eaa81f79a97/src/_pytest/config/__init__.py#L824).

Given that this adds `is_blocked` for this, we could also override `load_setuptools_entrypoints` in PytestPluginmanager and skip them there.

But like asked in #4518 already: how to enable this for all of pytest's inner tests then?
Should pytester forward this?  If so, should it use an env or default args?

src/_pytest/config/__init__.py
```diff
@@ -757,6 +782,11 @@ def _mark_plugins_for_rewrite(self, hook):
 
         self.pluginmanager.rewrite_hook = hook
 
+        load_entrypoint_plugins = self.getini("load_entrypoint_plugins")
+        if load_entrypoint_plugins is not notset and not len(load_entrypoint_plugins):
+            assert isinstance(load_entrypoint_plugins, list)
```
nicoddemus: Can we have a more friendly message than an `assert`? For example if a user writes `load_entrypoint_plugins=xdist`, the option will be read as a string and the assert will blow up without any useful information.

src/_pytest/config/__init__.py
```diff
@@ -238,6 +240,19 @@ def __init__(self):
         # Used to know when we are importing conftests after the pytest_configure stage
         self._configured = False
 
+    def is_blocked(self, name):
```
nicoddemus: It will be a bummer to write one, but I guess we should add a functional test for this?

src/_pytest/config/__init__.py
```diff
@@ -321,6 +336,13 @@ def hasplugin(self, name):
         """Return True if the plugin with the given name is registered."""
         return bool(self.get_plugin(name))
 
+    def pytest_addoption(self, parser):
+        parser.addini(
+            "load_entrypoint_plugins",
+            help="only load specified plugins via entrypoint",
```
nicoddemus: Please use `type="args"` here to force this to always be a list.

src/_pytest/config/__init__.py
nicoddemus: Actually see my tip above about using `type="args"` in the option declaration. 👍

src/_pytest/config/__init__.py
blueyed: Yeah, tests are fine.. :D

blueyed:
Rebased, and used type=args.

I came across this again, since pytest-pdb also gets used in inline runs and then throws an error if an attribute exists already.

nicoddemus:
Hey @blueyed,

Closing this for the same reasons as the other PRs, feel free to re-open it if you're still interested. 👍

Issue #4265, opened 2018-10-29
PYTEST_ADDOPTS=--cov pytest test_foo.py


blueyed:
pytest-cov has an argument with `nargs='?'`: https://github.com/pytest-dev/pytest-cov/blob/7205428bfb9ab9f8dbcaeefe23ce5fe426bb06b9/src/pytest_cov/plugin.py#L55-L58

I've just noticed that with `PYTEST_ADDOPTS=--cov pytest foo` "foo" becomes the arg for it, which really confused me.

It looks like the PYTEST_ADDOPTS handling injects the value, but it should get handled separately/isolated.

blueyed:
A failing test for this:

```diff
diff --git a/testing/test_config.py b/testing/test_config.py
index b7c68ea4..1b444ab6 100644
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -1109,6 +1109,28 @@ def test_addopts_from_ini_not_concatenated(self, testdir):
         )
         assert result.ret == _pytest.main.EXIT_USAGEERROR
 
+    @pytest.mark.xfail
+    def test_addopts_from_ini_with_default(self, testdir):
+        """addopts from ini should not take values from normal args (#4265)."""
+        testdir.makeconftest(
+            """
+            def pytest_addoption(parser):
+                group = parser.getgroup('cov')
+                group.addoption('--cov', action='append', default=[],
+                                dest="cov_source", nargs='?', const=True)
+            """)
+        testdir.makeini(
+            """
+            [pytest]
+            addopts=--cov
+        """)
+        testdir.makepyfile(test_test="""
+            def test_test(request):
+                assert request.config.getoption("cov_source") == [True]
+        """)
+        result = testdir.runpytest("test_test.py")
+        assert result.ret == _pytest.main.EXIT_OK
+
     def test_override_ini_does_not_contain_paths(self):
         """Check that -o no longer swallows all options after it (#3103)"""
         from _pytest.config import get_config
```

PR #4523, opened 2018-12-09
Ensure that PYTEST_ADDOPTS are valid by themselves

blueyed:
Ref https://github.com/pytest-dev/pytest/issues/4265.

### DIFF
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -1082,6 +1082,33 @@ class TestOverrideIniArgs(object):
         config._preparse([], addopts=True)
         assert config._override_ini == ["cache_dir=%s" % cache_dir]
 
+    def test_addopts_from_env_not_concatenated(self, monkeypatch):
+        """PYTEST_ADDOPTS should not take values from normal args (#4265)."""
+        from _pytest.config import get_config
+
+        monkeypatch.setenv("PYTEST_ADDOPTS", "-o")
+        config = get_config()
+        with pytest.raises(SystemExit) as excinfo:
+            config._preparse(["cache_dir=ignored"], addopts=True)
+        assert excinfo.value.args[0] == _pytest.main.EXIT_USAGEERROR
+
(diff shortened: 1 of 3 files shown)


testing/test_config.py
```diff
@@ -1082,6 +1082,33 @@ def test_addopts_before_initini(self, monkeypatch):
...
         assert config._override_ini == ["cache_dir=%s" % cache_dir]
 
+    def test_addopts_from_env_not_concatenated(self, monkeypatch):
+        """PYTEST_ADDOPTS should not take values from normal args (#4265)."""
+        from _pytest.config import get_config
+
+        monkeypatch.setenv("PYTEST_ADDOPTS", "-o")
+        config = get_config()
+        with pytest.raises(SystemExit) as excinfo:
+            config._preparse(["cache_dir=ignored"], addopts=True)
```
nicoddemus: Just curious, why do you need the `cache_dir=ignored`?

nicoddemus:
From #4265:

> I've just noticed that with PYTEST_ADDOPTS=--cov pytest foo "foo" becomes the arg for it, which really confused me.

`PYTEST_ADDOPTS` will be validated separately (which is correct), but `foo` will still be processed by `--cov`, no?

testing/test_config.py
blueyed: Not needed, amended.

blueyed: Well, it is kind of needed, since previously it would be concatenated and then pass.

nicoddemus: Ahh I see, thanks!

blueyed:
> but `foo` will still be processed by `--cov`, no?

Yeah.. :/
I've looked at bit at fixing this, and it might be possible by keeping the parsed namespace, instead of parsing all args altogether in the end, but not trivial.
I will change Fixes to Closed in the original comment.

PR #4525, opened 2018-12-10
tests: fix test_raises_exception_looks_iterable

blueyed:
Started to fail on py37-xdist between
https://travis-ci.org/pytest-dev/pytest/jobs/465498973 and
https://travis-ci.org/pytest-dev/pytest/jobs/465792343, but could not
find a diff in versions (from the tox header), and both commits failed
locally.


blueyed:
Well, should go into master also, of course.

Issue #4483, opened 2018-11-30
Running with --junitxml should report call times, not total times


aparamon:
Currently, [--junitxml](https://docs.pytest.org/en/latest/usage.html#creating-junitxml-format-files) reports the total test time, including fixture set-ups and tear-downs:

````py
import pytest
import time

@pytest.fixture
def requisite():
    time.sleep(2)
    yield
    time.sleep(3)

def test_this(requisite):
    time.sleep(1)
````

````bat
>pytest test.py --durations 0 --junitxml result.xml
============================= test session starts =============================
platform win32 -- Python 3.6.6, pytest-3.8.0, py-1.6.0, pluggy-0.7.1
rootdir: N:\pytest, inifile:
plugins: requests-mock-1.5.2, repeat-0.4.1, remotedata-0.3.0, ordering-0.5, openfiles-0.3.0, doctestplus-0.1.3, arraydiff-0.2
collected 1 item

test.py .                                                                [100%]

------------------ generated xml file: N:\pytest\result.xml -------------------
=========================== slowest test durations ============================
3.00s teardown test.py::test_this
2.00s setup    test.py::test_this
1.00s call     test.py::test_this
========================== 1 passed in 6.03 seconds ===========================
````

```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="0" name="pytest" skips="0" tests="1" time="6.026">
  <testcase classname="test" file="test.py" line="9" name="test_this" time="6.002954006195068"></testcase>
</testsuite>
````

There should be an option to force --junitxml report call times, not total times.
Please additionally consider making it the default behavior.

Zac-HD:
Unfortunately the JUnit XML spec ([e.g. here](https://www.ibm.com/support/knowledgecenter/en/SSQ2R2_14.1.0/com.ibm.rsar.analysis.codereview.cobol.doc/topics/cac_useresults_junit.html)) is pretty clear that the `time` field is for the actual time taken by executing the test, not time less setup.

aparamon:
@Zac-HD I cannot completely agree that definition from your link or in
https://github.com/windyroad/JUnit-Schema/blob/49e95a79cc0bfba7961aaf779710a43a4d3f96bd/JUnit.xsd#L127-L130
is completely unambiguous :-)

[...]

PR #4529, opened 2018-12-11
Add --junittime=call option

aparamon:
Implements https://github.com/pytest-dev/pytest/issues/4483
Please review and let me know if docs/changelog update is required for this simple change.
attn @MuchkoM

### DIFF
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -371,8 +371,9 @@ def mangle_test_address(address):
 
 
 class LogXML(object):
-    def __init__(self, logfile, prefix, suite_name="pytest", logging="no",
-                 report_duration=None):
+    def __init__(
+        self, logfile, prefix, suite_name="pytest", logging="no", report_duration=None
+    ):
         logfile = os.path.expanduser(os.path.expandvars(logfile))
         self.logfile = os.path.normpath(os.path.abspath(logfile))
         self.prefix = prefix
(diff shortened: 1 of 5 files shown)


src/_pytest/junitxml.py
```diff
@@ -361,12 +371,15 @@ def mangle_test_address(address):
 
 
 class LogXML(object):
-    def __init__(self, logfile, prefix, suite_name="pytest", logging="no"):
+    def __init__(
+        self, logfile, prefix, suite_name="pytest", logging="no", report_duration=None
```
nicoddemus: To make things simpler I propose to use `report_duration="total"`; this way you can remove the `not self.report_duration` from the check in `update_testcase_duration`

nicoddemus:
> this should be a ini option

I was wondering the same thing

src/_pytest/junitxml.py
aparamon: That was the initial version, but then I thought that duplicating const "total" at https://github.com/aparamon/pytest/blob/cfbd387a5d7a27a3e7dfb754d7cd4482e931822b/testing/test_junitxml.py#L748
would be less desirable. (Also overall, using `None` as default value looks good interface.)

aparamon:
Thanks for the review @nicoddemus, @RonnyPfannschmidt!
I'll try harder to install `pre-commit` -- for some reason it keeps failing on my box.
Will switch to ini option and update the docs/changelog/authors shortly.

PR #4580, opened 2018-12-29
Improve detailed summary report docs

adamchainz:
The existing examples had 0 tests collected so didn't show the actual summary report. Also I added a section explaining the difference between `p` and `P`.


adamchainz:
For future ref here's the test file I used to generate the output, commenting out different tests to make the examples minimal:

```python
import pytest


def test_pass():
    assert 1


def test_pass_with_output():
    print('Passing test')
    assert 1


def test_fail():
    assert 0


def test_error(unknown_fixture):
    pass


@pytest.mark.skip(reason="Example")
def test_skipped():
    assert 1


@pytest.mark.xfail
def test_xfail():
    assert 0


@pytest.mark.xfail
def test_xpass():
    assert 1
```

adamchainz:
@tolomea you might want to take a look at this since you recommended me to check `pytest -ra`

tolomea:
Looks good :)

PR #4598, opened 2019-01-04
Update copyright year

nicoddemus:
This for me has to be the least rewarding contribution yet... 😁


nicoddemus:
Codecov is clearly buggy, there's no way this code could have had any effect on the coverage.

Perhaps we should consider other alternatives?

nicoddemus:
Also trying to access the site to see the report more often than not just times out... very frustrating. 😕

Issue #4631, opened 2019-01-10
Assertion rewrite breaks for objects that reimplement `__getattr__`


AnjoMan:
Got stuck on this for a bit while I was testing an object I wrote which implemented `__getattr__` -- i'd forgotten to raise an `AttributeError` if I couldn't fetch the given attribute.

If you do an assertion like this with the property access directly in the assertion statement, pytest does a call to `object.__name__` to try and make a nicer assertion message, and when my `__getattr__` fails the stack trace ends up in the test error message, even though it wasn't the reason for the test to fail.

I've written a test and identified a fix: https://github.com/pytest-dev/pytest/pull/4632

PR #4632, opened 2019-01-10
Assertion rewrite breaks for objects that reimplement `__getattr__`

AnjoMan:
This fix addresses #4631.

### DIFF
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -180,6 +180,27 @@ class TestAssertionRewrite(object):
 
         assert getmsg(f, {"cls": X}) == "assert cls == 42"
 
+    def test_dont_rewrite_if_hasattr_fails(self):
+        class Y(object):
+            """ A class whos getattr fails, but not with `AttributeError` """
+
+            def __getattr__(self, attribute_name):
+                raise KeyError()
+
+            def __repr__(self):
+                return "Y"
+
+            def __init__(self):
+                self.foo = 3
+
+        def f():
+            assert cls().foo == 2  # noqa
+
+        message = getmsg(f, {"cls": Y})
+        assert "assert 3 == 2" in message
+        assert "+  where 3 = Y.foo" in message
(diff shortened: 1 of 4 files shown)


testing/test_assertrewrite.py
```diff
@@ -180,6 +180,27 @@ class X(object):
 
         assert getmsg(f, {"cls": X}) == "assert cls == 42"
 
+    def test_dont_rewrite_if_hasattr_fails(self):
```
nicoddemus: Interesting, this test passes on `features` just fine on `py36`, but indeed fails on `py27`. 👍

4631.bugfix.rst
```diff
@@ -0,0 +1 @@
+Don't rewrite assertion when __getattr__ is broken
```
nicoddemus: Small nitpick:

```rst
Don't rewrite assertion when ``__getattr__`` is broken.
```

testing/test_assertrewrite.py
asottile: This is due to the hasattr changes in python 3 - - perhaps we should make a note here so this gets cleaned up when dropping python 2

4631.bugfix.rst
AnjoMan: Thanks! Just realized i also did not put this file in the `changelog` folder.

testing/test_assertrewrite.py
AnjoMan: Can you explain a bit more about what we want to make a note of? As I understand it, this fixes behaviour that we'd only see after [python 3.2](https://docs.python.org/dev/whatsnew/3.2.html), when `hasattr` stopped catching errors that aren't `AttributeError`.

asottile: oh right, I was thinking the code was in the other direction -- carry on this is fine 👍

AnjoMan:
@nicoddemus I think this is ready to go now.

nicoddemus:
Yes, thanks again!

PR #4651, opened 2019-01-17
Display --help/--version with ArgumentErrors

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/2753.

TODO:

- [x] target features
- [x] changelog

### DIFF
--- a/src/_pytest/config/argparsing.py
+++ b/src/_pytest/config/argparsing.py
@@ -1,12 +1,10 @@
 import argparse
-import sys as _sys
 import warnings
-from gettext import gettext as _
 
 import py
 import six
 
-from ..main import EXIT_USAGEERROR
+from _pytest.config.exceptions import UsageError
 
 FILE_OR_DIR = "file_or_dir"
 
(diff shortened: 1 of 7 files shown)


src/_pytest/config/argparsing.py
```diff
@@ -337,14 +336,15 @@ def __init__(self, parser, extra_info=None, prog=None):
...
-
-        Prints a usage message incorporating the message to stderr and
-        exits.
-        Overrides the method in parent class to change exit code"""
-        self.print_usage(_sys.stderr)
+        """Transform argparse error message into UsageError."""
         args = {"prog": self.prog, "message": message}
-        self.exit(EXIT_USAGEERROR, _("%(prog)s: error: %(message)s\n") % args)
+        msg = _("%(prog)s: error: %(message)s\n") % args
+        msg = msg.rstrip("\n")  # to not require re-translation?!
```
blueyed: We're not using gettext translations, are we?
Then the "\n" in the msg could be dropped already.

blueyed: Found the reason: https://github.com/pytest-dev/pytest/pull/3925/files#r215195026.
I think it can be removed then.

nicoddemus: Indeed.

PR #4652, opened 2019-01-18
Add __repr__ for RunResult

blueyed:
TODO:

- [x] test

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -398,6 +398,12 @@ class RunResult(object):
         self.stderr = LineMatcher(errlines)
         self.duration = duration
 
+    def __repr__(self):
+        return (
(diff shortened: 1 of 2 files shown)


src/_pytest/pytester.py
```diff
@@ -402,6 +402,12 @@ def __init__(self, ret, outlines, errlines, duration):
         self.stderr = LineMatcher(errlines)
         self.duration = duration
 
+    def __repr__(self):
+        return (
+            "<RunResult ret=%r len(stdout.lines)=%d len(stderr.lines)=%d duration=%.2fs>"
+            % (self.ret, len(self.stdout.lines), len(self.stderr.lines), self.duration)
```
blueyed: len appears to 1 always (empty line(s)), but appears to be ok?!

nicoddemus: I guess that's fine. 👍

blueyed:
Added a test.

PR #4661, opened 2019-01-22
remove and ban py.io.BytesIO, py.process, py.path.local.sysfind

asottile:


nicoddemus:
👍

testing/test_capture.py
```diff
@@ -851,7 +853,7 @@ def test_write_bytes_to_buffer(self):
 
 
 def test_bytes_io():
-    f = py.io.BytesIO()
+    f = io.BytesIO()
```
thisch: This test only tests code from the stdlib and can be removed IMO.

asottile: lol true, want to make a patch for it?

thisch: Sure. I'll create a PR

Issue #4667, opened 2019-01-22
The short test summary info should use `pytest_report_teststatus` hook data


wimglenn:
https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_report_teststatus 

```
@hookspec(firstresult=True)
def pytest_report_teststatus(report):
    """ return result-category, shortletter and verbose word for reporting.

    Stops at first non-None result, see :ref:`firstresult` """
```

Short test summary info that you get (e.g. with `pytest -rap`) should use the hook result.

Instead it uses data from a weird place - some hardcoded strings in a seemingly unrelated module [`_pytest/skipping.py`](https://github.com/pytest-dev/pytest/blob/4.1.1/src/_pytest/skipping.py#L289-L298).

nicoddemus:
Hi @wimglenn,

Indeed, at first glance you are correct. `skipping.pytest_terminal_summary` should call `pytest_report_teststatus`, which can be implemented somewhere else to use that mapping.

wimglenn:
Well, I had an attempt at it. Unfortunately, there is a weird decision to make - should the symbols in short test summary be matching the symbols in the test report (i.e. what the hook has otherwise returned), or should we override hook again, at the cost of being inconsistent, in order to get backwards compatibility there. Because in the report it's currently defaulting to `FAILED` but in the short test summary it was `FAIL`.  Similarly, in the report there's `xfail` and in the short test summary it's `XFAIL`.

wimglenn:
Also https://github.com/pytest-dev/pytest/commit/9711e335d976b3030b667636a6666019c430b52c seems a violation of the hook spec - I think that should really be done differently.

PR #4668, opened 2019-01-23
short report summary to use report teststatus data

wimglenn:
`pytest_terminal_summary` uses result from `pytest_report_teststatus` hook, rather than hardcoded strings

Fix https://github.com/pytest-dev/pytest/issues/4667

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -376,8 +376,11 @@ class TerminalReporter(object):
             return
         running_xdist = hasattr(rep, "node")
         if markup is None:
-            if rep.passed:
+            was_xfail = hasattr(report, "wasxfail")
+            if rep.passed and not was_xfail:
                 markup = {"green": True}
+            elif rep.passed and was_xfail:
+                markup = {"yellow": True}
             elif rep.failed:
                 markup = {"red": True}
             elif rep.skipped:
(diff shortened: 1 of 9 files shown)


src/_pytest/terminal.py
```diff
@@ -806,8 +809,7 @@ def summary_errors(self):
             self.write_sep("=", "ERRORS")
             for rep in self.stats["error"]:
                 msg = self._getfailureheadline(rep)
-                if not hasattr(rep, "when"):
```
nicoddemus: 👍

testing/test_skipping.py
```diff
@@ -1084,7 +1084,7 @@ def test_func():
         result = testdir.runpytest("-rxs")
         result.stdout.fnmatch_lines(
             """
-            *XFAIL*
```
nicoddemus: Not sure about this change, all other verbose words are all caps, shouldn't we follow suit here?

wimglenn: Yes, I was meaning to ask about that. I would like to change to `XFAIL` too, but this would be a change in the test report aswell (currently no change there, I'm only changing the short test summary). I'm unaware of the history about why `xfail` is currently lowercase in test report and all the others are uppercase.

On a related note, should the short test summary be colorized too?

nicoddemus: > On a somewhat related note, should the short test summary be colorized too?

Hah you read my mind, I was thinking just yesterday that colorizing the uppercase word (`XFAIL`, `FAIL`, etc) in the summary would make it easier to read. Please go ahead and let's see how it looks. 😄

nicoddemus: > When I get home I'll add a commit to change the xfail to XFAIL

Awesome, thanks!

wimglenn:
@nicoddemus Changed xfail to XFAIL.  Actually only 1 test needed updating, in `testing/test_terminal.py`.

nicoddemus:
> Changed xfail to XFAIL. Actually only 1 test needed updating, in testing/test_terminal.py.

Great, LGTM. 👍 

Thanks again!

PR #4682, opened 2019-01-27
Document parametrizing conditional raises

### LINKED ISSUE #4324: document "raises/not raises" in parametersets via contextlib.supress / contextlib.ExitStack / contextlib.noopcontext

arel:
This PR addresses issues #4324 ~~and #1830~~, which documents how to use `pytest.raises` conditionally together with `pytest.mark.parametrize`.

I added documentation on how to use a null context manager to complement `pytest.raises` as well as added tests for the documented code.

Based on the discussion in #4679, I removed `does_not_raise` from the core library, and instead will rely on users copying and pasting the boilerplate.

I used:

```python
@contextmanager
def does_not_raise():
    yield
```

instead of

```python
from contextlib import nullcontext as does_not_raise
```

because it works from Python 2.5 and up without installing additional libraries or backporting, whereas `nullcontext` is only introduced in Python 3.7. 

Closes #4324

---

just a guideline):


doc/en/example/parametrize.rst
```diff
@@ -559,3 +559,37 @@ As the result:
...
+
+Use :func:`pytest.raises` with the
+:ref:`pytest.mark.parametrize ref` decorator to write parametrized tests
+in which some tests raise exceptions and others do not.
+
+It is helpful to define a function such as ``does_not_raise`` to serve
+as a complement to ``raises``. For example::
+
+    from contextlib import contextmanager
+    import pytest
```
nicoddemus: I think it is worth mentioning `nullcontext` for Python 3.7+ users here.

arel: Sounds good; will do.

arel:
@asottile I don't quite follow. The recipe you linked to (below) is more verbose and requires backporting `contextlib` and only [supports](https://pypi.org/project/contextlib2/) Python 2.6 and above

```python
if sys.version_info >= (3, 7):
    from contextlib import nullcontext
elif sys.version_info >= (3, 3):
    from contextlib import ExitStack as nullcontext
else:
    from contextlib2 import ExitStack as nullcontext
```

whereas, the recipe below uses only built-in features and supports Python 2.5 and above.

```python
@contextmanager
def does_not_raise():
    yield
```

doc/en/example/parametrize.rst
arel: Done in 8a1afe421357439dd15d9a16448ecbdd69fdab85

asottile:
@arel pytest doesn't support 2.6 and the recipe improves ootb support for other versions

nicoddemus:
IMHO the current example is good enough. We could also post @asottile's solution beside the current one, but I feel this would add complexity to what was supposed to be a simple example... but if @asottile feels strongly about it, let's add it alongside it then.

asottile:
I don't even think you need the recipe, but the important information is basically (something like this):

(rst syntax not necessarily valid, I can never remember how many `:`s I need)

```rst
If you need to parametrize based on "raising" you can implement a ``does_not_raise`` contextmanager using a noop contextmanager.

If you're only supporting python3.7+ you can use:

.. code-block: python

    from contextlib import nullcontext as does_not_raise

If you're only supporting python3.3+ you can use:

.. code-block: python

    from contextlib import ExitStack as does_not_raise

If you need to support older versions you can either use the `contextlib2` backport:

.. code-block: python

    from contextlib2 import ExitStack as does_not_raise

or you can define your own noop contextmanager

.. code-block: python

    from contextlib import contextmanager

    @contextmanager
    def does_not_raise():
        pass
```

I might even go so far as to put some opinions in there as well.  Something like "You might be over-engineering your tests with parametrize a bit -- it will likely read better if you split your testcases into successful cases and failure cases instead of trying to bake conditional logic into ``parametrize``"

nicoddemus:
@asottile perhaps you should mark this PR as "request changes" so it is clear in the Pull Requests view that this PR still has required work? Thanks!

asottile:
Correct, but it's better to use the builtin things where you have to write zero code.  pytest doesn't care about python 2.5 / 2.6 / 3.0 / 3.1 / 3.2 / 3.3 support and soon won't care about python2.x support and so documenting a pattern in a way that's soon to become obsolete should be the last goal.  Put the newest things first, and the "ok well if you absolutely must support old stuff use this" last please :)

arel:
Great! I don't have write access, so I'll leave merging up to one of you. Thanks!

asottile:
Oops I forgot

blueyed:
Thanks for your contribution!

I just wondered if / why this wasn't done for the features branch?

asottile:
> 
> 
> Thanks for your contribution!
> 
> I just wondered if / why this wasn't done for the features branch?

it's entirely docs / tests -- it'll make it into `features` when we merge `master` into features 👍

blueyed:
> it's entirely docs / tests

Oh, missed that afe9fd5 was "reverted"/removed again (when looking at new commits on master).

PR #4693, opened 2019-01-30
fixes for flake8 3.7

asottile:
I opted to _remove_ the PEP 484 type comment annotations for now since we're not linting for mypy

Also upgraded the pre-commit hooks since `virtualenv` has been released fixing the weird `black` issue that was happening before

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: doc/en/example/py2py3/test_py2.py
 repos:
 -   repo: https://github.com/ambv/black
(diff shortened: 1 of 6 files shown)


blueyed:
> I opted to _remove_ the PEP 484 type comment annotations for now since we're not linting for mypy

Does flake8 complain about them?

asottile:
> > I opted to _remove_ the PEP 484 type comment annotations for now since we're not linting for mypy
> 
> Does flake8 complain about them?

pyflakes 2.1 now recognizes PEP 484 `# type: ...` comments

Issue #3974, opened 2018-09-13
Confusion between message and match parameters in pytest.raises


rth:
After analyzing several code bases, it appears there is frequent confusion by users between the `match` and `message` parameters in `pytest.raises` (and consequentially `pytest.warns`). 

In most cases, users probably would want `match` but use `message` by accident, because is it visually similar and the naming does not differentiate their functionality enough. As a result the written test produces false positive which is quite dangerous.

For instance, here are instances of such errors in popular open-source projects,
 - https://github.com/pandas-dev/pandas/pull/22681#discussion_r217181448
 - https://github.com/scipy/scipy/pull/9264
 - https://github.com/scikit-learn/scikit-learn/pull/12001

the fact that these cases ended up in the code despite fairly thorough review is telling.

There was also an earlier issue illustrating this confusion in https://github.com/pytest-dev/pytest/issues/3350

Searching on github for `"pytest.raises(ValueError, message="` for instance, also shows [a number of examples](https://github.com/search?p=3&q=%22pytest.raises%28ValueError%2C+message%22&type=Code") that meant to use `"match"` instead of `"message"` (though sometimes it's hard to tell).

One possibility could be to deprecate the `message` parameter and rename it to something more explicit. I saw there were some deprecations related to the use of `node.warn` in 3.8, not sure if that's related. However, following the same logic even if `pytest.raises(ValueError('expected message'))` worked it would still not address all the existing code that currently uses the `message` parameter incorrectly.

RonnyPfannschmidt:
after checking we should deprecate the `message` name and rename it to something like `failure_message`

rth:
Thanks for the feedback @RonnyPfannschmidt !

Naming is tricky: at some level even `pytest.raises(ValueEror, failure_message='some message')` could be interpreted as the expected failure message of that `ValueError`. Something like `show_failure_mesage` might be bit more explicit, but it's probably too long and not very practical..

RonnyPfannschmidt:
@rth yes - the hardness of naming this the bane here ^^

asottile:
big +1 to this, we have an internal linter that looks for `pytest.raises(...message=` and just bans it -- maybe we should consider removing the parameter altogether

nicoddemus:
Hi @rth thanks for the excellent description and examples, they are extremely helpful in making the case about doing something about this matter.

So we have two proposals:

[...]

PR #4709, opened 2019-02-02
Document how to customize test failure message for missing warnings

namurphy:
I added a section to the docs on how to customize test failure messages for missing warnings when using `pytest.warns`, following some discussion in #3974 and adapting an example from @asottile.

### DIFF
--- a/doc/en/warnings.rst
+++ b/doc/en/warnings.rst
@@ -233,7 +233,7 @@ You can also use it as a contextmanager::
 .. _warns:
 
 Asserting warnings with the warns function
------------------------------------------------
+------------------------------------------
 
 .. versionadded:: 2.8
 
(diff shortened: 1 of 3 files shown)


doc/en/warnings.rst
```diff
@@ -329,6 +329,26 @@ warnings, or index into it to get a particular recorded warning.
...
+
+.. code-block:: python
+
+    def test():
+        with pytest.warns(Warning) as record:
+            f()
+            if not record:
+                pytest.fail('Expected a warning!')
+
+If no warnings are issued when calling ``f``, then ``not record`` will
```
asottile: ` ``True`` ` you mean ;)

nicoddemus: Fixed linting and this while at it. 👍

Issue #3711, opened 2018-07-23
Add ability to use globs when using --ignore


machinekoder:
There is no command line option for pytest to ignore a specific directory pattern.

For example:

```
-- code.py
-- tests
---- test_code.py
---- integration
------ test_code_integration.py
```

I would like to disable all integration tests.

The argument `--ignore=integration` does not work.

I have also tried `-k="not integration"` which works, but for some reason is ignored when executing pytest via `pytest.main`.

Note: Of couse I could specify the full path using the `--ignore` argument, but the actual project is much more complex and contains many `tests` and `integration` folders.

nicoddemus:
Thanks @machinekoder for the suggestion.

I've labeled this as an improvement, given that what you describe a feature request. Your title alone suggests that this is a bug though, could you please rephrase it? Something along the lines of *Add ability to use globs when using --ignore*, which better conveys that this is a feature request. Thanks!

asottile:
@machinekoder could you show how you're invoking `pytest.main` - - I expect it to work the same there too

PR #4722, opened 2019-02-05
Add ability to use globs when using --ignore

fetzerch:
This changes the `--ignore` option to allow Unix-style wildcards so
that `--ignore=integration*` excludes all tests that reside in files
starting with `integration`.

Fixes: #3711

### DIFF
--- a/doc/en/example/pythoncollection.rst
+++ b/doc/en/example/pythoncollection.rst
@@ -269,3 +269,17 @@ file will be left out:
     collected 0 items
 
     ======================= no tests ran in 0.12 seconds =======================
+
+It's also possible to ignore files based on Unix shell-style wildcards by adding
+patterns to ``collect_ignore_glob``.
+
+The following example ``conftest.py`` ignores the file ``setup.py`` and in
+addition all files that end with ``*_py2.py`` when executed with a Python 3
+interpreter::
+
+    # content of conftest.py
+    import sys
+
+    collect_ignore = ["setup.py"]
+    if sys.version_info[0] > 2:
+        collect_ignore_glob = ["*_py2.py"]
--- a/doc/en/example/pythoncollection.rst
+++ b/doc/en/example/pythoncollection.rst
@@ -41,6 +41,9 @@ you will see that ``pytest`` only collects test-modules, which do not match the 
 
     ========================= 5 passed in 0.02 seconds =========================
 
+The ``--ignore-glob`` option allows to ignore test file paths based on Unix shell-style wildcards.
+If you want to exclude test-modules that end with ``_01.py``, execute ``pytest`` with ``--ignore-glob='*_01.py'``.
+
 Deselect tests during test collection
 -------------------------------------
 
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -4,6 +4,7 @@ from __future__ import division
 from __future__ import print_function
 
 import contextlib
+import fnmatch
 import functools
 import os
 import pkgutil
@@ -116,6 +117,12 @@ def pytest_addoption(parser):
         action="append",
         metavar="path",
         help="ignore path during collection (multi-allowed).",
+    )
+    group.addoption(
+        "--ignore-glob",
+        action="append",
+        metavar="path",
+        help="ignore path pattern during collection (multi-allowed).",
     )
     group.addoption(
         "--deselect",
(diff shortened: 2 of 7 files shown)


src/_pytest/main.py
```diff
@@ -293,7 +293,9 @@ def pytest_ignore_collect(path, config):
     if excludeopt:
         ignore_paths.extend([py.path.local(x) for x in excludeopt])
 
-    if py.path.local(path) in ignore_paths:
+    if any(
+        py.path.local(path).fnmatch(str(ignore_path)) for ignore_path in ignore_paths
```
nicoddemus: Perhaps we should use `fnmatch.fnmatch(six.text_type(path))` instead? IIRC `py.path`'s `fnmatch` differs from `fnmatch.fnmatch` slightly, and given that we want to eventually get rid of `py.path`, we should avoid depending on its semantics further.

nicoddemus:
> even tho this is a really nice change, its breaking backward compatibility, i propose adding a `--ignore-glob` parameter

I agree; while I think few users are using `*` or `.` in that setting, `[]` seems more likely.

fetzerch:
I'll update the PR to introduce a new `--ignore-glob` parameter. While I'm at it, it appears to be relatively easy to also add support for `collect_ignore_glob` in `conftest.py`. Does this make sense to you?

nicoddemus:
> While I'm at it, it appears to be relatively easy to also add support for `collect_ignore_glob` in `conftest.py`. Does this make sense to you?

Sorry, what do you mean, a new option, or hook, ?

fetzerch:
PR is updated now with the `--ignore-glob` parameter and support for `collect_ignore_glob`.

> > While I'm at it, it appears to be relatively easy to also add support for `collect_ignore_glob` in `conftest.py`. Does this make sense to you?
> 
> Sorry, what do you mean, a new option, or hook, ?

See my [second commit](https://github.com/pytest-dev/pytest/pull/4722/commits/e899dcc968206b820aa762e915a4e7b6b507208d). I thought it might be useful and implemented it right away.

blueyed:
The option `collect_ignore_glob` could also be used via `-o …` then, i.e. there wouldn't be the need for a separate CLI option/arg.

testing/test_collection.py
```diff
@@ -374,6 +374,26 @@ def pytest_configure(config):
...
+            def pytest_configure(config):
+                if config.getvalue("XX"):
+                    collect_ignore_glob[:] = []
+        """
+        )
+        testdir.makepyfile(test_world="def test_hello(): pass")
+        testdir.makepyfile(test_welt="def test_hallo(): pass")
+        result = testdir.runpytest()
+        assert result.ret == EXIT_NOTESTSCOLLECTED
+        assert "passed" not in result.stdout.str()
```
nicoddemus: Probably better to explicitly use `result.stdout.fnmatch_lines('* no tests collected *')` as this is more explicit

testing/test_collection.py
```diff
@@ -374,6 +374,26 @@ def pytest_configure(config):
...
+        result = testdir.runpytest("--XX")
+        assert result.ret == 0
+        assert "passed" in result.stdout.str()
```
nicoddemus: Ditto here, probably better to check for `* 2 passed*` as this is more clear.

doc/en/example/pythoncollection.rst
```diff
@@ -266,3 +269,17 @@ file will be left out:
     collected 0 items
 
     ======================= no tests ran in 0.12 seconds =======================
+
+It's also possible to ignore files based on Unix shell-style wildcards by adding
+patterns to ``collect_ignore_glob``.
```
nicoddemus: Completely forgot about `collect_ignore`. 😛 

Could you please add `collect_ignore` and `collect_ignore_glob` to the [reference docs](https://docs.pytest.org/en/latest/reference.html)? They should go under the "Special Variables" section.

fetzerch: Sure, done.

nicoddemus:
> The option `collect_ignore_glob` could also be used via `-o …` then, i.e. there wouldn't be the need for a separate CLI option/arg.

I agree with this idea for new options, but given that we have `--collect-ignore` already, for consistency I think it is OK to have `--collect-ignore-glob` as well. 👍

blueyed:
@RonnyPfannschmidt 
Please re-review.

blueyed:
Thank you!

PR #4734, opened 2019-02-06
Refactor _setup_cli_logging code

thisch:
Change the indentation in _setup_cli_logging by moving the
self._log_cli_enabled check outside of the _setup_cli_logging method.

just a guideline):

### DIFF
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -420,54 +420,53 @@ class LoggingPlugin(object):
 
         self.log_cli_handler = None
 
-        self._setup_cli_logging()
-
-    def _setup_cli_logging(self):
-        terminal_reporter = self._config.pluginmanager.get_plugin("terminalreporter")
-
-        if self._log_cli_enabled() and terminal_reporter is not None:
-            # FIXME don't set verbosity level and derived attributes of
-            # terminalwriter directly
-            terminal_reporter.verbosity = self._config.option.verbose
-            terminal_reporter.showheader = terminal_reporter.verbosity >= 0
-            terminal_reporter.showfspath = terminal_reporter.verbosity >= 0
-            terminal_reporter.showlongtestinfo = terminal_reporter.verbosity > 0
-
-            capture_manager = self._config.pluginmanager.get_plugin("capturemanager")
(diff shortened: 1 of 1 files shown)


src/_pytest/logging.py
```diff
@@ -420,54 +420,53 @@ def __init__(self, config):
...
-            )
-            self.log_cli_handler = log_cli_handler
-            self.live_logs_context = lambda: catching_logs(
-                log_cli_handler, formatter=log_cli_formatter, level=log_cli_level
+        config = self._config
+        terminal_reporter = config.pluginmanager.get_plugin("terminalreporter")
+        capture_manager = config.pluginmanager.get_plugin("capturemanager")
+
+        if terminal_reporter is None or capture_manager is None:
+            # wouldn't it be better if we raise here?
```
nicoddemus: Not sure, those plugins might be disabled for other reasons (for example xdist disables terminal writer).

src/_pytest/logging.py
```diff
@@ -420,54 +420,53 @@ def __init__(self, config):
...
+            return
+
+        # FIXME don't set verbosity level and derived attributes of
```
nicoddemus: Please add a comment here pointing to #4733

thisch: I'll take care of #4733, once this is merged.

nicoddemus: Oh great, no need to update the comment right now then. 👍

thisch:
> Looks good, we also need a changelog entry. Seems to me like a "bugfix" about disabling logging automatically in case capture is also disabled, something which did not work previously anyway?

I just had a look again and checked what happens if the capturemanager plugin is disabled. The live logging still works (because the _LiveLoggingStreamHandler supports that self.capture_manager is None.)

So this PR actually does not fix a bug. Therefore, I decided not to write a bugfix changelog entry. 

I'll adapt the code s.t. the current behavior is not changed.

nicoddemus:
> So this PR actually does not fix a bug. Therefore, I decided not to write a bugfix changelog entry.

I agree, thanks for the explanation. 👍

PR #4741, opened 2019-02-07
Remove terminal_reporter workaround from logging.py

thisch:
The workaround was removed from the logging module by creating python
properties for verbosity related settings in the terminalreporter.

Closes: #4733

just a guideline):

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -222,12 +222,9 @@ class TerminalReporter(object):
         import _pytest.config
 
         self.config = config
-        self.verbosity = self.config.option.verbose
-        self.showheader = self.verbosity >= 0
-        self.showfspath = self.verbosity >= 0
-        self.showlongtestinfo = self.verbosity > 0
         self._numcollected = 0
         self._session = None
+        self._showfspath = None
 
         self.stats = {}
         self.startdir = py.path.local()
@@ -254,6 +251,28 @@ class TerminalReporter(object):
         if self.config.getoption("setupshow"):
             return False
         return self.config.getini("console_output_style") in ("progress", "count")
+
+    @property
+    def verbosity(self):
+        return self.config.option.verbose
+
+    @property
+    def showheader(self):
+        return self.verbosity >= 0
+
+    @property
+    def showfspath(self):
+        if self._showfspath is None:
+            return self.verbosity >= 0
+        return self._showfspath
+
+    @showfspath.setter
+    def showfspath(self, value):
+        self._showfspath = value
+
+    @property
+    def showlongtestinfo(self):
+        return self.verbosity > 0
 
     def hasopt(self, char):
         char = {"xfailed": "x", "skipped": "s"}.get(char, char)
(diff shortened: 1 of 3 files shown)


src/_pytest/terminal.py
```diff
@@ -255,6 +251,22 @@ def _determine_show_progress_info(self):
             return False
         return self.config.getini("console_output_style") in ("progress", "count")
 
+    @property
+    def verbosity(self):
+        return self.config.option.verbose
+
+    @property
+    def showheader(self):
+        return self.verbosity >= 0
```
thisch: This property is only used once in the codebase. Not sure if it is worth the effort to introduce this property. I would remove it and add a comment to the place where it is used. 
I doubt that we can remove it, since it is an API change, right?

src/_pytest/terminal.py
```diff
@@ -255,6 +251,22 @@ def _determine_show_progress_info(self):
...
+
+    @property
+    def showfspath(self):
+        return self.verbosity >= 0
+
+    @property
+    def showlongtestinfo(self):
+        return self.verbosity > 0
```
thisch: Same as above

src/_pytest/terminal.py
nicoddemus: Agree, let's keep it, it is not like it is a huge burden anyway. 👍

src/_pytest/terminal.py
nicoddemus: Agree, let's keep it, it is not like it is a huge burden anyway. 👍

nicoddemus:
Please add a "trivial" CHANGELOG entry that those internal attributes are now read-only properties. 

Hmmm this might break some code which was changing those properties. Of course that code would probably be working by accident like we saw was the case of the changes on the logging module.

src/_pytest/terminal.py
thisch: No, they are not really a burdern.

thisch:
Seems as if ``pytest-xdist`` needs to change the showfspath attribute:

```
INTERNALERROR>   File "/home/travis/build/pytest-dev/pytest/.tox/py37-xdist/lib/python3.7/site-packages/xdist/plugin.py", line 173, in pytest_configure
INTERNALERROR>     tr.showfspath = False
```

converting the read only properties to a read/write properties is unfortunately not as obvious as it may seem, since setting showfspath must not change the value of e.g. showheader.

What do you think about my fix for this? (If all tests pass I'll also add setter properties for the other show properties, but not for verbosity.

src/_pytest/terminal.py
```diff
@@ -255,6 +252,28 @@ def _determine_show_progress_info(self):
...
+    @property
+    def verbosity(self):
+        return self.config.option.verbose
+
+    @property
+    def showheader(self):
+        return self.verbosity >= 0
+
+    @property
+    def showfspath(self):
```
nicoddemus: If we want to allow this to change (as it seems to be the case), then we might as well turn this into a normal attribute initialized with `self.verbosity >= 0`. 😉

nicoddemus: Hmm actually nevermind, this has different semantics: unless someone changes `showfspath`, it will always be derived from `config.verbosity`. 👍

Issue #4344, opened 2018-11-08
StopIteration error in CircleCI using split tests pattern


Morendil:
These lines in release **3.10.0**:
https://github.com/pytest-dev/pytest/blob/5ac4eff09b8514a5b46bdff464605a60051abc83/src/_pytest/main.py#L550-L551

…are causing a [regression when running tests](https://circleci.com/gh/openfisca/openfisca-france/5746) under CircleCI (with the Python 3.7 image) with the recommended split test pattern and globbing:

```bash
pytest `circleci tests glob "tests/**/*.py" | circleci tests split`
```

(We are not seeing the same failure in Python 2 builds.)

A minimal reproduction consists simply of:
```bash
touch __init__.py
pytest __init__.py

============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.7.0, pytest-3.10.0, py-1.7.0, pluggy-0.8.0
rootdir: /app, inifile: setup.cfg
collected 0 items / 1 errors

=================================================================================================================== ERRORS ===================================================================================================================
_____________________________________________________________________________________________________________ ERROR collecting  ______________________________________________________________________________________________________________
/usr/local/lib/python3.7/site-packages/_pytest/main.py:568: in _collect
    yield next(m[0].collect())
E   StopIteration

The above exception was the direct cause of the following exception:
/usr/local/lib/python3.7/site-packages/_pytest/runner.py:212: in __init__
    self.result = func()
/usr/local/lib/python3.7/site-packages/_pytest/runner.py:272: in <lambda>
    call = CallInfo(lambda: list(collector.collect()), "collect")
/usr/local/lib/python3.7/site-packages/_pytest/main.py:478: in collect
    for x in self._collect(arg):
E   RuntimeError: generator raised StopIteration
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
========================================================================================================== 1 error in 0.30 seconds ===========================================================================================================
```

Our `pip list` is simply a fresh install of PyTest:
```
Package        Version
[...]

PR #4745, opened 2019-02-08
Fix handling of pkg init and test file via args

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/4344#issuecomment-441095934

TODO:

- [x] changelog

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -582,7 +582,7 @@ class Session(nodes.FSCollector):
                 col = self._node_cache[argpath]
             else:
                 collect_root = self._pkg_roots.get(argpath.dirname, self)
-                col = collect_root._collectfile(argpath)
+                col = collect_root._collectfile(argpath, handle_dupes=False)
                 if col:
                     self._node_cache[argpath] = col
             m = self.matchnodes(col, names)
--- a/testing/test_collection.py
+++ b/testing/test_collection.py
@@ -1157,3 +1157,32 @@ def test_collectignore_via_conftest(testdir, monkeypatch):
 
     result = testdir.runpytest()
     assert result.ret == EXIT_NOTESTSCOLLECTED
+
+
+def test_collect_pkg_init_and_file_in_args(testdir):
+    subdir = testdir.mkdir("sub")
+    init = subdir.ensure("__init__.py")
+    init.write("def test_init(): pass")
+    p = subdir.ensure("test_file.py")
+    p.write("def test_file(): pass")
+
+    # NOTE: without "-o python_files=*.py" this collects test_file.py twice.
+    # This changed/broke with "Add package scoped fixtures #2283" (2b1410895)
+    # initially (causing a RecursionError).
+    result = testdir.runpytest("-v", str(init), str(p))
+    result.stdout.fnmatch_lines(
+        [
+            "sub/test_file.py::test_file PASSED*",
+            "sub/test_file.py::test_file PASSED*",
+            "*2 passed in*",
+        ]
+    )
+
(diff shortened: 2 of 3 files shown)


blueyed:
Likely not the best fix, but appears to work for the added test at least.

src/_pytest/main.py
```diff
@@ -592,7 +592,7 @@ def filter_(f):
                 col = self._node_cache[argpath]
             else:
                 collect_root = self._pkg_roots.get(argpath.dirname, self)
-                col = collect_root._collectfile(argpath)
+                col = collect_root._collectfile(argpath, handle_dupes=False)
```
blueyed: That's wrong.. and will not handle dupes at all then.
Also the output of the test should be matched (includes `sub/test_file.py` twice).

blueyed:
This fails with 3.9.3 also already:
```
platform linux -- Python 3.7.2, pytest-3.9.3, py-1.7.0, pluggy-0.8.1 -- /home/daniel/Vcs/pytest/.venv/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/daniel/Vcs/pytest/.hypothesis/examples')
rootdir: /tmp/pytest-of-daniel/pytest-270/test_collect_pkg_init_and_file_in_args0, inifile:
plugins: forked-1.0.1, hypothesis-4.0.0
collecting ... collected 3 items

sub/test_file.py::test_file PASSED                                       [ 33%]
sub/test_file.py::test_file PASSED                                       [ 33%]
sub/test_file.py::test_file PASSED                                       [ 33%]

=========================== 3 passed in 0.01 seconds ===========================
```

With 3.6.4 it looks better:
```
sub/__init__.py::test_init PASSED                                        [ 33%]
sub/test_file.py::test_file PASSED                                       [ 66%]
sub/test_file.py::test_file PASSED                                       [ 66%]
```

But percentage is still off there, and it should not run `test_file` twice (due to not using `--keepduplicates`) I assume?

blueyed:
Re collection the file twice: this is documented, but could be fixed/improved likely: https://github.com/pytest-dev/pytest/blame/5ca81596bbbb35da0b8b742afaf7ed2ef075ccda/doc/en/example/pythoncollection.rst#L76-L84

blueyed:
Needs `-o python_files=*.py` to handle `__init__.py` after all..

src/_pytest/main.py
blueyed: Likely works better with `add_to_dupes` (https://github.com/pytest-dev/pytest/pull/4748).

blueyed:
I think this is a good small fix for now.

testing/test_collection.py
```diff
@@ -1157,3 +1157,20 @@ def test_collectignore_via_conftest(testdir, monkeypatch):
 
     result = testdir.runpytest()
     assert result.ret == EXIT_NOTESTSCOLLECTED
+
+
+def test_collect_pkg_init_and_file_in_args(testdir):
```
nicoddemus: Hmm this test is passing for me without your patch (`master`). Can you double check that this is reproducing the problem for you please?

blueyed: Oh, snap - thanks for catching.

The difference/fix is only without `-o python_files=*.py` then - extended the test.
Not sure if this is really expected, but better then not collection anything I guess.

nicoddemus: I see, thanks. I checked and the new test does fail with `master`. 👍

blueyed:
Added changelog entry.
Let me know if it is good and I'd like to squash it.

blueyed:
Squashed.

PR #4746, opened 2019-02-08
Use isolated_build option in tox.ini

nicoddemus:
As per the excellent article by @gaborbernat:

https://www.bernat.tech/pep-517-518/


nicoddemus:
Merging as the freeze failure is unrelated: #4750

PR #4757, opened 2019-02-09
Travis: use pypy from Xenial, add pypy3

blueyed:


blueyed:
Included in https://github.com/pytest-dev/pytest/pull/4753.

nicoddemus:
Not waiting on AppVeyor as it does not apply there.

PR #4766, opened 2019-02-11
consider_preparse: remove leading space

blueyed:
This handles "-p no:sugar" passed in as single arg via
`testdir.runpytest`.

Ref: https://github.com/Frozenball/pytest-sugar/issues/167

### DIFF
--- a/changelog/4766.bugfix.rst
+++ b/changelog/4766.bugfix.rst
@@ -1,0 +1,1 @@
+Handle ``-p no:fooplugin`` when passed as single argument.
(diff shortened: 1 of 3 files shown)


blueyed:
> Please add a CHANGELOG entry.

Sure, would have done already, but there wasn't a number before the PR yet.. :/

blueyed:
Not really sure about this after all..  `-p no:foo` is usually not a single argument.

nicoddemus:
> Not really sure about this after all.. `-p no:foo` is usually not a single argument.

Hmmm indeed. We probably should not support this weird use case.

blueyed:
Ok.. it was me after all using it like this.. ;)

PR #4767, opened 2019-02-11
AppVeyor: use xdist for py?? envs, drop py27/py37

blueyed:
This is meant to help with AppVeyor builds being being too much.

[skip travis]


blueyed:
Should we keep/use py27/py37 without xdist only?

nicoddemus:
It seems this saved us some good 20 mins, great job!

On a related note, I have been meaning to try Azure Pipelines at least for the Windows jobs.

PR #4768, opened 2019-02-11
Avoid pkg_resources import at the top-level.

horta:

just a guideline):

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -14,7 +14,6 @@ import warnings
 
 import py
 import six
-from pkg_resources import parse_version
 from pluggy import HookimplMarker
 from pluggy import HookspecMarker
 from pluggy import PluginManager
@@ -815,6 +814,7 @@ class Config(object):
 
     def _checkversion(self):
         import pytest
+        from pkg_resources import parse_version
 
         minver = self.inicfg.get("minversion", None)
         if minver:
--- a/changelog/4768.trivial.rst
+++ b/changelog/4768.trivial.rst
@@ -1,0 +1,1 @@
+Avoid pkg_resources import at the top-level.


horta:
Reduce from 0.3s to 0.1s import time on my machine.

src/_pytest/config/__init__.py
```diff
@@ -14,7 +14,6 @@
 
 import py
 import six
-from pkg_resources import parse_version
```
asottile: I'm -0

- this won't actually help at all as we the plugin system in the critical path imports setuptools / pkg_resources
- I'd rather keep imports at the top

nicoddemus: Are there other alternatives for parsing versions that don't require `pkg_resources`? 

I remember `distutils` has functionality for that, but @RonnyPfannschmidt was strictly against using it. Wouldn't make sense to use `disutils` just for this bit of functionality? I understand it should be avoided in general, but we are talking about a small utility function of something which is already in the stdlib, not about plugging into `distutils` in a big way or something.

I believe another point was that we require `pkg_resources` anyway because of the plugin system, but given https://github.com/pytest-dev/pluggy/pull/190 it seems doable for us to move away from `pkg_resources` entirely.

asottile: pkg_resources itself uses `packaging` (vendored)

nicoddemus: Interesting... perhaps we can consider adding it as dependency if that would mean we can drop `pkg_resources` from `pytest` and `pluggy`?

blueyed:
Merged, since it was approved, and https://github.com/pytest-dev/pytest/pull/4768#discussion_r255995719 could be done in a follow-up.

Thanks!

asottile:
This doesn't actually improve anything (and obscures an import) so I was against merging it.  @blueyed let's resolve the conversation next time before merging <3

asottile:
Also it further conflicts my branch which rips all of this out

blueyed:
> This doesn't actually improve anything (and obscures an import) so I was against merging it. @blueyed let's resolve the conversation next time before merging <3

You were -0.

> Also it further conflicts my branch which rips all of this out

Sorry, but you did not state that you were working on this, and it should be trivial to resolve.

I was just going through the approved PRs, waiting for someone to approve https://github.com/pytest-dev/pytest/pull/4887 - so that I could rebase branches on it.

blueyed:
@asottile 
The next time please request changes.

asottile:
> @asottile
> The next time please request changes.

I try to avoid "Request changes" because it produces a big angry red box

blueyed:
> I try to avoid "Request changes" because it produces a big angry red box

Now you have a lila one ("Merged").. ;)

PR #4774, opened 2019-02-12
Constrain more_itertools for Python 2.7 compatibility

sambrightman:
Fixes #4772, #4770.

There was some discussion in #4770 that maybe this can be written slightly differently but it worked for me like this.

I assume this doesn't require a changelog entry.

PR #4775, opened 2019-02-12
Replace flatten() with chain.from_iterable

asottile:
flatten is an alias in more-itertools anyway

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 
 import functools
 import inspect
+import itertools
 import sys
 import warnings
 from collections import defaultdict
@@ -13,7 +14,6 @@ from collections import OrderedDict
 import attr
 import py
 import six
-from more_itertools import flatten
 
 import _pytest
 from _pytest import nodes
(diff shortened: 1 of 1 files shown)


asottile:
We only use more-itertools in 3 places, this place provided zero benefit as far as I can tell.  given the pain in #4770 and our extremely sparse usage of this library maybe we should factor it out?

nicoddemus:
Not sure... the alternative to dropping `more-itertools` would be for us to implement that ourselves, which I'm not a big fan of.

And to be fair, #4770 is not `more-itertools`' fault, but setuptools', and is being solved already by #4774. I would say we can consider removing it in the future if it causes us pain again in the future, but I would like to give it another chance if possible.

What do others think?

asottile:
> 
> 
> Not sure... the alternative to dropping `more-itertools` would be for us to implement that ourselves, which I'm not a big fan of.
> 
> And to be fair, #4770 is not `more-itertools`' fault, but setuptools', and is being solved already by #4774. I would say we can consider removing it in the future if it causes us pain again in the future, but I would like to give it another chance if possible.
> 
> What do others think?

We currently use 2 functions which I think are trivially implemented (not sure how m-i implements them, but here's my best attempt)

```python
def always_iterable(x):
    if isinstance(x, (six.text_type, bytes)):
        return (x,)
    else:
        try:
            return tuple(x)
         except TypeError:
            return (x,)


def collapse(val, tps=(type, six.text_type, bytes)):
    def inner(x):
        if isinstance(x, tps):
            yield x
        else:
             try:
                 it = iter(x)
             except TypeError:
                 yield x
             else:
                 for y in it:
                     for z in inner(y):
                         yield z
    return tuple(inner(val))
```

PR #4779, opened 2019-02-12
Set up CI with Azure Pipelines [skip travis] [skip appveyor]

nicoddemus:
Just a few environments for now to see how it will behave for a few days


nicoddemus:
😓

nicoddemus:
I think we should merge this and see how it goes. We can ignore any failures with the the purposes of validating PRs until it is made official.

Issue #4782, opened 2019-02-13
Weird collecting error when run pytest


felipe3dfx:
```ini
[pytest]
DJANGO_SETTINGS_MODULE = app.settings
python_files = tests.py test_*.py
addopts = -n4 --nomigrations --cov=. --cov-report=html --cov-report=term:skip-covered
```

```shell
$ pytest
============================= test session starts ==============================
platform linux -- Python 3.6.7, pytest-4.2.1, py-1.7.0, pluggy-0.8.1
Django settings: app.settings (from ini file)
rootdir: /builds/xxxxx/xxxxx, inifile: pytest.ini
plugins: xdist-1.26.1, forked-1.0.1, django-3.4.7, cov-2.6.1
gw0 I / gw1 I / gw2 I / gw3 I
gw0 [1] / gw1 [1] / gw2 [1] / gw3 [1]

.                                                                        [100%]
==================================== ERRORS ====================================
_______________________ ERROR collecting app/__init__.py _______________________
/usr/local/lib/python3.6/site-packages/_pytest/runner.py:226: in from_call
    result = func()
/usr/local/lib/python3.6/site-packages/_pytest/runner.py:289: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
/usr/local/lib/python3.6/site-packages/_pytest/python.py:650: in collect
    for x in self._collectfile(path):
/usr/local/lib/python3.6/site-packages/_pytest/python.py:602: in _collectfile
    assert path.isfile()
E   AssertionError
_______________________ ERROR collecting app/__init__.py _______________________
/usr/local/lib/python3.6/site-packages/_pytest/runner.py:226: in from_call
    result = func()
/usr/local/lib/python3.6/site-packages/_pytest/runner.py:289: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
/usr/local/lib/python3.6/site-packages/_pytest/python.py:650: in collect
    for x in self._collectfile(path):
/usr/local/lib/python3.6/site-packages/_pytest/python.py:602: in _collectfile
    assert path.isfile()
E   AssertionError
_______________________ ERROR collecting app/__init__.py _______________________
/usr/local/lib/python3.6/site-packages/_pytest/runner.py:226: in from_call
    result = func()
/usr/local/lib/python3.6/site-packages/_pytest/runner.py:289: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
[...]

PR #4744, opened 2019-02-08
Fix handling of collect_ignore from parent conftest

blueyed:
`_collectfile` should be called on files only.

Fixes https://github.com/pytest-dev/pytest/issues/4592.

TODO:

- [x] squash
- ~~[ ] remove assertions?~~

PR #4784, opened 2019-02-13
collect: python: fix `AssertionError` with broken symlinks

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4782.

### DIFF
--- a/testing/test_collection.py
+++ b/testing/test_collection.py
@@ -1186,3 +1186,30 @@ def test_collect_pkg_init_and_file_in_args(testdir):
             "*2 passed in*",
         ]
     )
+
+
+@pytest.mark.skipif(
+    not hasattr(py.path.local, "mksymlinkto"),
+    reason="symlink not available on this platform",
+)
+@pytest.mark.parametrize("use_pkg", (True, False))
+def test_collect_sub_with_symlinks(use_pkg, testdir):
+    sub = testdir.mkdir("sub")
+    if use_pkg:
+        sub.ensure("__init__.py")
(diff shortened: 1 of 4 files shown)


testing/test_collection.py
```diff
@@ -1186,3 +1186,48 @@ def test_collect_pkg_init_and_file_in_args(testdir):
...
+        [
+            "pkg/test_file.py::test_file PASSED*",
+            "pkg/test_symlink.py::test_file PASSED*",
+            "*2 passed in*",
+        ]
+    )
+
+
+def test_collect_sub_with_symlinks(testdir):
+    """Similar to the test above, but without __init__.py.
```
blueyed:
```suggestion
    """Similar to test_collect_pkg_with_symlinks above, but without __init__.py.
```

blueyed:
Should we remove the asserts, but only keep them in features?
https://github.com/pytest-dev/pytest/pull/4744#issuecomment-463271301

testing/test_collection.py
```diff
@@ -1186,3 +1186,56 @@ def test_collect_pkg_init_and_file_in_args(testdir):
             "*2 passed in*",
         ]
     )
+
+
+@pytest.mark.skipif(
+    not hasattr(py.path.local, "mksymlinkto"),
+    reason="symlink not available on this platform",
+)
+def test_collect_pkg_with_symlinks(testdir):
```
blueyed: Will parametrize this test, but waiting for feedback from Windows users, where the issue might be something else.

blueyed: Done.

PR #4786, opened 2019-02-13
AppVeyor: use xdist with pypy, drop pluggymaster

blueyed:
pluggymaster is rather stable (not changing often), and currently the same as the released version anyway.

I think it is good enough to test this on Travis, and/or manually with new releases (for Windows).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,14 +6,13 @@ environment:
   - TOXENV: "py34-xdist"
   - TOXENV: "py35-xdist"
   - TOXENV: "py36-xdist"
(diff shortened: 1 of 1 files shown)


nicoddemus:
Fair enough. Hopefully Azure will allow us to test this on Windows again with minimal impact, I agree that AppVeyor times taking hours and hours are a real problem. 👍

nicoddemus:
Unfortunately pypy with xidst is buggy right now: https://github.com/pytest-dev/pytest-xdist/issues/142

PR #4787, opened 2019-02-13
Travis: use xdist for py?? envs, keeping py27/py37

blueyed:


blueyed:
TestEarlyRewriteBailout.test_cwd_changed failed on py34-xdist: https://travis-ci.org/pytest-dev/pytest/jobs/492822748#L281 - maybe only flaky, have not restarted it yet.

blueyed:
Also fails with `tox -e py34-xdist -- -n1 -k test_cwd_changed` locally.

blueyed:
Created https://github.com/pytest-dev/py/pull/207 for a fix in py.

It gets triggered with `-n1` because then `""` is in sys.path, but with `-n0` it is not.

blueyed:
Seems like codecov missed the immediate failure with `#py34` as of now.
Maybe after AppVeyor finishes.

blueyed:
Would be good to get the fix for write_fspath_result into master / a bugfix release, since it causes also problems with py's tests for the fix there..

I'm going to rebased this against master.

blueyed:
Cherry-picked xdist factor related commits from features.

PR #4789, opened 2019-02-13
Travis: add cron_only stage for py38-dev

blueyed:
The master and features branches are tested daily, so we can test
pluggymaster only therein.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ stages:
   if: repo = pytest-dev/pytest AND tag IS NOT present
 - name: deploy
   if: repo = pytest-dev/pytest AND tag IS present
+- name: cron_only
+  if: type = cron
 python:
   - '3.7'
 install:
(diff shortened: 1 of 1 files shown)


blueyed:
Canceled CI manually.  Travis job matrix looks OK, but should be reviewed/checked with the cron job after merging.

blueyed:
Should we also move py38 there?

nicoddemus:
I'm 👎 , if a PR breaks because it is incompatible with `pluggy-master` (which might not be the latest) I want to detect that as early as possible.

blueyed:
Hmm, ok.

What about moving py38-dev there then only for now? (amended like that)

And what about dropping py{27,37}-xdist, or py27/py37 then, but test them with the pluggymaster jobs (either like currently without xdist, or use xdist there also)?

blueyed:
@nicoddemus 
What about this PR as-is?

blueyed:
JFI: cron_only gets skipped/canceled if an earlier stage failed already: https://travis-ci.org/pytest-dev/pytest/builds/493572571
Can we improve this?

nicoddemus:
Don't know from the top of my head 🤔

blueyed:
Via conditional jobs: https://github.com/pytest-dev/pytest/pull/4803

PR #4799, opened 2019-02-14
codecov: use only linux/windows/osx flags

blueyed:
Using many flags are a reason for timeouts on their backend, and we do
not really need those.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ after_success:
       coverage combine
       coverage xml --ignore-errors
       coverage report -m --ignore-errors
-      bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml -F "${TOXENV//-/,},linux"
+      bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml -F $TRAVIS_OS_NAME
     fi
 
 notifications:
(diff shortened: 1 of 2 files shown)


blueyed:
@nicoddemus 
This appears to have helped with timeouts over at https://github.com/awesomeWM/awesome/pull/2648.

blueyed:
@nicoddemus 
Flags are being reported in the comment, and can be selected for coverage overlay in the file detail view.
Env vars appear to only show up with the build info.

blueyed:
Unfortunately https://codecov.io/gh/pytest-dev/pytest/pull/4799?src=pr&el=h1 still times out here.

blueyed:
Asked codecov support about this.
Let's hold this back for now.

.travis.yml
```diff
@@ -105,7 +105,7 @@ after_success:
       coverage combine
       coverage xml --ignore-errors
       coverage report -m --ignore-errors
-      bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml -F "${TOXENV//-/,},linux"
+      bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml -e TOXENV
```
blueyed: Re flags: we have macos here also.

blueyed:
Support indicated that number of uploads is also relevant likely (of course), and we're doing quite a lot.
They asked about 10, but we're doing 20 (https://codecov.io/gh/pytest-dev/pytest/commit/2106515f6d5757f2717efcdb71d6e7a55620f34b/build).

@nicoddemus 
What do you think about reducing number of uploads?
How many should we have, only around 10 maybe?

nicoddemus:
By "uploads" you mean builds that finish and upload coverage, correct?

blueyed:
Yes.

blueyed:
We might still upload more to Code Climate (https://github.com/pytest-dev/pytest/pull/4800), but maybe also not running with coverage all the time would benefit performance (of course).

PR #4800, opened 2019-02-14
[WIP] CI: Codeclimate integration [skip appveyor]

blueyed:
Followup of https://github.com/pytest-dev/pytest/pull/4459.

TODO:

- [ ] appveyor (or straight into Azure), @nicoddemus ?
- [x] remove debug output (`-d`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ env:
     - TOXENV=py37-pexpect,py37-trial,py37-numpy
     - TOXENV=py37-pluggymaster
     - TOXENV=py37-freeze PYTEST_NO_COVERAGE=1
+  global:
+    CC_TEST_REPORTER_ID=17da631beaabf819740e82ef7edee551fe9a4e41409ffdd81284ad6544a51782
 
 matrix:
   allow_failures:
@@ -91,6 +93,15 @@ before_script:
       export COVERAGE_PROCESS_START="$PWD/.coveragerc"
       export _PYTEST_TOX_COVERAGE_RUN="coverage run -m"
       export _PYTEST_TOX_EXTRA_DEP=coverage-enable-subprocess
+
+      # Codeclimate.
+      if [[ "$TRAVIS_OS_NAME" == osx ]]; then
+        curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-darwin-amd64 > ./cc-test-reporter
+      else
+        curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+      fi
+      chmod +x ./cc-test-reporter
+      ./cc-test-reporter before-build
     fi
 
 script: tox --recreate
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,83 +14,27 @@ install:
...
     if [[ "$PYTEST_NO_COVERAGE" != 1 ]]; then
       export COVERAGE_FILE="$PWD/.coverage"
       export COVERAGE_PROCESS_START="$PWD/.coveragerc"
       export _PYTEST_TOX_COVERAGE_RUN="coverage run -m"
       export _PYTEST_TOX_EXTRA_DEP=coverage-enable-subprocess
+
+      # Codeclimate.
+      curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+      chmod +x ./cc-test-reporter
+      ./cc-test-reporter before-build -d
```
blueyed: Might need to get removed - not in the example at https://docs.codeclimate.com/docs/travis-ci-test-coverage (the 2nd one).

blueyed:
Report was uploaded: https://codeclimate.com/repos/5bf9a4be8d88b928f6000b35/settings/test_reports/5c65a8f25ce3550257009a03

Activated PR comments at https://codeclimate.com/repos/5bf9a4be8d88b928f6000b35/settings/github/edit.

.travis.yml
```diff
@@ -105,8 +112,12 @@ after_success:
       coverage xml --ignore-errors
       coverage report -m --ignore-errors
       bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml -F "${TOXENV//-/,},linux"
+
+      ./cc-test-reporter format-coverage -d -t coverage.py -o codeclimate.json coverage.xml
+      ./cc-test-reporter upload-coverage -d -i codeclimate.json
     fi
 
+
```
blueyed:
```suggestion
```

asottile:
I'm really not a fan of codeclimate -- for a lot of the same reasons I don't use `pylint` (which seems to power a lot of its suggestions)

here's an example useless suggestion: https://codeclimate.com/github/pytest-dev/pytest/pull/4804

I'm fine using it to replace coveralls / codecov, but please turn off the rest of it <3

nicoddemus:
I'm on the fence here. I lean towards @asottile suggestion that most of the suggestions are not really useful, but I'm OK with we activating it and see how it goes. A month or two later we then decide if we want to keep using it (and taking the suggestions seriously) or just drop it altogether.

What do you guys think?

asottile:
If the first time a long error is brought up is after the PR is created that's way too late, especially if the error isn't actionable.

nicoddemus:
> If the first time a long error is brought up is after the PR is created that's way too late, especially if the error isn't actionable.

So you are 👎 on adding it?

asottile:
If we can get the coverage bits without any of the rest of it that's great :)

nicoddemus:
Let's hear what @blueyed has to say then. 👍

blueyed:
Yeah, only added it for coverage mainly - which cannot be used alone it seems.
The status is not required, so should be OK.

asottile:
I fear it'll be confusing for newcomers and annoying to have all of our PRs red. I definitely use the PR icon as to whether to look at something or not

blueyed:
I've pinged Codeclimate support about this (only having coverage related things cause a red X).

PR #4802, opened 2019-02-15
Fix code-block in Node docstring

gyermolenko:
before 
image

after
image


nicoddemus:
Oh thanks for catching this!

PR #4812, opened 2019-02-20
Logging: Make pytest_runtest_logreport() hook available for logging

mitzkia:
Signed-off-by: Andras Mitzki <andras.mitzki@balabit.com>

### DIFF
--- a/changelog/4810.bugfix.rst
+++ b/changelog/4810.bugfix.rst
@@ -1,0 +1,1 @@
+Logging messages inside ``pytest_runtest_logreport()`` are now properly captured and displayed.
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -567,6 +567,11 @@ class LoggingPlugin(object):
         with self._runtest_for(None, "finish"):
             yield
 
+    @pytest.hookimpl(hookwrapper=True)
+    def pytest_runtest_logreport(self):
+        with self._runtest_for(None, "logreport"):
+            yield
+
     @pytest.hookimpl(hookwrapper=True, tryfirst=True)
     def pytest_sessionfinish(self):
         with self.live_logs_context():
--- a/testing/logging/test_reporting.py
+++ b/testing/logging/test_reporting.py
@@ -1004,6 +1004,40 @@ def test_log_in_hooks(testdir):
         assert "sessionfinish" in contents
 
 
+def test_log_in_runtest_logreport(testdir):
+    log_file = testdir.tmpdir.join("pytest.log").strpath
+
+    testdir.makeini(
+        """
+        [pytest]
+        log_file={}
+        log_file_level = INFO
+        log_cli=true
+        """.format(
+            log_file
+        )
+    )
+    testdir.makeconftest(
+        """
+        import logging
+        logger = logging.getLogger(__name__)
+
+        def pytest_runtest_logreport(report):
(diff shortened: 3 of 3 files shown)


mitzkia:
@nicoddemus, Thank you for the help. Sorry, but yesterday I was in a little hurry (just wanted to run CI, for the code). In this PR I also wanted to check if any other hook is related here.

mitzkia:
As I check the following hooks can be also affected:
```
* pytest_runtestloop()
* pytest_runtest_makereport()
* pytest_runtest_protocol()
```

After I have added the missing methods to logging.py, the logging will work from the following hooks also:
```
* pytest_runtest_makereport()
* pytest_runtest_protocol()
```

Finally I would rewrite this PR title and extend the fix/tests/doc to the following hooks:
```
* pytest_runtest_logreport()
* pytest_runtest_makereport()
* pytest_runtest_protocol()
```

mitzkia:
About the unit-tests:
- may I extend an already existing testcase (with my new tests): test_reporting.py: test_log_in_hooks(), or should I create a new testcase?
- in this testcase (test_log_in_hooks()) I found that pytest_runtestloop() hook should also work (but I reported, I does not work). I will check this again.

mitzkia:
ok, sorry it did work. So pytest_runtestloop() hook is already fine.

mitzkia:
After understanding that there are still other hooks, I just check again all of them (there are at all 52 hooks).
Logging is not work in the following hooks (except I already found):
```
* pytest_addhooks
* pytest_addoption
* pytest_assertrepr_compare
* pytest_cmdline_main
* pytest_cmdline_parse
* pytest_cmdline_preparse
* pytest_collect_directory
* pytest_configure
* pytest_deselected
* pytest_doctest_prepare_content
* pytest_enter_pdb
* pytest_fixture_post_finalizer
* pytest_fixture_setup
* pytest_ignore_collect
* pytest_internalerror
* pytest_itemstart
* pytest_keyboard_interrupt
* pytest_leave_pdb
* pytest_load_initial_conftests
* pytest_logwarning
* pytest_make_parametrize_id
* pytest_report_teststatus
* pytest_unconfigure
* pytest_warning_captured
```

mitzkia:
May I fix all of them?

mitzkia:
After checking some of the hooks, the fix will not easy (for all of them). There are some hooks which are not belongs to any class...

nicoddemus:
Hi @mitzkia,

> May I fix all of them?

Many hook calls are nested (for example, `pytest_runtest_loop` calls `pytest_runtest_setup` and others), so if we are already capturing logs for a calling hook it is not needed to also enable log capturing for a hook called by it.

nicoddemus:
A quick glance shows that `pytest_runtest_logreport` indeed was missing, giving that we are log capturing `pytest_runtest_logstart` and `pytest_runtest_logfinish`.

Other hooks might be difficult/impossible because they are called very early on the plugin system, and would require a deeper look.

I suggest we just wrap `pytest_runtest_logreport` for now then. 👍

mitzkia:
Hi @nicoddemus,

Thanks for the explanations. I have updated my PR. I hope the CI will passed :)

mitzkia:
I hope the change/PR is correct (final) for now. 
As I checked the CI, it fails on different tests / different places (not related to this PR).
Could you please validate it? 

Thank you.

PR #4813, opened 2019-02-20
Doc update - recommend use venv instead of virtualenv

discdiver:
Recommend use venv instead of virtualenv

From the Python docs:
"Changed in version 3.5: The use of venv is now recommended for creating virtual environments."
-https://docs.python.org/3/library/venv.html

### DIFF
--- a/doc/en/goodpractices.rst
+++ b/doc/en/goodpractices.rst
@@ -8,7 +8,7 @@ Install package with pip
 -------------------------------------------------
 
 For development, we recommend you use venv_ for virtual environments 
-(or virtualenv_ for Python 2.7 and 3.4) and 
+(or virtualenv_ for Python 2.7) and 
 pip_ for installing your application and any dependencies, 
 as well as the ``pytest`` package itself. 
 This ensures your code and dependencies are isolated from your system Python installation.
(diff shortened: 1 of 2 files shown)


doc/en/goodpractices.rst
```diff
@@ -7,12 +7,12 @@ Good Integration Practices
 Install package with pip
 -------------------------------------------------
 
-For development, we recommend to use virtualenv_ environments and pip_
+For development, we recommend you use venv_ for virtual environments and pip_
```
nicoddemus: Good suggestion, but because we still support Python 2.7, I believe we should mention both, perhaps like this:

```
we recommend you use venv[link] for virtual environments (or virtualenv[link] for Python 2.7 and 3.4) and pip ...
```

What do you think?

discdiver: Good thought!

discdiver: Updated PR to mention virtualenv for Python  2.7. 

Python 3.4 comes with venv, so I just mentioned 2.7 needing virtualenv. [Source](https://docs.python.org/3/library/venv.htm).

nicoddemus:
Merging as docs have passed already. 👍

PR #4828, opened 2019-02-25
Clarify docs on --strict marks

Zac-HD:
I've been looking into the behaviour of unknown marks today, and discovered that the docs for the `--strict` option are a bit misleading:  even with `--strict`, you don't need to list marks in `pytest.ini` if they come from pytest or are added by a plugin.  IMO this is the right thing to do, as `--strict` is really to catch typos, so I've just fixed the docs.


nicoddemus:
Thanks @Zac-HD!

PR #4834, opened 2019-02-26
ci: Travis: remove py27/py37, run it on macos

blueyed:
I think it is good enough to run py27/py37 only once, which can be done
on macos then.

py27-xdist/py37-xdist is still tested on Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ install:
   - pip install --upgrade --pre tox
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


blueyed:
Canceled AppVeyor, Travis is good.

nicoddemus:
Hmm I kind disagree, it is important to run at least the base line on each platform, and without xdist.

blueyed:
ok, closing.

PR #4835, opened 2019-02-26
Travis: improve macos setup (py37)

blueyed:


blueyed:
> [skip azure]

Not supported yet: https://developercommunity.visualstudio.com/idea/420289/add-support-ci-skip-and-azure-skip-to-azure-pipeli.html?childToView=443074#comment-443074

Would it work to remove the azure-pipelines.yml file?

.travis.yml
```diff
@@ -29,45 +18,21 @@ matrix:
...
+      before_script:
+        - test $(python -c 'import sys; print("%d%d" % sys.version_info[0:2])') = 27
       env: TOXENV=py27-xdist
     - <<: *test-macos
       env: TOXENV=py37-xdist
       before_install:
-        - brew update
-        - brew upgrade python
-        - brew unlink python
-        - brew link python
```
blueyed: The "brew update/upgrade" is very slow.
This uses pyenv instead to activate/use the 3.7.1 Python that should be available.

.travis.yml
```diff
@@ -29,45 +18,24 @@ matrix:
...
-      if: type = cron
-
-    - stage: baseline
-      env: TOXENV=py27-pexpect,py27-trial,py27-numpy
-    - env: TOXENV=py37-xdist
-    - env: TOXENV=linting,docs,doctesting
-      python: '3.7'
+        - which python3
+        - python3 --version
+        - ln -sfn "$(which python3)" /usr/local/bin/python
```
blueyed: It has 3.7.0 in /usr/local/bin/python3.
I think that is good?!

blueyed:
Works now: https://travis-ci.org/pytest-dev/pytest/jobs/498708715

blueyed:
We could also use `_PYTEST_TOX_COVERAGE_RUN=python3`, but I think it is better to setup `python`.

Issue #4807, opened 2019-02-18
Include py37-freeze on Azure Pipelines


nicoddemus:
It was failing due to not being able to import `distutils`.

nicoddemus:
No longer applies.

PR #4836, opened 2019-02-26
ci: xdist with pluggymaster

blueyed:


blueyed:
Relevant CI jobs passed, canceled others.

blueyed:
- Included dc8c2703 from features.
- Also use xdist on Azure, and fix the env for py27-pluggymaster (it was just "pluggymaster" before, likely using the default testenv then?!).

appveyor.yml
```diff
@@ -6,14 +6,13 @@ environment:
   - TOXENV: "py34-xdist"
   - TOXENV: "py35-xdist"
   - TOXENV: "py36-xdist"
+  # NOTE: pypy-xdist is buggy currently (https://github.com/pytest-dev/pytest-xdist/issues/142).
   - TOXENV: "pypy"
     PYTEST_NO_COVERAGE: "1"
   # Specialized factors for py27.
   - TOXENV: "py27-trial,py27-numpy,py27-nobyte"
-  - TOXENV: "py27-pluggymaster"
```
nicoddemus: I would prefer not to touch this; when we adopt Azure for good, we will disable AppVeyor entirely, so it doesn't make much sense to change things mid way now.

blueyed: This is in features already: https://github.com/pytest-dev/pytest/commit/dc8c27037a7e6e6cd24dd5e3fd2ee6bc56db3594

nicoddemus: Oh nevermind then! 😋 

Besides https://github.com/pytest-dev/pytest/issues/4807, we could adopt Azure as our official Windows build right away. I will try to get to it later.

nicoddemus:
[Latest build from master](https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=121): 

* `py27-pluggymaster`: 8m 27s (which was incorrect as you point out)
* `py37-pluggymaster`: 7m 46s

This PR:

* `py27-pluggymaster-xdist`: 7m 57s
* `py37-pluggymaster-xdist`: 4m 38s

I'm disappointed we couldn't get better timings for p2y7, but for py37 it seems worth it (although I suspect this is due to the difference between the machines rather than a difference between Python versions).

nicoddemus:
> * Also use xdist on Azure, and fix the env for py27-pluggymaster (it was just "pluggymaster" before, likely using the default testenv then?!).

Oops!

PR #4841, opened 2019-02-26
coverage: use run.include, remove --ignore-errors, send TOXENV as name to codecov

blueyed:
This appears to improve performance - ~4s with `tox -e py37-coverage --
testing/test_collection.py`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,10 @@
 [run]
-source = pytest,_pytest,testing/
+source = .
 parallel = 1
 branch = 1
+
+[report]
+include = src/*, testing/*
 
 [paths]
 source = src/
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,9 @@ branch = 1
 include = src/*, testing/*
 
 [paths]
-source = src/
-  .tox/*/lib/python*/site-packages/
-  .tox\*\Lib\site-packages\
+source = src/_pytest
+  .tox/*/lib/python*/site-packages/_pytest/
+  .tox\*\Lib\site-packages\pytest\
+source_pytest_py = src/pytest.py
+  .tox/*/lib/python*/site-packages/pytest.py
+  .tox\*\Lib\site-packages\pytest.py
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,15 +1,15 @@
 [run]
-source = .
+include =
+  src/*
+  testing/*
+  .tox/*/lib/python*/site-packages/_pytest/*
+  .tox/*/lib/python*/site-packages/pytest.py
+  .tox\*\Lib\site-packages\_pytest\*
+  .tox\*\Lib\site-packages\pytest.py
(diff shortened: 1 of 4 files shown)


blueyed:
With `tox -e py37-coverage`:

- master (9dcd6f2): 232.63s user 94.48s system 87% cpu 6:13.17 total
- this PR: 219.65s user 91.22s system 86% cpu 5:57.47 total

The `.coverage` file itself is also a bit smaller (old: 298K, new: 220K).

blueyed:
Re timing: currently again at "234.82s user 92.08s system 88% cpu 6:11.41 total", 223K.

blueyed:
Names are showing up now: https://codecov.io/gh/pytest-dev/pytest/commit/31174f3f8306471dc41d266f1f2831f647b344c2/build

blueyed:
src/pytest went missing: https://codecov.io/gh/pytest-dev/pytest/tree/ecddf69524e7917bc9b87d18c7d178ffa8a12b02

@nicoddemus 
btw: have you changed the required checks?  At least e.g. the WIP should be in there to prevent accidental merging.  I would also add Travis, Azure, and codecov/patch.

nicoddemus:
> btw: have you changed the required checks? At least e.g. the WIP should be in there to prevent accidental merging. I would also add Travis, Azure, and codecov/patch.

I did not change anything, it has never been enabled actually. The only check we have in place is to require reviewer approval.

I will enable WIP; I will hold on enabling the others because we have merged PRs that had one or more checks failing because the change was unrelated to that PR (for example a docs PR to `master` and `master` itself is broken for other reasons). On those cases having to ask the original developer to rebase after the issue is fixed is kind of a pain.

blueyed:
> On those cases having to ask the original developer to rebase after the issue is fixed is kind of a pain.

Admins (we) could always override it (this is also a setting).

blueyed:
Merging, https://codecov.io/gh/pytest-dev/pytest/tree/ee95d666f8c875a7b4d46b2f4e0ce08cb05a0007 says +1.15%.

nicoddemus:
> Admins (we) could always override it (this is also a setting).

Yep, but @RonnyPfannschmidt has asked us to disable that rule even for admins. I'm OK with letting admins override that rule, the rule is mostly to prevent mistakes, but we are already well versed with the workflow.

blueyed:
> > Admins (we) could always override it (this is also a setting).
> 
> Yep, but @RonnyPfannschmidt has asked us to disable that rule even for
> admins. I'm OK with letting admins override that rule, the rule is mostly to
> prevent mistakes, but we are already well versed with the workflow.

@RonnyPfannschmidt
Are you still wanting this?

I think it is better to e.g. clearly have an indicator if Travis/Azure fails,
but allow us to merge nonetheless.

I agree that it would be better to still have the "Require pull request reviews
before merging" enforced though.
It is a pity that it is not possible to only override status checks currently.

PR #4847, opened 2019-02-27
tox: split default posargs, use --lsof with single job only

blueyed:
This is required for combining.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ jobs:
 
     - stage: baseline
       env: TOXENV=py27-pexpect,py27-trial
-    - env: TOXENV=py37-numpy-xdist
+    - env: TOXENV=py37-lsof-numpy-xdist PYTEST_COVERAGE=1
     - env: TOXENV=linting,docs,doctesting
       python: '3.7'
 
--- a/tox.ini
+++ b/tox.ini
@@ -21,12 +21,12 @@ envlist =
 
 [testenv]
 commands =
-    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest {posargs:{env:_PYTEST_TOX_DEFAULT_POSARGS}}
+    {env:_PYTEST_TOX_COVERAGE_RUN:} pytest {posargs:{env:_PYTEST_TOX_DEFAULT_POSARGS:}}
     coverage: coverage combine
     coverage: coverage report
 passenv = USER USERNAME COVERAGE_* TRAVIS PYTEST_ADDOPTS
 setenv =
-    _PYTEST_TOX_DEFAULT_POSARGS=--lsof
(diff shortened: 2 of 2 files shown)


blueyed:
The plan is to use `lsof` with only a few jobs.
This should now not use it all, and hopefully codecov will show.

nicoddemus:
> This should now not use it all, and hopefully codecov will show.

You think `lsof` was the cause for the timeouts?

I can reach codecov by clicking on the image (https://codecov.io/gh/pytest-dev/pytest/pull/4847?src=pr&el=tree), but all other links still time out to me.

blueyed:
> You think `lsof` was the cause for the timeouts?

No, I've meant that it will show missing coverage.

blueyed:
I will compare it manually/rebase it when codecov works again.

nicoddemus:
OK, thanks!

blueyed:
Waiting for codecov status.
Good according to https://codecov.io/github/pytest-dev/pytest/commit/3452460090c7233a78baed6557a65af432b2422f.

azure-pipelines.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         tox.env: 'linting,docs,doctesting'
       py37-xdist:
         python.version: '3.7'
-        tox.env: 'py37-xdist'
+        tox.env: 'py37-lsof-xdist'
```
nicoddemus: `lsof` tests are unix only, and are skipped on Windows:

```
SKIPPED [2] D:\a\1\s\testing\test_capture.py:952: could not run 'lsof'
```

(From https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=193)

blueyed: Thanks for catching this, fixed.  Moved to Travis.

blueyed:
@nicoddemus 
Changed/simplified, please re-review.

blueyed:
Needs the fix from https://github.com/pytest-dev/pytest/pull/4851, i.e. should also get done against features then.

blueyed:
Backed out to previous 9186d62 and rebased (due to the PYTEST_ADDOPTS issue).
I would like to use the lsof factor with #4865 then.

.travis.yml
```diff
@@ -65,7 +65,7 @@ jobs:
 
     - stage: baseline
       env: TOXENV=py27-pexpect,py27-trial
-    - env: TOXENV=py37-numpy-xdist
+    - env: TOXENV=py37-lsof-numpy-xdist PYTEST_COVERAGE=1
```
blueyed: Given that `lsof` is only used to cover the code for it (we do not react / handle its warnings anyway), this could also be done with a faster job for less overhead maybe.

PR #4848, opened 2019-02-27
Fix test_argcomplete: use python -m pytest

blueyed:
Previously it was not run with a) xdist ("-c"), and b) "python -m
pytest" ("…/pytest.py", not executable).

We still need an env to acutally have argcomplete installed.

Where should it go?
Thought about `py27-trial` (what does trial stand for there?), but it appears to only run `test_unittests`?!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ def main():
         # fmt: off
         extras_require={
             "testing": [
+                "argcomplete",
                 "hypothesis>=3.56",
                 "nose",
                 "requests",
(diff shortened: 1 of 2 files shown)


blueyed:
Hmm, Travis does not run for draft PRs, but Azure does..

nicoddemus:
> Where should it go?

Hmmm can't we install `argcomplete` in every environment, like we do with `hypothesis` and `mock`? Seems like it shouldn't be a big dependency.

> Thought about `py27-trial` (what does trial stand for there?), but it appears to only run `test_unittests`?!

"trial" because it tests the "twisted trial" functionality, which is the builtin facilities to test "twisted" applications. I agree it is strange, I believe it would be better to rename to `py27-twisted` instead.

nicoddemus:
Now that I think about it, why we don't just install `numpy` and `twisted` in every environment as well? I think the cost of doing that would greatly offset the extra jobs we have just for those dependencies, specially with wheels available for `numpy` (which I suspect was the cause for splitting those environments in the first place).

blueyed:
Sounds good, will do.

PR #4846, opened 2019-02-27
WIP: ci: Travis: revisit/fix coverage reporting

blueyed:
This should bring coverage back that got missing with 9dcd6f2.

Continuation of https://github.com/pytest-dev/pytest/pull/4839.

PR #4851, opened 2019-02-27
ci: PYTEST_ADDOPTS=-vv

blueyed:
This appears to break some tests, which we should fix.

Ref: https://github.com/pytest-dev/pytest/pull/4846#issuecomment-467949008

### DIFF
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -127,7 +127,7 @@ class TestAssertionRewrite(object):
         result = testdir.runpytest_subprocess()
         assert "warnings" not in "".join(result.outlines)
 
-    def test_name(self):
+    def test_name(self, request):
         def f():
             assert False
 
@@ -147,17 +147,41 @@ class TestAssertionRewrite(object):
         def f():
             assert sys == 42
 
-        assert getmsg(f, {"sys": sys}) == "assert sys == 42"
+        verbose = request.config.getoption("verbose")
+        msg = getmsg(f, {"sys": sys})
+        if verbose > 0:
+            assert msg == (
+                "assert <module 'sys' (built-in)> == 42\n"
+                "  -<module 'sys' (built-in)>\n"
+                "  +42"
+            )
+        else:
+            assert msg == "assert sys == 42"
 
         def f():
-            assert cls == 42  # noqa
+            assert cls == 42  # noqa: F821
 
         class X(object):
             pass
 
-        assert getmsg(f, {"cls": X}) == "assert cls == 42"
+        msg = getmsg(f, {"cls": X}).splitlines()
+        if verbose > 0:
+            if six.PY2:
+                assert msg == [
+                    "assert <class 'test_assertrewrite.X'> == 42",
+                    "  -<class 'test_assertrewrite.X'>",
+                    "  +42",
+                ]
+            else:
+                assert msg == [
+                    "assert <class 'test_...e.<locals>.X'> == 42",
+                    "  -<class 'test_assertrewrite.TestAssertionRewrite.test_name.<locals>.X'>",
+                    "  +42",
+                ]
+        else:
+            assert msg == ["assert cls == 42"]
 
-    def test_dont_rewrite_if_hasattr_fails(self):
+    def test_dont_rewrite_if_hasattr_fails(self, request):
         class Y(object):
             """ A class whos getattr fails, but not with `AttributeError` """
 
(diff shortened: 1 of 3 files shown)


blueyed:
I find `-vv` useful in general (e.g. especially with hanging tests), but we should maybe use different combinations?

nicoddemus:
It seems the problem is that `testdir` inherits verbosity from the calling pytest config.

blueyed:
(Hopefully) fixed pytester.  Rebased against features.

blueyed:
This now fails with the assert-rewrite tests:
```
FAILED testing/test_assertrewrite.py::TestAssertionRewrite::test_dont_rewrite_if_hasattr_fails
FAILED testing/test_assertrewrite.py::TestAssertionRewrite::test_len
FAILED testing/test_assertrewrite.py::TestAssertionRewrite::test_custom_repr
FAILED testing/test_assertrewrite.py::TestAssertionRewrite::test_name
```

The problem seems to be that those tests themselves get rewritten differently when using verbose mode or not.

testing/test_assertrewrite.py
```diff
@@ -173,10 +190,16 @@ def __init__(self):
...
+        # XXX: looks like the "where" should also be there in verbose mode?!
+        message = getmsg(f, {"cls": Y}).splitlines()
+        if request.config.getoption("verbose"):
+            assert message == ["assert 3 == 2", "  -3", "  +2"]
+        else:
+            assert message == [
+                "assert 3 == 2",
+                " +  where 3 = Y.foo",
+                " +    where Y = cls()",
+            ]
```
blueyed: This is not clear to me (see the "XXX") - an uncovered bug?

blueyed: @asottile 
Any idea?  (since it is about assertion rewriting)

blueyed:
This is good in general, but might have uncovered some bug (see above).

testing/test_assertrewrite.py
asottile: that is quite weird!

Looking at the rich compare it looks like the verbose diff mode skips the `where` metalanguage entirely: https://github.com/pytest-dev/pytest/blob/77643122a8d090f951d227ccc7c672deb4e8466c/src/_pytest/assertion/util.py#L154-L155

testing/test_assertrewrite.py
```diff
@@ -147,17 +147,41 @@ def f():
         def f():
             assert sys == 42
 
-        assert getmsg(f, {"sys": sys}) == "assert sys == 42"
+        verbose = request.config.getoption("verbose")
+        msg = getmsg(f, {"sys": sys})
+        if verbose:
```
nicoddemus: I'm not sure we should be having different checks based on the verbosity the user used in the command-line. This might cause confusion, for example if I execute my tests manually without verbosity and it works fine, but breaks on CI.

Is there a way for us to reset the verbosity in `testdir`, so that it will execute with the standard verbosity by default, unless explicitly told otherwise in the `runpytest` call?

blueyed: > Is there a way for us to reset the verbosity in `testdir`, so that it will execute with the standard verbosity by default, unless explicitly told otherwise in the `runpytest` call?

Not sure - after all the PYTEST_ADDOPTS are unset already.
This appears due to the outer pytest already rewriting the code to be verbose.

I think it is OK to have a few conditions like this, but it should not be the common practice, of course - assert rewriting is kind of special in that sense.

nicoddemus: > Not sure - after all the PYTEST_ADDOPTS are unset already.
This appears due to the outer pytest already rewriting the code to be verbose.

I meant by running `pytest` manually... I do it frequently when developing.

> assert rewriting is kind of special in that sense.

Hmmm right, now that I look closer I see that those are inline tests, not run by `pytester`...

Well I guess it is fine to keep that to only a few tests if we don't find a way around it, I believe.

blueyed: > Is there a way for us to reset the verbosity in `testdir`, so that it will
> execute with the standard verbosity by default, unless explicitly told
> otherwise in the `runpytest` call?

> > Not sure - after all the PYTEST_ADDOPTS are unset already.
> > This appears due to the outer pytest already rewriting the code to be verbose.
> 
> I meant by running `pytest` manually... I do it frequently when developing.

The inner runpytestcall?

There are only args to control it, so `testdir.runpytest("--verbosity=1")` ?

blueyed: btw: just found this https://github.com/blueyed/pytest/blob/f43c54e8b2a6d6975f74f7374d87f6b3fe42cfad/testing/test_assertion.py#L20-L29

blueyed:
Can this fly?

testing/test_assertrewrite.py
blueyed: => https://github.com/pytest-dev/pytest/issues/5192

PR #4854, opened 2019-02-28
pdb: add option to skip `pdb.set_trace()`

blueyed:
TODO:

- [ ] doc ?
- [ ] better name(s) ?

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -59,6 +59,13 @@ def pytest_addoption(parser):
         action="store_true",
         help="Immediately break when running each test.",
     )
+    group._addoption(
+        "--pdb-skip",
+        "--pdb-ignore-set_trace",
+        dest="pdb_ignore_set_trace",
+        action="store_true",
+        help="Ignore calls to pdb.set_trace().",
+    )
 
 
 def pytest_configure(config):
@@ -202,6 +209,9 @@ class pytestPDB(object):
     @classmethod
     def set_trace(cls, *args, **kwargs):
         """Invoke debugging via ``Pdb.set_trace``, dropping any IO capturing."""
+        if pytestPDB._config:  # Might not be available when called directly.
+            if pytestPDB._config.getoption("pdb_ignore_set_trace"):
+                return
         frame = sys._getframe().f_back
         _pdb = cls._init_pdb(*args, **kwargs)
         _pdb.set_trace(frame)
--- a/changelog/4854.feature.rst
+++ b/changelog/4854.feature.rst
@@ -1,0 +1,5 @@
+The ``--pdb-skip`` option can now be used to ignore calls to
+``pdb.set_trace()`` (and ``pytest.set_trace()``).
+
+This is meant to help while debugging, where you want to use e.g. ``--pdb`` or
+``--trace`` only, or just run the tests again without any interruption.
--- a/testing/test_pdb.py
+++ b/testing/test_pdb.py
@@ -10,6 +10,7 @@ import sys
 import _pytest._code
 import pytest
 from _pytest.debugging import _validate_usepdb_cls
+from _pytest.main import EXIT_NOTESTSCOLLECTED
 
 try:
     breakpoint
@@ -1112,3 +1113,16 @@ def test_pdb_suspends_fixture_capturing(testdir, fixture):
     assert child.exitstatus == 0
     assert "= 1 passed in " in rest
     assert "> PDB continue (IO-capturing resumed for fixture %s) >" % (fixture) in rest
+
+
+def test_pdb_skip_option(testdir):
+    p = testdir.makepyfile(
+        """
+        print("before_set_trace")
+        __import__('pdb').set_trace()
+        print("after_set_trace")
+        """
+    )
+    result = testdir.runpytest_inprocess("--pdb-ignore-set_trace", "-s", p)
+    assert result.ret == EXIT_NOTESTSCOLLECTED
+    result.stdout.fnmatch_lines(["*before_set_trace*", "*after_set_trace*"])


blueyed:
It is meant ease running tests, where the code is (full of) `__import__("pdb").set_trace()`.

src/_pytest/debugging.py
```diff
@@ -154,6 +161,9 @@ def setup(self, f, tb):
     @classmethod
     def set_trace(cls, *args, **kwargs):
         """Invoke debugging via ``Pdb.set_trace``, dropping any IO capturing."""
+        if pytestPDB._config:  # Might not be available when called directly.
+            if pytestPDB._config.getoption("pdb_ignore_set_trace"):
```
jeffreyrack: Reading the option name, I would've expected this option to only override `pdb.set_trace()`.

However this will cause it to also skip `pytest.set_trace()`, which I'm not sure is an intended behavior.

blueyed: Ah, true.

Are you using `pytest.set_trace()`?

Do you think it makes sense to not skip it?
Might be helpful to still insert a `set_trace` when using the option after all.

jeffreyrack: I just poked at the docs for `pytest.set_trace` and it states:

`This is no longer needed and one can use the native import pdb;pdb.set_trace() call directly.`.

This is stating that the behavior between the two should be the same, so I think it actually makes sense here to have them behave the same way.

blueyed: The docs could be adjusted after all.

I am not sure myself, but cnsidered it a good hack to being able to still `__import__("pytest").set_trace()` if needed.

Anyway, I would also be happy to not having to change docs of course.. :)

jeffreyrack: While I can think of some fringe cases where you might want to set a breakpoint when using this option, I think it would be preferred to have this option skip setting a breakpoint whenever possible.

As a side-note, I believe that `breakpoint()` uses `pdb.set_trace()` by default, so this will also disable `breakpoint()` (I don't have Python 3.7 installed currently to test).

nicoddemus: IMHO it should work the same for both `pytest.set_trace()` as well as `pdb.set_trace()`: `pytest.set_trace()` was the first implementation, afterwards it was improved to work on `pdb.set_trace()`, so it makes sense for both to do the same thing.

blueyed:
What about the option name(s)?

I've used/added a longer run for clarify, but the typical use would rather be to append `-skip` to an existing `--pdb` (the short version).
Since there is `--pdbcls` it should maybe be `--pdbskip` instead?!

nicoddemus:
> Since there is --pdbcls it should maybe be --pdbskip instead?!

Personally I prefer `--pdb-skip`, but I'm fine either way. 👍

blueyed:
> To disable break points set via pdb, use --pdb-skip.

Not a bad idea - but in general it just adds noise - typically you would a `set_trace` to trigger.

I consider this new feature rather as a gem for advanced users, which are also messy by needing it in the first place.. ;)

Similar to how e.g. pdbpp also has an `edit` command to edit the source in-place, but does not add a note about it when entering interactive mode - but maybe it could/should? :p

blueyed:
@nicoddemus 
I've added a method to pass `stdin` to `MultiCapture`, which can be used to interact with pdb in tests.
What do you think?

nicoddemus:
> I've added a method to pass stdin to MultiCapture, which can be used to interact with pdb in tests.

The code looks good, but I'm not sure I understand the user case, can you please elaborate?

blueyed:
> The code looks good, but I'm not sure I understand the user case, can you please elaborate?

It allows to use inprocess pytest instead of pexpect.
Extended the example a bit.

One downside currently is that you do not get the "stdin gets echoed to stdout" that terminals normally do.

blueyed:
Waiting for coverage improvements through master to rebase this against.

asottile:
Would it be better to _not_ implement this and wait for `PYTHONBREAKPOINT=0` to become ubiquitous

blueyed:
> Would it be better to _not_ implement this and wait for `PYTHONBREAKPOINT=0` to become ubiquitous

Oh, wasn't aware of it.

Should we just respect / look at it here instead?

asottile:
> > Would it be better to _not_ implement this and wait for `PYTHONBREAKPOINT=0` to become ubiquitous
> 
> Oh, wasn't aware of it.
> 
> Should we just respect / look at it here instead?

:+1: sg2m

blueyed:
Well, almost finished it - but then thought that having to set `PYTHONBREAKPOINT=0` is much more involving than just appending `--pdb-skip` (which also can be completed through your shell).
I therefore think we should have both.

asottile:
My leaning is that pytest shouldn't reimplement stuff that's standard python (then you have to learn two sets of tools), curious to hear what the others have to say though

blueyed:
Well, `PYTHONBREAKPOINT` is only supported with `breakpoint()` - given that we should probably not look at it at all with `pdb.set_trace`.

asottile:
That's a good point, which makes this feature seem even less of a good idea since it won't handle other debuggers like `pudb` / `ipdb` / the dozen others

blueyed:
Yes, it is specific to `pdb`, but that's implied through the name alread.
pytest only monkeypatches `pdb.set_trace`, not more.  Since I am using pdbpp that is fine, since it monkeypatches `pdb` itself, but AFAIK there are plugins for `ipdb` etc to make pytest work with them.  That sounds like a situation to be improved in general though (not related to this PR currently).

asottile:
My hope for the debugging plugins was that as we eventually drop <3.7 we can just use `breakpoint` and not need all these options and plugins

blueyed:
Will move the stdin improvements (https://github.com/pytest-dev/pytest/pull/4854/commits/e2cd88455a54830d69827ee5c77a99f326f3608c) to https://github.com/pytest-dev/pytest/issues/4937 / a new PR.

blueyed:
Just found that `--pdbcls unittest.mock:Mock` can be used already for skipping `set_trace`, except for that it would still print "PDB set_trace (IO-capturing turned off)", and obviously also breaks `--pdb`.
But maybe proving a better no-op-for-set_trace class is a good idea?

Otherwise I think this is good for now, but could only need some addition to the docs maybe?

asottile:
:S

blueyed:
?

asottile:
My impression was we _weren't_ going to merge this as is given these comments:

- https://github.com/pytest-dev/pytest/pull/4854#issuecomment-469886994
- https://github.com/pytest-dev/pytest/pull/4854#issuecomment-469897713
- https://github.com/pytest-dev/pytest/pull/4854#issuecomment-469900770

blueyed:
@asottile 
I see - but this is another case for a review that has been approved (hit my search for that as a single result), and then getting not feedback via my last comments.
I've came back to this due to having a need for it mysel, but hoped for  "hopefully we're not having to deprecate anything later" when merging it.

blueyed:
As for looking at PYTHONBREAKPOINT: I think that this is not really that much relevant here, since this is about skipping `pdb.set_trace()` alone, but certainly something good to consider in a future PR - I'd be happy to review something in this regard from you, after all I've noticed you getting the improved "Quitting debugger" outcome during the screencast that I've watched.. :)

asottile:
> @asottile
> I see - but this is another case for a review that has been approved (hit my search for that as a single result), and then getting not feedback via my last comments.
> I've came back to this due to having a need for it mysel, but hoped for "hopefully we're not having to deprecate anything later" when merging it.

heh well I didn't think I had to block it because it was `[WIP]` when I looked at it -- usually when moving something from WIP -> not WIP you get another review

blueyed:
> usually when moving something from WIP -> not WIP you get another review

Not here..  WIP is just a silly blocking status.

asottile:
> after all I've noticed you getting the improved "Quitting debugger" outcome during the screencast that I've watched.. :)

have greatly appreciated this btw :+1: I used to `^\` pytest but not any more!

blueyed:
@asottile 
> I used to ^\ pytest but not any more!

Interesting - did not know that exists..

blueyed:
Reverting it in https://github.com/pytest-dev/pytest/pull/5173.

PR #4855, opened 2019-02-28
--pdbcls: improve validation, and allow for "mod:attr.class"

blueyed:


blueyed:
> will decrease coverage by 0.1%.

Started looking at this, but stopped when I've found that codecov considers a "def" not being covered, although the function is: https://codecov.io/gh/pytest-dev/pytest/src/c4c979fce27442c4e4e92ded61f100aef0762119/testing/test_warnings.py#L699

blueyed:
`coverage report -m` from the build, i.e. it is not missing:

```
testing/test_warnings.py                     183      8     20      0    95%   164-186, 577-595
```

The missing-def issue appears to affect more lines there.

In the base (current features) the def-lines are covered from 2 builds only, while others are from 44 (there are so many due to https://github.com/pytest-dev/pytest/pull/4856).

src/_pytest/debugging.py
```diff
@@ -11,6 +12,31 @@
 from _pytest.config import hookimpl
 
 
+def _validate_usepdb_cls(value):
```
nicoddemus: I think this is an excellent candidate for unittests instead of integration tests... what do you think?

blueyed: Ok, there is one unit test already, will move one of the integration tests there also.

testing/test_pdb.py
```diff
@@ -688,6 +690,23 @@ def test_pdb_custom_cls(self, testdir, custom_pdb_calls):
...
+        result.stderr.fnmatch_lines(
+            [
+                "*: error: argument --pdbcls: 'invalid' is not in the format 'modname:classname'"
+            ]
+        )
+
+    def test_pdb_validate_usepdb_cls(self, testdir):
+        assert _validate_usepdb_cls("os.path:dirname.__name__") == "dirname"
+
+        with pytest.raises(argparse.ArgumentTypeError) as exc:
```
nicoddemus: Just a reminder that you can use `match=` here 😉

`match` will accept a regex, which you can use to make the test to pass on py27

blueyed:
Canceled Travis manually.  Should be good.

blueyed:
Thanks, will merge it later.

blueyed:
@nicoddemus 
JFI: I've meant to merge it later (to not block CI, especially since "features" is much slower still).

nicoddemus:
Hmm sorry, I didn't get that. You said:

> Canceled Travis manually. Should be good.

I then "Approved" and would immediately merge it, but I got called off my desk for a few minutes, that's why I did merge a few minutes later only.

> JFI: I've meant to merge it later

Feel free to leave a comment making that explicit; usually we can merge right after a single approve. 👍

blueyed:
I've left that comment:

> Thanks, will merge it later.

Anyway.. :)

nicoddemus:
Heh you are right. I got the comment more along the lines of "thanks for the approval, I will merge later since you didn't already". 👍

PR #4856, opened 2019-02-28
ci: Travis: skip coverage with cron runs

blueyed:
The current commit on features has 50+ uploads already:
https://codecov.io/gh/pytest-dev/pytest/commit/c7bbb2a78873a63b9df9df08f8dc77cde30debd3/build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,6 +87,11 @@ jobs:
 
 before_script:
   - |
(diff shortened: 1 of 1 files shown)


blueyed:
@nicoddemus 
This might be part of the puzzle.

blueyed:
Canceled CI.
Should also get done against features already maybe?

PR #4858, opened 2019-02-28
ci: codecov: remove flags completely for now

blueyed:
This appears to be one of the reasons for timeouts on their backend.


blueyed:
https://codecov.io/gh/pytest-dev/pytest/pull/4858?src=pr&el=h1 is not timing out.. \o/

PR #4859, opened 2019-02-28
tox: add generic nobyte and numpy factors

blueyed:
Remove `py27-nobyte` from tox.ini, which was using xdist already.
Therefore this also removes `py27-xdist` from Travis.

`py27-xdist` is still run on Azure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,10 @@ env:
   matrix:
     - TOXENV=py27 PYTEST_COVERAGE=1
     # Specialized factors for py27.
-    - TOXENV=py27-nobyte PYTEST_COVERAGE=1
-    - TOXENV=py27-xdist PYTEST_COVERAGE=1
+    - TOXENV=py27-nobyte-numpy-xdist PYTEST_COVERAGE=1
     - TOXENV=py27-pluggymaster-xdist
     # Specialized factors for py37.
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -14,8 +14,7 @@ env:
   matrix:
     - TOXENV=py27 PYTEST_COVERAGE=1
     # Specialized factors for py27.
-    - TOXENV=py27-nobyte PYTEST_COVERAGE=1
-    - TOXENV=py27-xdist PYTEST_COVERAGE=1
+    - TOXENV=py27-nobyte-xdist PYTEST_COVERAGE=1
```
nicoddemus: But I think there's a reason why we test xdist with byte code writing enabled, and another with it disabled. It is hard to tell at this point... we might consider dropping `nobyte` altogether actually, I can't see why it needs a special test in pytest.

blueyed: I've kept py27-xdist with Azure.
It was added in 036557ac to test that test_assertrewrite.py works then.

So I think it makes sense to keep a single job with it for now (which can use xdist, so I will change Azure also to use it).

I'll merge this with numpy job(s) then.

nicoddemus: Thanks for digging this up!

Your suggestion makes sense, let's merge this. 👊

PR #4860, opened 2019-03-01
getfixturevalue does not correctly declare dependency with the calling fixture

### LINKED ISSUE #1895: Incorrect finalize/cleanup order of fixtures when using request.getfixturevalue()
Here's a test file that can be run to verify the issue: http://hastebin.com/exozenenas.py

Basically, when you have 2 fixtures, one session and one function scoped that both require another session scoped fixture via request.getfixturevalue(), the cleanup order of the fixtures is incorrect. The shared resource gets cleaned up before the session fixture that relies on it.

nicoddemus:
Fix #1895


nicoddemus:
Ready for review

nicoddemus:
This is related to #4871. Any chance @blueyed and/or @asottile could take a look?

testing/python/fixtures.py
```diff
@@ -562,6 +562,44 @@ def test_func(something):
         reprec = testdir.inline_run()
         reprec.assertoutcome(passed=1)
 
+    def test_getfixturevalue_teardown(self, testdir):
+        """
+        Issue #1895
+
+        `test_inner` requests `inner` fixture, which in turns requests `resource`
```
blueyed:
```suggestion
        `test_inner` requests `inner` fixture, which in turn requests `resource`
```

testing/python/fixtures.py
```diff
@@ -562,6 +562,44 @@ def test_func(something):
...
+        using getfixturevalue. `test_func` then requests `resource`.
```
blueyed:
```suggestion
        using `getfixturevalue`. `test_func` then requests `resource`.
```

testing/python/fixtures.py
```diff
@@ -562,6 +562,44 @@ def test_func(something):
...
+
+        `resource` is teardown before `inner` because the fixture mechanism won't consider
+        `inner` dependent on `resource` when it is get via `getfixturevalue`: `test_func`
```
blueyed:
```suggestion
        `inner` dependent on `resource` when it is used via `getfixturevalue`: `test_func`
```

nicoddemus:
> it should be against master, no?

While it is a bug fix, it does change the teardown order of fixtures, so I want to play safe here and release this on 4.4. 👍

PR #4863, opened 2019-03-01
Move import of  _format_explanation in _pytest.assertion.rewrite

blueyed:
Moved/added there in 250160b4, but apparently not intentionally?!

`# noqa` was only added later in two steps.

/cc @benjaminp

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -21,6 +21,9 @@ import six
 
 from _pytest._io.saferepr import saferepr
 from _pytest.assertion import util
+from _pytest.assertion.util import (  # noqa: F401
+    format_explanation as _format_explanation,
+)
 from _pytest.compat import spec_from_file_location
 from _pytest.pathlib import fnmatch_ex
 from _pytest.pathlib import PurePath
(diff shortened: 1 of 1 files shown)


blueyed:
Fails quite badly, e.g.
```
_______________________________ test_hello[2-5] ________________________________
arg1 = 2, arg2 = 5
    def test_hello(arg1, arg2):
>       assert 0, (arg1, arg2)
E       AttributeError: module '_pytest.assertion.rewrite' has no attribute '_format_explanation'
test_parametrize_functional2.py:5: AttributeError
=========================== 4 failed in 0.07 seconds ===========================
```

It is not clear to me how this works / is being called from - but at least it can be moved to the top it seems.

nicoddemus:
Oh then it is probably being used by the rewritten AST.

nicoddemus:
And yes it can be moved to the top. 👍

blueyed:
> Oh then it is probably being used by the rewritten AST.

Ah.. via https://github.com/blueyed/pytest/blob/9e9b140806b2a525ef0f830b627511b78830aea7/src/_pytest/assertion/rewrite.py#L857.

PR #4866, opened 2019-03-01
tests: install pexpect via testing extra always

blueyed:
This removes the pexpect tox factor.


blueyed:
Downside: slows down all jobs that use it now, of course.

PR #4869, opened 2019-03-01
Add missing ref docs to tmp_path and tmp_path_factory

nicoddemus:


nicoddemus:
Docs passed already, so merging. 👍

blueyed:
@nicoddemus 
btw: at least with Travis we could skip jobs/stages based on commit message content, e.g. `skip tests`.

blueyed:
(OTOH that is bad then when merging, where the base coverage might be missing then)

nicoddemus:
> we could skip jobs/stages based on commit message content, e.g. `skip tests`

I guess you mean we could implement this logic in our `.travis.yml` file.

I'm not sure it will save us much because the builds now are quite faster, plus I guess most people contributing docs won't know about the `skip tests` trick... I don't know. 🤷‍♂️

PR #4877, opened 2019-03-03
tox: generic pluggymaster factor, remove env hack

blueyed:


blueyed:
I hope codecov is just wrong about the increased coverage for the code removed in e1f97e41 (features)..

Issue #2370, opened 2017-04-17
IOError: close() called during concurrent operation on the same file object.


blockjon:
Occasionally, I get `IOError: close() called during concurrent operation on the same file object.` when my tests run. I'm not sure what I should do to resolve this.

PIP LIST:
https://gist.github.com/blockjon/787895c8f5aab3b2587051b919e4e037

PYTEST:
pytest (3.0.5)

OS: 
Debian Jessie running Docker Container

TRACE:
```
00:03:04.535 Traceback (most recent call last):
00:03:04.535   File "runtests.py", line 107, in <module>
00:03:04.535     main()
00:03:04.535   File "runtests.py", line 99, in main
00:03:04.535     status = pytest.main(pytest_args)
00:03:04.535   File "/usr/local/lib/python2.7/dist-packages/_pytest/config.py", line 57, in main
00:03:04.549     return config.hook.pytest_cmdline_main(config=config)
00:03:04.551   File "/usr/local/lib/python2.7/dist-packages/_pytest/vendored_packages/pluggy.py", line 745, in __call__
00:03:04.564     return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
00:03:04.565   File "/usr/local/lib/python2.7/dist-packages/_pytest/vendored_packages/pluggy.py", line 339, in _hookexec
00:03:04.565     return self._inner_hookexec(hook, methods, kwargs)
00:03:04.566   File "/usr/local/lib/python2.7/dist-packages/_pytest/vendored_packages/pluggy.py", line 334, in <lambda>
00:03:04.566     _MultiCall(methods, kwargs, hook.spec_opts).execute()
00:03:04.566   File "/usr/local/lib/python2.7/dist-packages/_pytest/vendored_packages/pluggy.py", line 614, in execute
00:03:04.566     res = hook_impl.function(*args)
00:03:04.566   File "/usr/local/lib/python2.7/dist-packages/_pytest/main.py", line 127, in pytest_cmdline_main
00:03:04.574     return wrap_session(config, _main)
00:03:04.574   File "/usr/local/lib/python2.7/dist-packages/_pytest/main.py", line 123, in wrap_session
00:03:04.574     config._ensure_unconfigure()
00:03:04.575   File "/usr/local/lib/python2.7/dist-packages/_pytest/config.py", line 897, in _ensure_unconfigure
00:03:04.577     fin()
[...]

PR #4878, opened 2019-03-03
tests: make test_crash_on_closing_tmpfile_py27 more reliable

blueyed:
It fails reliable for me now without the fix from 9517c3a2a, and had
actually a wrong assertion in the end.

Ref: #2370

### DIFF
--- a/testing/test_capture.py
+++ b/testing/test_capture.py
@@ -1403,28 +1403,36 @@ def test_dontreadfrominput_has_encoding(testdir):
 
 
 def test_crash_on_closing_tmpfile_py27(testdir):
-    testdir.makepyfile(
+    p = testdir.makepyfile(
         """
         from __future__ import print_function
-        import time
         import threading
         import sys
 
+        printing = threading.Event()
+
         def spam():
             f = sys.stderr
-            while True:
-                print('.', end='', file=f)
+            print('SPAMBEFORE', end='', file=f)
+            printing.set()
 
-        def test_silly():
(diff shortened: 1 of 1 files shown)


testing/test_capture.py
```diff
@@ -1402,30 +1402,42 @@ def test_capattr():
...
 
         def test_silly():
+            global printing
+
             t = threading.Thread(target=spam)
             t.daemon = True
             t.start()
-            time.sleep(0.5)
 
+            while not printing.isSet():
```
nicoddemus: Perhaps `printing.wait()` instead?

nicoddemus: Hmm noticed this failed on Linux. You might try to only set the event after printing a number of times?

blueyed: For reference: all failed with `AssertionError: assert 'SPAMSPAMSPAM' in ''`.

Pushed a fixup based on your suggestions.

blueyed:
It was not failing on CI anymore, but looked like it was not covered.  added a todo for diff coverage.

But maybe it would be better to just monkeypatch `f.close` to  raise IOError reliably?

nicoddemus:
> But maybe it would be better to just monkeypatch f.close to raise IOError reliably?

Hmm that's an idea. Let's see, if this gets flaky we might go down that route, otherwise I'm fine with the code as it is.

blueyed:
Ok, squashed.
Let's merge it when it stays green.

PR #4872, opened 2019-03-01
WIP: [merge later] cacheprovider: _ensure_supporting_files: remove unused branches

blueyed:
It is only called with empty/new dirs since 0385c273.

PR #4880, opened 2019-03-04
[WIP] debug/fix race with _ensure_supporting_files

blueyed:
Related to / because of https://github.com/pytest-dev/pytest/pull/4872.


blueyed:
Looks like there is no race here.

PR #4883, opened 2019-03-04
.coveragerc: fix include for pypy

blueyed:
PyPy uses "site-packages" directly.


blueyed:
Worked as expected according to https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=320.

PR #4884, opened 2019-03-04
Travis: run `pytest --help` for (stable) TTY coverage

blueyed:
According to https://codecov.io/gh/pytest-dev/pytest/pull/4881/changes
the "isatty" branches in _pytest/terminal.py are flaky.

This tries to fix it.

Maybe running it in a subprocess would also help, but it might be more about the outer term in the first place.

Travis should behave as a TTY AFAIK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,6 +99,7 @@ before_script:
     fi
 
 script:
+  - python -c 'import sys; assert sys.stdout.isatty()'
   - tox --recreate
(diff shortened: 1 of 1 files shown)


blueyed:
Since it was covered via https://github.com/pytest-dev/pytest/pull/4883 again, PyPy might be covering this part after all.

blueyed:
Also increased/covered without.

PR #4893, opened 2019-03-07
Simplify 'obj' property definition in PyobjMixin

nicoddemus:
This uses modern property definition syntax, declaring both getter
and setter as obj() functions. That definition has always bothered me a little. 🙃 

It was defined this way so the getter and setter wouldn't be a part of the API. The modern property declaration also has the same characteristics.

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -243,25 +243,24 @@ class PyobjMixin(PyobjContext):
     def __init__(self, *k, **kw):
         super(PyobjMixin, self).__init__(*k, **kw)
 
-    def obj():
-        def fget(self):
-            obj = getattr(self, "_obj", None)
-            if obj is None:
-                self._obj = obj = self._getobj()
-                # XXX evil hack
-                # used to avoid Instance collector marker duplication
-                if self._ALLOW_MARKERS:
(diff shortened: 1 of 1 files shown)


src/_pytest/python.py
```diff
@@ -243,25 +243,24 @@ class PyobjMixin(PyobjContext):
...
+            if self._ALLOW_MARKERS:
+                self.own_markers.extend(get_unpacked_marks(self.obj))
+        return obj
+
+    @obj.setter
+    def obj(self, value):
+        self._obj = value
 
     def _getobj(self):
+        """Gets the underlying python object. May be overwritten by subclasses."""
```
blueyed:
```suggestion
        """Gets the underlying Python object. May be overwritten by subclasses."""
```

src/_pytest/python.py
```diff
@@ -243,25 +243,24 @@ class PyobjMixin(PyobjContext):
...
-
-        def fset(self, value):
-            self._obj = value
-
-        return property(fget, fset, None, "underlying python object")
-
-    obj = obj()
+    @property
+    def obj(self):
+        """underlying python object"""
```
blueyed:
```suggestion
        """Underlying Python object."""
```

nicoddemus:
Done, thanks!

Issue #4896, opened 2019-03-08
Some plugin hooks missing from docs


altendky:
The hooks listed below are not listed in the generated documentation.  I have low confidence I'm organizing them correctly but I'll submit a PR in case that eases discussion of where they do belong.

- `pytest_plugin_registered`
- `pytest_make_collect_report`
- `pytest_pycollect_makemodule`
- `pytest_collection_finish`
- `pytest_pyfunc_call`

Tadaboody:
Is there still work tbd on this or was this closed by #4897?

altendky:
@Tadaboody, thanks for the bump.

PR #4897, opened 2019-03-08
Add missing plugin hooks to docs

altendky:
pytest-dev/pytest#4896

### DIFF
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -584,6 +584,8 @@ Initialization hooks called for plugins and ``conftest.py`` files.
 .. autofunction:: pytest_sessionstart
 .. autofunction:: pytest_sessionfinish
 
+.. autofunction:: pytest_plugin_registered
+
 Test running hooks
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


altendky:
As mentioned in #4896, I'm really not confident where these additions should go.  So, tear it apart...

doc/en/reference.rst
```diff
@@ -629,12 +634,15 @@ items, delete or otherwise amend the test items:
 
 .. autofunction:: pytest_collection_modifyitems
 
+.. autofunction:: pytest_collection_finish
```
nicoddemus: They all seem properly placed to me, good work!

nicoddemus:
Anything else you feel needs adding to this @altendky? Otherwise I think we should change from "WIP" status and merge this. 👍

altendky:
@nicoddemus, ah well, I totally missed your comment.  Sorry.

altendky:
@nicoddemus, do you want a changelog entry?

nicoddemus:
> @nicoddemus, do you want a changelog entry?

Don't think we need for this case. Thanks! Merging.

PR #4902, opened 2019-03-10
[WIP/RFC] pytester: allow for passing in env

blueyed:


src/_pytest/pytester.py
```diff
@@ -1046,23 +1046,24 @@ def collect_by_name(self, modcol, name):
...
         env = os.environ.copy()
-        env["PYTHONPATH"] = os.pathsep.join(
-            filter(None, [os.getcwd(), env.get("PYTHONPATH", "")])
-        )
+        env_update = kw.get("env", {})
+        if "PYTHONPATH" not in env_update:
+            env_update["PYTHONPATH"] = os.pathsep.join(
+                filter(None, [os.getcwd(), env.get("PYTHONPATH", "")])
+            )
+        env.update(env_update)
```
blueyed: I came across this again (to pass `$LINES` and `$COLUMNS` for a pty), and think it might be better to use "env" completely, if provided, but copy only the global env otherwise.  I.e. the caller is responsible for copying it then.
On top, "PYTHONPATH" would still be handled always like currently - although I I can imagine someone wanting to not have it adjust/set..
Makes sense?

src/_pytest/pytester.py
```diff
@@ -790,6 +790,14 @@ def inline_genitems(self, *args):
         items = [x.item for x in rec.getcalls("pytest_itemcollected")]
         return items, rec
 
+    def _get_isolated_env(self):
+        tmpdir = str(self.tmpdir)
+        return (
+            # Do not load user config.
+            ("HOME", tmpdir),
+            ("USERPROFILE", tmpdir),
```
blueyed: I plan to add `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1` here later (in another PR).

src/_pytest/pytester.py
blueyed: Updated.

nicoddemus:
The idea and implementation looks good to me. 👍

nicoddemus:
Hi @blueyed,

This has been on WIP for a long time now, and has some conflicts. I'm closing this for now to clear up our PR queue a bit, but please do re-open this when you have the time.

Cheers!

Issue #2064, opened 2016-11-15
Error when attempting to use IPython debugger


abalkin:
## Problem description

From `py.test` help:

```
  --pdb                 start the interactive Python debugger on errors.
  --pdbcls=modulename:classname
                        start a custom interactive Python debugger on errors. For example:
                        --pdbcls=IPython.terminal.debugger:TerminalPdb
```

I created a simple test file containing just

```
def test_pdb():
    assert 0
```

And ran `py.test` as follows:

```
$ py.test  test_pdb.py \
--pdbcls=IPython.terminal.debugger:TerminalPdb \
--pdb
```

The result was an internal error with the following traceback:

```
Traceback (most recent call last):
  File ".../_pytest/main.py", line 96, in wrap_session
    session.exitstatus = doit(config, session) or 0
  ...
  File ".../IPython/terminal/debugger.py", line 38, in pt_init
    self.pt_cli = CommandLineInterface(self._pt_app, eventloop=self.shell._eventloop)
AttributeError: 'TerminalInteractiveShell' object has no attribute '_eventloop'
```

## Installed packages

```
$ pip freeze
appnope==0.1.0
decorator==4.0.10
entrypoints==0.2.2
ipykernel==4.5.0
ipython==5.1.0
ipython-genutils==0.1.0
ipywidgets==5.2.2
Jinja2==2.8
jsonschema==2.5.1
jupyter==1.0.0
jupyter-client==4.4.0
jupyter-console==5.0.0
jupyter-core==4.2.0
MarkupSafe==0.23
mistune==0.7.3
nbconvert==4.2.0
nbformat==4.1.0
notebook==4.2.3
pexpect==4.2.1
pickleshare==0.7.4
pluggy==0.4.0
prompt-toolkit==1.0.9
ptyprocess==0.5.1
py==1.4.31
Pygments==2.1.3
pytest==3.0.4
pyzmq==16.0.1
qtconsole==4.2.1
simplegeneric==0.8.1
six==1.10.0
terminado==0.6
tornado==4.4.2
traitlets==4.3.1
wcwidth==0.1.7
widgetsnbextension==1.2.6
```

## Pytest and operating system versions

I am using pytest 3.0.4 and

```
$ python
Python 3.5.2+ (3.5:da2ac103d326+, Nov 15 2016, 14:14:18)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
```

## Issue checklist
[...]

PR #4908, opened 2019-03-11
pdb: trigger pytest_enter_pdb hook with post-mortem

blueyed:
This is required for pytest-pdb to be called with `--pdb`.

TODO:

- [x] test
- [ ] ~~pass mode to hook, e.g. "post_mortem" in this case?~~ => later
- [ ] skip "Find last non-hidden frame" in non-postmortem mode?

### DIFF
--- a/changelog/4908.bugfix.rst
+++ b/changelog/4908.bugfix.rst
@@ -1,0 +1,1 @@
+The ``pytest_enter_pdb`` hook gets called with post-mortem (``--pdb``).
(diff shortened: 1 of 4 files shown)


blueyed:
> pass mode to hook, e.g. "post_mortem" in this case?

Should be done separate if needed.

> skip "Find last non-hidden frame" in non-postmortem mode?

This might change behavior, but is not clear to me really.

blueyed:
I would like to get this in, to import/load the pdbcls later, which is required to work around the issue with ipython (https://github.com/pytest-dev/pytest/issues/2064, https://github.com/ipython/ipython/issues/11745).

Issue #4913, opened 2019-03-12
monkeypatching PYTHONPATH breaks real PYTHONPATH


hroncok:
In Fedora, we invoke tests with custom PATH and PYTHONPATH set:

```
PATH=%{buildroot}%{_bindir}:${PATH} \
PYTHONPATH=%{buildroot}%{python2_sitelib} \
  %{buildroot}%{_bindir}/pytest-%{python2_version} -r s testing

PATH=%{buildroot}%{_bindir}:${PATH} \
PYTHONPATH=%{buildroot}%{python3_sitelib} \
  %{buildroot}%{_bindir}/pytest-%{python3_version} -r s testing
```

The monkeypatching in https://github.com/pytest-dev/pytest/blob/51bf7c3aef8f018f44e7c36378be894aa13e32b5/testing/test_collection.py#L1111
breaks that:

```
=================================== FAILURES ===================================
______________________ test_collect_pyargs_with_testpaths ______________________
testdir = <Testdir local('/tmp/pytest-of-mockbuild/pytest-0/test_collect_pyargs_with_testpaths0')>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f9018546b50>
    def test_collect_pyargs_with_testpaths(testdir, monkeypatch):
        testmod = testdir.mkdir("testmod")
        # NOTE: __init__.py is not collected since it does not match python_files.
        testmod.ensure("__init__.py").write("def test_func(): pass")
        testmod.ensure("test_file.py").write("def test_func(): pass")
    
        root = testdir.mkdir("root")
        root.ensure("pytest.ini").write(
            textwrap.dedent(
                """
            [pytest]
            addopts = --pyargs
            testpaths = testmod
        """
            )
        )
        monkeypatch.setenv("PYTHONPATH", str(testdir.tmpdir))
        with root.as_cwd():
            result = testdir.runpytest_subprocess()
>       result.stdout.fnmatch_lines(["*1 passed in*"])
E       Failed: remains unmatched: '*1 passed in*'
/builddir/build/BUILD/pytest-4.3.0/testing/test_collection.py:1114: Failed
----------------------------- Captured stdout call -----------------------------
running: /usr/bin/python2 -mpytest --basetemp=/tmp/pytest-of-mockbuild/pytest-0/test_collect_pyargs_with_testpaths0/runpytest-0
     in: /tmp/pytest-of-mockbuild/pytest-0/test_collect_pyargs_with_testpaths0/root
[...]

PR #4914, opened 2019-03-12
Fix pytest tests invocation with custom PYTHONPATH

hroncok:
Fixes https://github.com/pytest-dev/pytest/issues/4913

- [x] Create a new changelog file in the `changelog` folder...
- [x] Include new tests or update existing tests when applicable (this is a test update).
- [x] Add yourself to `AUTHORS` in alphabetical order (already there)

### DIFF
--- a/testing/test_collection.py
+++ b/testing/test_collection.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import os
 import pprint
 import sys
 import textwrap
@@ -1108,7 +1109,7 @@ def test_collect_pyargs_with_testpaths(testdir, monkeypatch):
     """
         )
     )
-    monkeypatch.setenv("PYTHONPATH", str(testdir.tmpdir))
+    monkeypatch.setenv("PYTHONPATH", str(testdir.tmpdir), prepend=os.pathsep)
     with root.as_cwd():
         result = testdir.runpytest_subprocess()
     result.stdout.fnmatch_lines(["*1 passed in*"])
--- a/changelog/4913.trivial.rst
+++ b/changelog/4913.trivial.rst
@@ -1,0 +1,1 @@
+Fix pytest tests invocation with custom ``PYTHONPATH``.


changelog/4913.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix pytest tests invocation with custom ``PYTHONPATH``.
```
asottile: The `bugfix` type is ~usually for externally facing bugs -- I'd use the `trivial` changelog type here since this is internal to pytest's testsuite

testing/test_collection.py
```diff
@@ -1108,7 +1109,9 @@ def test_collect_pyargs_with_testpaths(testdir, monkeypatch):
     """
         )
     )
-    monkeypatch.setenv("PYTHONPATH", str(testdir.tmpdir))
+    pp = os.getenv("PYTHONPATH")
+    pp = ":".join((str(testdir.tmpdir), pp)) if pp else str(testdir.tmpdir)
```
nicoddemus: On Windows the path separator is actually `;`. Also, `setenv` has a `prepend` option which we can use to simplify this code to only this:

```python
monkeypatch.setenv("PYTHONPATH", str(testdir.tmpdir), prepend=os.pathsep)
```

👍

nicoddemus: Actually I think we can get away with using `monkeypatch.syspath_prepend` instead; I took the liberty of editing the file directly, I hope @hroncok doesn't mind. 😁

hroncok: I don't mind, I just want this fixed. This is much cleaner now, thanks.

hroncok: `monkeypatch.syspath_prepend()` only changes `sys.path`, not `PYTHONPATH` and this tests uses subprocess to call Python. Your fix doesn't seem to work.

hroncok: setenv with prepend should, will amend

hroncok: Back to:

```python
monkeypatch.setenv("PYTHONPATH", str(testdir.tmpdir), prepend=os.pathsep)
```

hroncok:
I'm running a check to see if the current patch actually fixes the issue. It might take  a few minutes.

hroncok:
It does.

hroncok:
@blueyed don't worry about it, it's pretty nonstandard setup, it breaks all the time :D

nicoddemus:
Thanks everyone!

PR #4829, opened 2019-02-25
Remove dead-code related to yield tests

nicoddemus:
Just noticed some code that no longer is needed when we removed yield-tests

PR #4917, opened 2019-03-13
ci: Travis: add PyPy to allowed failures

blueyed:


blueyed:
Should probably rather wait a bit / fix it instead.
We could also remove it on Travis, since it runs on Azure?!  /cc @nicoddemus

nicoddemus:
Sure, as we discussed in https://github.com/pytest-dev/pytest/pull/4829, let's see if we can wait a few more hours. 👍

PR #4919, opened 2019-03-13
Optimize TracebackEntry.ishidden

blueyed:
The expected behavior is that there is no "`__tracebackhide__`" attribute,
so use `getattr` instead of multiple try/except.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -241,25 +241,20 @@ class TracebackEntry(object):
 
     def ishidden(self):
         """ return True if the current frame has a var __tracebackhide__
-            resolving to True
+            resolving to True.
 
             If __tracebackhide__ is a callable, it gets called with the
             ExceptionInfo instance and can decide whether to hide the traceback.
 
             mostly for internal use
         """
-        try:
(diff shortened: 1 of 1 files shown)


src/_pytest/_code/code.py
```diff
@@ -241,25 +241,20 @@ def getsource(self, astcache=None):
...
-            tbh = self.frame.f_locals["__tracebackhide__"]
-        except KeyError:
-            try:
-                tbh = self.frame.f_globals["__tracebackhide__"]
-            except KeyError:
-                return False
-
-        if callable(tbh):
+        f = self.frame
+        tbh = f.f_locals.get(
```
nicoddemus: LGTM, but out of curiosity, did you find this to be a bottleneck somewhere?

blueyed: No, premature/untested performance fix - mostly also due to readability.
I came across if when looking for how `__tracebackhide__` gets handled - the debugging module does not check if it is callable, for example (via #4908).

nicoddemus: I see, thanks

blueyed: Numbers were wrong: actually 1079 hits ,6463 misses.

Issue #1367, opened 2016-02-09
Support for Python 3.4 unittest subtests


PiDelport:
Unittest in Python 3.4 added support for [subtests](https://docs.python.org/3.4/library/unittest.html#distinguishing-test-iterations-using-subtests), a lightweight mechanism for recording parameterised test results.

At the moment, pytest does not support this functionality: when a test that uses `subTest()` is run with pytest, it simply stops at and reports the first failure.

Pytest already supports its own, pytest-specific parametrizing mechanism, but it would be convenient if the pytest runner also supports the standard library's `subTest()` functionality.

pjdelport:
There's some relevant but old discussion on the testing-in-python mailing list: "[unittest subtests](http://lists.idyll.org/pipermail/testing-in-python/2013-January/005356.html)" (January 2013)

RonnyPfannschmidt:
It's important to note that pytest parametrize and subtests are fundamentally completely different

The-Compiler:
While this is a nice feature, it shouldn't block 3.0. I also think this would be some more serious work due to pytest's separation of collection/running, so I'm removing the milestone here.

aksenov007:
Any updates? Any plans to add support of this feature in nearest months?

The-Compiler:
@aksenov007 As far as I know, nobody is working on it currently - though a PR for it would certainly be appreciated! :wink:

shreyashah:
Any idea if subtest feature is supported in pytest? It is an important feature to have.

nicoddemus:
@shreyashah pytest supports [parametrization](https://docs.pytest.org/en/latest/parametrize.html) and in general is a perfect replacement for a test using `subtest`, but it has a few differences in implementation and philosophy.

nicoddemus:
I thought this days that we could use a mechanism similar to what [`pytest-rerunfailures`](https://github.com/pytest-dev/pytest-rerunfailures) does, where it would be probably implemented by a fixture:

```python
def test(subtests):
    for i in range(10):
        with subtests.test():
            assert i % 2 == 0
```

The fixture would trap assertion errors and trigger extra `pytest_runtest_logreport` for each error. Of course the devil is in the details, but it is an idea.

shreyashah:
[...]

PR #4920, opened 2019-03-13
Internal refactorings required for *external* pytest-subtests plugin

nicoddemus:
#1367 is about adding unittest subtest support to pytest. After long discussions, it was decided to introduce a new report type and new hook into pytest. After some experiments, I realized we only needed a few internal changes in order to support unittest subtests and a new `subtests` fixture with similar semantics. 

An external plugin is a better approach here IMO because it allows it make more frequent iterations on the details and mature outside of pytest's release cycle. After the plugin matures, we can consider introducing it into the core.

This PR showcases:

* The internal refactorings required to make the plugin work as I intended.
* The `pytest-subtests` plugin implementation. I'm committing the plugin files (`subtests.py` and `test_subtests.py`) just for context about the other internal changes. Feel free to comment on those, but I realize they lack docs and more tests so no need to point that out. 

After reviews, I will remove the plugin files and rebase so we can merge it.

### unittest subtest ###

```python
import unittest

class T(unittest.TestCase):

    def test_foo(self):
        for i in range(5):
            with self.subTest('custom message', i=i):
                self.assertEqual(i % 2, 0)

if __name__ == '__main__':
    unittest.main()
```

```
λ python .tmp\test-unit-subtest.py

======================================================================
FAIL: test_foo (__main__.T) [custom message] (i=1)
----------------------------------------------------------------------
  File ".tmp\test-unit-subtest.py", line 9, in test_foo
    self.assertEqual(i % 2, 0)
AssertionError: 1 != 0

======================================================================
FAIL: test_foo (__main__.T) [custom message] (i=3)
----------------------------------------------------------------------
  File ".tmp\test-unit-subtest.py", line 9, in test_foo
    self.assertEqual(i % 2, 0)
AssertionError: 1 != 0

----------------------------------------------------------------------
Ran 1 test in 0.001s

FAILED (failures=2)
```

```
λ pytest .tmp\test-unit-subtest.py
...
collected 1 item

.tmp\test-unit-subtest.py FF.                                  [100%]

============================= FAILURES ==============================
_________________ T.test_foo [custom message] (i=1) _________________

self = <test-unit-subtest.T testMethod=test_foo>

    def test_foo(self):
        for i in range(5):
            with self.subTest('custom message', i=i):
>               self.assertEqual(i % 2, 0)
E               AssertionError: 1 != 0

.tmp\test-unit-subtest.py:9: AssertionError
_________________ T.test_foo [custom message] (i=3) _________________

self = <test-unit-subtest.T testMethod=test_foo>

    def test_foo(self):
        for i in range(5):
            with self.subTest('custom message', i=i):
>               self.assertEqual(i % 2, 0)
E               AssertionError: 1 != 0

.tmp\test-unit-subtest.py:9: AssertionError
================ 2 failed, 1 passed in 0.07 seconds =================
```

### subtest fixture ###

```python
def test(subtests):
    for i in range(5):
        with subtests.test(msg='custom message', i=i):
            assert i % 2 == 0
```

```
λ pytest .tmp\test-subtest.py
...
collected 1 item

.tmp\test-subtest.py .F.F..                                    [100%]

============================= FAILURES ==============================
____________________ test [custom message] (i=1) ____________________

    def test(subtests):
        for i in range(5):
            with subtests.test(msg='custom message', i=i):
>               assert i % 2 == 0
E               assert (1 % 2) == 0

.tmp\test-subtest.py:4: AssertionError
____________________ test [custom message] (i=3) ____________________

    def test(subtests):
        for i in range(5):
            with subtests.test(msg='custom message', i=i):
>               assert i % 2 == 0
E               assert (3 % 2) == 0

.tmp\test-subtest.py:4: AssertionError
================ 2 failed, 1 passed in 0.07 seconds =================
```

### DIFF
--- a/src/_pytest/reports.py
+++ b/src/_pytest/reports.py
@@ -1,6 +1,8 @@
 import py
 
+from _pytest._code.code import ExceptionInfo
 from _pytest._code.code import TerminalRepr
+from _pytest.outcomes import skip
 
 
 def getslaveinfoline(node):
@@ -20,6 +22,7 @@ def getslaveinfoline(node):
 
 class BaseReport(object):
     when = None
+    location = None
 
     def __init__(self, **kw):
         self.__dict__.update(kw)
@@ -97,6 +100,43 @@ class BaseReport(object):
     def fspath(self):
         return self.nodeid.split("::")[0]
 
+    @property
+    def count_towards_summary(self):
+        """
+        **Experimental**
+
+        Returns True if this report should be counted towards the totals shown at the end of the
+        test session: "1 passed, 1 failure, etc".
+
+        .. note::
+
+            This function is considered **experimental**, so beware that it is subject to changes
+            even in patch releases.
+        """
+        return True
+
+    @property
+    def head_line(self):
+        """
+        **Experimental**
+
+        Returns the head line shown with longrepr output for this report, more commonly during
+        traceback representation during failures::
+
+            ________ Test.foo ________
+
+
+        In the example above, the head_line is "Test.foo".
+
+        .. note::
+
+            This function is considered **experimental**, so beware that it is subject to changes
+            even in patch releases.
+        """
+        if self.location is not None:
+            fspath, lineno, domain = self.location
+            return domain
+
 
 class TestReport(BaseReport):
     """ Basic test report object (also used for setup and teardown calls if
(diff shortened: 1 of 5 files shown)


src/_pytest/reports.py
```diff
@@ -159,6 +199,49 @@ def __repr__(self):
             self.outcome,
         )
 
+    @classmethod
+    def from_item_and_call(cls, item, call):
+        """
+        Factory method to create and fill a TestReport with standard item and call info.
```
nicoddemus: This is moved verbatim from `pytest_runtest_makereport`.

src/_pytest/subtest.py
```diff
@@ -0,0 +1,86 @@
+from contextlib import contextmanager
```
nicoddemus: This file will be removed when I rebase this.

src/_pytest/terminal.py
```diff
@@ -874,18 +874,23 @@ def summary_stats(self):
...
+                unknown_type_seen = True
     parts = []
-    for key in keys:
-        val = stats.get(key, None)
-        if val:
-            parts.append("%d %s" % (len(val), key))
+    for key in known_types:
+        reports = stats.get(key, None)
+        if reports:
+            count = sum(
```
nicoddemus: This is the only actual change in this function; the rest was cosmetic change to the variable names to make it clear (IMO of course).

testing/test_subtests.py
```diff
@@ -0,0 +1,80 @@
+import six
```
nicoddemus: This file will also be removed when I rebase this PR.

nicoddemus:
> I think (?) this PR would also enable someone to revive yield-tests as a plugin as well

You mean `pytest-subtests`? Definitely. 😈 

I guess one could also use the same tricks as `pytest-subtests` will, but building on it is definitely possible (perhaps with a slightly different implementation though).

PR #4903, opened 2019-03-10
Allow tests to pass after 2038

bmwiedemann:
- [x] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>.<TYPE>.rst`. See [changelog/README.rst](https://github.com/pytest-dev/pytest/blob/master/changelog/README.rst) for details.
- [X] Target the `master` branch for bug fixes, documentation updates and trivial changes.
- [ ] Target the `features` branch for new features and removals/deprecations.
- [ ] Include documentation when adding new features.
- [ ] Include new tests or update existing tests when applicable.

Unless your change is trivial or a small documentation fix (e.g.,  a typo or reword of a small section) please:

- [ ] Add yourself to `AUTHORS` in alphabetical order;

Without this change, the python-apache-libcloud tests failed
in the year 2039 with

```
     fp.write(struct.pack("<ll", mtime, size))
 E   error: 'l' format requires -2147483648 <= number <= 2147483647
```

Note: 2nd hunk is untested

modeled after https://github.com/python/cpython/pull/9892/files

PR #4921, opened 2019-03-14
Add test for mtime issue in #4903

nicoddemus:
As suggested in https://github.com/pytest-dev/pytest/pull/4903#issuecomment-472759958.

I verified this manually against `features` (which still doesn't contain the original patch) and verified that the test breaks as expected with the original exception.

### DIFF
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -1197,6 +1197,29 @@ class TestIssue2121:
         result.stdout.fnmatch_lines("*E*assert (1 + 1) == 3")
 
 
+@pytest.mark.parametrize("offset", [-1, +1])
+def test_source_mtime_long_long(testdir, offset):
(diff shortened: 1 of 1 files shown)


testing/test_assertrewrite.py
```diff
@@ -1197,6 +1197,27 @@ def test_simple_failure():
...
+        E   struct.error: argument out of range
+    """
+    p = testdir.makepyfile(
+        """
+        def test(): pass
+    """
+    )
+    # use unsigned long timestamp which overflows signed long,
+    # which was the cause of the bug
+    timestamp = 2 ** 32 - 1
```
bmwiedemann: You could add another test with `2**32 + 1` - that is where the added 0xFFFFFFFF of my patch comes into play.

nicoddemus: Good suggestion, done. 👍

blueyed:
Thank you!

PR #4925, opened 2019-03-14
pdb: post_mortem: use super()

blueyed:
This is good practice in general, and I've seen it cause problems (MRO)
with pdb++.

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -263,9 +263,9 @@ def _find_last_non_hidden_frame(stack):
 
 
 def post_mortem(t):
-    class Pdb(pytestPDB._pdb_cls):
+    class Pdb(pytestPDB._pdb_cls, object):
         def get_stack(self, f, t):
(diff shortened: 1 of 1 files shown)


src/_pytest/debugging.py
```diff
@@ -263,9 +263,9 @@ def _find_last_non_hidden_frame(stack):
 
 
 def post_mortem(t):
-    class Pdb(pytestPDB._pdb_cls):
+    class Pdb(pytestPDB._pdb_cls, object):
```
nicoddemus: Wondering, is it safe to mix base metaclasses like this in Python 2?

blueyed: Not sure what you mean with metaclasses.. the inheritance from `object` is required here to make it a new-style class.. `pdb.Pdb` does not have it.

nicoddemus: > pdb.Pdb does not have it.

Yes, which makes it an old-style class... well if it works then it works I guess.

Issue #4928, opened 2019-03-15
Wrong line offset with ScopeMismatch error


blueyed:
Given the following file, the line numbers are not reported correctly:
```python
import pytest


@pytest.fixture(scope="session")
def fix1(request, fix2):
    pass


@pytest.fixture
def fix2(fix1):
    pass


def test_1(fix1):
    pass
```

```
platform linux -- Python 3.7.2, pytest-4.3.1.dev29+ga0207274, py-1.7.0, pluggy-0.9.0
…
collected 1 item

t_fail_lnum.py E                                                                                                                                                       [100%]

=================================================================================== ERRORS ===================================================================================
__________________________________________________________________________ ERROR at setup of test_1 __________________________________________________________________________
ScopeMismatch: You tried to access the 'function' scoped fixture 'fix2' with a 'session' scoped request object, involved factories
t_fail_lnum.py:3:  def fix1(request, fix2)
t_fail_lnum.py:8:  def fix2(fix1)
========================================================================== short test summary info ===========================================================================
ERROR t_fail_lnum.py::test_1
========================================================================== 1 error in 0.07 seconds ===========================================================================
```

The following fixes it:
```diff
import pytest


@pytest.fixture(scope="session")
def fix1(request, fix2):
    pass


@pytest.fixture
def fix2(fix1):
    pass


def test_1(fix1):
    pass
```

Will create a PR later.

asottile:
This is in fact "correct" as is: the ast points us at those line numbers

In python3.8 the ast changes where that points so be careful trying to fix this by +1 / -1 hacks :)

PR #4929, opened 2019-03-15
Fix line offsets with `ScopeMismatch` errors

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4928.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -612,7 +612,7 @@ class FixtureRequest(FuncargnamesCompatAttr):
             fs, lineno = getfslineno(factory)
             p = self._pyfuncitem.session.fspath.bestrelpath(fs)
             args = _format_args(factory)
-            lines.append("%s:%d:  def %s%s" % (p, lineno, factory.__name__, args))
+            lines.append("%s:%d:  def %s%s" % (p, lineno + 1, factory.__name__, args))
         return lines
 
     def _getscopeitem(self, scope):
--- a/testing/python/fixture.py
+++ b/testing/python/fixture.py
@@ -992,8 +992,8 @@ class TestFixtureUsages(object):
         result.stdout.fnmatch_lines(
             [
                 "*ScopeMismatch*involved factories*",
-                "* def arg2*",
-                "* def arg1*",
+                "test_receives_funcargs_scope_mismatch.py:6:  def arg2(arg1)",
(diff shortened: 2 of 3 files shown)


src/_pytest/fixtures.py
```diff
@@ -612,7 +612,7 @@ def _factorytraceback(self):
             fs, lineno = getfslineno(factory)
             p = self._pyfuncitem.session.fspath.bestrelpath(fs)
             args = _format_args(factory)
-            lines.append("%s:%d:  def %s%s" % (p, lineno, factory.__name__, args))
+            lines.append("%s:%d:  def %s%s" % (p, lineno + 1, factory.__name__, args))
```
asottile: this is incorrect for multi-decorated things and in python3.8 (where the ast now points us at the function itself)

blueyed: The test works for me with Python 3.8.0a2+ also.
Was this changed in the last days only?

Have to check multi-decorated things, but in general this is not the first time that line numbers are off, because they are zero-based internally..

asottile: hmmm I was expecting https://github.com/python/cpython/pull/9731 to change the outcome here :thinking:

blueyed: Worth investigating probably, but since tests are passing also with 3.8-dev, I think this is fine for now.

testing/python/fixture.py
```diff
@@ -992,8 +992,8 @@ def test_add(arg2):
         result.stdout.fnmatch_lines(
             [
                 "*ScopeMismatch*involved factories*",
-                "* def arg2*",
-                "* def arg1*",
+                "test_receives_funcargs_scope_mismatch.py:6:  def arg2(arg1)",
+                "test_receives_funcargs_scope_mismatch.py:2:  def arg1()",
```
nicoddemus: Hmm strange, `def arg1` to me looks like in line 3, and `def arg2` on line 7... 🤔

blueyed: It is referring to the decorator.
I've just tried with py38 again (given @asottile's comment), but there it is also line 6 and 2.
I think this is fine, since the error comes from the `@fixture` after all.
(But it refers to the first fixture)
The -1 is coming from https://github.com/vim/vim/commit/461a7fcfce3cd6414f990037e6468af3b5ccf119#diff-de9ab8906fef668ed8042046e0a1eb02L8145 btw.

PR #4931, opened 2019-03-15
pytester: LineMatcher: assert lines

blueyed:
This can be helpful when passing a set accidentally.

### DIFF
--- a/testing/test_pytester.py
+++ b/testing/test_pytester.py
@@ -17,6 +17,7 @@ from _pytest.main import EXIT_OK
 from _pytest.main import EXIT_TESTSFAILED
 from _pytest.pytester import CwdSnapshot
 from _pytest.pytester import HookRecorder
+from _pytest.pytester import LineMatcher
 from _pytest.pytester import SysModulesSnapshot
 from _pytest.pytester import SysPathsSnapshot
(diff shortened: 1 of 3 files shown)


blueyed:
Yes.
I came across this when using `{ … }` instead of `( … )` accidentally, which made the test flaky, since the order was changing.

blueyed:
Moved the assertion to `_match_lines` and added tests.

asottile:
I think you've rebased / merged incorrectly?

blueyed:
Have forgotten to change the base on GH.

testing/test_pytester.py
```diff
@@ -453,3 +454,15 @@ def test_timeout():
...
     with pytest.raises(testdir.TimeoutExpired):
         testdir.runpytest_subprocess(testfile, timeout=1)
+
+
+def test_linematcher_with_set():
+    """Test LineMatcher with regard to passing in a set (accidentally)."""
+    lm = LineMatcher([])
+
+    with pytest.raises(AssertionError):
+        lm.fnmatch_lines({})
```
asottile: this is a dict, it does validate the test but it isn't what the docstring says ;)

PR #4932, opened 2019-03-15
Pin sphinx-removed-in to >= 0.2.0 to support Sphinx 2.0

hroncok:
Fixes https://github.com/pytest-dev/pytest/issues/4912


blueyed:
Should we also unpin Sphinx again then?

nicoddemus:
You mean back to `sphinx>=1.8.2` instead of `sphinx>=1.8.2,<2.1`?

blueyed:
Oh, no - missed that this was addressed here already.
`sphinx>=1.8.2,<2.1` is fine.

PR #4934, opened 2019-03-15
[WIP/RFC] docs: nit-picky i.e. fail on missing references

blueyed:
I think we should not have any missing references, which usually indicates something overseen, and results in docs not being linked.

TODO:

- [ ] add a separate docs-strict env, that would use this?


nicoddemus:
Hi @blueyed,

Closing this for now to clean up our PR queue a bit, feel free to re-open this if you want to tackle this again. 👍

PR #4936, opened 2019-03-16
Handle `-p plug` after `-p no:plug`

blueyed:
This is useful to override e.g. "addopts" config from the command line.

Might be worth adding `set_unblocked` to pluggy.


blueyed:
Added a changelog entry.
Please merge if it is ok.

nicoddemus:
Great, thanks!

PR #4941, opened 2019-03-18
pytester: testdir: set $HOME to tmpdir

blueyed:
This avoids loading user configuration, which might interfere with test
results, e.g. a `~/.pdbrc.py` with pdb++.

TODO:

- [x] changelog

### DIFF
--- a/changelog/4941.feature.rst
+++ b/changelog/4941.feature.rst
@@ -1,0 +1,3 @@
+``pytester``'s ``Testdir`` sets ``$HOME`` and ``$USERPROFILE`` to the temporary directory.
+
+This ensures to not load configuration files from the real user's home directory.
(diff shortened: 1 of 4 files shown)


blueyed:
Pushed an update for @asottile (via https://bugs.python.org/issue36264) ;)
Does $USERPROFILE work like $HOME in this regard?

asottile:
> Pushed an update for @asottile (via https://bugs.python.org/issue36264) ;)
> Does $USERPROFILE work like $HOME in this regard?

yep

blueyed:
> > Pushed an update for @asottile (via https://bugs.python.org/issue36264) ;)
> > Does $USERPROFILE work like $HOME in this regard?
> 
> yep

Might be worth having a helper in monkeypatch for this, handling any future changes in this regard maybe.. what do you think?

asottile:
> > > Pushed an update for @asottile (via https://bugs.python.org/issue36264) ;)
> > > Does $USERPROFILE work like $HOME in this regard?
> > 
> > 
> > yep
> 
> Might be worth having a helper in monkeypatch for this, handling any future changes in this regard maybe.. what do you think?

mayyyybe, seems pretty niche though

should this code do:

```python
if sys.platform == 'nt':
    ...
else:
    ...
```

blueyed:
Yeah - basically it would help with what Python expects to see I guess; which also means that it should set `$HOME` with older versions (before 3.8), if it has preference there.

asottile:
> Yeah - basically it would help with what Python expects to see I guess; which also means that it should set `$HOME` with older versions (before 3.8), if it has preference there.

shrugs, I don't think we should add it (increase our maintenance surface area) until someone asks for it

Issue #4942, opened 2019-03-18
Do not monkeypatch logging.raiseExceptions?!


blueyed:
I have found the following code, which was added in 2c7d00579 by @hpk42, but apparently without a test:
https://github.com/pytest-dev/pytest/blob/b3eb5d1eb78f544d27b82be4fe1cb75cccf793c0/src/_pytest/capture.py#L59-L64

A later commit (07196025) indicates that this might be because logging itself tries to close some fd.

I think it is bad practice to change module attributes, especially if it might indicate that pytest is/was doing something wrong itself.

PR #4943, opened 2019-03-18
capture: do not set logging.raiseExceptions = False

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/4942

### DIFF
--- a/changelog/4942.trivial.rst
+++ b/changelog/4942.trivial.rst
@@ -1,0 +1,1 @@
+``logging.raiseExceptions`` is not set to ``False`` anymore.
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -56,13 +56,6 @@ def pytest_load_initial_conftests(early_config, parser, args):
     # make sure that capturemanager is properly reset at final shutdown
     early_config.add_cleanup(capman.stop_global_capturing)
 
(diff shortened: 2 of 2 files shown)


nicoddemus:
Sorry for the delay!

Hmmm are you getting an error because of this code? We want to add a CHANGELOG error explaining in detail why we are reverting this code, and get a fix back quickly if this issue comes back.

blueyed:
> Hmmm are you getting an error because of this code?

No, just found it confusing - likely when looking at coverage reports.

I'll add a changelog.

blueyed:
Maybe related to https://github.com/pytest-dev/pytest/pull/4988.

nicoddemus:
Would you like to include this one in 4.4 @blueyed?

blueyed:
Rather not.. too  much in there already.. ;)

nicoddemus:
Gentle ping in case you want to pick this up again. 👍

blueyed:
Added changelog, and rebased it.

nicoddemus:
I noticed you added this to the 5.0 milestone, is that correct?

blueyed:
Yes, before - but only since it was the next after 4.4.

nicoddemus:
OK to move to 4.5 then?

blueyed:
Sure, I'd just merge it already.

PR #4952, opened 2019-03-18
pytester: lower default pexpect timeout to 1s

blueyed:


blueyed:
Not worth the API change probably - and usually tests are not hanging/failing.

PR #4957, opened 2019-03-20
config: handle `-p no:plugin` with default plugins

blueyed:


blueyed:
And likely fixed tests.. ;) - I assume this is failing somehow somewhere..

Came across this when covering https://github.com/pytest-dev/pytest/pull/4951 for "not capman", and noticed that `-p no:capture` fails hard currently.

blueyed:
That turned out to be easier than thought.. :)

Added changelog, and removed some dead/unnecessary code, as shown by https://codecov.io/gh/pytest-dev/pytest/compare/da81c1e49ad9771b672d579d72503fea21690d4c...327afd81f017a9009e69ea6e1e9730ab1512d9ab/diff#D2-459, and manually checking the code.  First I though about adding an assertion instead, not sure if that would be better?!

nicoddemus:
> First I though about adding an assertion instead, not sure if that would be better?!

Up to you, I'm fine either way. 😁

PR #4958, opened 2019-03-20
ci: Travis: allow to skip jobs without coverage

blueyed:
This skips all jobs without coverage reporting if
"ci-skip-without-coverage" is found in the commit message.

This is meant for quicker results on PRs - currently only 5 jobs are run
on Travis then, instead of 17.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,23 +35,30 @@ jobs:
         - ln -sfn "$(which python3)" /usr/local/bin/python
         - python -V
         - test $(python -c 'import sys; print("%d%d" % sys.version_info[0:2])') = 37
+      if: commit_message !~ /(ci-skip-without-coverage)/ AND type = pull_request
 
     # Full run of latest (major) supported versions, without xdist.
     - env: TOXENV=py27
       python: '2.7'
+      if: commit_message !~ /(ci-skip-without-coverage)/ AND type = pull_request
     - env: TOXENV=py37
       python: '3.7'
+      if: commit_message !~ /(ci-skip-without-coverage)/ AND type = pull_request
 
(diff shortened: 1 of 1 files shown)


blueyed:
On Azure Build.SourceVersionMessage could be used for this probably.. https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml

blueyed:
@nicoddemus 
What is your take on having only coverage reporting jobs for PRs in general, but the full run on master/branch builds?

nicoddemus:
> What is your take on having only coverage reporting jobs for PRs in general, but the full run on master/branch builds?

Not sure, I think one of the main benefits of having coverage is to be able to see that new code is not properly covered. By the time it reaches master, it is too late and probably nobody will want to touch it...

TBH I'm happy with the current build times we now have with Azure. We have results in under 10 minutes now, which is perfectly acceptable I think. 👍

nicoddemus:
> This is meant for quicker results on PRs - currently only 5 jobs are run
on Travis then, instead of 17.

How do we decide if we want to skip coverage? Only for WIP commits and such, while the "final" commit would not contain that string as to get a full coverage report?

blueyed:
> Not sure, I think one of the main benefits of having coverage is to be able to see that new code is not properly covered. By the time it reaches master, it is too late and probably nobody will want to touch it...

?
It would run _only_ / all coverage related jobs, skipping those that do not report coverage anyway (and therefore won't affect master).
But uncovered code might break master.

I've used the special string on occasion, but can also keep removing them manually from .travis.yml.

Closing for now.

PR #4959, opened 2019-03-20
[WIP] ci: Azure split jobs into coverage/no-coverage

blueyed:


azure-pipelines.yml
```diff
@@ -133,3 +104,78 @@ jobs:
       PYTHON: $(python.exe)
       CODECOV_TOKEN: $(CODECOV_TOKEN)
       PYTEST_CODECOV_NAME: $(tox.env)
+
+- job: 'test_without_coverage'
```
blueyed: This gets run as a single job/build?!  https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=492&view=logs

blueyed: @nicoddemus 
Do you know how to fix this?
The idea is a) to have stages (coverage/important jobs first), and b) then allow to skip the non-coverage jobs similar to #4958 for Travis.

nicoddemus: I think you got the right idea in https://github.com/pytest-dev/pytest/pull/4958#issuecomment-474668726 👍 

But I'm not fully convinced that this is a good idea yet. 😁

blueyed: ok :)

PR #4962, opened 2019-03-20
tests: add test_report_collect_after_half_a_second

blueyed:
This is meant for stable coverage with "collecting X item(s)".

I've thought about monkeypatching the duration from inside the first collected test, since the test is rather slow (1s).  Do you think that's a good idea?

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -26,6 +26,8 @@ from _pytest.main import EXIT_OK
 from _pytest.main import EXIT_TESTSFAILED
 from _pytest.main import EXIT_USAGEERROR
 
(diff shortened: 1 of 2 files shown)


nicoddemus:
> I've thought about monkeypatching the duration from inside the first collected test, since the test is rather slow (1s). Do you think that's a good idea?

How about monkeypatching `REPORT_COLLECTING_RESOLUTION` to zero? This way we guarantee we will write an update for each collected item always.

blueyed:
> How about monkeypatching `REPORT_COLLECTING_RESOLUTION` to zero? This way we guarantee we will write an update for each collected item always.

ok!

blueyed:
Adjusted.

nicoddemus:
Thanks!

PR #4963, opened 2019-03-20
ci: rename "trial" tox factor to "twisted"

blueyed:
Ref: https://github.com/pytest-dev/pytest/pull/4848#issuecomment-467909204

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ jobs:
     # Specialized factors for py37.
     # Coverage for:
     # - test_sys_breakpoint_interception (via pexpect).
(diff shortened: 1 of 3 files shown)


nicoddemus:
I don't believe we need a CHANGELOG btw

blueyed:
JFI: Azure failed (https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=497&view=results):

> Could not retrieve file content for /azure-pipelines.yml from repository self hosted on https://api.github.com using commit dddbd38f51903f91ac694d373decce38d8c71da3. GitHub reported the error, "Maximum threads for service reached"

Could not restart it, since my account has been suspended?!
Pushed a new commit.

PR #4966, opened 2019-03-21
config: fix consider_preparse with missing argument to -p

blueyed:
This is only required after/with 415899d4 - otherwise argparse ensures
there is an argument already.

### DIFF
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -313,6 +313,9 @@ class TestPytestPluginManagerBootstrapming(object):
         assert '"hello123"' in excinfo.value.args[0]
         pytestpm.consider_preparse(["-pno:hello123"])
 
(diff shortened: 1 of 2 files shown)


testing/test_pluginmanager.py
```diff
@@ -313,6 +313,8 @@ def test_preparse_args(self, pytestpm):
         assert '"hello123"' in excinfo.value.args[0]
         pytestpm.consider_preparse(["-pno:hello123"])
 
+        pytestpm.consider_preparse(["-p"])
```
nicoddemus: I suggest to add a comment in the lines "just make sure single -p doesn't blow up"

nicoddemus:
> I think a changelog is not required - it was not used without argparse in a release yet.

Oh good point. 👍

PR #4967, opened 2019-03-21
Fix some issues related to "-p no:X" with default_plugins

blueyed:
This excludes some plugins for now, which are not trivial (but not impossible) to fix.

Some of them, e.g. "main" should be disallowed to be disabled though maybe?  (e.g. internal/core plugin versus interal default plugin)


nicoddemus:
Another plugin which can't be ignored is Python, as it defines the `request` fixture (IIRC) which is essential to the core.

PR #4972, opened 2019-03-21
[WIP/RFC] test_pdb: restore_settrace auto-use fixture

blueyed:
Only really relevant for when `pdb.set_trace` is used.

This is expected to increase coverage.


blueyed:
Well, we're using pexpect.. ;)

PR #4975, opened 2019-03-22
Fix usages of "verbose" option

blueyed:
With `-qq` `bool(config.getoption("verbose"))` is True; it needs to be
checked for `> 0`.

Found via `rg 'if .*verbose[^<>=]*$' `.

### DIFF
--- a/changelog/4975.bugfix.rst
+++ b/changelog/4975.bugfix.rst
@@ -1,0 +1,1 @@
+Fix the interpretation of ``-qq`` option where it was being considered as ``-v`` instead.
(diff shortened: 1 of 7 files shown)


blueyed:
> LGTM, need a CHANGELOG entry only.

Yes?
Can you suggest one then, please?
I think it is more internal and for code health / clarity in general.

nicoddemus:
IIUC this means that `-qq` option was being interpreted as `-v` in several places in the codebase, so this is certainly a bug fix.

How about:

```rst
Fix the interpretation of ``-qq`` option were it was being considered as ``-v`` instead.
```

blueyed:
Canceled Travis.

nicoddemus:
Thanks!

PR #4978, opened 2019-03-22
Do not swallow outcomes.Exit in assertrepr_compare

blueyed:
It is annoying if `q` does not work.

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -9,6 +9,7 @@ import six
 
 import _pytest._code
 from ..compat import Sequence
+from _pytest import outcomes
 from _pytest._io.saferepr import saferepr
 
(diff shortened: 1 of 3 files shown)


src/_pytest/assertion/util.py
```diff
@@ -102,38 +103,45 @@ def _format_lines(lines):
...
-            iter(obj)
-            return not istext(obj)
-        except TypeError:
-            return False
+def isiterable(obj):
+    try:
+        iter(obj)
+        return not istext(obj)
+    except TypeError:
+        return False
```
blueyed: This is moved out of the function only, which I'd like to have done anyway - but allows for easy monkeypatching.

blueyed:
Canceled Travis.

PR #4980, opened 2019-03-22
monkeypatch.syspath_prepend: call fixup_namespace_packages

blueyed:
Without the patch the test fails as follows:

            # Prepending should call fixup_namespace_packages.
            monkeypatch.syspath_prepend("world")
    >       import ns_pkg.world
    E       ModuleNotFoundError: No module named 'ns_pkg.world'


blueyed:
There is also `testdir.syspathinsert`, which needs the same.. :/
Added no test for it there.

PR #4987, opened 2019-03-23
CollectError.repr_failure: honor explicit tbstyle option

blueyed:


blueyed:
Does it require a changelog?

nicoddemus:
I think so, this has external effects so it is nice to let users know of the change. 👍

blueyed:
Canceled Travis.

nicoddemus:
Thanks!

PR #4988, opened 2019-03-23
logging: close log_file_handler

blueyed:
See https://github.com/pytest-dev/pytest/pull/4981/commits/92ffe42b45973bce478f25bb8acbb5e8e791aafa / #4981 for more information.

TODO:

- [x] changelog
- [ ] test

It means that coverage with src/_pytest/capture.py is likely not flaky anymore, but decreases - it was only accidentally covered.

### DIFF
--- a/changelog/4988.bugfix.rst
+++ b/changelog/4988.bugfix.rst
@@ -1,0 +1,1 @@
+Close logging's file handler explicitly when the session finishes.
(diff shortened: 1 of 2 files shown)


nicoddemus:
Interesting find.

blueyed:
Changed it to close in "finally", but a test is rather non-trivial - it woud involve mocking the `close` function, or is there a better way?

Anyway, I think this is good now.

PR #4990, opened 2019-03-24
Session repr

blueyed:


src/_pytest/main.py
```diff
@@ -441,6 +441,15 @@ def __init__(self, config):
...
         self.config.pluginmanager.register(self, name="session")
 
+    def __repr__(self):
+        return "<%s %s exitstatus=%r testsfailed=%d testscollected=%d>" % (
+            self.__class__.__name__,
+            self.name,
+            getattr(self, "exitstatus", "<UNSET>"),
+            self.testsfailed,
+            self.testscollected,
+        )
```
blueyed: Previously it was `<Session name>` (via `Node`).

PR #5002, opened 2019-03-26
skipping: factor out _get_pos, pass only config to _get_report_str

blueyed:
This is for "short summary".

I've started looking into making the format there configurable, to support e.g. "fname:lineno" instead of nodeids.

### DIFF
--- a/src/_pytest/skipping.py
+++ b/src/_pytest/skipping.py
@@ -207,20 +207,22 @@ def pytest_terminal_summary(terminalreporter):
 def show_simple(terminalreporter, lines, stat):
     failed = terminalreporter.stats.get(stat)
     if failed:
+        config = terminalreporter.config
         for rep in failed:
-            verbose_word = _get_report_str(terminalreporter, rep)
-            pos = terminalreporter.config.cwd_relative_nodeid(rep.nodeid)
+            verbose_word = _get_report_str(config, rep)
+            pos = _get_pos(config, rep)
             lines.append("%s %s" % (verbose_word, pos))
 
 
 def show_xfailed(terminalreporter, lines):
     xfailed = terminalreporter.stats.get("xfailed")
     if xfailed:
+        config = terminalreporter.config
         for rep in xfailed:
-            verbose_word = _get_report_str(terminalreporter, rep)
-            pos = terminalreporter.config.cwd_relative_nodeid(rep.nodeid)
-            reason = rep.wasxfail
+            verbose_word = _get_report_str(config, rep)
+            pos = _get_pos(config, rep)
             lines.append("%s %s" % (verbose_word, pos))
+            reason = rep.wasxfail
             if reason:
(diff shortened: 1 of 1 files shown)


nicoddemus:
> I've started looking into making the format there configurable, to support e.g. "fname:lineno" instead of nodeids.

I'm curious, in order to jump to the code using your editor?

blueyed:
> I'm curious, in order to jump to the code using your editor?

Yeah, `fname:lineno` is very common.

It was mainly inspired when looking at https://github.com/Frozenball/pytest-sugar/issues/173.

I am using a Vim pluging that handles pytest's nodeids also (https://github.com/wsdjeg/vim-fetch), but there then the lineno could be useful also.

I was thinking about changing the default to "nodeid, line X" at least - but allow for configuring this through an ini option, so that e.g. `nodeid (fname:lineno)` could also be used.

nicoddemus:
> I was thinking about changing the default to "nodeid, line X" at least - but allow for configuring this through an ini option, so that e.g. nodeid (fname:lineno) could also be used.

I'm 👍 to change the default, but I don't think we need an option if that default is good enough for:

1. Quickly selecting the node id to run again; for me this works if there's a space between nodeid and (fname:lineno);
2. Configure your editor to automatically parse that line so you can jump to the code directly.

1) is the main use case I see for today's behavior, and I believe it can get alongside 2 just fine.

blueyed:
I agree with not having unnecessary options, especially since this would involve a mini template engine in this case ("lineno" might be not available, and then you do not want "fname:" I guess).

I think we should retain `nodeid`, but having `fname:lineno` after it might make it very long (since the fname is basically duplicated then).
But I think just appending " (fname:lineno)" would be a good improvement nonetheless as a first step.

nicoddemus:
> I think we should retain nodeid, but having fname:lineno after it might make it very long (since the fname is basically duplicated then).
But I think just appending " (fname:lineno)" would be a good improvement nonetheless as a first step.

Oh good point, it will probably be too long indeed...

blueyed:
So not a good default then?
It would be easier if pytest could handle `fname:lineno` for collection (then we could use just this here), but OTOH the nodeid is nice, and more stable (with changing files) in general.

I've also thought about extending the nodeid, i.e. `fname::func:lnum`, but this would e.g. not work with the Vim plugin I am using (but could be handled there, of course), or other tools handling only the old format.
Better would be `fname:lineno::func` as an extended nodeid, i.e. have `:lineno` optionally after the fname.  Editors that support nodeids already could be extended to use this as an extra hint (or just ignore it, if they are looking for the test class/name; but it would be a good hint for when there are multiple / where to start), and also allows for using this with tools that only handle `fname:lineno` - with the drawback of not having the extra space there (for double click selection).

blueyed:
Merging as-is already, happy to follow up the discussion here (or in a new issue, whatever you prefer).

PR #5005, opened 2019-03-27
Split out list of essential plugins

blueyed:
No changelog, since this is meant to be internal/private?  (rename to `_essential_plugins` maybe?)

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -112,13 +112,18 @@ def directory_arg(path, optname):
     return path
 
 
-default_plugins = (
+# Plugins that cannot be disabled via "-p no:X" currently.
+essential_plugins = (
     "mark",
     "main",
-    "terminal",
     "runner",
     "python",
     "fixtures",
+    "helpconfig",  # Provides -p.
+)
+
+default_plugins = essential_plugins + (
+    "terminal",  # Has essential options, but xdist uses -pno:terminal.
     "debugging",
     "unittest",
     "capture",
@@ -127,7 +132,6 @@ default_plugins = (
     "monkeypatch",
     "recwarn",
     "pastebin",
-    "helpconfig",
     "nose",
     "assertion",
     "junitxml",
@@ -142,7 +146,6 @@ default_plugins = (
     "logging",
     "reports",
 )
-
 
 builtin_plugins = set(default_plugins)
 builtin_plugins.add("pytester")
@@ -496,6 +499,9 @@ class PytestPluginManager(PluginManager):
     def consider_pluginarg(self, arg):
         if arg.startswith("no:"):
             name = arg[3:]
+            if name in essential_plugins:
+                raise UsageError("plugin %s cannot be disabled" % name)
+
             # PR #4304 : remove stepwise if cacheprovider is blocked
             if name == "cacheprovider":
                 self.set_blocked("stepwise")
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -173,6 +173,7 @@ def getreportopt(config):
     return reportopts
 
 
+@pytest.hookimpl(trylast=True)  # after _pytest.runner
 def pytest_report_teststatus(report):
     if report.passed:
         letter = "."
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -1205,20 +1205,12 @@ def test_config_does_not_load_blocked_plugin_from_args(testdir):
     [
         x
         for x in _pytest.config.default_plugins
-        if x
-        not in [
-            "fixtures",
-            "helpconfig",  # Provides -p.
-            "main",
-            "mark",
-            "python",
-            "runner",
-            "terminal",  # works in OK case (no output), but not with failures.
-        ]
+        if x not in _pytest.config.essential_plugins
     ],
 )
 def test_config_blocked_default_plugins(testdir, plugin):
     if plugin == "debugging":
+        # Fixed in xdist master (after 1.27.0).
         # https://github.com/pytest-dev/pytest-xdist/pull/422
         try:
             import xdist  # noqa: F401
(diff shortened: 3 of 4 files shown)


src/_pytest/config/__init__.py
```diff
@@ -112,6 +112,17 @@ def directory_arg(path, optname):
...
+    "mark",
+    "main",
+    "terminal",  # Required to display/handle failures.
+    "runner",
+    "python",
+    "fixtures",
+    "helpconfig",  # Provides -p.
+)
+
 default_plugins = (
```
nicoddemus: I think we should remove explicitly listing `essential_plugins` here, mostly for consistency, and then use `essential_plugins` to initialize `default_plugins`

```
default_plugins = essential_plugins + (
    "debugging",
    "unittest",
    ...
```

blueyed: Like this?
```
index 2b3faa2e..253c61f0 100644
@@ -123,7 +123,7 @@ def directory_arg(path, optname):
     "helpconfig",  # Provides -p.
 )

-default_plugins = (
+default_plugins = essential_plugins + (
     "mark",
     "main",
     "terminal",
@@ -153,7 +153,7 @@ def directory_arg(path, optname):
     "logging",
 )

-builtin_plugins = set(essential_plugins + default_plugins)
+builtin_plugins = set(default_plugins)


```

nicoddemus: Yes, while also removing the duplicated names from `default_plugins` which are already in `essential_plugins`

blueyed: Oh, sure - have re-added that accidentally when comparing the list again - order there appears to matter (at least for some other test in the file), and I've used the order from the test at first.

src/_pytest/config/__init__.py
```diff
@@ -112,13 +112,18 @@ def directory_arg(path, optname):
     return path
 
 
-default_plugins = (
+# Plugins that cannot be disabled via "-p no:X" currently.
+essential_plugins = (
     "mark",
     "main",
-    "terminal",
+    "terminal",  # Required to display/handle failures.
```
nicoddemus: I wonder if this will this affect xdist? https://github.com/pytest-dev/pytest-xdist/blob/master/xdist/remote.py#L252

blueyed: Yeah, certainly.
I had it excluded in the test, because there it failed to report.
It should be not in there then I think.

nicoddemus: Yeah it seems this will break xdist:

https://github.com/pytest-dev/pytest-xdist/blob/ba434848f5afdfab117191ec304f70c3846185b9/xdist/remote.py#L252

```python
def remote_initconfig(option_dict, args):
    from _pytest.config import Config

    option_dict["plugins"].append("no:terminal")
```

https://github.com/pytest-dev/pytest/blob/0fe114253afbfc2f81491bdc1c3dadbebcb214a2/src/_pytest/config/__init__.py#L192-L197


https://github.com/pytest-dev/pytest/blob/0fe114253afbfc2f81491bdc1c3dadbebcb214a2/src/_pytest/config/__init__.py#L498-L502

nicoddemus: > It should be not in there then I think.

Agreed. 👍

blueyed: But it must come before "runner" apparently for a recently added test:
```
==================== FAILURES ====================
_ TestPytestPluginManager.test_pluginmanager_ENV_startup _

self = <test_pluginmanager.TestPytestPluginManager object at 0x7faab37cb320>
testdir = <Testdir local('/tmp/pytest-of-user/pytest-839/test_pluginmanager_ENV_startup0')>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7faab1cdd940>

    def test_pluginmanager_ENV_startup(self, testdir, monkeypatch):
        testdir.makepyfile(pytest_x500="#")
        p = testdir.makepyfile(
            """
            import pytest
            def test_hello(pytestconfig):
                plugin = pytestconfig.pluginmanager.get_plugin('pytest_x500')
                assert plugin is not None
        """
        )
        monkeypatch.setenv("PYTEST_PLUGINS", "pytest_x500", prepend=",")
        result = testdir.runpytest(p, syspathinsert=True)
        assert result.ret == 0
>       result.stdout.fnmatch_lines(["*1 passed*"])
E       Failed: nomatch: '*1 passed*'
E           and: '============== test session starts ==============='
E           and: 'platform linux -- Python 3.7.2, pytest-4.3.2.dev151+g6e6d3b74.d20190327, py-1.8.0, pluggy-0.9.0'
E           and: "hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('…/Vcs/pytest/.hypothesis/examples')"
E           and: 'rootdir: /tmp/pytest-of-user/pytest-839/test_pluginmanager_ENV_startup0'
E           and: 'plugins: hypothesis-4.14.0'
E           and: 'collected 1 item'
E           and: ''
E           and: 'test_pluginmanager_ENV_startup.py ...      [100%]'
E           and: ''
E           and: '============ 3 passed in 0.01 seconds ============'
E           and: ''
E       remains unmatched: '*1 passed*'

…/Vcs/pytest/testing/test_pluginmanager.py:268: Failed
-------------- Captured stdout call --------------
============== test session starts ===============
platform linux -- Python 3.7.2, pytest-4.3.2.dev151+g6e6d3b74.d20190327, py-1.8.0, pluggy-0.9.0
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('…/Vcs/pytest/.hypothesis/examples')
rootdir: /tmp/pytest-of-user/pytest-839/test_pluginmanager_ENV_startup0
plugins: hypothesis-4.14.0
collected 1 item

test_pluginmanager_ENV_startup.py ...      [100%]

============ 3 passed in 0.01 seconds ============
============ short test summary info =============
FAILED testing/test_pluginmanager.py::TestPytestPluginManager::test_pluginmanager_ENV_startup
```

blueyed: Might be worth checking if https://github.com/pytest-dev/pytest/pull/5006 helps with that.

PR #5006, opened 2019-03-27
tests: ensure cleanup with configs via get_config()

blueyed:
Also done in test_pluginmanager, although no resource warnings are
there at least.

Fixes https://github.com/pytest-dev/pytest/issues/4355.

### DIFF
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -9,7 +9,6 @@ import sys
 import types
 
 import pytest
-from _pytest.config import get_config
 from _pytest.config import PytestPluginManager
 from _pytest.main import EXIT_NOTESTSCOLLECTED
 from _pytest.main import Session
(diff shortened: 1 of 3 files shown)


testing/test_pluginmanager.py
```diff
@@ -20,8 +19,19 @@ def pytestpm():
...
 
+@pytest.fixture
+def config_for_test():
+    from _pytest.config import get_config
+
+    config = get_config()
+    try:
+        yield config
+    finally:
+        config._ensure_unconfigure()  # cleanup, e.g. capman closing tmpfiles.
```
blueyed: Would be good to share the fixture, but there is no conftest.. :D - any advice?

nicoddemus: I think we can move that to `pytester` perhaps?

testing/test_pluginmanager.py
```diff
@@ -20,8 +19,19 @@ def pytestpm():
     return PytestPluginManager()
 
 
+@pytest.fixture
+def config_for_test():
+    from _pytest.config import get_config
+
+    config = get_config()
+    try:
```
nicoddemus: You don't need the `try/finally` btw, fixture teardown is always executed even if the `yield` block raises

testing/test_pluginmanager.py
blueyed: Ok. Moved to `_config_for_test` therein.

nicoddemus: Good idea to use `_`. 👍

PR #5007, opened 2019-03-27
runner: runtestprotocol: remove unused reports

blueyed:
Apparently never used - first occurred in 2009 in ed216e77.

### DIFF
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -85,19 +85,17 @@ def runtestprotocol(item, log=True, nextitem=None):
     if hasrequest and not item._request:
         item._initrequest()
     rep = call_and_report(item, "setup", log)
-    reports = [rep]
     if rep.passed:
         if item.config.getoption("setupshow", False):
             show_test_item(item)
         if not item.config.getoption("setuponly", False):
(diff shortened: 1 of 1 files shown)


src/_pytest/runner.py
```diff
@@ -85,19 +85,17 @@ def runtestprotocol(item, log=True, nextitem=None):
...
     rep = call_and_report(item, "setup", log)
-    reports = [rep]
     if rep.passed:
         if item.config.getoption("setupshow", False):
             show_test_item(item)
         if not item.config.getoption("setuponly", False):
-            reports.append(call_and_report(item, "call", log))
-    reports.append(call_and_report(item, "teardown", log, nextitem=nextitem))
+            call_and_report(item, "call", log)
+    call_and_report(item, "teardown", log, nextitem=nextitem)
```
nicoddemus: Probably there are plugins which might be using this. Although not public API, we shouldn't remove this without a good reason I think.

blueyed: Indeed - https://github.com/pytest-dev/pytest-forked/blob/42d7146d65749b54571480db51a8461aa509b3ac/src/pytest_forked/__init__.py#L50.

PR #5009, opened 2019-03-28
[WIP] coverage: try timid=1 (PyTracer)

blueyed:


blueyed:
/cc @nedbat (via IRC already).
Closing for now.

nicoddemus:
@blueyed can you give a quick summary of what's the purpose of this PR?

blueyed:
Trying out timid=1 (PyTracer).

blueyed:
I plan to minkeypatch `sys.settrace` to always call the coverage tracer, and thought that it might be easier with PyTracer (but probably not).

testing/code/test_excinfo.py
```diff
@@ -238,7 +238,10 @@ def f(n):
                 n += 1
             f(n)
 
+        # Restore trace function (https://bugs.python.org/issue36474).
+        oldtrace = sys.gettrace()
         excinfo = pytest.raises(RuntimeError, f, 8)
+        sys.settrace(oldtrace)
```
blueyed: Do you think it makes sense to let `pytest.raises` handle this, maybe optionally by default only?

blueyed: The problem here is that Python forgets about the trace function with RecursionErrors (https://bugs.python.org/issue36474).

PR #5013, opened 2019-03-29
Display message from reprcrash in short test summary

blueyed:
This is useful to see common patterns easily, but also for single
failures already.

There might be a better API to get to the message?!

See https://github.com/pytest-dev/pytest/pull/5002#issuecomment-477078420 for some discussion/ideas to change the format in general / make it configurable.


asottile:
What does this look like for longer messages?  The example in the test looks fine for a trivial `assert 0` but I suspect for something like `raise TypeError('very long message blah blah blah')` this might get unwieldy

nicoddemus:
> this might get unwieldy

Indeed, we use very long messages (even multi-line ones) at work. How about truncating up until the terminal width?

blueyed:
It uses the first line currently already only.
Truncating up until the terminal width sounds good (reminded me of the not so good working truncation of ini-options in `pytest --help`, which appears to use a fixed width of 80 (likely an argparse issue)).

blueyed:
Updated.  Good for adding a changelog?

src/_pytest/skipping.py
```diff
@@ -211,7 +211,25 @@ def show_simple(terminalreporter, lines, stat):
...
+            else:
+                # Only use the first line.
+                # Might be worth having a short_message property, which
+                # could default to this behavior.
+                i = msg.find("\n")
+                if i != -1:
+                    msg = msg[:i]
+                max_len = terminalreporter.writer.fullwidth - len(line) - 2
+                if len(msg) > max_len:
+                    msg = msg[: (max_len - 1)] + "…"
```
nicoddemus: Let's avoid writing unicode characters to the terminal, `cmd.exe` for example does not support this and will likely blow up.

src/_pytest/skipping.py
```diff
@@ -211,7 +211,25 @@ def show_simple(terminalreporter, lines, stat):
...
+            except AttributeError:
+                pass
+            else:
+                # Only use the first line.
+                # Might be worth having a short_message property, which
+                # could default to this behavior.
+                i = msg.find("\n")
+                if i != -1:
+                    msg = msg[:i]
+                max_len = terminalreporter.writer.fullwidth - len(line) - 2
```
nicoddemus: For sake of legibility, I suggest:

```python
sep = ": "
max_len = terminalreporter.writer.fullwidth - len(line) - len(sep)
...
line += sep + msg
```

But you probably need to adjust for my comment to avoid using `…`, probably in favor of `...`.

Also can we move this code to a function and unittest it?

src/_pytest/skipping.py
blueyed: Ok, changed to "...".

src/_pytest/skipping.py
blueyed: Done - but needs to be revisited.

blueyed:
Noticed that having ":" at the end of the nodeid is bad for copy'n'pasing them..
Change to "VERB node::id - msg" maybe?

nicoddemus:
> Change to "VERB node::id - msg" maybe?

```
FAILED testing/test_cacheprovider.py::TestNewAPI::test_config_cache_makedir: assert 0
```

On my terminal (Cmder on Windows), double clicking on the test line only selects the node id, the `:` at the end is left out. For you double clicking also includes the `:`?

Anyway, I'm fine with using ` - ` as separator. 👍

blueyed:
Offtopic:

> reminded me of the not so good working truncation of ini-options in `pytest --help`, which appears to use a fixed width of 80

This is due to me using `pytest --help | less`, where stdout then is redirected, and `shutil.get_terminal_size()` fails (https://github.com/python/cpython/pull/12697).

src/_pytest/skipping.py
```diff
@@ -204,14 +204,46 @@ def pytest_terminal_summary(terminalreporter):
...
+        if i != -1:
+            msg = msg[:i]
+        len_msg = len(msg)
+
+        sep, len_sep = " - ", 3
+        max_len_msg = termwidth - len_line - len_sep
+        if max_len_msg >= len_ellipsis:
+            if len_msg > max_len_msg:
+                msg = msg[: (max_len_msg - len_ellipsis)] + ellipsis
+            line += sep + msg
```
nicoddemus: Oh just realized this: is `msg` unicode-safe in Python 2? because we will be joining a `str/bytes` with a potential unicode message, so there's a good chance this might blow up

blueyed: Would appreciate a failing example..  do you mean we should prefix `sep` etc also with `u`?

nicoddemus: Hmm actually no, this is safe: if `msg` is unicode, Python 2 will try to elevate `sep` to unicode using ascii (which is fine), not the other way around. Sorry for the brain fart. 👍

nicoddemus: If you want to play it safe and ensure we don't regress though, you can change the message to use a smiling face emoji instead (😄), but is up to you. 👍

blueyed: Added.. but it shows that we're not handling terminal cells there really - i.e. it would wrap due to this.

nicoddemus: Oh. Hmm not even sure what the solution would be, TBH.

blueyed: added a commit using wcwidth.. not sure if it is worth the extra dependency, but it worked well.. ;)

src/_pytest/skipping.py
```diff
@@ -204,14 +204,52 @@ def pytest_terminal_summary(terminalreporter):
...
+        len_msg = wcswidth(msg)
+
+        sep, len_sep = " - ", 3
+        max_len_msg = termwidth - len_line - len_sep
+        if max_len_msg >= len_ellipsis:
+            if len_msg > max_len_msg:
+                max_len_msg -= len_ellipsis
+                msg = msg[:max_len_msg]
+                while wcswidth(msg) > max_len_msg:
+                    msg = msg[:-1]
```
blueyed: This could be optimized probably if we're taking it - there is also `wcwidth`.

blueyed:
I think in the long run the terminal writer should become aware of cells instead of chars.

src/_pytest/skipping.py
nicoddemus: I think adding `wcwidth` is fine: small, license is OK, seems to be maintained

blueyed:
Hmm, py2 only fails on Windows and MacOS..

nicoddemus:
> Hmm, py2 only fails on Windows and MacOS..

Oh right, I saw Travis and thought it was Linux, didn't realize it was the MacOS job. 

I can take a look at the Windows failure tomorrow if you like

blueyed:
@nicoddemus 
Yes, please take a look.

src/_pytest/skipping.py
```diff
@@ -204,14 +207,60 @@ def pytest_terminal_summary(terminalreporter):
...
+                if six.PY2:
```
nicoddemus: Here's why it was failing on some Python 2 systems (I'm guessing). This fixes this for me on my 32-bit Python 2.7, let's see how it fares on MacOS.

nicoddemus:
@asottile and @wimglenn, any other comments or can we merge this?

src/_pytest/terminal.py
```diff
@@ -956,6 +956,57 @@ def _get_pos(config, rep):
...
+                msg = msg[:max_len_msg]
+                while wcswidth(msg) > max_len_msg:
+                    msg = msg[:-1]
+                if six.PY2:
+                    # on python 2 systems with narrow unicode compilation, trying to
+                    # get a single character out of a multi-byte unicode character such as
+                    # u'😄' will result in a High Surrogate (U+D83D) character, which is
+                    # rendered as u'�'; in this case we just strip that character out as it
+                    # serves no purpose being rendered
+                    while msg.endswith(u"\uD83D"):
```
wimglenn: msg.rstrip not working for some reason...?

nicoddemus: Hmmm you are right, not sure why it didn't occur to me. Let me try that. 👍

nicoddemus: Done, thanks!

blueyed:
> changelog entry mentioning the new dependency?

This appears to be missing still.

src/_pytest/terminal.py
```diff
@@ -956,6 +956,56 @@ def _get_pos(config, rep):
...
+        len_msg = wcswidth(msg)
+
+        sep, len_sep = " - ", 3
+        max_len_msg = termwidth - len_line - len_sep
+        if max_len_msg >= len_ellipsis:
+            if len_msg > max_len_msg:
+                max_len_msg -= len_ellipsis
+                msg = msg[:max_len_msg]
+                while wcswidth(msg) > max_len_msg:
+                    msg = msg[:-1]
```
blueyed: Said this earlier already, but this could certainly be done better, e.g. by using `wcwidth` from the end/tail (substracting it), but well..  I think when we use this in other places it can be improved then still.

nicoddemus:
> This appears to be missing still.

Ahh thanks for remembering this.

I've pushed my suggestion, feel free to edit it as you see fit. 👍

changelog/5013.trivial.rst
```diff
@@ -0,0 +1 @@
+pytest now depends on `wcwidth <https://pypi.org/project/wcwidth>`__ to properly track unicode character sizes for more precise terminal output.
```
blueyed: Thanks! :heart:

blueyed:
Let's :sheep: :it:.

The-Compiler:
```
ERROR tests/end2end/features/test_open_bdd.py::test_opening_a_quickmark - ass...
```

well, thanks pytest, I guess? :laughing:

blueyed:
At least we've not used ", " as delimiter here.. ;)

PR #5015, opened 2019-03-31
Pre-PR for warnings on unknown markers

Zac-HD:
This pull request is designed to make #4935 much smaller and easier to work on (or review).  The four commits each make a helpful change that is also worth doing independently.

- Typo fixes should be clear.
- `pytester` was using an unregistered marker - adding it to the help is both useful for users and will avoid making usage a warning later.
- We needed to document how a mark could be registered, so I've done that and linked to it from the section on `--strict` and marks.
- Our test suite already uses both `mark.issue` and `mark.issueNNN`; standardising on the former makes selecting issue-relevant tests uniform - and means we won't need to register every number.

I'm targeting `features` so that I can rebase #4935 as soon as this is merged :smile:

### DIFF
--- a/doc/en/mark.rst
+++ b/doc/en/mark.rst
@@ -43,7 +43,7 @@ Marks can be registered in ``pytest.ini`` like this:
 When the ``--strict`` command-line flag is passed, any unknown marks applied
 with the ``@pytest.mark.name_of_the_mark`` decorator will trigger an error.
 Marks added by pytest or by a plugin instead of the decorator will not trigger
-this error.  To enforce a limited set of markers, add ``--strict`` to ``addopts``:
+this error.  To enforce validation of markers, add ``--strict`` to ``addopts``:
 
 .. code-block:: ini
 
--- a/doc/en/mark.rst
+++ b/doc/en/mark.rst
@@ -26,12 +26,10 @@ which also serve as documentation.
     :ref:`fixtures <fixtures>`.
 
 
-Raising errors on unknown marks: --strict
------------------------------------------
+.. _unknown-marks:
 
-When the ``--strict`` command-line flag is passed, any unknown marks applied
-with the ``@pytest.mark.name_of_the_mark`` decorator will trigger an error.
-Marks defined or added by pytest or by a plugin will not trigger an error.
+Raising errors on unknown marks
+-------------------------------
 
 Marks can be registered in ``pytest.ini`` like this:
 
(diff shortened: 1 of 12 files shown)


doc/en/mark.rst
```diff
@@ -42,8 +40,10 @@ Marks can be registered in ``pytest.ini`` like this:
         slow
         serial
 
-This can be used to prevent users mistyping mark names by accident. Test suites that want to enforce this
-should add ``--strict`` to ``addopts``:
+When the ``--strict`` command-line flag is passed, any unknown marks applied
+with the ``@pytest.mark.name_of_the_mark`` decorator will trigger an error.
+Marks added by pytest or by a plugin instead of the decorator will not trigger
+this error.  To enforce a limited set of markers, add ``--strict`` to ``addopts``:
```
blueyed:
```suggestion
this error.  To enforce validation of markers, add ``--strict`` to ``addopts``:
```

?

tox.ini
```diff
@@ -166,6 +166,8 @@ filterwarnings =
     # Do not cause SyntaxError for invalid escape sequences in py37.
     default:invalid escape sequence:DeprecationWarning
 pytester_example_dir = testing/example_scripts
+markers =
+    issue
```
blueyed: s/issue/gh_issue/ maybe?

Zac-HD: We're already using `issue`, so I'd rather leave any changes to a future PR and just standardise here.

blueyed: Well, a) I do not know how this is used actually, but b) you're touching all those lines already (and therefore it would be good to do it right away).

nicoddemus: > We're already using issue, so I'd rather leave any changes to a future PR and just standardise here.

I never used it myself. I think we can actually remove them in the future, unless someone steps up and mentions he/she uses it on their workflow.

doc/en/mark.rst
```diff
@@ -42,8 +40,10 @@ Marks can be registered in ``pytest.ini`` like this:
         slow
         serial
 
-This can be used to prevent users mistyping mark names by accident. Test suites that want to enforce this
-should add ``--strict`` to ``addopts``:
+When the ``--strict`` command-line flag is passed, any unknown marks applied
```
nicoddemus: @Zac-HD how about using this PR and add `--strict-marks` as an alias to `--strict`, while also updating the docs to use the former?

Zac-HD: I'd prefer to leave functional changes to the other PR - will discuss there.

nicoddemus: Sure, thanks!

Issue #5018, opened 2019-04-01
Fail doc/en/example/assertion/test_failures.py::test_failure_demo_fails_properly on python-2.7


mimi1vx:
- [x ] Include a detailed description of the bug or suggestion
- [ ] `pip list` of the virtual environment you are using
- [x] pytest and operating system versions
- [ ] Minimal example if possible

pytest-4.4.0, python-2.7 on openSUSE:Tumbleweed

pytest doc testsuite uses python 3.6 + syntax and fails on python2.7

```
  397s] _______________________ test_failure_demo_fails_properly _______________________
[  397s] 
[  397s] testdir = <Testdir local('/tmp/pytest-of-abuild/pytest-0/test_failure_demo_fails_properly0')>
[  397s] 
[  397s]     def test_failure_demo_fails_properly(testdir):
[  397s]         target = testdir.tmpdir.join(failure_demo.basename)
[  397s]         failure_demo.copy(target)
[  397s]         failure_demo.copy(testdir.tmpdir.join(failure_demo.basename))
[  397s]         result = testdir.runpytest(target, syspathinsert=True)
[  397s] >       result.stdout.fnmatch_lines(["*44 failed*"])
[  397s] E       Failed: nomatch: '*44 failed*'
[  397s] E           and: u'============================= test session starts =============================='
[  397s] E           and: u'platform linux2 -- Python 2.7.15, pytest-4.4.0, py-1.8.0, pluggy-0.9.0'
[  397s] E           and: u"hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/abuild/rpmbuild/BUILD/pytest-4.4.0/.hypothesis/examples')"
[  397s] E           and: u'rootdir: /tmp/pytest-of-abuild/pytest-0/test_failure_demo_fails_properly0'
[  397s] E           and: u'plugins: hypothesis-4.13.0'
[  397s] E           and: u'collected 0 items / 1 errors'
[  397s] E           and: u''
[  397s] E           and: u'==================================== ERRORS ===================================='
[  397s] E           and: u'_______________________ ERROR collecting failure_demo.py _______________________'
[  397s] E           and: u'/usr/lib/python2.7/site-packages/_pytest/python.py:511: in _importtestmodule'
[  397s] E           and: u'    mod = self.fspath.pyimport(ensuresyspath=importmode)'
[  397s] E           and: u'/usr/lib/python2.7/site-packages/py/_path/local.py:701: in pyimport'
[...]

PR #5020, opened 2019-04-01
ci: use Python 2.7 for doctesting on Azure

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5018

Expected to fail.


blueyed:
Meh, it uses "python3" explicitly (via d52ea4b6).

PR #5029, opened 2019-04-03
Minor: doc, import

blueyed:


blueyed:
This removes `_pytest.runner.skip` - is that OK?

nicoddemus:
> This removes _pytest.runner.skip - is that OK?

Yep that's fine. 👍

PR #5031, opened 2019-04-03
pytester: use monkeypatch with Testdir

blueyed:


src/_pytest/pytester.py
```diff
@@ -508,7 +512,7 @@ def finalize(self):
         self._sys_modules_snapshot.restore()
```
nicoddemus: While at it, would you like to replace those "snapshot" implementations by your `monkeypatch` instance?

If not, please go ahead and merge it as this is already an improvement 

(every time I looked at this code I had the same idea to use monkeypatch for that, but kept forgetting to actually do it 🙃)

testing/test_pytester.py
```diff
@@ -469,3 +469,13 @@ def test_linematcher_with_nonlist():
 
     assert lm._getlines({}) == {}
     assert lm._getlines(set()) == set()
+
+
+def test_pytester_addopts(request, monkeypatch):
+    monkeypatch.setenv("PYTEST_ADDOPTS", "--orig-unused")
+
+    testdir = request.getfixturevalue("testdir")
+    assert "PYTEST_ADDOPTS" not in os.environ
```
nicoddemus: You should surround this `assert` with a `try/finally` to ensure `testdir.finalize()` gets called

src/_pytest/pytester.py
blueyed: Not really now - basically got side-tracked by this already for #5032.

blueyed: btw: to me it looked just like code that could be improved, assuming that `os.environ` would get restored, but that's actually not the case..  therefore targetting master.

blueyed:
Azure failure is unrelated.

nicoddemus:
Thanks!

PR #5034, opened 2019-04-03
run-last-failure: improve reporting

blueyed:


src/_pytest/cacheprovider.py
```diff
@@ -196,18 +189,35 @@ def pytest_collection_modifyitems(self, session, config, items):
...
-                    # of the set of tests currently executing
-                    return
-                if self.config.getoption("lf"):
-                    items[:] = previously_failed
-                    config.hook.pytest_deselected(items=previously_passed)
+                    # Running a subset of all tests with recorded failures
+                    # only outside of it.
+                    self._report_status = "%d known failures not in selected tests" % (
+                        len(self.lastfailed),
+                    )
```
blueyed: This could have a hint to use `--cache-show cache/lastfailed` to see them.

(I am having a lot there from files that do not exist anymore etc - it could also have an improved message for when no args are passed really then)

nicoddemus: If it doesn't get too verbose, I'm fine with it. Up to you. 👍

blueyed: Maybe / rather in a followup.

PR #5035, opened 2019-04-03
Support glob argument with ``--cache-show``

blueyed:
TODO:

- [x] changelog

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -292,9 +292,13 @@ def pytest_addoption(parser):
     )
     group.addoption(
         "--cache-show",
-        action="store_true",
+        action="append",
+        nargs="?",
         dest="cacheshow",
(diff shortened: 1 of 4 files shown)


src/_pytest/cacheprovider.py
```diff
@@ -282,9 +282,13 @@ def pytest_addoption(parser):
     )
     group.addoption(
         "--cache-show",
-        action="store_true",
+        action="append",
+        nargs="?",
```
blueyed: There might be a better way?

nicoddemus: I've tried a few things (action="store" with default="*") but that didn't work, so I guess this is fine.

blueyed: Ok, borrowed / inspired by pytest-cov's `--cov`.

nicoddemus:
Great work! 👍

PR #5037, opened 2019-04-03
docs: revisit/fix regen

blueyed:


doc/en/assert.rst
```diff
@@ -29,8 +29,8 @@ you will see the return value of the function call:
     $ pytest test_assert1.py
     =========================== test session starts ============================
     platform linux -- Python 3.x.y, pytest-4.x.y, py-1.x.y, pluggy-0.x.y
-    cachedir: $PYTHON_PREFIX/.pytest_cache
-    rootdir: $REGENDOC_TMPDIR
+    cachedir: /tmp/tox/home/daniel/Vcs/pytest/regen/.pytest_cache
+    rootdir: /path/to/example
```
blueyed: Where was the `$PYTHON_PREFIX` coming from previously?!

blueyed: Via lost releases?!  https://github.com/pytest-dev/regendoc/issues/4

doc/en/capture.rst
```diff
@@ -71,7 +71,7 @@ of the failing function and hide the other one:
     =========================== test session starts ============================
     platform linux -- Python 3.x.y, pytest-4.x.y, py-1.x.y, pluggy-0.x.y
     cachedir: $PYTHON_PREFIX/.pytest_cache
-    rootdir: $REGENDOC_TMPDIR
+    rootdir: /path/to/example
```
blueyed: Those changes are due to master / https://github.com/pytest-dev/regendoc/pull/8 being used, can update from 0.6.1, but I think it is good and the mess with the missed releases is hopefully merged/fixed soon (i.e. we would get this output also with release versions).

blueyed: Anyway, we could have a custom replacement for this also..

doc/en/Makefile
```diff
@@ -13,6 +13,7 @@ PAPEROPT_letter = -D latex_paper_size=letter
 ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
 REGENDOC_ARGS := \
+	--normalize "/[ \t]+\n/\n/" \
```
blueyed: This avoids having to run pre-commit afterwards.

doc/en/capture.rst
blueyed: Added a custom replacement.

blueyed:
Closing in favor of #5042.

PR #5051, opened 2019-04-04
Fix test_conftest when run via pytest-randomly

blueyed:


testing/test_conftest.py
```diff
@@ -15,13 +16,24 @@
 
 @pytest.fixture(scope="module", params=["global", "inpackage"])
 def basedir(request, tmpdir_factory):
+    # Do not write bytecode to make tests not fail when run randomly
+    # (ConftestImportFailure with TestConftestValueAccessGlobal).
+    orig_env = os.environ.get("PYTHONDONTWRITEBYTECODE")
```
nicoddemus: I suggest to create a `MonkeyPatch()` instance manually and call `undo` at the end

blueyed: Turned out it was related to sys.modules instead - used SysModulesSnapshot there now.

blueyed:
I wonder if we should have a strict CI job, which then would use this:
```diff
diff --git c/testing/conftest.py i/testing/conftest.py
new file mode 100644
index 00000000..b4856828
--- /dev/null
+++ i/testing/conftest.py
@@ -0,0 +1,13 @@
+import sys
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _check_syspath():
+    """Ensure that tests do not mess with sys.path."""
+    oldpath = sys.path[:]
+    yield
+    assert (
+        oldpath == sys.path
+    ), "sys.path was changed in test, consider using _sys_snapshot"
```

nicoddemus:
> I wonder if we should have a strict CI job

Excellent idea, but do you mean a separate job? I think that fixture should be included in every normal run, it is cheap and catches mistakes early. 👍

blueyed:
Wrapping every test is not that cheap, is it? :)
But if it's OK we could just have it by default already (i.e. not only on CI).

nicoddemus:
> Wrapping every test is not that cheap, is it? :)
But if it's OK we could just have it by default already (i.e. not only on CI).

I would say extremely cheap if compared to all the other things that are happening behind the scenes; just to put things into perspective, consider how much works is done just to call a hook... having an extra fixture which will cost a few ms to the end of the test run is definitely worth it if we are avoiding mistakes in the process. 😉

blueyed:
I think the check should be added separately, and IMHO it is enough to run this from time to time manually (like I did here).

PR #5052, opened 2019-04-05
Add the junit_log_passing_tests ini flag to skip logging output for passing tests.

### LINKED ISSUE #4559: junit-xml: add option for disabling logs for passed tests
We have near 130k tests for some web-API, and in these tests we log a lot of http requests with its request/response headers, data, query and etc. We are logging it with logging module (`logging.getLogger('api_logger')`). There is a lot of logs, really. We are using Jenkins to run these tests and `--junit-xml` option for creating junit-xml style report.

So the problem is connected with  junit-xml creation. Junit-xml reporter captures not only failed tests (<1%) but passed tests too (>99%) and it can crash the Jenkins process and freeze the runner-server for a while.

There are some ways to disable logs for failed tests like:
- `--show-capture=no` Controls how captured stdout/stderr/log is shown on failed tests. Default is 'all',
- `--no-print-logs` disable printing caught logs on failed tests,

but no possibility to disable logs for passed test.

Right now we handle this with custom `pytest_runtest_logreport` that checks `if report.failed and report.when == 'call'`

So, maybe pytest needs an option for junit-xml that disables logs for passed tests and logs only failed, what do you think?

jeffreyrack:
Closes #4559 by adding the junit_log_passing_tests ini flag which can be turned off to disable logging test output and logging for passing tests in the junit XML report.

### DIFF
--- a/changelog/4559.feature.rst
+++ b/changelog/4559.feature.rst
@@ -1,0 +1,1 @@
+Added the ``junit_log_passing_tests`` ini value which can be used to enable and disable logging passing test output in the Junit XML file.
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -166,6 +166,9 @@ class _NodeReporter(object):
         self.append(node)
 
     def write_captured_output(self, report):
+        if not self.xml.log_passing_tests and report.passed:
+            return
+
         content_out = report.capstdout
         content_log = report.caplog
         content_err = report.capstderr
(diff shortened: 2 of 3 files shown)


andkononykhin:
@jeffreyrack @nicoddemus Hello. Thank you for that option but seems the change introduces one drawback: it disables any junit logging for failed tests which don't fail during `teardown` phase.

thisch:
@andkononykhin Hi, thank you very much for the report. Can you create a ticket/issue for this regression with a minimal example please? Thx

andkononykhin:
@thisch  I created https://github.com/pytest-dev/pytest/issues/5430. Sorry for the delay.

PR #5056, opened 2019-04-05
Inject width via pylib to argparse formatter

blueyed:
`argparse.HelpFormatter` looks at `$COLUMNS` only, falling back to a default of 80.

`py.io.get_terminal_width()` is smarter there, and could even work better with https://github.com/pytest-dev/py/pull/219.

This ensures to use a consistent value for formatting the ini values etc.

1: https://github.com/pytest-dev/py/pull/219

TODO:

- [x] changelog
- [x] coverage

### DIFF
--- a/src/_pytest/config/argparsing.py
+++ b/src/_pytest/config/argparsing.py
@@ -405,6 +405,12 @@ class DropShorterLongHelpFormatter(argparse.HelpFormatter):
     - cache result on action object as this is called at least 2 times
     """
 
+    def __init__(self, *args, **kwargs):
+        """Use more accurate terminal width via pylib."""
+        if "width" not in kwargs:
+            kwargs["width"] = py.io.get_terminal_width()
+        super().__init__(*args, **kwargs)
+
     def _format_action_invocation(self, action):
         orgstr = argparse.HelpFormatter._format_action_invocation(self, action)
         if orgstr and orgstr[0] != "-":  # only optional arguments
--- a/changelog/5056.trivial.rst
+++ b/changelog/5056.trivial.rst
@@ -1,0 +1,1 @@
+The HelpFormatter uses ``py.io.get_terminal_width`` for better width detection.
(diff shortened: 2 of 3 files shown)


blueyed:
Together with the py patch, it makes `pytest --help | less` look much nicer now (I have a `HH` shell suffix alias, which I usually use to look at `--help` output).

src/_pytest/config/argparsing.py
```diff
@@ -405,6 +405,12 @@ class DropShorterLongHelpFormatter(argparse.HelpFormatter):
     - cache result on action object as this is called at least 2 times
     """
 
+    def __init__(self, *args, **kwargs):
+        """Use more accurate terminal width via pylib."""
+        if "width" not in kwargs:
+            kwargs["width"] = py.io.get_terminal_width()
```
asottile: should we use `setdefault(...)` and `shutil.get_terminal_width`?

blueyed: `setdefault` maybe, but otherwise it is meant to use an improved version (https://github.com/pytest-dev/py/pull/219).

asottile: hmmm I'm firmly in the camp that we should stop adding things to pylib

blueyed: sure, but better then to pytest - assuming that Python does not fix it itself.  IIRC there's an old issue / PR idling there.
Anyway, this is merged, and pylib's PR is also idling.  So no worries.. ;)

asottile: I'd rather put it in pytest, I've been trying to factor out pylib and this does the opposite!

blueyed: ok, maybe later.
Feel free to fix this one up like you suggested initially then if it bothers you.

nicoddemus: Just FTR, I'm also in the camp that we should avoid adding new things to pylib at all costs, even though dropping it as a dependency is a *far* away dream.

PR #5059, opened 2019-04-06
pytester: allow passing in stdin to run/popen

blueyed:
This makes it more flexible to use `testdir.run` / `testdir.popen`.

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1034,7 +1034,14 @@ class Testdir(object):
             if colitem.name == name:
                 return colitem
 
-    def popen(self, cmdargs, stdout, stderr, stdin=CLOSE_STDIN, **kw):
+    def popen(
+        self,
+        cmdargs,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        stdin=CLOSE_STDIN,
+        **kw
+    ):
         """Invoke subprocess.Popen.
 
         This calls subprocess.Popen making sure the current working directory
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -36,8 +36,6 @@ IGNORE_PAM = [  # filenames added when obtaining details about the current user
     u"/var/lib/sss/mc/passwd"
 ]
 
-CLOSE_STDIN = object
-
 
 def pytest_addoption(parser):
     parser.addoption(
@@ -474,6 +472,8 @@ class Testdir(object):
        the method using them so refer to them for details.
 
     """
+
+    CLOSE_STDIN = object
 
     class TimeoutExpired(Exception):
         pass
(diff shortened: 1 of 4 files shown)


src/_pytest/pytester.py
```diff
@@ -1032,7 +1034,7 @@ def collect_by_name(self, modcol, name):
             if colitem.name == name:
                 return colitem
 
-    def popen(self, cmdargs, stdout, stderr, **kw):
+    def popen(self, cmdargs, stdout, stderr, stdin=CLOSE_STDIN, **kw):
```
blueyed: Also making `stdout` and `stderr` have a default here would be nice.  What do you think?
They could just default `subprocess.PIPE`.

src/_pytest/pytester.py
```diff
@@ -1086,8 +1099,15 @@ def run(self, *cmdargs, **kwargs):
...
-                cmdargs, stdout=f1, stderr=f2, close_fds=(sys.platform != "win32")
+                cmdargs,
+                stdout=f1,
+                stderr=f2,
+                close_fds=(sys.platform != "win32"),
+                **popen_kwargs
             )
+            if stdin is not CLOSE_STDIN:
+                popen.stdin.write(stdin)
+                popen.stdin.close()
```
blueyed: This is a bit wonky here (closing with "not CLOSE_STDIN"), but it is due to compatibility with `popen`'s behavior mostly.

blueyed: Improved by handling `bytes` in a special way.

src/_pytest/pytester.py
blueyed: Done.

src/_pytest/pytester.py
```diff
@@ -36,6 +36,8 @@
     u"/var/lib/sss/mc/passwd"
 ]
 
+CLOSE_STDIN = object
```
nicoddemus: Perhaps move this constant to the class for convenience? Then users can use `testdir.CLOSE_STDIN` instead of having to import from `_pytest.pytester`.

src/_pytest/pytester.py
```diff
@@ -1065,15 +1082,24 @@ def run(self, *cmdargs, **kwargs):
...
 
         """
         __tracebackhide__ = True
 
         timeout = kwargs.pop("timeout", None)
+        stdin = kwargs.pop("stdin", CLOSE_STDIN)
         raise_on_kwargs(kwargs)
 
+        popen_kwargs = {"stdin": stdin}
+        if isinstance(stdin, bytes):
```
nicoddemus: It doesn't seem like you need to handle this explicitly, you may pass `stdin` directly through `self.popen` and it will do the right thing.

testing/test_pytester.py
```diff
@@ -482,3 +483,79 @@ def test_pytester_addopts(request, monkeypatch):
...
+
+
+def test_run_stdin(testdir):
+    with pytest.raises(testdir.TimeoutExpired):
+        testdir.run(
+            sys.executable,
+            "-c",
+            "import sys; print(sys.stdin.read())",
+            stdin=subprocess.PIPE,
+            timeout=0.1,
```
nicoddemus: This doesn't seem reliable, better add a sleep here like you do with the block below this one?

src/_pytest/pytester.py
blueyed: Yes, only closing is needed then.  Thanks!

Issue #5060, opened 2019-04-06
Mixed colors within lines


blueyed:
1. I think it would be helpful if the "1 passed" would be green in `= 1 failed, 1 passed in 0.25 seconds =` (same for other statuses), and only the rest (filler, timing) in the dominating (current) color.

2. It would also be nice in other cases to not just use red for the whole line that starts with "E ", but e.g. there could be only the "E " highlighted in red for the second and follow lines.
I think this would look better with e.g. `fnmatch_lines` failures, which currently can often result in several screens of red colored output.
Having just the "headline" (first line) highlighted there would make it easier to find them visually.

PR #5061, opened 2019-04-06
Multiple colors with terminal summary_stats

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5060

This is quite a hack, and should probably go with better support in py itself, or some more generic wrapping.

55671348-86675f80-588f-11e9-9997-b6efed19514a

TODO:

- [x] (adjust) tests
- [x] changelog

### DIFF
--- a/changelog/5061.feature.rst
+++ b/changelog/5061.feature.rst
@@ -1,0 +1,1 @@
+Use multiple colors with terminal summary statistics.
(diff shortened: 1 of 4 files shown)


nicoddemus:
This is what it looks like on my terminal:

image

I'm OK with it; at first it does seem too colorful, but after using it for awhile I can see that it clearly brings attention to the important part of the message.

Let's see what others think.

blueyed:
No other feedback?

blueyed:
Adjusted tests, added changelog.

PR #5064, opened 2019-04-07
[WIP] revisit style with terminal reporting

blueyed:
1. use less bold/red with location representation

## Using `--tb=short` before:

less-bold-before

## After:

less-bold-after

TODO:

- [ ] tests
- [ ] changelog

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -773,10 +773,15 @@ class FormattedExcinfo:
             extraline = None
 
         last = traceback[-1]
+        crashentry = traceback.getcrashentry()
         entries = []
         for index, entry in enumerate(traceback):
             einfo = (last == entry) and excinfo or None
             reprentry = self.repr_traceback_entry(entry, einfo)
+
+            # Store this for toterminal.
+            reprentry._is_crashentry = entry == crashentry
+
             entries.append(reprentry)
         return ReprTraceback(entries, extraline, style=self.style)
 
@@ -963,6 +968,8 @@ class ReprEntryNative(TerminalRepr):
 
 
 class ReprEntry(TerminalRepr):
+    _is_crashentry = None
+
     def __init__(self, lines, reprfuncargs, reprlocals, filelocrepr, style):
         self.lines = lines
         self.reprfuncargs = reprfuncargs
@@ -972,11 +979,14 @@ class ReprEntry(TerminalRepr):
 
     def toterminal(self, tw):
         if self.style == "short":
-            self.reprfileloc.toterminal(tw)
+            self.reprfileloc.toterminal(tw, style="short")
             for line in self.lines:
-                red = line.startswith("E   ")
-                tw.line(line, bold=True, red=red)
+                is_error = line.startswith("E   ")
+                red = is_error
+                bold = is_error or self._is_crashentry
+                tw.line(line, bold=bold, red=red)
             return
+
         if self.reprfuncargs:
             self.reprfuncargs.toterminal(tw)
         for line in self.lines:
@@ -1002,15 +1012,16 @@ class ReprFileLocation(TerminalRepr):
         self.lineno = lineno
         self.message = message
 
-    def toterminal(self, tw):
+    def toterminal(self, tw, style=None):
         # filename and lineno output for each entry,
         # using an output format that most editors unterstand
         msg = self.message
         i = msg.find("\n")
         if i != -1:
             msg = msg[:i]
-        tw.write(self.path, bold=True, red=True)
-        tw.line(":{}: {}".format(self.lineno, msg))
+        bold = style != "short"
+        tw.write("%s" % self.path, bold=bold)
+        tw.line(":%s: %s" % (self.lineno, msg))
 
 
 class ReprLocals(TerminalRepr):
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -773,14 +773,10 @@ class FormattedExcinfo:
             extraline = None
 
         last = traceback[-1]
-        crashentry = traceback.getcrashentry()
         entries = []
         for index, entry in enumerate(traceback):
             einfo = (last == entry) and excinfo or None
             reprentry = self.repr_traceback_entry(entry, einfo)
-
-            # Store this for toterminal.
-            reprentry._is_crashentry = entry == crashentry
 
             entries.append(reprentry)
         return ReprTraceback(entries, extraline, style=self.style)
@@ -968,8 +964,6 @@ class ReprEntryNative(TerminalRepr):
 
 
 class ReprEntry(TerminalRepr):
-    _is_crashentry = None
-
     def __init__(self, lines, reprfuncargs, reprlocals, filelocrepr, style):
         self.lines = lines
         self.reprfuncargs = reprfuncargs
@@ -982,9 +976,7 @@ class ReprEntry(TerminalRepr):
             self.reprfileloc.toterminal(tw, style="short")
             for line in self.lines:
                 is_error = line.startswith("E   ")
-                red = is_error
-                bold = is_error or self._is_crashentry
-                tw.line(line, bold=bold, red=red)
+                tw.line(line, bold=is_error, red=is_error)
             return
 
         if self.reprfuncargs:
@@ -1021,7 +1013,7 @@ class ReprFileLocation(TerminalRepr):
             msg = msg[:i]
         bold = style != "short"
         tw.write("%s" % self.path, bold=bold)
-        tw.line(":%s: %s" % (self.lineno, msg))
+        tw.line(":{}: {}".format(self.lineno, msg))
 
 
 class ReprLocals(TerminalRepr):
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -609,7 +609,7 @@ class TerminalReporter:
 
         if self.config.getoption("collectonly"):
             if self.stats.get("failed"):
-                self._tw.sep("!", "collection failures")
+                self.write_sep("!", "collection failures")
                 for rep in self.stats.get("failed"):
                     rep.toterminal(self._tw)
 
@@ -807,7 +807,7 @@ class TerminalReporter:
             if showcapture != "all" and showcapture not in secname:
                 continue
             if "teardown" in secname:
-                self._tw.sep("-", secname)
+                self.write_sep("-", secname)
                 if content[-1:] == "\n":
                     content = content[:-1]
                 self._tw.line(content)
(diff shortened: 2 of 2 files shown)


RonnyPfannschmidt:
im wondering how it looked if filenames/line numers where bold and the code line was pygmentized

blueyed:
Yes, using real syntax highlighting would be nice, but I consider this to be a next step (currently just working on small iterations in this area).

blueyed:
pdbpp also uses pygments, but it seems like it is not maintained anymore?
What would you recommend?

Also, @asottile would the Sphinx syntax renderer you did need to be adjusted, or does it use actual output?

blueyed:
pdbpp uses quite a hack for example to set the bg for the current line: https://github.com/antocuni/pdb/blob/1d5a1f1f881baf5fe9126ff146dfca0f29a133ee/pdb.py#L108-L129

asottile:
> pdbpp also uses pygments, but it seems like it is not maintained anymore?
> What would you recommend?
> 
> Also, @asottile would the Sphinx syntax renderer you did need to be adjusted, or does it use actual output?

Yeah this will break that

blueyed:
I see, found it: https://github.com/asottile/pygments-pytest/blob/master/pygments_pytest.py#L20

nicoddemus:
On my terminal currently:

image

With this PR:

image

With @RonnyPfannschmidt's suggestion of making the filenames bold:

image

I like the one from this PR, and like @RonnyPfannschmidt's suggestion a little better: it highlights the error portion nicely, while providing an aid to distinguish line numbers from code lines in the rest of the traceback. 

So I'm 👍 on this PR, and +1.1 on keeping filenames as bold as suggested by @RonnyPfannschmidt. 😁

blueyed:
> why is just the last frame red? that's going to be really hard to implement in pygments and seems arbitrary :(

It's the crashing one (might not be the last necessarily in theory)

Wouldn't pytest-pygments work better by using the actual output from the test run?  IIRC regendoc runs pytest in place, and that could be used then in the docs.
(I see that pytest-pygments might also be used in general outside of pytest though)

Anyway, I don't think we should put limits on improving the output due to this for now.

> keeping filenames as bold as suggested by @RonnyPfannschmidt 

I think this was more about using pygments, but the suggestion also included line numbers - have you left those out on purpose?

I think bold should be used less, and I'd rather would highlight the part after the line number then ("in …"), because it is shorter, and usually it is more important to see the function names in the traceback.
So when parsing bold things visually, there is less to parse.

nicoddemus:
> I think this was more about using pygments, but the suggestion also included line numbers - have you left those out on purpose?

Yes, just did a quick change to see what it would look like.

> I think bold should be used less, and I'd rather would highlight the part after the line number then ("in …"), because it is shorter, and usually it is more important to see the function names in the traceback.

I see, your suggestion makes sense. Feel free to take a stab at how it would look like, or just skip it if it is too much a bother. Let's see if @RonnyPfannschmidt would like to weight in again too.

asottile:
> Wouldn't pytest-pygments work better by using the actual output from the test run? 

I think you're missing the point of pygments.  pygments takes plain text and applies highlighting to it

blueyed:
@asottile 
Yeah, what I've meant is that we should use something that would translate the existing terminal escape codes into colored output.  Maybe https://bitbucket.org/birkenfeld/sphinx-contrib/src/default/autorun/ ?

nicoddemus:
IMHO I like the changes in the output we have decided here, I think they make it easier to see the interesting points in the tracebacks (if everything is highlighted, nothing is highlighted).

If we decide on the changes here based on their merits only, we should definitely discuss how to improve/adapt our docs output highlighting because they certainly do make for more readable output.

> Yeah, what I've meant is that we should use something that would translate the existing terminal escape codes into colored output.

This would be ideal as it is future proof. Adding syntax highlighting to the code is definitely something that users would absolutely love. 😁

asottile:
> > Yeah, what I've meant is that we should use something that would translate the existing terminal escape codes into colored output.
> 
> This would be ideal as it is future proof. Adding syntax highlighting to the code is definitely something that users would absolutely love. grin


no! this is completely the wrong direction!  The point of sphinx is to have **human readable** text in the docs.  If you can't copy paste the plain text output into a doc and have it highlight properly that's broken.  Jamming weird color sigils into the docs is both something that external users can't do and will make the plaintext mangled and unreadable.

I really think that highlighting the "cause frame" is confusing at best.  If it's hard for a computer to parse, imagine what the poor humans will think of it.  I'm fine with the other changes but please make all of the frames highlighted consistently.

nicoddemus:
> no! this is completely the wrong direction! The point of sphinx is to have human readable text in the docs. If you can't copy paste the plain text output into a doc and have it highlight properly that's broken. Jamming weird color sigils into the docs is both something that external users can't do and will make the plaintext mangled and unreadable.

Oh that's not what I mean (and I can see I didn't express myself clearly). I meant that we could use not have output in the docs at all, but it always be generated with `autorun` at "html-build" time.

But indeed I see your point: it is useful to have the output in the `.rst` source as this helps detect problems in the output. Many times we have detected problems by looking at the diff produced by regendoc.

> I really think that highlighting the "cause frame" is confusing at best. If it's hard for a computer to parse, imagine what the poor humans will think of it. I'm fine with the other changes but please make all of the frames highlighted consistently.

I disagree, but how about we compromise. Taken this picture:

img

How about if:

* We only use red bold on the `E` line
* Only use white bold for the code lines (`names = os.listdir(path)`, `onerror(os.listdir, path, sys.exc_info())`, etc).

This would still highlight only the error line and reduce the overall bold red used in the rest of the image, while still making it easy for `pygments-pytest` to parse it (please correct me if I'm wrong @asottile).  

What do you guys think?

asottile:
> What do you guys think?

:+1: sg2m

blueyed:
> no! this is completely the wrong direction! The point of sphinx is to have **human readable** text in the docs. If you can't copy paste the plain text output into a doc and have it highlight properly that's broken. Jamming weird color sigils into the docs is both something that external users can't do and will make the plaintext mangled and unreadable.

What I've meant is that the source is not changed, but only the generated html - similar to how pytest-pygments does it also, no?  Except for that it would not use a lexer, but the output from the pytest run (converting terminal escape sequences to colors, but only for the html output).
Doesn't that make sense?

As for the suggestion: it is better already, but I'd rather only see the "in rmtree" and "in _rmtree_unsafe" being bold there, not the whole paths, which given theirs length act very much like section headers visually.

Also it is not clear to me why pygments-pytest can handle "all bold, last red" better - isn't it still a special case then?

asottile:
> > no! this is completely the wrong direction! The point of sphinx is to have **human readable** text in the docs. If you can't copy paste the plain text output into a doc and have it highlight properly that's broken. Jamming weird color sigils into the docs is both something that external users can't do and will make the plaintext mangled and unreadable.
> 
> What I've meant is that the source is not changed, but only the generated html - similar to how pytest-pygments does it also, no? Except for that it would not use a lexer, but the output from the pytest run (converting terminal escape sequences to colors, but only for the html output).
> Doesn't that make sense?

I don't think I understand.  Are you suggesting to run `pytest` as a side-effect of building docs?


> As for the suggestion: it is better already, but I'd rather only see the "in rmtree" and "in _rmtree_unsafe" being bold there, not the whole paths, which given theirs length act very much like section headers visually.
> 
> Also it is not clear to me why pygments-pytest can handle "all bold, last red" better - isn't it still a special case then?

That's not what @nicoddemus is suggestion, bold-red for only the `E` line(s)

blueyed:
> I don't think I understand. Are you suggesting to run `pytest` as a side-effect of building docs?

Yes, or at least integrate it with, and keep the output/info from when `regendoc` is run.

> That's not what @nicoddemus is suggesting, bold-red for only the `E` line(s)

I see now, I've thought he updated the image also.
Sounds good then.

asottile:
> 
> 
> > I don't think I understand. Are you suggesting to run `pytest` as a side-effect of building docs?
> 
> Yes, or at least integrate it with, and keep the output/info from when `regendoc` is run.
> 

ok I thought that's what you're suggesting, that seems like a bad idea to run pytest as a side-effect of building docs.  as for the second part: we already keep the plaintext output from regendoc, I think you're suggesting keeping the escapes somehow as well?  This would again break the human readable / copy pastable / plain text invariant of sphinx docs

blueyed:
> I think you're suggesting keeping the escapes somehow as well?

Yes.

> This would again break the human readable / copy pastable / plain text invariant of sphinx docs

It would not be in the plaintext docs, but only the generated output from there - i.e. in the same step as pytest-pygments does its transformations, except that it is not based on rules outside of pytest (i.e. its lexer), it would come from running pytest, and using its output.
This is very similar to regendoc, but could work separately (in a later step), but it might be feasible to extend/use regendoc for this (except for that it would generate necessary markup for Sphinx' build).

asottile:
Yeah, like I said, you're missing the point of pygments :P

blueyed:
Well, I _think_ I get it, but the issue here is that we're having two places where style/markup is defined, and that there should be only one at best.

asottile:
> Well, I _think_ I get it, but the issue here is that we're having two places where style/markup is defined, and that there should be only one at best.

there's tests to keep them in sync in pygments-pytest -- I guess I just disagree with this attempt to over-DRY.  The syntax highlighting for cpython isn't in the cpython repository for instance.

blueyed:
Like this?

2019-05-09-220030_744x751_escrotum

Will leave the "startswith('src/')" hack out for now, but it shows that we can later highlight relevant things easier.

So if I understand it correctly the "is this a crash related entry/location" property is not used anymore - which has the benefit of not needing the hacks to get it there in the first place.

blueyed:
@asottile @nicoddemus 
Ping.

asottile:
Can you update the screenshots with up-to-date outputs?

blueyed:
The last comment has an updated screenshot.

asottile:
> Will leave the "startswith('src/')" hack out for now, but it shows that we can later highlight relevant things easier.

but it's highlighted in the screenshot (?)

blueyed:
..I think you can imagine that not being highlighted? :)

asottile:
ignoring that highlight it looks fine :)  I'm still interested in seeing what a multi-traceback looks like with this (and one with locals)

nicoddemus:
image

image

I like it. 👍

blueyed:
Rebased (from faedc85, removing is-crash related things).
@thisch please consider reviewing/trying it :)

thisch:
Thx for improving the formatting of the traceback @blueyed! I like it.

nicoddemus:
Between https://github.com/pytest-dev/pytest/pull/6029 and this, I'm like this one better. 

So if it is up to me, we can add a CHANGELOG, test, and merge it. 👍

blueyed:
> Between #6029 and this, I'm like this one better.

They are not exclusive.

nicoddemus:
Indeed, I meant just that I'm +0 on #6029 but 👍 on this.

blueyed:
I do not feel like finishing this here (test, coverage) for now, and it conflicts with #6029.
Might come back to it later.

PR #5068, opened 2019-04-07
Add support for reportchars=A (`-rA`)

blueyed:
TODO:

- [x] changelog, docs

### DIFF
--- a/changelog/5068.feature.rst
+++ b/changelog/5068.feature.rst
@@ -1,0 +1,1 @@
+The ``-r`` option learnt about ``A`` to display all reports (including passed ones) in the short test summary.
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -83,7 +83,7 @@ def pytest_addoption(parser):
         metavar="chars",
         help="show extra test summary info as specified by chars: (f)ailed, "
         "(E)rror, (s)kipped, (x)failed, (X)passed, "
-        "(p)assed, (P)assed with output, (a)ll except passed (p/P). "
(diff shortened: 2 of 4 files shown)


blueyed:
Added doc/changelog.

src/_pytest/terminal.py
```diff
@@ -166,10 +166,13 @@ def getreportopt(config):
...
     if reportchars:
         for char in reportchars:
-            if char not in reportopts and char != "a":
-                reportopts += char
-            elif char == "a":
+            if char == "a":
                 reportopts = "sxXwEf"
+            elif char == "A":
+                reportopts = "sxXwEfpP"
+                break
```
nicoddemus: Should we `break` on `if char == "a":` above?

changelog/5068.feature.rst
```diff
@@ -0,0 +1 @@
+The ``-r`` option learnt about ``A`` to display all reports (including passed ones) in the short test summary.
```
nicoddemus: Git-like, I like it. 😁

src/_pytest/terminal.py
blueyed: That would not work for `aA` then.. ;)

changelog/5068.feature.rst
blueyed: Heh, noticed this myself.. :)

src/_pytest/terminal.py
nicoddemus: Ahh right, thanks! 👍

PR #5071, opened 2019-04-07
Add a conftest to prefer faster tests

blueyed:
This uses pytest_collection_modifyitems for pytest's own tests to order
them, preferring faster ones via quick'n'dirty heuristics only for now.


blueyed:
Could become its own plugin maybe, not enabled by default, but with an option to do so.

This would be useful for test suites in general (using the "slow" mark by convention).

blueyed:
Thanks for the review, let's rather have it in features first.. :)

PR #5074, opened 2019-04-08
pdb: add test for --trace with --pdbcls

blueyed:


src/_pytest/debugging.py
```diff
@@ -256,7 +256,7 @@ def _test_pytest_function(pyfuncitem):
     _pdb = pytestPDB._init_pdb()
     testfunction = pyfuncitem.obj
     pyfuncitem.obj = _pdb.runcall
-    if "func" in pyfuncitem._fixtureinfo.argnames:  # noqa
+    if "func" in pyfuncitem._fixtureinfo.argnames:
```
blueyed: @nicoddemus 
This was from you.. am I missing something?

blueyed: Changed to "# pragma …" - when quickly checking before I've thought the whole function wasn't covered.

Was "# noqa" actually picked up?

blueyed:
> It's not really clear from the previous commit that this is fixed.

I think the fix is that it previously called the original `Pdb.set_trace`, which we added our `set_trace` (on the class) there, which handles ``--pdbcls`.

> Given that an issue was opened up for this change, would it make sense to add a bugfix changelog entry for this?

Maybe, but since it was fixed before already, this would be retrospective, and it is in the git log after all.
So I'd rather skip it here.

PR #5082, opened 2019-04-10
pytester: improve/fix kwargs validation

blueyed:


blueyed:
Coverage diff looks OK.
Changes might be something flaky?  https://codecov.io/gh/pytest-dev/pytest/compare/10fa66e5b5a5ba25ddc8c1633f3956ba4a28a8d6...148f2fc72cae5eb4af2e1ecf0ccd95aebec03b33/changes

src/_pytest/pytester.py
```diff
@@ -1213,7 +1220,7 @@ def runpytest_subprocess(self, *args, **kwargs):
         if plugins:
             args = ("-p", plugins[0]) + args
         args = self._getpytestargs() + args
-        return self.run(*args, timeout=kwargs.get("timeout"))
+        return self.run(*args, timeout=timeout)
```
blueyed: Note that this test only works accidentally, and is not testing the original issue anymore (see https://github.com/pytest-dev/pytest/issues/5084), but that should not hold this PR back.

PR #5083, opened 2019-04-10
capture: store _suspended state

blueyed:
TODO:

- [x] squash

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -456,6 +456,7 @@ CaptureResult = collections.namedtuple("CaptureResult", ["out", "err"])
 
 class MultiCapture(object):
     out = err = in_ = None
+    _state = None
 
     def __init__(self, out=True, err=True, in_=True, Capture=None):
         if in_:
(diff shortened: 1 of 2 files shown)


blueyed:
This is useful for debugging, and is a starting point for making assertions (no nested suspend) in the future.

src/_pytest/capture.py
```diff
@@ -495,6 +504,7 @@ def suspend_capturing(self, in_=False):
             self._in_suspended = True
 
     def resume_capturing(self):
+        self._suspended = False
```
nicoddemus: I noticed `stop_capturing` doesn't set `self._suspended = True` and I can see why, it is not really suspended after all, it has stopped. But that might mean we don't want a boolean, but a "state" string with `"capturing"`, `"suspended"` and `"stopped"`. What do you think?

blueyed: Makes sense.  Could also be set to `None` there though.

nicoddemus: Yep, but in that case `"stopped"` might be better for debugging purposes, and we can even start it with a proper string `"not initialized"` or something.

blueyed: Yeah.. note however that MultiCapture has `_reset` also already.
I suggest adding `_state` (string), and removing `_reset` then.

nicoddemus: Sounds good. 👍

blueyed: Done.

PR #5091, opened 2019-04-11
Improve output of ini options in --help

blueyed:
Do not cut long help texts, but wrap them the same way as argparse wraps
the other help items.

TODO:

- [x] changelog

### DIFF
--- a/changelog/5091.feature.rst
+++ b/changelog/5091.feature.rst
@@ -1,0 +1,1 @@
+The output for ini options in ``--help`` has been improved.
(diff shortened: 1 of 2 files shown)


nicoddemus:
Previous:

image

New:

image

Nice. 😁

blueyed:
No test.  Ok?

Noticed that `rsyncdirs …` should be on a single line (from your screenshot), fixed.

nicoddemus:
Without even a unittest for this though we are vulnerable to regressions though...

blueyed:
> Without even a unittest for this though we are vulnerable to regressions though...

Yeah.. I am not happy about it myself really - but it is also frustrating to have so many things not being tested in the first place..  might add a test later (next days).

PR #5096, opened 2019-04-12
 blacken-docs more code samples in docs 

asottile:
noticed a few ugly ones so I marked them as python and let blacken-docs do its thing


nicoddemus:
Thanks!

PR #5098, opened 2019-04-12
monkeypatch.syspath_prepend: invalidate import cache

blueyed:
This was done with testdir only, and uses the fixed monkeypatch method
there now.

Fixes https://github.com/pallets/flask/issues/3151.

### DIFF
--- a/src/_pytest/monkeypatch.py
+++ b/src/_pytest/monkeypatch.py
@@ -271,6 +271,18 @@ class MonkeyPatch(object):
         # https://github.com/pypa/setuptools/blob/d8b901bc/docs/pkg_resources.txt#L162-L171
         fixup_namespace_packages(str(path))
 
+        # A call to syspathinsert() usually means that the caller wants to
+        # import some dynamically created files, thus with python3 we
+        # invalidate its import caches.
+        # This is especially important when any namespace package is in used,
+        # since then the mtime based FileFinder cache (that gets created in
+        # this case already) gets not invalidated when writing the new files
+        # quickly afterwards.
+        if sys.version_info >= (3, 3):
+            from importlib import invalidate_caches
+
+            invalidate_caches()
+
     def chdir(self, path):
         """ Change the current working directory to the specified path.
         Path can be a string or a py.path.local object.
--- a/changelog/5098.bugfix.rst
+++ b/changelog/5098.bugfix.rst
@@ -1,0 +1,1 @@
+Invalidate import caches with ``monkeypatch.syspath_prepend``, which is required with namespace packages being used.
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -616,27 +616,10 @@ class Testdir(object):
         This is undone automatically when this object dies at the end of each
         test.
         """
-        from pkg_resources import fixup_namespace_packages
-
         if path is None:
             path = self.tmpdir
 
-        dirname = str(path)
-        sys.path.insert(0, dirname)
-        fixup_namespace_packages(dirname)
-
-        # a call to syspathinsert() usually means that the caller wants to
-        # import some dynamically created files, thus with python3 we
-        # invalidate its import caches
-        self._possibly_invalidate_import_caches()
-
-    def _possibly_invalidate_import_caches(self):
-        # invalidate caches if we can (py33 and above)
-        try:
-            from importlib import invalidate_caches
-        except ImportError:
-            return
-        invalidate_caches()
+        self.monkeypatch.syspath_prepend(str(path))
 
     def mkdir(self, name):
         """Create a new (sub)directory."""
--- a/testing/test_monkeypatch.py
+++ b/testing/test_monkeypatch.py
@@ -462,3 +462,10 @@ def test_syspath_prepend_with_namespace_packages(testdir, monkeypatch):
     import ns_pkg.world
 
     assert ns_pkg.world.check() == "world"
+
+    # Should invalidate caches via importlib.invalidate_caches.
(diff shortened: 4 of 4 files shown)


blueyed:
Tried to come up with a test, which would basically have to trigger an invalid cache I guess, but could not..

For reference (trying to mimic Flask's tests):
```python
def test_syspath_prepend_invalidates_cache(testdir):
    p1 = testdir.makepyfile(
        """
        import pytest
        from _pytest import monkeypatch

        @pytest.fixture(scope='session', autouse=True)
        def _standard_os_environ():
            mp = monkeypatch.MonkeyPatch()
            yield
            mp.undo()

        @pytest.fixture
        def purge_module(request):
            def inner(name):
                request.addfinalizer(lambda: sys.modules.pop(name, None))

            return inner

        @pytest.fixture
        def modules_tmpdir(tmpdir, monkeypatch):
            rv = tmpdir.mkdir('modules_tmpdir')
            print("tmpdir", rv)
            monkeypatch.syspath_prepend(str(rv))
            return rv

        def test1(modules_tmpdir):
            import sys
            import main_app
            sys.modules.pop("main_app")
            # with pytest.raises(ImportError):
            #     import main_app

        def test2(modules_tmpdir):
            import os
            print(os.getcwd())
            app = modules_tmpdir.join('main_app.py')
            app.write('print("main_app_imported"); app = True')

            from main_app import app
        """,
        main_app="""
        """
    )
    result = testdir.runpytest("-s", str(p1))
    assert result.ret == 0
    result.stdout.fnmatch_lines([
        "*main_app_imported",
```

blueyed:
Got a test.. \o/
(it requires namespace packages to be in use already)

src/_pytest/monkeypatch.py
```diff
@@ -271,6 +271,19 @@ def syspath_prepend(self, path):
         # https://github.com/pypa/setuptools/blob/d8b901bc/docs/pkg_resources.txt#L162-L171
         fixup_namespace_packages(str(path))
 
+        # A call to syspathinsert() usually means that the caller wants to
+        # import some dynamically created files, thus with python3 we
+        # invalidate its import caches.
+        # This is especially important when any namespace package is in used,
+        # since then the mtime based FileFinder cache (that gets created in
+        # this case already) gets not invalidated when writing the new files
+        # quickly afterwards.
```
nicoddemus:
```suggestion
        # quickly afterwards. See pallets/flask#3151 for reference.
```

I think it is worth adding a reference to the issue; although one can use "blame" to see the commit originally, it is very easy for this information to be lost by a rename or an auto-formatter (not really lost, but much harder to find). Adding a comment is a pretty painless way to avoid that IMHO. 👍

src/_pytest/monkeypatch.py
```diff
@@ -271,6 +271,19 @@ def syspath_prepend(self, path):
         # https://github.com/pypa/setuptools/blob/d8b901bc/docs/pkg_resources.txt#L162-L171
         fixup_namespace_packages(str(path))
 
+        # A call to syspathinsert() usually means that the caller wants to
```
nicoddemus: Great explanation!

src/_pytest/monkeypatch.py
blueyed: It can also be found via the changelog (and this PR).. isn't that one of the reasons to have them? :)
I'd not add it, but feel free to do so, and merge it then.

src/_pytest/monkeypatch.py
```diff
@@ -271,6 +271,19 @@ def syspath_prepend(self, path):
...
+        # invalidate its import caches.
+        # This is especially important when any namespace package is in used,
+        # since then the mtime based FileFinder cache (that gets created in
+        # this case already) gets not invalidated when writing the new files
+        # quickly afterwards.
+        try:
+            from importlib import invalidate_caches
+        except ImportError:
+            return
+        invalidate_caches()
```
asottile: to make this more future-proof (in case we add code after this) I'd do this:

```
if sys.version_info >= (...):
    from importlib import invalidate_caches
    invalidate_caches()
```

this avoids general frustration that can be caused by backport modules which aren't intended to be run in this way

it also gives us good documentation on when we can remove the version-specific hax

blueyed: Good point.
I've just moved this though basically, and would not know the version to use without checking.

asottile: fortunately the python docs are very helpful here

https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches

> New in version 3.3.

blueyed: Yeah, let's add it in features after this is merged - no functional changes on master.

blueyed:
> please make the change I've suggested before merging -- it is not a functional change

Well, like you said, with that change backported modules might behave different then, no?

asottile:
> > please make the change I've suggested before merging -- it is not a functional change
> 
> Well, like you said, with that change backported modules might behave different then, no?

yeah, your current code is much more likely to crash

blueyed:
> yeah, your current code is much more likely to crash

This is moved only.

asottile:
> > yeah, your current code is much more likely to crash
> 
> This is moved only.

All the more reason to improve it at the same time :)

blueyed:
But not necessarily for a bugfix.
(I've thought using the existing monkeypatch in the first place should be avoided, but skipped that then)

nicoddemus:
IMHO: this code is fine to go into `master` as it is a bug fix, and the change requested by @asottile is reasonable as well. 😁

PR #5109, opened 2019-04-14
summary_passes: use bold green for report headers

blueyed:


asottile:
code seems fine, want to attach screenshots showing a before and after so it's easy to see from the code review what's changing?

nicoddemus:
Before:

image

After:

image

PR #5117, opened 2019-04-14
Minor coverage improvements for terminal plugin

blueyed:


src/_pytest/terminal.py
```diff
@@ -852,7 +851,8 @@ def summary_errors(self):
                     msg = "ERROR collecting " + msg
                 elif rep.when == "setup":
                     msg = "ERROR at setup of " + msg
-                elif rep.when == "teardown":
```
asottile: is it possible a plugin adds additional `.when`s or is this an internal field?

blueyed: Wondered the same.

When checking again I found https://github.com/blueyed/pytest/blob/9374114370ad509169745548ffcc1176d430daf4/testing/test_terminal.py#L1549, so changed it.

PR #5119, opened 2019-04-14
reportopts: A: put "Pp" in front

blueyed:
No changelog, followup to #5068.

Have not really checked if this is sane yet, but "P" means to provide the output of the passed tests, and "p" to include it in the summary - where it should come first also then.


nicoddemus:
Ping @blueyed to solve the conflicts

PR #5126, opened 2019-04-15
terminal/reports: use get_verbose_word method

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5047#issuecomment-483109956

Should become more flexible with regard to other properties also, this is just a request for comments.

### DIFF
--- a/src/_pytest/reports.py
+++ b/src/_pytest/reports.py
@@ -148,6 +148,12 @@ class BaseReport(object):
             fspath, lineno, domain = self.location
             return domain
 
+    def _get_verbose_word(self, config):
+        _category, _short, verbose = config.hook.pytest_report_teststatus(
+            report=self, config=config
+        )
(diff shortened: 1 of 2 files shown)


src/_pytest/reports.py
```diff
@@ -148,6 +148,12 @@ def head_line(self):
             fspath, lineno, domain = self.location
             return domain
 
+    def get_verbose_word(self, config):
```
nicoddemus: We probably better make this private

blueyed: Makes sense for now, but the idea with my commnet on #5047 is to make this a public API in the end.

But we can interate towards this internally, which makes sesnse.

So I'll change this to `_get_verbose_word` for now then?

nicoddemus: Sounds good

PR #5138, opened 2019-04-17
Fix dependencies on 'terminal' plugin

### LINKED ISSUE #5139: notify_exception crashes with -p no:terminal
e.g. run with `pytest -p no:terminal test.py` where test.py is
```python
raise Exception
```

ikonst:
Fix #5139

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -105,6 +105,7 @@ Hugo van Kemenade
 Hui Wang (coldnight)
 Ian Bicking
 Ian Lesperance
+Ilya Konstantinov
 Ionuț Turturică
 Iwan Briquemont
 Jaap Broekhuizen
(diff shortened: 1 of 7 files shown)


blueyed:
Thanks for working on this!

I wonder what @RonnyPfannschmidt has to say about the terminal plugin options being coupled like that - shouldn't only the terminal part/plugin itself know about it in the best case?

ikonst:
I agree this coupling is not optimal. I only corrected the code with the assumption that this coupling is a necessary evil for now.

blueyed:
@ikonst 
That good, yeah - just wondered myself how to improve this in the future.

blueyed:
Since this is green on CI now, please squash-merge it (it does not need 7 commits, and we have the squash-merge button disabled here).

blueyed:
Thank you!

PR #5149, opened 2019-04-20
[WIP/RFC] pdb: set sys.last_{type,value,traceback} in _enter_pdb

blueyed:
This affects pytest_exception_interact (`--pdb`) currently only, but
makes sense for all calls to `post_mortem` probably, or even more in
general.

Ref: https://github.com/python/cpython/blame/9f316bd9684d27b7e21fbf43ca86dc5e65dac4af/Doc/library/sys.rst#L891-L904

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -289,6 +289,7 @@ def _enter_pdb(node, excinfo, rep):
     tw.sep(">", "entering PDB")
     tb = _postmortem_traceback(excinfo)
     rep._pdbshown = True
+    sys.last_type, sys.last_value, sys.last_traceback = excinfo._excinfo
     post_mortem(tb)
     return rep
 
--- a/testing/test_pdb.py
+++ b/testing/test_pdb.py
@@ -89,6 +89,7 @@ class TestPDB(object):
         return pdblist
 
     def test_pdb_on_fail(self, testdir, pdblist):
+        sys.last_type, sys.last_value, sys.last_traceback = (None, None, None)
         rep = runpdb_and_get_report(
             testdir,
             """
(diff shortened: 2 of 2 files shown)


src/_pytest/debugging.py
```diff
@@ -289,6 +289,7 @@ def _enter_pdb(node, excinfo, rep):
     tw.sep(">", "entering PDB")
     tb = _postmortem_traceback(excinfo)
     rep._pdbshown = True
+    sys.last_type, sys.last_value, sys.last_traceback = excinfo._excinfo
     post_mortem(tb)
     return rep
```
blueyed: Initially I wanted this to use `Pdb.user_exception` instead, but that would require some adjustments of (mocking) tests.
`user_exception` would set `__exception__` into locals (but only type and value currently (https://bugs.python.org/issue36663)).

blueyed:
From the description it sounds like this should be set in pytest core tiself maybe even:

>    These three variables are not always defined; they are set when an exception is
   not handled and the interpreter prints an error message and a stack traceback.


(via https://github.com/python/cpython/blame/9f316bd9684d27b7e21fbf43ca86dc5e65dac4af/Doc/library/sys.rst#L891-L904)

blueyed:
I still think it should be more generic, and not via the debugging plugin probably, i.e. everytime Python "creates" an exception (failing test, error) it should make this info available for inspection.
But maybe having it via the debugging plugin is the easiest to achieve this - but it should certainly be in a more central place there then.

blueyed:
Found it being done here already: https://github.com/blueyed/pytest/blob/73b74c74c9674947c8383f657c3a78b1d73ffe1b/src/_pytest/runner.py#L118-L131

blueyed:
Reworked in https://github.com/blueyed/pytest/pull/16, in case you are interested.

PR #5153, opened 2019-04-22
docs: remove extra characters in the command line prompt

hroncok:

just a guideline):

### DIFF
--- a/changelog/5153.trivial.rst
+++ b/changelog/5153.trivial.rst
@@ -1,0 +1,2 @@
+Remove extra characters in the shown command line prompts in the test discovery
+documentation.
--- a/doc/en/example/pythoncollection.rst
+++ b/doc/en/example/pythoncollection.rst
@@ -206,7 +206,7 @@ You can always peek at the collection tree without running tests like this:
 
 .. code-block:: pytest
 
-    . $ pytest --collect-only pythoncollection.py
+    $ pytest --collect-only pythoncollection.py
     =========================== test session starts ============================
     platform linux -- Python 3.x.y, pytest-4.x.y, py-1.x.y, pluggy-0.x.y
     cachedir: $PYTHON_PREFIX/.pytest_cache
@@ -266,7 +266,7 @@ leave out the ``setup.py`` file:
 
 .. code-block:: pytest
 
-    #$ pytest --collect-only
+    $ pytest --collect-only
     ====== test session starts ======
     platform linux2 -- Python 2.7.10, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
     rootdir: $REGENDOC_TMPDIR, inifile: pytest.ini


hroncok:
I've used the GitHub web Interface to create this PR. Please squash when merging.

Note that I *think*those characters there are redundant, but I'm not sure if the bare no special meaning.

blueyed:
> Please squash when merging.

Not possible (https://github.com/pytest-dev/pytest/issues/4361).

I think those examples should be auto-generated, see `tox -e regen`.

changelog/5153.trivial.rst
```diff
@@ -0,0 +1,2 @@
+Remove extra characters in the shown command line prompts in the test discovery
+documentation.
```
blueyed: btw: I do not think we need a changelog for this.

blueyed:
> I think those examples should be auto-generated, see `tox -e regen`.

Probably means that they are currently not updated then, and that this PR is good to do so in the future.

doc/en/example/pythoncollection.rst
```diff
@@ -206,7 +206,7 @@ You can always peek at the collection tree without running tests like this:
 
 .. code-block:: pytest
 
-    . $ pytest --collect-only pythoncollection.py
```
nicoddemus: They are there to *not* run regendoc, because regendoc uses the line starting as `$` to mean to execute the shell command:

https://github.com/pytest-dev/regendoc/blob/27e3652723659d8b2d7a5311567480fccc49cbe2/regendoc/parse.py#L82-L84

changelog/5153.trivial.rst
hroncok: me neither. creating this was pain via github.

doc/en/example/pythoncollection.rst
hroncok: So this PR is probably bogus. Should I open an issue to have a mechanism for this that doesn't uglify the docs?

nicoddemus:
> Probably means that they are currently not updated then, and that this PR is good to do so in the future.

Yep, missed your comment before commenting earlier.

Thanks a lot for the quick and thoughtful contribution @hroncok, but I guess we can close this. 👍

changelog/5153.trivial.rst
hroncok: https://github.com/pytest-dev/pytest/blob/master/changelog/README.rst

> `trivial`: fixing a small typo or internal change that might be noteworthy.

Note sure if the "might be noteworthy" part applies to fixing typos, but I think this list should be updated to say not to create this just for a trivial typo fix.

doc/en/example/pythoncollection.rst
nicoddemus: Good idea, I would kindly ask to create it in https://github.com/pytest-dev/regendoc though. 👍

changelog/5153.trivial.rst
nicoddemus: Oh excellent point, thanks for pointing that out! We should update that templates to mean that trivial is meant to actual code or dependency changes, but not to small documentation changes in general. 👍

doc/en/example/pythoncollection.rst
hroncok: https://github.com/pytest-dev/regendoc/issues/9

hroncok: Note that studying the code, a dot with a space should not stop regendoc:

https://github.com/pytest-dev/regendoc/blob/27e3652723659d8b2d7a5311567480fccc49cbe2/regendoc/parse.py#L85-L88

blueyed: > They are there to _not_ run regendoc

Why though?  Shouldn't we have working examples for those also?

nicoddemus: You are right, probably they should

PR #5155, opened 2019-04-22
Adding test cases for unrolling an iterable #5062

danielx123:
Adds tests that will check if the pytest terminal output from using assertion statements with "all" or "any" will unroll and produce the same output as assertion statements inside of a for-loop. These tests are related to issue #5062 and two of them currently fail. Once the implementation in #5103 is finished, all three of these tests should pass.

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -9,6 +9,7 @@ from __future__ import print_function
 import argparse
 import collections
 import platform
+import subprocess
 import sys
 import time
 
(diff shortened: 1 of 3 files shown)


src/_pytest/terminal.py
```diff
@@ -574,6 +575,11 @@ def pytest_sessionstart(self, session):
         ):
             msg += " -- " + str(sys.executable)
         self.write_line(msg)
+        pipe = subprocess.Popen("pip list", shell=True, stdout=subprocess.PIPE).stdout
+        package_msg = pipe.read()
+        package_msg = package_msg[:-2]
+        if package_msg:
+            self.write_line(package_msg)
```
blueyed: Is this meant for debugging / this PR only?

PR #5166, opened 2019-04-25
Remove special-case for PyPy CI

vtbassmatt:
Hi there, PM from Azure Pipelines here. We recently added `pypy2` and `pypy3` as versions for the `UsePythonVersion` task. That means you no longer have to special-case PyPy versions in your build matrix.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,6 @@ trigger:
 variables:
   PYTEST_ADDOPTS: "--junitxml=build/test-results/$(tox.env).xml -vv"
   python.needs_vc: False
-  python.exe: "python"
   COVERAGE_FILE: "$(Build.Repository.LocalPath)/.coverage"
   COVERAGE_PROCESS_START: "$(Build.Repository.LocalPath)/.coveragerc"
   PYTEST_COVERAGE: '0'
@@ -43,14 +42,12 @@ jobs:
         #   But no exception with py27-pexpect,py27-twisted,py27-numpy.
         PYTEST_COVERAGE: '1'
       pypy:
-        python.version: 'pypy'
+        python.version: 'pypy2'
         tox.env: 'pypy'
-        python.exe: 'pypy'
       # NOTE: pypy3 fails to install pip currently due to an interal error.
       # pypy3:
       #   python.version: 'pypy3'
       #   tox.env: 'pypy3'
-      #   python.exe: 'pypy3'
       py34-xdist:
         python.version: '3.4'
         tox.env: 'py34-xdist'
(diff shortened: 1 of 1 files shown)


nicoddemus:
Thanks a lot once again @vtbassmatt!

Unfortunately the Azure Pipelines build did not start. Do you have an idea of what might it be?

vtbassmatt:
Not sure - that's quite unexpected. You aren't seeing this on any other PRs, are you?

nicoddemus:
This one I opened 2 hours ago didn't trigger a build either. 🤔

vtbassmatt:
That's really weird. I see that the last PR build was on 4/23. I'm assuming nothing changed on your end (let me know if that's a wrong assumption :smile:) and will have someone take a look.

vtbassmatt:
We're digging into what happened here. I may have a request for you in a little while, but we're gathering a bit of data first.

vtbassmatt:
OK, we have a couple of suspected culprits. We're rolling out fixes for them, after which I can try pushing another commit here to see if it fires the build.

I also noticed that you're using an OAuth connection rather than the [GitHub app](https://github.com/marketplace/azure-pipelines). If it's possible to [switch over to the app](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops#integrate-using-the-github-app), we sometimes get more traceability than with plain OAuth tokens. As a bonus, the app works with GitHub Checks so you'd get your Azure Pipelines on the Checks tab.

nicoddemus:
Hi @vtbassmatt,

Thanks for the follow up! It seems to be working now on `master`. 👍 

Could you please rebase on `master` and push-force? Thanks

---

> I also noticed that you're using an OAuth connection rather than the GitHub app

Strange, it seems we are already connected through the app:

image

image

image

Do you see anything amiss? Thanks!

vtbassmatt:
I'll rebase, thanks!

On the other stuff - @davidstaheli do you know why they'd have the app installed but we'd be getting their events as if they were web hooks?

nicoddemus:
@vtbassmatt thanks for pointing this out:

image

Probably those are there because we configured Azure at some point using OAuth and never realize we should also remove the hooks.

Should I remove them then?

vtbassmatt:
Probably, but I don't want your PR builds to get interrupted again. Let's see if David has any guidance.

vtbassmatt:
Rebased. Hmm... it's not liking something in the tox config for PyPy. I'd fix it, but I don't quite understand what the error is telling me,

nicoddemus:
I had to keep using `pypy` as an executable to make `tox` happy (#5209).

Thanks again @vtbassmatt for the PR, we appreciate it! 😁

Issue #5167, opened 2019-04-25
pip 19.1 prevents pytest from being used in editable mode


nicoddemus:
pip 19.1 prevents pytest to use editable mode (https://github.com/pypa/pip/pull/6370) due to our `pyproject.toml` file.

This is a problem for us for two reasons:

1. `tox.ini` uses `usedevelop` in a `release` and `docs` environments. I think we can get away from using it in `docs`, not so sure about release.

2. We use editable mode ourselves.

I will open a PR shortly trying to address 1, but for 2 I don't know of any workaround other than downgrading pip manually.

cc @RonnyPfannschmidt

gaborbernat:
https://github.com/pypa/pip/issues/6434 issue report and discussion, things should be better once pip 19.1.1 gets out (hopefully soon)

bskinn:
For (2), try removing `-e .` (or `pip install -e .`, or whatever mechanism) for the pip-mediated editable install, and instead running `python setup.py develop` to achieve an editable install via `setuptools.

This has worked fine so far for both of my PEP517 projects.

PR #5168, opened 2019-04-25
Workaround for editable installs with pip 19.1

nicoddemus:
Related to #5167


nicoddemus:
If Travis is happy, I think we can get this merged, proves the `docs` environment is working at least.

PR #5178, opened 2019-04-27
Further "unknown marks warning" improvements

### LINKED ISSUE #5023: Further "unknown marks warning" improvements
As discussed in https://github.com/pytest-dev/pytest/pull/4935 and https://github.com/pytest-dev/pytest/pull/4935#discussion_r271099528:

* Add `--strict-marks` as an alias to `--strict`, and use `--strict-marks` in the documentation and examples;
* Remove the various marks that are used through testing: `foo`, `bar`, multiple spellings of parametrize, etc.
* After the above, we should try to get rid of `UnknownMarksWarning` in favor of `PytestWarning`; if not, we should then create more subclasses of `PytestWarning` for the other warnings emitted by pytest.
* Add integration tests.

nicoddemus:
Fix #5023

### DIFF
--- a/changelog/5023.feature.rst
+++ b/changelog/5023.feature.rst
@@ -1,1 +1,5 @@
-``--strict`` is now called ``--strict-markers`` as it is more explicit about what it does. The old name still works for backward compatibility.
+New flag ``--strict-markers`` that triggers an error when unknown markers (e.g. those not registered using the `markers option`_ in the configuration file) are used in the test suite.
+
+The existing ``--strict`` option has the same behavior currently, but can be augmented in the future for additional checks.
+
+.. _`markers option`: https://docs.pytest.org/en/latest/reference.html#confval-markers
--- a/changelog/5023.feature.rst
+++ b/changelog/5023.feature.rst
@@ -1,0 +1,1 @@
+``--strict`` is now called ``--strict-markers`` as it is more explicit about what it does. The old name still works for backward compatibility.
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1261,18 +1261,23 @@ passed multiple times. The expected format is ``name=value``. For example::
 
 .. confval:: markers
 
-    When the ``--strict`` command-line argument is used, only known markers -
+    When the ``--strict-markers`` command-line argument is used, only known markers -
     defined in code by core pytest or some plugin - are allowed.
-    You can list additional markers in this setting to add them to the whitelist.
 
-    You can list one marker name per line, indented from the option name.
+    You can list additional markers in this setting to add them to the whitelist,
+    in which case you probably want to add ``--strict-markers`` to ``addopts``
+    to avoid future regressions:
 
     .. code-block:: ini
 
         [pytest]
+        addopts = --strict-markers
         markers =
             slow
             serial
+
+    **Note**: This option was previously called ``--strict``, which is now an
+    alias preserved for backward compatibility.
 
 .. confval:: minversion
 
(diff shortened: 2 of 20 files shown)


changelog/5023.feature.rst
```diff
@@ -0,0 +1 @@
+``--strict`` is now called ``--strict-markers`` as it is more explicit about what it does. The old name still works for backward compatibility.
```
blueyed: With regard to `--strict` I would keep it anyway, being able to select also other strict modes later..

blueyed: And therefore it could still be used in default config (when the intention here is to enforce strict checkings in general).

nicoddemus: > With regard to --strict I would keep it anyway, being able to select also other strict modes later..

It has not been removed, I only added `--strict-markers` as the now preferred name, `--strict` is still present and works as before.

blueyed: I know.. it just sounded like it would be only kept for backward compatibility, while I think it makes sense to keep it in general.

doc/en/reference.rst
```diff
@@ -1261,19 +1261,24 @@ passed multiple times. The expected format is ``name=value``. For example::
...
     .. code-block:: ini
 
         [pytest]
+        addopts = --strict-markers
         markers =
             slow
             serial
 
+    **Note**: This option was previously called ``--strict``, which is now an
+    alias preserved for backward compatibility.
```
blueyed: I.e. this could be rephrased.

changelog/5023.feature.rst
blueyed: And `--strict` should be used in some test probably still.

nicoddemus: Indeed, I thought the same thing, please see `test_strict_prohibits_unregistered_markers`

> I know.. it just sounded like it would be only kept for backward compatibility, while I think it makes sense to keep it in general.

Not sure if we should ever change the meaning of `--strict` in the future as well. I mean, suppose we decide to turn "cannot collect classes with __init__" into an error using `--strict`. If we do that, then test suites which used `--strict` to turn marks into errors will now break.

In the future we probably should grow another option for that kind of check and leave this option alone?

nicoddemus: @blueyed any thoughts on this? If you feel strongly about changing the meaning of `--strict` and wording, I don't mind changing it back.

nicoddemus:
@blueyed gentle ping. 😁

changelog/5023.feature.rst
blueyed: > If we do that, then test suites which used `--strict` to turn marks into errors will now break.

I'd argue that this is OK - and easy to fix then.
It is easier to have `--strict` for CI etc, than having to add multiple `--strict-*` variants then.
New strictness checks could still get their own option, of course (if it makes sense).

blueyed:
I'm fine with keeping it like this, but having it clearer would allow problems with deprecations in the future.  I.e. if we say now that `--strict` is an alias, instead of used for strict checking in general we have to adhere to it.

nicoddemus:
> I'm fine with keeping it like this, but having it clearer would allow problems with deprecations in the future. 

Makes sense, I don't have strong feelings either way to be honest. Can you suggest the changes you would like to see then (either here as "suggestions" or a commit yourself)?

nicoddemus:
> I'm fine with keeping it like this, but having it clearer would allow problems with deprecations in the future. I.e. if we say now that --strict is an alias, instead of used for strict checking in general we have to adhere to it.

Went ahead and implemented the suggested changes to the docs, please let me know what you think. 👍

PR #5179, opened 2019-04-28
Warning subclasses

### LINKED ISSUE #5177: Introduce warning subclasses for each type of warning we emit
As we did with `UnknownMarkWarning` in #4935, we should create various subclasses for all the types of warnings we emit: not collecting classes with `__init__`, unknown markers, etc.

nicoddemus:
Fix #5177

### DIFF
--- a/doc/en/warnings.rst
+++ b/doc/en/warnings.rst
@@ -415,10 +415,20 @@ The following warning types ares used by pytest and are part of the public API:
 
 .. autoclass:: pytest.PytestWarning
 
-.. autoclass:: pytest.PytestDeprecationWarning
+.. autoclass:: pytest.PytestAssertRewriteWarning
 
-.. autoclass:: pytest.RemovedInPytest4Warning
+.. autoclass:: pytest.PytestCacheWarning
+
+.. autoclass:: pytest.PytestCollectionWarning
+
+.. autoclass:: pytest.PytestConfigWarning
+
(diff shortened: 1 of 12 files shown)


doc/en/warnings.rst
```diff
@@ -415,8 +415,20 @@ The following warning types ares used by pytest and are part of the public API:
...
+
+.. autoclass:: pytest.PytestDeprecationWarning
 
 .. autoclass:: pytest.PytestExperimentalApiWarning
+
+.. autoclass:: pytest.PytestUnhandledCoroutineWarning
+
+.. autoclass:: pytest.PytestUnknownMarkWarning
+
+.. autoclass:: pytest.RemovedInPytest4Warning
```
blueyed: RemovedInPytest4Warning is not used anymore really, is it?

blueyed: Ah, seen https://github.com/pytest-dev/pytest/issues/5180 now - but wonder why/if it is being kept now?

nicoddemus: I understand your point, but we really shouldn't remove that symbol in a minor release, it might break some configuration or other (it might be being referenced in a warnings filter for example).

It doesn't hurt keeping it for now and removing it on 5.0 to avoid breaking code needlessly.

blueyed: Makes sense, thanks for explaining!

PR #5183, opened 2019-04-28
Review doctest docs

nicoddemus:
* Add pytest.skip() example
* Add blurb about Sybil
* Create a subsection for doctest-options
* Create a subsection for pytest-specific features
* Fix examples

cc @cjw296

### DIFF
--- a/doc/en/doctest.rst
+++ b/doc/en/doctest.rst
@@ -31,10 +31,10 @@ then you can just invoke ``pytest`` directly:
     =========================== test session starts ============================
     platform linux -- Python 3.x.y, pytest-4.x.y, py-1.x.y, pluggy-0.x.y
     cachedir: $PYTHON_PREFIX/.pytest_cache
-    rootdir: $REGENDOC_TMPDIR, inifile: pytest.ini
+    rootdir: $REGENDOC_TMPDIR
     collected 1 item
 
-    mymodule.py .                                                        [100%]
+    test_example.txt .                                                   [100%]
 
     ========================= 1 passed in 0.12 seconds =========================
 
@@ -57,6 +57,16 @@ and functions, including from test modules:
 .. code-block:: bash
 
     $ pytest --doctest-modules
+    =========================== test session starts ============================
+    platform linux -- Python 3.x.y, pytest-4.x.y, py-1.x.y, pluggy-0.x.y
+    cachedir: $PYTHON_PREFIX/.pytest_cache
+    rootdir: $REGENDOC_TMPDIR
+    collected 2 items
+
+    mymodule.py .                                                        [ 50%]
+    test_example.txt .                                                   [100%]
+
+    ========================= 2 passed in 0.12 seconds =========================
 
 You can make these changes permanent in your project by
 putting them into a pytest.ini file like this:
--- a/doc/en/doctest.rst
+++ b/doc/en/doctest.rst
@@ -10,61 +10,20 @@ can change the pattern by issuing:
 
     pytest --doctest-glob='*.rst'
 
-on the command line. Since version ``2.9``, ``--doctest-glob``
-can be given multiple times in the command-line.
-
-.. versionadded:: 3.1
-
-    You can specify the encoding that will be used for those doctest files
-    using the ``doctest_encoding`` ini option:
-
-    .. code-block:: ini
-
-        # content of pytest.ini
-        [pytest]
-        doctest_encoding = latin1
-
-    The default encoding is UTF-8.
-
-You can also trigger running of doctests
-from docstrings in all python modules (including regular
-python test modules):
-
-.. code-block:: bash
-
-    pytest --doctest-modules
-
-You can make these changes permanent in your project by
-putting them into a pytest.ini file like this:
-
-.. code-block:: ini
-
-    # content of pytest.ini
-    [pytest]
-    addopts = --doctest-modules
+on the command line. ``--doctest-glob`` can be given multiple times in the command-line.
 
 If you then have a text file like this:
 
 .. code-block:: text
 
-    # content of example.rst
+    # content of test_example.txt
 
     hello this is a doctest
     >>> x = 3
     >>> x
     3
 
-and another like this::
-
-    # content of mymodule.py
-    def something():
-        """ a doctest in a docstring
-        >>> something()
-        42
-        """
-        return 42
-
-then you can just invoke ``pytest`` without command line options:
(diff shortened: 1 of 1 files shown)


doc/en/doctest.rst
```diff
@@ -64,36 +54,58 @@ and another like this::
...
-when executing text doctest files.
+.. note::
 
-The standard ``doctest`` module provides some setting flags to configure the
-strictness of doctest tests. In pytest, you can enable those flags using the
-configuration file. To make pytest ignore trailing whitespaces and ignore
+    If you are making serious use of doctests through your code, please checkout
+    `Sybil <https://sybil.readthedocs.io/en/latest/index.html>`__, which besides
+    supporting doctests also supports Sphinx directives, plus other features.
+    It provides pytest integration out of the box.
```
cjw296: Might be better to phrase it like this:

```
If you're looking to check that examples in your documentation are correct, including doctests 
`.. codeblock:: python` and any other examples your documentation include, you may wish to
consider `Sybil <https://sybil.readthedocs.io/en/latest/index.html>`__. 
It provides pytest integration out of the box.
```

doc/en/doctest.rst
```diff
@@ -110,16 +122,7 @@ Python 3 unchanged:
 * ``ALLOW_BYTES``: when enabled, the ``b`` prefix is stripped from byte strings
   in expected doctest output.
 
-As with any other option flag, these flags can be enabled in ``pytest.ini`` using
```
cjw296: Where's the code for `ALLOW_UNICODE` and `ALLOW_BYTES`? Sybil has these for its doctest support, but I had to put some big caveats in because parsing `u'...'` and `b'...'` out of expected output when nested quotes are involved turns out to be really very hard.

doc/en/doctest.rst
```diff
@@ -128,20 +131,58 @@ itself:
...
+
+Using fixtures
+^^^^^^^^^^^^^^
+
+It is possible to use fixtures using the ``getfixture`` helper:
+
+.. code-block:: text
+
+    # content of example.rst
+    >>> tmp = getfixture('tmpdir')
```
cjw296: Hah! I did this slightly differently in Sybil, not sure which way I prefer but I'd be interested in seeing the implementation and wiring for `getfixture`!

doc/en/doctest.rst
```diff
@@ -171,20 +212,14 @@ Note that like the normal ``conftest.py``, the fixtures are discovered in the di
...
 
-    pytest --doctest-modules --doctest-report none
-    pytest --doctest-modules --doctest-report udiff
-    pytest --doctest-modules --doctest-report cdiff
-    pytest --doctest-modules --doctest-report ndiff
-    pytest --doctest-modules --doctest-report only_first_failure
+    >>> import sys, pytest
+    >>> if sys.platform.startswith('win'):
+    ...     pytest.skip('this doctest does not work on Windows')
+    ...
```
cjw296: Right, but this is ugly to include in documentation; my approach with Sybil has been to always make the documentation the key and just work around whatever's needed to check the examples it contains are correct (hence [capture](https://sybil.readthedocs.io/en/latest/parsers.html#capture), [skip](https://sybil.readthedocs.io/en/latest/parsers.html#skip) and [invisible-code-block](https://sybil.readthedocs.io/en/latest/parsers.html#codeblock)!).

doc/en/doctest.rst
nicoddemus: Hmm the first sentence ("If you're looking to check that examples in your documentation are correct") gives the impression that the builtin support does not support this, which reads weird (up until this point we have just saying that it does have some support). Would you mind changing that to "advanced" support or something along those lines? 

This is the documentation for pytest's doctest module, so it makes sense to suggest something which is supported here *plus* a bunch of other features.

```
The builtin pytest doctest supports only `doctest` blocks, but if you are looking 
for more advanced checking over all your documentation examples, 
including doctests, `.. codeblock:: python`, and any other examples your documentation may...
```

What do you think? Please feel free to provide an updated blurb. 😁

doc/en/doctest.rst
nicoddemus: You can find it here: https://github.com/pytest-dev/pytest/blob/9c700d1fd5496c2f844f3dd20835ab204a822945/src/_pytest/doctest.py#L452-L504

doc/en/doctest.rst
nicoddemus: Sure, here it is:

https://github.com/pytest-dev/pytest/blob/9c700d1fd5496c2f844f3dd20835ab204a822945/src/_pytest/doctest.py#L189-L197

Basically we inject `request.getfixturevalue` as `getfixture` into the namespace.

doc/en/doctest.rst
nicoddemus: Oh I agree it is not ideal, it certainly can be improved! 👍

nicoddemus:
> It would be great if doctests were opt-in, like a normal plugin, or at the very least, completely opt-out so that I could advise Sybil users to do that.

The plugin can be completely disabled with `-p no:doctest`, but I believe that has been mentioned in that thread already. Did that not work? 🤔

cjw296:
I haven't tried, it was unclear to me what version of pytest `-p no:doctest` arrived in and if that'd been released? I mean, my view is that should be the default...

doc/en/doctest.rst
cjw296: Yes, use Sybil and hide everything that the user doesn't need to see ;-)

doc/en/doctest.rst
cjw296: Cool, if I ever get desperate, I'll borrow those regexes, thanks! (but only a few more months of Py2, right?!)

doc/en/doctest.rst
cjw296: Yep, your new version is great, go with that :-)

nicoddemus:
> I haven't tried, it was unclear to me what version of pytest -p no:doctest arrived in and if that'd been released? I mean, my view is that should be the default...

I actually agree that doctesting should be opt-in, but we can really change that in a major version...

doc/en/doctest.rst
nicoddemus: Yes, after that only critical bug fixes

doc/en/doctest.rst
nicoddemus: Updated. 👍

PR #5184, opened 2019-04-28
Remove all version references to obsolete pytest versions

nicoddemus:
Remove version references to pytest 2 and 3.

Just like Python 3 no longer has references to Python 2, I think
we should follow the same approach in pytest.


nicoddemus:
I did a quick find/replace and Sublime leaves it like that, so I'm gonna keep it for now if you don't mind. 😁

PR #5185, opened 2019-04-28
--lf now skips colletion of files without failed tests

### LINKED ISSUE #5172: last-failed: smarter collection
With `--lf` it should not run a full collection, but use the information from the cache.

nicoddemus:
Fix #5172

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -158,6 +158,33 @@ class LFPlugin(object):
         self.lastfailed = config.cache.get("cache/lastfailed", {})
         self._previously_failed_count = None
         self._report_status = None
+        self._skipped_files = 0  # count skipped files during collection due to --lf
+
+    def last_failed_paths(self):
+        """Returns a set with all Paths()s of the previously failed nodeids (cached).
+        """
+        result = getattr(self, "_last_failed_paths", None)
+        if result is None:
+            rootpath = Path(self.config.rootdir)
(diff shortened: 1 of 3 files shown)


src/_pytest/cacheprovider.py
```diff
@@ -158,6 +158,33 @@ def __init__(self, config):
...
+        Ignore this file path if we are in --lf mode and it is not in the list of
+        previously failed files.
+        """
+        if not self.active or not self.config.getoption("lf") or not path.isfile():
+            return
+        last_failed_paths = self.last_failed_paths()
+        if not last_failed_paths:
+            # if no previous failures, we want to run all items regardless
+            return
+        result = Path(path) not in self.last_failed_paths()
```
blueyed:
```suggestion
        result = Path(path) not in last_failed_paths
```

nicoddemus: Good catch, thanks; done slightly different, I check for `self.lastfailed` on the `if`

blueyed:
Thanks again.. this will help me with working on the pytest plugin / conftest to use pytest with Django's own tests (> 6000 IIRC), and basically all unittest based.
Ref: https://code.djangoproject.com/ticket/30415

nicoddemus:
Awesome!

blueyed:
I think this might have caused https://github.com/pytest-dev/pytest/issues/5301 - started looking into it, but might not finish it today anymore.

PR #5188, opened 2019-04-30
Improve help for --runxfail flag

allanlewis:
As per 528eea7:

> The help for the '--runxfail' flag is incorrect. The default behaviour is to run tests marked as 'xfail' but to ignore the results. This flag alters that behaviour in two ways:
>
> 1. *Only* tests marked 'xfail' are run.
> 2. The results of these tests are reported as if they weren't marked 'xfail'.
>
> This commit updates the help to describe point 1 above; point 2 is implied.

- [ ] ~~Target the `features` branch for new features and removals/deprecations.~~
- [ ] ~~Include documentation when adding new features.~~
- [ ] ~~Include new tests or update existing tests when applicable.~~

### DIFF
--- a/changelog/5188.doc.rst
+++ b/changelog/5188.doc.rst
@@ -1,0 +1,1 @@
+Improve help for ``--runxfail`` flag.
--- a/src/_pytest/skipping.py
+++ b/src/_pytest/skipping.py
@@ -17,7 +17,7 @@ def pytest_addoption(parser):
         action="store_true",
         dest="runxfail",
         default=False,
-        help="run tests even if they are marked xfail",
+        help="report the results of xfail tests as if they were not marked",
     )
 
     parser.addini(


nicoddemus:
Hi @allanlewis,

First of all thanks for the PR.

> Only tests marked 'xfail' are run.

I don't think this is true, consider:

```python 
import pytest

@pytest.mark.xfail
def test_x():
    assert 0

def test_y():
    pass
``` 

```
λ pytest .tmp\test-foo.py -q
x.
1 passed, 1 xfailed in 0.06 seconds


λ pytest .tmp\test-foo.py -q --runxfail
F.
============================================ FAILURES =============================================
_____________________________________________ test_x ______________________________________________

    @pytest.mark.xfail
    def test_x():
>       assert 0
E       assert 0

.tmp\test-foo.py:6: AssertionError
1 failed, 1 passed in 0.06 seconds
```

As you can see, the run with `--runxfail` still executed `test_y`, even if it was not marked as `xfail`.

So I think the help information is correct, can you tell why did you think otherwise?

allanlewis:
Hi @nicoddemus, thanks - you're absolutely right, I wrote this in haste, bear with me...

allanlewis:
I've rephrased the help message and the commit message.

PR #5204, opened 2019-05-03
Emit a warning for record_property when used with xunit2

nicoddemus:
`property` elements cannot be children of `testsuite` according to the schema, so it is incompatible with xunit2

Related to #5202

cc @danilomendesdias

### DIFF
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -281,6 +281,21 @@ class _NodeReporter(object):
         self.to_xml = lambda: py.xml.raw(data)
 
 
+def _warn_incompatibility_with_xunit2(request, fixture_name):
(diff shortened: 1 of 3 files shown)


src/_pytest/junitxml.py
```diff
@@ -307,27 +323,22 @@ def record_xml_attribute(request):
...
-    if xml is not None and xml.family != "xunit1":
-        request.node.warn(
-            PytestWarning(
-                "record_xml_attribute is incompatible with junit_family: "
-                "%s (use: legacy|xunit1)" % xml.family
-            )
-        )
-    elif xml is not None:
+    xml = getattr(request.config, "_xml", None)
+    if xml is not None:
```
blueyed: The false-branch is not covered.

nicoddemus: Nice, thanks for pointing that out. Added a new test. 👍

PR #5222, opened 2019-05-06
Restore usedevelop=True now that pip 19.1.1 restored the old behavior

### LINKED ISSUE #5167: pip 19.1 prevents pytest from being used in editable mode

nicoddemus:
Fix #5167

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -73,8 +73,7 @@ commands = pre-commit run --all-files --show-diff-on-failure
 
 [testenv:docs]
 basepython = python3
-# broken due to pip 19.1 (#5167)
-# usedevelop = True
+usedevelop = True
 changedir = doc/en
 deps = -r{toxinidir}/doc/en/requirements.txt
 
@@ -128,8 +127,7 @@ commands =
 [testenv:release]
 decription = do a release, required posarg of the version number
 basepython = python3.6
-# broken due to pip 19.1 (#5167)
-# usedevelop = True
+usedevelop = True
 passenv = *
 deps =
     colorama


blueyed:
Azure does not have/use pip 19.1.1 apparently? (for some envs only)
https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=1073&view=logs&jobId=e38d3164-b5f2-578d-f4d9-260ae069b587&taskId=8198004e-82e3-5b60-c14f-a5b1a78c79f7&lineStart=74&lineEnd=75&colStart=1&colEnd=1

asottile:
this is part of the changes in latest tox which defaults to the bundled version of pip w/ virtualenv instead of unconfigurably reaching to pypi for the latest version

we can opt into the unpinned behaviour by setting `download=true` on the testenv -- or I can bug @gaborbernat to make a virtualenv release with pip 19.1.1 :laughing:

asottile:
actually, looks like that's already in motion! https://github.com/pypa/virtualenv/commit/be8261235d949ad16f8e4b6ffbc2635487bbc692

gaborbernat:
Doing the new release is a WIP... Will probably do it in the morning.

gaborbernat:
Done via https://github.com/pypa/virtualenv/pull/1357

blueyed:
Thanks!

PR #5229, opened 2019-05-08
Require pluggy>=0.11

sitaktif:
Pluggy 0.10.0 introduced a new dependency `importlib-metadata` but that package cannot be imported when installed as an egg and that is causing problems for users relying on `setup.py` to install packages present in `test_require`. This change was quickly reverted in pluggy 0.11.0.

See https://github.com/pytest-dev/pluggy/issues/205

### DIFF
--- a/changelog/5229.bugfix.rst
+++ b/changelog/5229.bugfix.rst
@@ -1,0 +1,2 @@
+Require ``pluggy>=0.11.0`` which reverts a dependency to ``importlib-metadata`` added in ``0.10.0``.
+The ``importlib-metadata`` package cannot be imported when installed as an egg and causes issues when relying on ``setup.py`` to install test dependencies.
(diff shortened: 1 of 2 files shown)


sitaktif:
Should I create a changelog item for this (small) change?

nicoddemus:
Excellent, thanks @sitaktif!

changelog/5229.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Require ``pluggy>=0.11.0`` which reverts a dependency to ``importlib-metadata`` added in ``0.10.0``.
```
nicoddemus: Excellent CHANGELOG. 👍

blueyed:
Strictly speaking we should require `!= 0.10`, i.e. `0.9` is still fine, isn't it?

sitaktif:
> Strictly speaking we should require `!= 0.10`, i.e. `0.9` is still fine, isn't it?

Indeed you're right; I was not aware of the `!=` operator.

nicoddemus:
> Strictly speaking we should require != 0.10, i.e. 0.9 is still fine, isn't it?

In part, because we required `>=0.9` before, so it could be `>=0.9,!=0.10` I guess.

Issue #5235, opened 2019-05-08
pdb: DontReadFromInput vs EOFError


blueyed:
`DontReadFromInput` only throws `IOError` when `read` is being used.

`pyrepl` for example would raise `termios.error` when it tries to use `tcgetattr` - it does not `read` directly:

```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/daniel/Vcs/pyrepl-git/pyrepl/reader.py", line 482, in prepare
INTERNALERROR>     self.console.prepare()
INTERNALERROR>   File "/home/daniel/Vcs/pyrepl-git/pyrepl/unix_console.py", line 357, in prepare
INTERNALERROR>     self.__svtermstate = tcgetattr(self.input_fd)
INTERNALERROR>   File "/home/daniel/Vcs/pyrepl-git/pyrepl/fancy_termios.py", line 34, in tcgetattr
INTERNALERROR>     return TermState(termios.tcgetattr(fd))
INTERNALERROR> termios.error: (25, 'Inappropriate ioctl for device')
```

I've tried changing this to `EOFError`, which I found more appropriate (https://github.com/blueyed/pyrepl/pull/7), but then noticed that this causes pytest to consider this to be a quitting-the-debugger event (same as `Ctrl-d`).

This "silent" exit was confusing when I've noticed this while stepping to pytest itself, where it would run code with `set_trace` and apparently capturing output around it (another/a different issue).

I think we should maybe not use `Exit` for `do_EOF` (via `set_quit`), but only with `do_quit` as a first step.  This would also give the benefit of having `Ctrl-d` work differently than `q` (displaying a traceback, and test summaries).

https://github.com/pytest-dev/pytest/blob/2051e30b9b596e944524ccb787ed20f9f5be93e3/src/_pytest/capture.py#L704-L738

PR #5236, opened 2019-05-08
pdb: only use outcomes.exit via do_quit

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5235.

### DIFF
--- a/testing/test_pdb.py
+++ b/testing/test_pdb.py
@@ -6,6 +6,8 @@ import os
 import platform
 import sys
 
+import six
+
 import _pytest._code
 import pytest
 from _pytest.debugging import _validate_usepdb_cls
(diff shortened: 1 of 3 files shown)


testing/test_pdb.py
```diff
@@ -1163,3 +1163,28 @@ def runcall(self, *args, **kwds):
...
+    p1 = testdir.makepyfile(
+        """
+
+        def input_without_read(*args, **kwargs):
+            raise EOFError()
+
+
+        def test(monkeypatch):
+            try:
+                import __builtin__
```
nicoddemus: I suggest we make this obvious that is Python 2 compatibility code, so when we drop Python 2 we also remove the related code. Suggestion:

```python
def test_pdb_eoferror_without_read(testdir):
    builtin_module = '__builtin__' if six.PY2 else 'builtin'
    p1 = testdir.makepyfile(
        """
        def input_without_read(*args, **kwargs):
            raise EOFError()
        def test(monkeypatch):
            import {builtin_module}
            monkeypatch.setattr({builtin_module}, "input", input_without_read)
            __import__('pdb').set_trace()
        """.format(builtin_module=builtin_module)
    )
```

blueyed: Thanks, amended.

PR #5238, opened 2019-05-09
[RFC] _get_line_with_reprcrash_message: remove duplicate prefix

blueyed:
Before:

> FAILED foo/test.py::test_bar - Failed: Database access not allowed, …

After:

> FAILED foo/test.py::test_bar - Database access not allowed, …

No changelog, not yet released.. \o/ ;)

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -19,6 +19,7 @@ import pluggy
 import py
 import six
 from more_itertools import collapse
+from wcwidth import wcswidth
 
 import pytest
 from _pytest import nodes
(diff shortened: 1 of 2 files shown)


blueyed:
Too much of a hack (due to not having the orig exception there anymore) - it does not work when `pytest.fail` is used during teardown, i.e. would still display "ERROR … - Failed:" then.
Maybe a new property to be used with summaries could be added?

nicoddemus:
Not sure, you really think it worth it? 

Otherwise, a new parameter to `TestReport` as you suggest might be a good solution. Make sure to mark the property as "experimental" though.

blueyed:
Yeah, maybe later.

Issue #5241, opened 2019-05-10
Enhanced nodeids to include line numbers


blueyed:
Pytest should handle an extended nodeid format of `fname:lnum::func`, which is would then use with the short test summary (for easier selection of tests in an editor).

The format currently is `fname::func`, which is good in terms of being stable across file changes and when referring to tests themselves, but for errors / test failures it should be more specific (i.e. include the line number).

An alternative might be `fname::func:lnum`, but this is not that easy for editors to parse (but basically you would have to ignore the "::func" therin.

As for the short test summary, it could also just display `fname:lineno` there, but I think it is useful to have the test name there also.

_Originally posted by @blueyed in https://github.com/pytest-dev/pytest/pull/5002#issuecomment-477233062_

blueyed:
Actually for the short test summary we could just go with `fname:lnum: test: …`:
```
FAILED testing/test_pdb.py:1107: test_sticky_dunder_exception - assert False
FAILED testing/test_pdb.py:1141: test_sticky_dunder_exception_with_highlight - assert False
```

blueyed:
But that makes it difficult for pytest to pick it up then, since `fname:lnum` cannot be used for test selection still.

blueyed:
Done in my fork: https://github.com/blueyed/pytest/commit/c94722492058f2595b8e0aa1b43cf082c427af85

PR #5242, opened 2019-05-10
[RFC/WIP] Display line numbers in short test summary

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5241.

TODO:

- [ ] test
- [ ] changelog

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -960,17 +960,24 @@ def _get_pos(config, rep):
     if not testname:
         return nodeid
 
+    # Append location (line number).
+    # This uses the first traceback entry for the location in the test itself
+    # (rather than reprcrash, which might be less relevant for going to
+    # directly, e.g. pexpect failures in pytest itself).
     try:
-        reprcrash = rep.longrepr.reprcrash
+        testloc = rep.longrepr.reprtraceback.reprentries[0].reprfileloc
     except AttributeError:
         return nodeid
 
-    if isinstance(reprcrash.path, six.string_types):
-        crash_path = py.path.local(reprcrash.path)
+    if isinstance(testloc.path, six.string_types):
+        testloc_path = py.path.local(testloc.path)
     else:
-        crash_path = reprcrash.path
-    rel_crash_path = config.invocation_dir.bestrelpath(crash_path)
-    return "%s::%s:%s" % (rel_crash_path, testname, reprcrash.lineno)
+        testloc_path = testloc.path
+    rel_testloc_path = config.invocation_dir.bestrelpath(testloc_path)
+    if rel_testloc_path == path:
+        return "%s (line %d)" % (nodeid, testloc.lineno)
+
+    return "%s (%s:%d)" % (nodeid, rel_testloc_path, testloc.lineno)
 
 
 def _get_line_with_reprcrash_message(config, rep, termwidth):
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -9,6 +9,7 @@ from __future__ import print_function
 
 import argparse
 import collections
+import os
 import platform
 import sys
 import time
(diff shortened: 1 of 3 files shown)


blueyed:
This currently does not display the location of the crash.
Should use `rep.longrepr.reprcrash` - but that might be a hidden frame.. :/

blueyed:
One idea would be for pytest to accept `fname:lnum::test` as an argument, but ignore the `:lnum` part therein.

Related: https://github.com/pytest-dev/pytest/pull/3282

blueyed:
Also having the crash location explicitly appended might be good:

> FAILED testing/test_pdb.py::TestPDB::test_pdb_interaction_capturing_simple (.venv/lib/python3.7/site-packages/pexpect/expect.py:82) - pexpect.exceptions.TIMEOUT: Timeout e...

> FAILED testing/test_pdb.py::TestDebuggingBreakpoints::test_pdb_custom_cls (line 908) - Failed: nomatch: '*CustomDebugger*'

But for the first case I think it is better to have the location / line from the test itself - so I've changed it to do that now.

src/_pytest/terminal.py
```diff
@@ -956,7 +956,28 @@ def show_skipped(lines):
...
+    # directly, e.g. pexpect failures in pytest itself).
+    try:
+        testloc = rep.longrepr.reprtraceback.reprentries[0].reprfileloc
+    except AttributeError:
+        return nodeid
+
+    if isinstance(testloc.path, six.string_types):
+        testloc_path = py.path.local(testloc.path)
+    else:
+        testloc_path = testloc.path
```
blueyed: I thought I've seen this to be necessary when testing, but only one branch is covered here.

src/_pytest/terminal.py
```diff
@@ -956,7 +956,28 @@ def show_skipped(lines):
...
+    rel_testloc_path = config.invocation_dir.bestrelpath(testloc_path)
+    if rel_testloc_path == path:
+        return "%s (line %d)" % (nodeid, testloc.lineno)
+
+    return "%s (%s:%d)" % (nodeid, rel_testloc_path, testloc.lineno)
```
blueyed: Does this make sense like this?
Is there a better API (to get to the entry in the test)?

src/_pytest/terminal.py
```diff
@@ -956,7 +956,28 @@ def show_skipped(lines):
...
+    # Append location (line number).
+    # This uses the first traceback entry for the location in the test itself
+    # (rather than reprcrash, which might be less relevant for going to
+    # directly, e.g. pexpect failures in pytest itself).
+    try:
+        testloc = rep.longrepr.reprtraceback.reprentries[0].reprfileloc
+    except AttributeError:
+        return nodeid
+
+    if isinstance(testloc.path, six.string_types):
```
nicoddemus: I think you can do just this:

```python
testloc_path = py.path.local(testloc.path)
```

If `testloc.path` is already a `py.path.local` instance, then it should be a noop (or very cheap).

src/_pytest/terminal.py
nicoddemus: In which situation does this fall into? Won't this generate very long summary lines? I fear as it is the message already takes quite all the terminal space and most of the times it gets trimmed anyway, adding this information to the summary line will make it even less useful, as more of exception message will be trimmed.

How do you feel if we include this information only when `-v` is passed?

src/_pytest/terminal.py
blueyed: Yeah, this was more to see if it is needed after all (via coverage).
IIRC I've seen some inconsistency when testing, but only one case is used IIRC^2.

src/_pytest/terminal.py
blueyed: You mean the last case?
Ok with verbose > 0 here.
Adding only " (line XX)" is useful.

It is probably not hit anyway I would imagine (have not checked).

nicoddemus: About `-v`, I meant if we only added line numbers of `-v` is passed. In retrospect this could also be the case for the message summary as well I guess.

blueyed: Well, I think those are very useful to have by default and only add like ~10 extra chars.

blueyed:
Done in my fork: https://github.com/blueyed/pytest/commit/c94722492058f2595b8e0aa1b43cf082c427af85

@asfaltboy (since you +1'd): feel free to pick it up from there.  I've gone with a slightly different format there.

Issue #5228, opened 2019-05-08
pdb: cls._pluginmanager, config etc might be None


blueyed:
This can happen when debugging threads after pytest finished already (and the debugging plugin was unconfigured).

A failing test:

```python
def test_pdb_in_thread_after_exit(testdir):
    """It should be imported in pytest_configure or later only."""
    p1 = testdir.makepyfile(
        """
        import threading


        def test():
            evt = threading.Event()
            evt2 = threading.Event()

            def target():
                evt.set()
                evt2.wait()
                print("target_set_trace")
                __import__('pdb').set_trace()
                print("target_end")

            thread = threading.Thread(target=target)
            thread.start()

            evt.wait()
            evt2.set()
        """,
    )
    child = testdir.spawn_pytest(str(p1) + " -s")
    child.expect("target_set_trace")
    rest = child.read().decode("utf8")
    assert "Exception in thread" not in rest
    assert child.exitstatus == 0
```

I think it should not pop the last set of attributes, but hard to say when the last time is maybe?!
IIRC it does not use pytest_unconfigure intentionally.

The test causes this:
```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/tmp/pytest-of-user/pytest-61/test_pdb_in_thread_after_exit0/test_pdb_in_thread_after_exit.py", line 12, in target
    __import__('pdb').set_trace()
  File "…/Vcs/pytest/src/_pytest/debugging.py", line 221, in set_trace
    _pdb = cls._init_pdb(*args, **kwargs)
  File "…/Vcs/pytest/src/_pytest/debugging.py", line 212, in _init_pdb
    cls._pluginmanager.hook.pytest_enter_pdb(config=cls._config, pdb=_pdb)
AttributeError: 'NoneType' object has no attribute 'hook'
```

I've seen this initially:
```
(Pdb++[sync-to-async-_0]) c
Exception in worker
Traceback (most recent call last):
  File "/usr/lib/python3.7/concurrent/futures/thread.py", line 79, in _worker
    if work_item is not None:
  File "/usr/lib/python3.7/concurrent/futures/thread.py", line 79, in _worker
    if work_item is not None:
  File "/usr/lib/python3.7/bdb.py", line 88, in trace_dispatch
[...]

PR #5244, opened 2019-05-11
WIP/RFC pdb: fix usage in child thread after main thread exited

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5228.

(not done against master to avoid conflicts)

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -77,9 +77,7 @@ def pytest_configure(config):
     if config.getvalue("usepdb"):
         config.pluginmanager.register(PdbInvoke(), "pdbinvoke")
 
-    pytestPDB._saved.append(
-        (pdb.set_trace, pytestPDB._pluginmanager, pytestPDB._config, pytestPDB._pdb_cls)
-    )
+    pytestPDB._saved.append(pdb.set_trace)
     pdb.set_trace = pytestPDB.set_trace
     pytestPDB._pluginmanager = config.pluginmanager
     pytestPDB._config = config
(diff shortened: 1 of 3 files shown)


blueyed:
Fixed the non-related test in https://github.com/pytest-dev/pytest/pull/5245.

src/_pytest/debugging.py
```diff
@@ -77,9 +77,7 @@ def pytest_configure(config):
     if config.getvalue("usepdb"):
         config.pluginmanager.register(PdbInvoke(), "pdbinvoke")
 
-    pytestPDB._saved.append(
-        (pdb.set_trace, pytestPDB._pluginmanager, pytestPDB._config, pytestPDB._pdb_cls)
-    )
+    pytestPDB._saved.append(pdb.set_trace)
```
nicoddemus: I'm a little worried about this because this leaves `config` and `pluginmanager` alive after `pytest.main()`, which might be a problem for someone embedding `pytest` I guess.

I wonder if we can only pass what is is actually needed to `pytestPDB` instead of the entire `config` and `pluginmanager` objects?

blueyed: Well, the main point here is to keep them alive after pytest exits, isn't it?.. ;)

(The whole wrapping can be improved in general - I'm doing this basically step by step when running into issues)

nicoddemus: > Well, the main point here is to keep them alive after pytest exits, isn't it?.. ;)

Yep, and I'm not sure that's a good thing 😅, specially if I'm not using threads at all.

blueyed: An alternative would be to fix the debugging plugin to not crash in this case I think.

PR #5245, opened 2019-05-11
tests: fix test_trace_after_runpytest

blueyed:
It was not really testing what it was supposed to test (e.g. the inner
test was not run in the first place).

### DIFF
--- a/testing/test_pdb.py
+++ b/testing/test_pdb.py
@@ -1027,24 +1027,28 @@ def test_trace_after_runpytest(testdir):
         from _pytest.debugging import pytestPDB
 
         def test_outer(testdir):
-            from _pytest.debugging import pytestPDB
-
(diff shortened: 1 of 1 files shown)


blueyed:
Would rebase this against features then after the release.

testing/test_pdb.py
```diff
@@ -1027,24 +1027,28 @@ def test_trace_after_runpytest(testdir):
...
-            testdir.runpytest("-k test_inner")
+            testdir.makepyfile(
+                \"""
+                from _pytest.debugging import pytestPDB
 
-            __import__('pdb').set_trace()
+                def test_inner():
+                    assert len(pytestPDB._saved) == 2
+                    print()
+                    print("test_inner_" + "end")
```
nicoddemus: 😉

PR #5250, opened 2019-05-12
Documentation: add setenv/delenv examples to monkeypatch docs

EvanKepner:
This is a documentation update for `monkeypatch` to include an example of `setenv` and `delenv`. Targeting the `master` branch since it is documentation, `tox -e docs` builds locally for me and I've set my RST hard-wrapping at 88 based on guidance from `black` defaults. Will push the commit for the changelog shortly referencing this PR number.

File changes:
 - docs/en/monkeypatch.rst: added a new example section related to environment variables.
 - AUTHORS: added myself in alphabetical order

### DIFF
--- a/doc/en/monkeypatch.rst
+++ b/doc/en/monkeypatch.rst
@@ -86,32 +86,39 @@ Monkeypatching environment variables
 
 If you are working with environment variables you often need to safely change the values
 or delete them from the system for testing purposes. ``Monkeypatch`` provides a mechanism
-to do this using the ``setenv`` and ``delenv`` method. Our example code to test::
+to do this using the ``setenv`` and ``delenv`` method. Our example code to test:
+
+.. code-block:: python
 
     # contents of our original code file e.g. code.py
     import os
 
+
     def get_os_user_lower():
-    """Simple retrieval function.
-    Returns lowercase USER or raises EnvironmentError."""
-    username = os.getenv("USER")
+        """Simple retrieval function.
+        Returns lowercase USER or raises EnvironmentError."""
+        username = os.getenv("USER")
 
-    if username is None:
-        raise EnvironmentError("USER environment is not set.")
+        if username is None:
+            raise EnvironmentError("USER environment is not set.")
 
-    return username.lower()
+        return username.lower()
 
 There are two potential paths. First, the ``USER`` environment variable is set to a
 value. Second, the ``USER`` environment variable does not exist. Using ``monkeypatch``
-both paths can be safely tested without impacting the running environment::
+both paths can be safely tested without impacting the running environment:
+
+.. code-block:: python
 
     # contents of our test file e.g. test_code.py
     import pytest
+
 
     def test_upper_to_lower(monkeypatch):
         """Set the USER env var to assert the behavior."""
         monkeypatch.setenv("USER", "TestingUser")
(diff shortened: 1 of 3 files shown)


doc/en/monkeypatch.rst
```diff
@@ -81,6 +81,80 @@ so that any attempts within tests to create http requests will fail.
...
+        monkeypatch.setenv("USER", "TestingUser")
+        assert get_os_user_lower() == "testinguser"
+
+
+    def test_raise_exception(monkeypatch):
+        """Remove the USER env var and assert EnvironmentError is raised."""
+        monkeypatch.delenv("USER", raising=False)
+
+        with pytest.raises(EnvironmentError):
+            _ = get_os_user_lower()
```
blueyed: I wonder why there is an assignment here?

doc/en/monkeypatch.rst
```diff
@@ -81,6 +81,80 @@ so that any attempts within tests to create http requests will fail.
...
+
+
+    # Notice the tests reference the fixtures for mocks
+    def test_upper_to_lower(mock_env_user):
+        assert get_os_user_lower() == "testinguser"
+
+
+    def test_raise_exception(mock_env_missing):
+        with pytest.raises(EnvironmentError):
+            _ = get_os_user_lower()
```
blueyed: Also here.

doc/en/monkeypatch.rst
nicoddemus: Some linters complain that the statement has no effect unless you are assigning it to a variable. I'm fine keeping it like this, but feel free to fix it if you prefer. 👍

doc/en/monkeypatch.rst
nicoddemus: Ditto

doc/en/monkeypatch.rst
EvanKepner: Yeah this is me being explicit since `get_os_user_lower()` has the intent of returning a value, but I know it will be an unused value (and never actually returned) if the test executes as expected. Agreed that if it's confusing I have no issue with removing it, and happy to make that commit myself on the fork.

EvanKepner:
@nicoddemus @blueyed thanks for taking a look and for the `blacken-docs` formatting commit! I wondered about the code-blocks in RST but was following the convention in the earlier unchanged statements (e.g. line 23). I'll catch that on any future contributions :+1:

nicoddemus:
@EvanKepner sure, no worries, it was not obvious at all!

Merging as the Azure failure is unrelated. 👍

doc/en/monkeypatch.rst
blueyed: It's ok (no additional change required) - using `_` here shows the intention.

@nicoddemus 
What linters complain about this?

nicoddemus: Hmmm I might be confusing things. 

```python
a, b = foo()
```

Where you don't to use `b`, PyCharm marks `b` as "unused", but it won't complain if you write:

```python
a, _ = foo()
```

instead.

Of course this situation does not apply here.

blueyed: Yeah.. it would only work if the linter knows that the function returns something in the first place usually, e.g. via type information - otherwise every `print()` would be marked already.. :)

PR #5252, opened 2019-05-12
Fix invalid Python file encoding "utf8"

tambry:
Since Python 3 it must be "utf-8", which is the official name.
This is backwards compatible with Python 2.


blueyed:
Interesting - do you have a reference?

There are more occurences of this however, e.g. in testing/test_warnings.py.

tambry:
I unfortunately do not have a reference. It seems to be a bit flaky whether the error happens. Though you can find cases of this problem on Google.

I fixed only this occurence, because in my initial search only it showed up on Chromium CodeSearch.

asottile:
unless there's a stacktrace I'm inclined to leave this as-is -- [PEP 263](https://www.python.org/dev/peps/pep-0263/) seems to allow what we have currently and `utf8` is properly regisetered as an alias:

```pycon
>>> codecs.lookup('utf8')
<codecs.CodecInfo object for encoding utf-8 at 0x7faf98f0fca0>
```

nicoddemus:
Thanks @tambry!

Normalized all encoding declarations to the same format, it doesn't hurt being consistent. 👍

src/_pytest/skipping.py
```diff
@@ -1,4 +1,4 @@
-# coding=utf8
```
asottile: let's use `fix-encoding-pragma` from `pre-commit/pre-commit-hooks`

(yep, there's a pre-commit hook for everything!)

nicoddemus: Sure, sounds good! Could you please contribute that in place of this PR then? 😁

asottile: done!

asottile:
hmmm, that's weird it's showing a diff in CI -- let me see what I can do there, sorry about that!

src/_pytest/skipping.py
blueyed: Can it only do it if necessary maybe?

asottile: for consistency it's all or none, that way you don't have to think about it

asottile:
> 
> 
> hmmm, that's weird it's showing a diff in CI -- let me see what I can do there, sorry about that!

I suspect crlf since it's windows and that's not implemented yet -- easy patch though: https://github.com/pre-commit/pre-commit-hooks/pull/384

tambry:
Thanks for continuing work on this!

But I think this misses all the cases that are in the middle of code somewhere. Might be worth running grep and fixing those manually?

asottile:
> 
> 
> Thanks for continuing work on this!
> 
> But I think this misses all the cases that are in the middle of code somewhere. Might be worth running grep and fixing those manually?

yep, did that already -- did I miss some?

tambry:
> > Thanks for continuing work on this!
> > But I think this misses all the cases that are in the middle of code somewhere. Might be worth running grep and fixing those manually?
> 
> yep, did that already -- did I miss some?

Ah, probably fine then. 👍

nicoddemus:
Waiting for @blueyed's approval/change requests before merging.

src/_pytest/skipping.py
blueyed: Isn't the main point of pre-commit already that "you don't have to think about it", i.e. you do not have to care, and it will insert/request it if necessary?

blueyed:
> Waiting for @blueyed's approval/change requests before merging.

I don't like changing that many files unnecessarily (https://github.com/pytest-dev/pytest/pull/5252/files#r285079054).  But feel free to ignore me, of course.. :)

src/_pytest/skipping.py
asottile: yeah so now you don't have to futz with whether a file has it or not, it always has it

nicoddemus:
Personally I don't mind when the change is a single line too much; wide sweeping changes like formatting annoy me more as they make 'git blame' harder to use. 👍 😁

PR #5263, opened 2019-05-14
Merge master to fix CI

blueyed:


asottile:
lolrip, rng gave you the creepy cat

PR #5267, opened 2019-05-15
Fix/improve comparison of byte strings

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5260.

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -291,8 +291,11 @@ def _compare_eq_sequence(left, right, verbose=0):
         if left[i] != right[i]:
             explanation += [u"At index %s diff: %r != %r" % (i, left[i], right[i])]
             break
-    len_diff = len_left - len_right
 
+    if isinstance(left, bytes):
+        return explanation
+
+    len_diff = len_left - len_right
     if len_diff:
         if len_diff > 0:
             dir_with_more = "Left"
--- a/testing/test_assertion.py
+++ b/testing/test_assertion.py
@@ -354,6 +354,18 @@ class TestAssert_reprcompare(object):
         assert "- spam" in diff
         assert "+ eggs" in diff
 
+    def test_bytes_diff(self):
+        diff = callequal(b"spam", b"eggs")
+        if PY3:
+            assert diff == [
+                "b'spam' == b'eggs'",
+                "At index 0 diff: 115 != 101",
+                "Use -v to get the full diff",
+            ]
+        else:
+            # Handled as text on Python 2.
+            assert diff == ["'spam' == 'eggs'", "- spam", "+ eggs"]
+
     def test_list(self):
         expl = callequal([0, 1], [0, 2])
         assert len(expl) > 1
--- a/changelog/5260.bugfix.rst
+++ b/changelog/5260.bugfix.rst
@@ -1,0 +1,1 @@
+Improve/fix comparison of byte strings with Python 3.


src/_pytest/assertion/util.py
```diff
@@ -147,7 +147,9 @@ def assertrepr_compare(config, op, left, right):
     explanation = None
     try:
         if op == "==":
-            if istext(left) and istext(right):
+            if isinstance(left, bytes) and isinstance(right, bytes):
+                explanation = _diff_text(left, right, verbose)
+            elif istext(left) and istext(right):
```
blueyed: Note that this compares bytes and unicode on py2 still..

blueyed:
Azure failure is unrelated.

src/_pytest/assertion/util.py
blueyed: Maybe this as a follow up?
But it changes `test_nonascii_text`, since the type is the derived class then (https://github.com/pytest-dev/pytest/pull/5268/files#diff-14ad89cac128f1dd3aaa927c656035b3R597).

Anyway, should not block this PR.

```diff
commit 0b9d2657e (HEAD -> cmp-bytes)
Date:   Wed May 15 12:29:14 2019 +0200

    Do not mix unicode/bytes on Python 2

diff --git a/changelog/5267.bugfix.rst b/changelog/5267.bugfix.rst
new file mode 100644
index 000000000..749360195
--- /dev/null
+++ b/changelog/5267.bugfix.rst
@@ -0,0 +1 @@
+Comparisons with assertion rewriting are now skipped with mixed unicode/bytes on Python 2.
diff --git a/src/_pytest/assertion/util.py b/src/_pytest/assertion/util.py
index 23bbe49b7..7e45735e3 100644
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -147,9 +147,7 @@ def assertrepr_compare(config, op, left, right):
     explanation = None
     try:
         if op == "==":
-            if isinstance(left, bytes) and isinstance(right, bytes):
-                explanation = _diff_text(left, right, verbose)
-            elif istext(left) and istext(right):
+            if type(left) == type(right) and (isinstance(left, bytes) or istext(left)):
                 explanation = _diff_text(left, right, verbose)
             else:
                 if issequence(left) and issequence(right):
diff --git a/testing/test_assertion.py b/testing/test_assertion.py
index 9f4244640..501bf6ae5 100644
--- a/testing/test_assertion.py
+++ b/testing/test_assertion.py
@@ -361,6 +361,11 @@ def test_bytes_diff(self):
         else:
             assert diff == ["'spam' == 'eggs'", "- spam", "+ eggs"]
 
+    def test_str_and_bytes_diff(self):
+        """pytest_assertrepr_compare does nothing with mixed bytes + string."""
+        diff = callequal(u"spam", b"eggs")
+        assert diff is None
+
     def test_list(self):
         expl = callequal([0, 1], [0, 2])
         assert len(expl) > 1
```

src/_pytest/assertion/util.py
```diff
@@ -147,7 +147,9 @@ def assertrepr_compare(config, op, left, right):
     explanation = None
     try:
         if op == "==":
-            if istext(left) and istext(right):
+            if isinstance(left, bytes) and isinstance(right, bytes):
+                explanation = _diff_text(left, right, verbose)
```
asottile: I don't think we should be treating binary data as text

blueyed: That's what it does on py2 and from all the comments there it looks to just be
an oversight with py3.

So if not doing it for py3 it should be done with unicode strings on py2
probably then also only?

Current:

```
b'\\x00\\x00\\x00' == b'\\x00\\x01\\x02'
At index 1 diff: 0 != 1
Full diff:
- b'\\x00\\x00\\x00'
?          ^   ^
+ b'\\x00\\x01\\x02'
?          ^   ^
```

This:

```
b'\\x00\\x00\\x00' == b'\\x00\\x01\\x02'
- \\x00\\x00\\x00
?        ^   ^
+ \\x00\\x01\\x02
?        ^   ^
```

So the current version is actually more useful.

asottile: The second version loses the `b'` which is arguably worse

blueyed: That would be easy to not do though - I've done it to match what the function is intended to do (with regard to the py2 behavior).

blueyed: Let's go with your suggestion then?

> we should maybe just omit the "contains more items" messaging for bytes objects?

blueyed: And should the py2 behavior be kept then?

asottile: Yeah, sounds good. We can keep the py2 behavior as strings and bytes are more interchangeable there

blueyed: Ok, done.

testing/test_assertion.py
```diff
@@ -354,6 +354,18 @@ def test_multiline_text_diff(self):
         assert "- spam" in diff
         assert "+ eggs" in diff
 
+    def test_bytes_diff(self):
+        diff = callequal(b"spam", b"eggs")
+        if PY3:
+            assert diff == [
+                "b'spam' == b'eggs'",
+                "At index 0 diff: 115 != 101",
```
asottile: hmmm I thought that we *wouldn't* see this line, but only see the `- b'spam'` `+ b'eggs'` bit?

blueyed: No - it just skips the "contains more items" now.
I think it is fine / useful, although it could mention that those a bytes / ordinals.
If so, how should it sound like then?
Or should we skip any handling (i.e. as a sequence) for bytes?

asottile: hmmm I'd like to still see the reprdiff since I think that's useful

nicoddemus: How about using the slice operator instead?

```
>>> b'foo'[0]
102
>>> b'foo'[0:1]
b'f'
```

We will then be able to see `"At index 0 diff: b's' != b'e'"`, which is readable and has no downsides as far as I can see.

blueyed: As a special case for bytes on py3 then, or for sequences in general?

nicoddemus: special case for bytes on py3, for sequences it is not clear to me it would be better than what we have now, at least for me.

blueyed:
I do not feel strong enough about this myself currently, closing for now.

nicoddemus:
> Let's keep it for later - almost there.. ;)

By "later" you mean after dropping Python 2 in 4.6/4.7?

nicoddemus:
Gentle ping :)

blueyed:
Busy pong.. ;)
Feel free to pick it up or close it to get off your radar.. :)

nicoddemus:
Superseded by #5495

PR #5271, opened 2019-05-15
Handle lone surrogate unicode character not being representable in Jython

### LINKED ISSUE #5256: Invalid unicode in _pytest/terminal.py with Jython
With: Jython 2.7.1 and pytest 4.5.0, I'm having the error below:

```
  File "C:\bin\jython2.7.1\Lib\runpy.py", line 161, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\bin\jython2.7.1\Lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\bin\jython2.7.1\Lib\site-packages\pytest.py", line 101, in <module>
    raise SystemExit(pytest.main())
  File "C:\bin\jython2.7.1\Lib\site-packages\_pytest\config\__init__.py", line 60, in main
  File "C:\bin\jython2.7.1\Lib\site-packages\_pytest\config\__init__.py", line 60, in main
  File "C:\bin\jython2.7.1\Lib\site-packages\_pytest\config\__init__.py", line 60, in main
  File "C:\bin\jython2.7.1\Lib\site-packages\_pytest\config\__init__.py", line 190, in _prepareconfig
    config = get_config(args)
  File "C:\bin\jython2.7.1\Lib\site-packages\_pytest\config\__init__.py", line 164, in get_config
    pluginmanager.import_plugin(spec)
  File "C:\bin\jython2.7.1\Lib\site-packages\_pytest\config\__init__.py", line 562, in import_plugin
    __import__(importspec)
UnicodeDecodeError: 'unicodeescape' codec can't decode bytes in position 2-8: illegal Unicode character
```

The error happens because `terminal._get_line_with_reprcrash_message` has the following unicode literal: `u"\uD83D"`.

I'm not sure about the proper fix... in that function `msg` seems to be doing a `msg.find("\n")` and later does `msg.rstrip(u"\uD83D")`, so, I'm not even sure if we're dealing with `unicode` or `bytes` there on Python 2 because it's mixed (my guess is bytes, but I'm not really sure).

If it's bytes, doing it an actual `str` literal instead of a `unicode` literal -- i.e.: remove the `u` -- does make Jython happy (would that be an appropriate solution?)

nicoddemus:
No tests for this because we don't test with Jython currently.

Fix #5256

### DIFF
--- a/changelog/5256.bugfix.rst
+++ b/changelog/5256.bugfix.rst
@@ -1,0 +1,1 @@
+Handle internal error due to a lone surrogate unicode character not being representable in Jython.
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -998,7 +998,15 @@ def _get_line_with_reprcrash_message(config, rep, termwidth):
                     # u'😄' will result in a High Surrogate (U+D83D) character, which is
                     # rendered as u'�'; in this case we just strip that character out as it
                     # serves no purpose being rendered
-                    msg = msg.rstrip(u"\uD83D")
+                    try:
+                        surrogate = six.unichr(0xD83D)
+                        msg = msg.rstrip(surrogate)
+                    except ValueError:  # pragma: no cover
+                        # Jython cannot represent this lone surrogate at all (#5256):
+                        # ValueError: unichr() arg is a lone surrogate in range
+                        #     (0xD800, 0xDFFF) (Jython UTF-16 encoding)
+                        # ignore this case as it shouldn't appear in the string anyway
+                        pass
                 msg += ellipsis
             line += sep + msg
     return line


blueyed:
@asottile 
I think this should have waited on https://github.com/pytest-dev/pytest/issues/5256#issuecomment-493034326.

nicoddemus:
@blueyed I think it is fine, we can always do a follow-up. 👍 

And after giving this some more thought, we needed this fix because *our* test suite used 😄  emoji in the messages (my suggestion), so it might not even happen in the wild as I doubt people are putting emojis in exception messages.

Either way I think a follow up later (including a CHANGELOG adjustment) is fine. 👍

blueyed:
Sure.  But there was also no reason to get this intermediate fix in fast - since we're not testing against Jython.

Anyway, it was just a remark - nothing too bad.

fabioz:
The current change is already ok for me (Jython works)... 

As a note, what I came up with in the end is below (where `obj_repr` is a `unicode` string and `left_count` the index to be sliced):

```
try:
    if left_count > 0 and unichr(0xD800) <= obj_repr[left_count - 1] <= unichr(0xDBFF):
        left_count -= 1
except ValueError:
    # On Jython unichr(0xD800) will throw an error:
    # ValueError: unichr() arg is a lone surrogate in range (0xD800, 0xDFFF) (Jython UTF-16 encoding)
    # Just ignore it in this case.
    pass
```

I'm also a bit wary of `pytest` doing `msg.rstrip()` with a unicode identifier where `msg` is `bytes` on Python 2 (that could fail depending on the encoding used to automatically make the conversion from bytes to unicode) -- in: https://github.com/pytest-dev/pytest/commit/d94b4b031ffb1e1d47a6498c1552d0d5cb7a6d6a#diff-606d01f030119020f3e820547882b6b0R1003 (but as @nicoddemus mentioned, it's quite rare that exceptions have unicode at all, so, this may not happen all that much in practice).

nicoddemus:
You are right @fabioz, that could lead to problems. 

I propose we wait a bit though, I'm not sure this will really be a problem in the wild. If it does we can always fix it with an actual problem as guide. 👍 

Thanks everyone!

PR #5273, opened 2019-05-16
Fixed double `be` word on monkeypatch docs page.

lsnk:
Do i really need to create issue and changelog record for such a small fix?


nicoddemus:
Hi @lsnk, thanks for the contribution!

> Do i really need to create issue and changelog record for such a small fix?

No, but you are right we should improve our template to make that explicit. 👍

PR #5274, opened 2019-05-16
Disable PyPy on Azure

nicoddemus:
Disabling pypy2 and pypy3 for now on Azure so we can get green PRs again (#5279)


nicoddemus:
Disabling pypy2 and pypy3 for now on Azure so we can get green PRs again (#5279)

PR #5276, opened 2019-05-16
Make it clear that small typos in docs don't require a CHANGELOG entry

nicoddemus:
From: https://github.com/pytest-dev/pytest/pull/5273#issuecomment-493076587


nicoddemus:
Oops indeed, my bad!

Issue #5277, opened 2019-05-16
Can't use `-p no:python` since 4.5.0


drothlis:
I have a custom pytest plugin that runs a test remotely. I run pytest with `-p my_plugin -p no:python` so that it doesn't also run it locally. Since pytest 4.5.0, pytest raises `UsageError: plugin python cannot be disabled`.

This behaviour was introduced in #5005. I have read through that PR, and its issue #4976, and PR #4965 where the change was originally suggested, but I still don't understand the reason for preventing `p -no:python`.

The reason I use pytest for this, is because the PyCharm IDE provides pytest integration. With my configuration I can click on a "play" icon next to the test name, and PyCharm runs pytest. PyCharm doesn't allow configuring a custom shell command, hence I've been using pytest.

blueyed:
This is mainly due to pytest itself crashing without it (which could be fixed by using default for options etc):

```
Traceback (most recent call last):
  File "…/Vcs/pytest/src/_pytest/config/__init__.py", line 931, in getini
    return self._inicache[name]
KeyError: 'python_files'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "…/Vcs/pytest/src/_pytest/config/__init__.py", line 938, in _getini
    description, type, default = self._parser._inidict[name]
KeyError: 'python_files'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "…/Vcs/pytest/src/pytest.py", line 101, in <module>
    raise SystemExit(pytest.main())
  File "…/Vcs/pytest/src/_pytest/config/__init__.py", line 60, in main
    config = _prepareconfig(args, plugins)
  File "…/Vcs/pytest/src/_pytest/config/__init__.py", line 203, in _prepareconfig
    pluginmanager=pluginmanager, args=args
  File "…/Vcs/pytest/.venv/lib/python3.7/site-packages/pluggy/hooks.py", line 289, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "…/Vcs/pytest/.venv/lib/python3.7/site-packages/pluggy/manager.py", line 68, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
[...]

PR #5278, opened 2019-05-16
Allow disabling of python plugin

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5277

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -113,16 +113,16 @@ def directory_arg(path, optname):
 
 
 # Plugins that cannot be disabled via "-p no:X" currently.
-essential_plugins = (
+essential_plugins = (  # fmt: off
     "mark",
     "main",
     "runner",
-    "python",
     "fixtures",
     "helpconfig",  # Provides -p.
-)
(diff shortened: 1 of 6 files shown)


src/_pytest/config/__init__.py
```diff
@@ -761,6 +761,18 @@ def _initini(self, args):
         self._parser.extra_info["inifile"] = self.inifile
         self._parser.addini("addopts", "extra command line options", "args")
         self._parser.addini("minversion", "minimally required pytest version")
+        self._parser.addini(
+            "python_files",
```
nicoddemus: Not a big fan of having this options here, as they are clearly related to the `python` plugin.

If the cause is because of assertion rewrite (https://github.com/pytest-dev/pytest/issues/5277#issuecomment-493232024), can we just use a default there (using an empty `[]` if not defined)? Would that allow `-p no:python` to work?

blueyed: Done, using the actual default.
And moved `usefixtures` to the fixtures plugin.

blueyed: Please re-review.

PR #5297, opened 2019-05-21
Add link to tidelift to the main docs

nicoddemus:
As commented in:

https://github.com/pytest-dev/pytest/pull/5240#issuecomment-494175451

### DIFF
--- a/doc/en/contents.rst
+++ b/doc/en/contents.rst
@@ -50,6 +50,7 @@ Full pytest documentation
    projects
    faq
    contact
+   tidelift
 
 .. only:: html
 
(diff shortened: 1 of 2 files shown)


hugovk:
I spotted there's a bit of extra (~$46.84/month) Tidelift income available for the https://github.com/pytest-dev/py project, I think it'd make sense for the pytest team to collect this rather than leaving it unclaimed:

* https://tidelift.com/lifter/search/pypi/py

nicoddemus:
Thanks @hugovk,

I personally don't feel comfortable receiving that as I'm not doing much for `py` other than the _very_ occasional fix. If someone wants to step in to give it a bit of love and receive those funds it would be great; I'm not saying revive the project or anything, but some small maintenance work, like making sure it runs in the latest pytest for example, would be a welcome change.

PR #5306, opened 2019-05-24
pytester: use temporary HOME with spawn

blueyed:
Followup to https://github.com/pytest-dev/pytest/issues/4956.

### DIFF
--- a/changelog/4956.feature.rst
+++ b/changelog/4956.feature.rst
@@ -1,0 +1,1 @@
+pytester's ``testdir.spawn`` uses ``tmpdir`` as HOME/USERPROFILE directory.
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1233,7 +1233,13 @@ class Testdir(object):
         if sys.platform.startswith("freebsd"):
             pytest.xfail("pexpect does not work reliably on freebsd")
         logfile = self.tmpdir.join("spawn.out").open("wb")
-        child = pexpect.spawn(cmd, logfile=logfile)
+
+        # Do not load user config.
+        env = os.environ.copy()
+        env["HOME"] = str(self.tmpdir)
+        env["USERPROFILE"] = env["HOME"]
+
+        child = pexpect.spawn(cmd, logfile=logfile, env=env)
         self.request.addfinalizer(logfile.close)
         child.timeout = expect_timeout
         return child
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -508,6 +508,10 @@ class Testdir(object):
         # Discard outer pytest options.
         mp.delenv("PYTEST_ADDOPTS", raising=False)
 
+        # Environment (updates) for inner runs.
+        tmphome = str(self.tmpdir)
+        self._env_run_update = {"HOME": tmphome, "USERPROFILE": tmphome}
+
     def __repr__(self):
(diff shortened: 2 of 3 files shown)


changelog/4956.feature.rst
```diff
@@ -0,0 +1 @@
+pytester's ``testdir.spawn`` uses a temporary HOME directory.
```
nicoddemus:
```suggestion
pytester's ``testdir.spawn`` uses ``tmpdir`` as HOME/USERPROFILE directory.
```

blueyed:
Refactored it, which also allows for adjusting it via `testdir._env_run_update`.

blueyed:
This would allow for the following now, which works around pdb++ hijacking pdb and adding source code highlighting, which fails some (3) tests currently.  Not sure if it should be added though, given that it is rather specific (pdbpp being installed), and tests could be adjusted to cope with it (i.e. loser/other assertions).

```diff
diff --git i/testing/test_pdb.py w/testing/test_pdb.py
index c0ba7159b..81e455b13 100644
--- i/testing/test_pdb.py
+++ w/testing/test_pdb.py
@@ -76,6 +76,13 @@ def set_trace(self, frame):
     del _pytest._CustomDebugger


+@pytest.fixture(autouse=True)
+def pdb_env(request):
+    if "testdir" in request.funcargnames:
+        testdir = request.getfixturevalue("testdir")
+        testdir._env_run_update["PDBPP_HIJACK_PDB"] = "0"
+
+
 class TestPDB(object):
     @pytest.fixture
     def pdblist(self, request):
```

blueyed:
Note that the above fixture is different from setting `PDBPP_HIJACK_PDB=0` before running pytest: it would still use pdbpp then in the outer tests.

nicoddemus:
> This would allow for the following now, which works around pdb++ hijacking pdb and adding source code highlighting, which fails some (3) tests currently. Not sure if it should be added though, given that it is rather specific (pdbpp being installed), and tests could be adjusted to cope with it (i.e. loser/other assertions).

I'm fine if this helps people with pdb++ installed to run the pytest suite (such as yourself), just please make sure to add an informative docstring to the fixture though.

Also feel free to merge this now if you prefer to introduce this fixture in a future PR instead.

blueyed:
Test was skipped - needs https://github.com/pytest-dev/pytest/pull/5313.

tox.ini
```diff
@@ -40,7 +40,7 @@ setenv =
     lsof: _PYTEST_TOX_POSARGS_LSOF=--lsof
 
     pexpect: _PYTEST_TOX_PLATFORM=linux|darwin
-    pexpect: _PYTEST_TOX_POSARGS_PEXPECT=testing/test_pdb.py testing/test_terminal.py testing/test_unittest.py
+    pexpect: _PYTEST_TOX_POSARGS_PEXPECT=-m uses_pexpect
```
blueyed: Switched to using the new marker here.

blueyed: Meh.. that will fail when combined with twisted etc I guess.
Also with using pexpect for macos in the other PR.
Should get back to https://github.com/pytest-dev/pytest/pull/5313 first...

blueyed: ...it is not being combined though, so that should be good.

blueyed: Pulled that into the fix-pexpect PR (6896dbc).

PR #5307, opened 2019-05-24
pdb: import pdbcls lazily

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/2064.

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -49,42 +49,18 @@ def pytest_addoption(parser):
     )
 
 
-def _import_pdbcls(modname, classname):
-    try:
-        __import__(modname)
-        mod = sys.modules[modname]
-
-        # Handle --pdbcls=pdb:pdb.Pdb (useful e.g. with pdbpp).
-        parts = classname.split(".")
-        pdb_cls = getattr(mod, parts[0])
-        for part in parts[1:]:
-            pdb_cls = getattr(pdb_cls, part)
-
-        return pdb_cls
-    except Exception as exc:
-        value = ":".join((modname, classname))
-        raise UsageError("--pdbcls: could not import {!r}: {}".format(value, exc))
-
-
 def pytest_configure(config):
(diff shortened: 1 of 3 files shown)


testing/test_pdb.py
```diff
@@ -1157,12 +1157,13 @@ def runcall(self, *args, **kwds):
...
-    result.stderr.fnmatch_lines(
+    result.stdout.fnmatch_lines(
         [
-            "ERROR: --pdbcls: could not import 'really.invalid:Value': No module named *really*"
+            "*= FAILURES =*",
+            "E * --pdbcls: could not import 'really.invalid:Value': No module named *really*",
         ]
     )
-    assert result.ret == 4
+    assert result.ret == 1
```
blueyed: Note that this delays the UsageError now, which is not handled too well by pytest (e.g. via exitcode).
IIRC UsageErrors are only handled during config parsing/setup.

src/_pytest/debugging.py
```diff
@@ -144,7 +145,9 @@ def _init_pdb(cls, *args, **kwargs):
                     else:
                         tw.sep(">", "PDB set_trace")
 
-            class PytestPdbWrapper(cls._pdb_cls, object):
+            pdb_cls = cls._import_pdb_cls()
+
+            class PytestPdbWrapper(pdb_cls, object):
```
blueyed: I'd also like to move the PytestPdbWrapper class out of the method, likely into a method that gets it cached based on `pdb_cls` then (typically only one cache entry then).
This is good for less overhead in general, but allows for testing if the returned instance is the same as before (useful for pdbpp's global pdb handling).
This should be done before 5.0 to avoid conflicts in this area.

nicoddemus: Sure, sounds reasonable. I suspect this started as a small class so it was put inside the method for tersiness, but it has grown quite a lot since then.

Please do it right after this gets merged. 👍

blueyed: After https://github.com/pytest-dev/pytest/pull/5319.

PR #5312, opened 2019-05-25
logging: Improve formatting of multiline message

thisch:
This PR improves the formatting of captured and live log messages.

```
LOGGER.info("test1\nasdfdf")
LOGGER.warning("test2\nasdfkkk")
LOGGER.warning("single line")
````

Before this PR:
```
INFO     test_multiline:test_multiline.py:11 test1
asdfdf
WARNING  test_multiline:test_multiline.py:12 test2
asdfkkk
WARNING  test_multiline:test_multiline.py:13 single line
```

With this PR:
```
INFO     test_multiline:test_multiline.py:11 test1
                                             asdfdf
WARNING  test_multiline:test_multiline.py:12 test2
                                             asdfkkk
WARNING  test_multiline:test_multiline.py:13 single line
```

### DIFF
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -77,6 +77,36 @@ class ColoredLevelFormatter(logging.Formatter):
         return super(ColoredLevelFormatter, self).format(record)
 
 
+if not six.PY2:
+    # Formatter classes don't support format styles in PY2
+
+    class PercentStyleMultiline(logging.PercentStyle):
+        """A logging style with special support for multiline messages.
+
+        If the message of a record consists of multiple lines, this style
+        formats the message as if each line were logged separately.
+        """
+
+        @staticmethod
+        def _update_message(record_dict, message):
+            tmp = record_dict.copy()
+            tmp["message"] = message
+            return tmp
+
+        def format(self, record):
+            if "\n" in record.message:
+                lines = record.message.splitlines()
+                formatted = self._fmt % self._update_message(record.__dict__, lines[0])
+                # TODO optimize this by introducing an option that tells the
+                # logging framework that the indentation doesn't
+                # change. This allows to compute the indentation only once.
+                indentation = _remove_ansi_escape_sequences(formatted).find(lines[0])
+                lines[0] = formatted
+                return ("\n" + " " * indentation).join(lines)
+            else:
+                return self._fmt % record.__dict__
+
+
 def get_option_ini(config, *names):
     for name in names:
         ret = config.getoption(name)  # 'default' arg won't work as expected
@@ -444,6 +474,9 @@ class LoggingPlugin(object):
             )
         else:
             formatter = logging.Formatter(log_format, log_date_format)
+
+        if not six.PY2:
+            formatter._style = PercentStyleMultiline(formatter._style._fmt)
         return formatter
 
     def _setup_cli_logging(self):
--- a/testing/logging/test_formatter.py
+++ b/testing/logging/test_formatter.py
@@ -2,7 +2,9 @@
 import logging
 
 import py.io
+import six
 
+import pytest
 from _pytest.logging import ColoredLevelFormatter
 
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: testing/logging/test_formatter.py [hunk 1]
```diff
@@ -34,3 +36,31 @@ class option(object):
...
+        exc_info=False,
+    )
+    # this is called by logging.Formatter.format
+    record.message = record.getMessage()
+
+    style = PercentStyleMultiline(logfmt)
+    output = style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\n"
+        "dummypath                   10 INFO     line2"
```
blueyed: Hmm, I think it would be better for this to just be indented, but not have the prefix for the second line - it is rather confusing, since it looks like two separate log messages (although there is an indicator (`\n`)).

thisch: I also think it would be better to just indent this line. The problem is that it is not so easy to determine the indentation in a generic way.  A year ago I wrote [this](https://github.com/thisch/pytest/commit/4c846c8559bae7e9322d670928e5e90acbc02cfe), which  I don't really like.

thisch: Here is a simplified approach of my commit from last year:

```diff
diff --git a/src/_pytest/logging.py b/src/_pytest/logging.py
index 1f2d5e1d4..70ce15ad9 100644
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -17,6 +17,7 @@ from _pytest.pathlib import Path
 
 DEFAULT_LOG_FORMAT = "%(levelname)-8s %(name)s:%(filename)s:%(lineno)d %(message)s"
 DEFAULT_LOG_DATE_FORMAT = "%H:%M:%S"
+_ESCAPE_SEQ = re.compile(r'\x1b\[[\d;]+m')
 
 
 class ColoredLevelFormatter(logging.Formatter):
@@ -88,15 +89,13 @@ if not six.PY2:
 
         def format(self, record):
             if "\n" in record.message:
-                fmt = self._fmt
-                dct = record.__dict__
-
-                data_seq = [
-                    # fmt % {**dct, "message": line}
-                    fmt % self._update_message(dct, line)
-                    for line in record.message.splitlines()
-                ]
-                return "\n".join(data_seq)
+                lines = record.message.splitlines()
+                formatted = self._fmt % self._update_message(
+                    record.__dict__, lines[0])
+                indentation = (_ESCAPE_SEQ.sub('', formatted)
+                               .find(lines[0]))
+                lines[0] = formatted
+                return ("\n" + ' '*indentation).join(lines)
             else:
                 return self._fmt % record.__dict__
 
diff --git a/testing/logging/test_formatter.py b/testing/logging/test_formatter.py
index 2fa4b3157..0879875df 100644
--- a/testing/logging/test_formatter.py
+++ b/testing/logging/test_formatter.py
@@ -62,5 +62,5 @@ def test_multiline_message():
     output = style.format(record)
     assert output == (
         "dummypath                   10 INFO     Test Message line1\n"
-        "dummypath                   10 INFO     line2"
+        "                                        line2"
     )
```

blueyed: Doesn't look too bad to me..!

FWIW: I'm using loguru for my current project, where log messages end up on stdout/stderr instead - should check how it behaves with regard to multiple lines there.

src/_pytest/logging.py
```diff
@@ -427,6 +457,22 @@ def __init__(self, config):
...
+            log_format
+        ):
+            formatter = ColoredLevelFormatter(
+                create_terminal_writer(self._config), log_format, log_date_format
+            )
+        else:
+            formatter = logging.Formatter(log_format, log_date_format)
+
+        if not six.PY2:
+            formatter._style = PercentStyleMultiline(formatter._style._fmt)
```
nicoddemus: ~~Can you write a comment here please explaining why this is needed only for Python 3? I didn't look deep into it, but it feels surprising that Python 3 does not support this.~~

I suggest to write your skip reason here then: `Formatter classes don't support format styles in PY2` 😁

nicoddemus:
I've tried and it looks great, thanks @thisch!

Lothiraldan:
If you got broken test because you try to check if a multi-line log message is present in the capture log using `assert msg in caplog.text`, see https://github.com/iterative/dvc/pull/2082 workaround, TLDR use `caplog.messages` instead of `caplog.text`

PR #5313, opened 2019-05-27
[WIP] ci: tox/Travis: use pexpect for dep only

blueyed:
For https://github.com/pytest-dev/pytest/pull/5306 I would have to add
`testing/test_pytester.py` to the default pos args, but that would run
all other tests therein additionally.

This also adds pexpect tests to the MacOS job, which appears to have
caused trouble at least in the past, and likely needs fixing here before
this works.

TODO:

- [ ] might hang on py37-linux job

PR #5314, opened 2019-05-27
conftest: add uses_pexpect mark

blueyed:
Pulled out of https://github.com/pytest-dev/pytest/pull/5313 - is this something we want to have?

### DIFF
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -11,9 +11,10 @@ def pytest_collection_modifyitems(config, items):
     """
     fast_items = []
     slow_items = []
+    slowest_items = []
     neutral_items = []
 
-    slow_fixturenames = ("testdir",)
+    spawn_names = {"spawn_pytest", "spawn"}
 
     for item in items:
         try:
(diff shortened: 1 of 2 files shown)


testing/conftest.py
```diff
@@ -23,7 +24,12 @@ def pytest_collection_modifyitems(config, items):
             # (https://github.com/pytest-dev/pytest/issues/5070)
             neutral_items.append(item)
         else:
-            if any(x for x in fixtures if x in slow_fixturenames):
+            if "testdir" in fixtures and spawn_names.intersection(
+                item.function.__code__.co_names
+            ):
+                slow_items.append(item)
+                item.add_marker(pytest.mark.uses_pexpect)
+            elif any(x for x in fixtures if x in slow_fixturenames):
```
blueyed: Could be simplified given that `testdir` is the only slow fixture.

blueyed:
I think a generic feature like `-k using_fixture(foo)`, or `--using-fixture foo` would be better.

tox.ini
```diff
@@ -175,6 +175,7 @@ markers =
     baz
     # conftest.py reorders tests moving slow ones to the end of the list
     slow
+    uses_pexpect
```
nicoddemus: I would add a quick description here 👍

nicoddemus:
> --using-fixture foo would be better.

Yep, I've seen the need for this myself... should we open a new issue for it?

blueyed:
> Yep, I've seen the need for this myself... should we open a new issue for it?

Yeah.

tox.ini
blueyed: Well, it is rather self-explaining, isn't it?

blueyed:
Marking WIP to not merge it yet - there is no use for it currently really, given that we're likely to change the pexpect factor for CI (#5313, it would make sense to use `-m uses_pexpect` if jobs are being kept separate).

tox.ini
blueyed: Ok, done - also added a new "slowest_items" category.

PR #5315, opened 2019-05-27
Documentation: expansion of Monkeypatch to include mocked classes and dictionaries

EvanKepner:
Following up on #5250 I've added another expansion to the `monkeypatch` documentation:

1. Included the reference output from `pytest --fixtures` for `monkeypatch` to list methods.
2. All Python code-blocks are in `code-block` RST sections across the file.
3. Added conceptual scenarios for the major functions to the intro.
4. Modified the "simple example" to include broader explanatory comments.
5. Added "Monkeypatching returned objects: building mock classes" section.
6. Added "Monkeypatching dictionaries" section.
7. Applied changes from the `blacken-docs` `pre-commit` hook.

### PR Checklist

- [x] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>`.

### DIFF
--- a/doc/en/monkeypatch.rst
+++ b/doc/en/monkeypatch.rst
@@ -63,40 +63,38 @@ testing, you do not want your test to depend on the running user. ``monkeypatch`
 can be used to patch functions dependent on the user to always return a
 specific value.
 
-In this example, :py:meth:`monkeypatch.setattr` is used to patch ``os.path.expanduser``
-so that the known testing string ``"/abc"`` is always used when the test is run.
+In this example, :py:meth:`monkeypatch.setattr` is used to patch ``Path.home``
+so that the known testing path ``Path("/abc")`` is always used when the test is run.
 This removes any dependency on the running user for testing purposes.
 :py:meth:`monkeypatch.setattr` must be called before the function which will use
 the patched function is called.
-After the test function finishes the ``os.path.expanduser`` modification will be undone.
+After the test function finishes the ``Path.home`` modification will be undone.
 
 .. code-block:: python
 
     # contents of test_module.py with source code and the test
-    # os.path is imported for reference in monkeypatch.setattr()
-    import os.path
+    from pathlib import Path
 
 
     def getssh():
         """Simple function to return expanded homedir ssh path."""
-        return os.path.expanduser("~/.ssh")
+        return Path.home() / ".ssh"
 
 
     def test_getssh(monkeypatch):
-        # mocked return function to replace os.path.expanduser
-        # given a path, always return '/abc'
-        def mockreturn(path):
-            return "/abc"
+        # mocked return function to replace Path.home
+        # always return '/abc'
+        def mockreturn():
+            return Path("/abc")
 
-        # Application of the monkeypatch to replace os.path.expanduser
+        # Application of the monkeypatch to replace Path.home
         # with the behavior of mockreturn defined above.
-        monkeypatch.setattr(os.path, "expanduser", mockreturn)
+        monkeypatch.setattr(Path, "home", mockreturn)
 
-        # Calling getssh() will use mockreturn in place of os.path.expanduser
+        # Calling getssh() will use mockreturn in place of Path.home
         # for this test with the monkeypatch.
         x = getssh()
-        assert x == "/abc/.ssh"
-
+        assert x == Path("/abc/.ssh")
 
 Monkeypatching returned objects: building mock classes
 ------------------------------------------------------
--- a/doc/en/monkeypatch.rst
+++ b/doc/en/monkeypatch.rst
@@ -8,46 +8,217 @@ Sometimes tests need to invoke functionality which depends
 on global settings or which invokes code which cannot be easily
 tested such as network access.  The ``monkeypatch`` fixture
 helps you to safely set/delete an attribute, dictionary item or
-environment variable or to modify ``sys.path`` for importing.
+environment variable, or to modify ``sys.path`` for importing.
+
+The ``monkeypatch`` fixture provides these helper methods for safely patching and mocking
+functionality in tests:
+
+.. code-block:: python
+
+    monkeypatch.setattr(obj, name, value, raising=True)
+    monkeypatch.delattr(obj, name, raising=True)
+    monkeypatch.setitem(mapping, name, value)
+    monkeypatch.delitem(obj, name, raising=True)
+    monkeypatch.setenv(name, value, prepend=False)
+    monkeypatch.delenv(name, raising=True)
+    monkeypatch.syspath_prepend(path)
+    monkeypatch.chdir(path)
+
+All modifications will be undone after the requesting
+test function or fixture has finished. The ``raising``
+parameter determines if a ``KeyError`` or ``AttributeError``
+will be raised if the target of the set/deletion operation does not exist.
+
+Consider the following scenarios:
+
+1. Modifying the behavior of a function or the property of a class for a test e.g.
+there is an API call or database connection you will not make for a test but you know
+what the expected output should be. Use :py:meth:`monkeypatch.setattr` to patch the
+function or property with your desired testing behavior. This can include your own functions.
+Use :py:meth:`monkeypatch.delattr` to remove the function or property for the test.
+
+2. Modifying the values of dictionaries e.g. you have a global configuration that
+you want to modify for certain test cases. Use :py:meth:`monkeypatch.setitem` to patch the
+dictionary for the test. :py:meth:`monkeypatch.delitem` can be used to remove items.
+
+3. Modifying environment variables for a test e.g. to test program behavior if an
+environment variable is missing, or to set multiple values to a known variable.
+:py:meth:`monkeypatch.setenv` and :py:meth:`monkeypatch.delenv` can be used for
+these patches.
+
+4. Use :py:meth:`monkeypatch.syspath_prepend` to modify the system ``$PATH`` safely, and
+:py:meth:`monkeypatch.chdir` to change the context of the current working directory
+during a test.
+
 See the `monkeypatch blog post`_ for some introduction material
 and a discussion of its motivation.
 
 .. _`monkeypatch blog post`: http://tetamap.wordpress.com/2009/03/03/monkeypatching-in-unit-tests-done-right/
 
-
 Simple example: monkeypatching functions
 ----------------------------------------
 
-If you want to pretend that ``os.expanduser`` returns a certain
-directory, you can use the :py:meth:`monkeypatch.setattr` method to
-patch this function before calling into a function which uses it::
+Consider a scenario where you are working with user directories. In the context of
+testing, you do not want your test to depend on the running user. ``monkeypatch``
+can be used to patch functions dependent on the user to always return a
+specific value.
 
-    # content of test_module.py
+In this example, :py:meth:`monkeypatch.setattr` is used to patch ``os.path.expanduser``
+so that the known testing string ``"/abc"`` is always used when the test is run.
+This removes any dependency on the running user for testing purposes.
+:py:meth:`monkeypatch.setattr` must be called before the function which will use
+the patched function is called.
+After the test function finishes the ``os.path.expanduser`` modification will be undone.
+
+.. code-block:: python
+
+    # contents of test_module.py with source code and the test
+    # os.path is imported for reference in monkeypatch.setattr()
     import os.path
-    def getssh(): # pseudo application code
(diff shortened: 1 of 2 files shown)


doc/en/monkeypatch.rst
```diff
@@ -8,46 +8,222 @@ Sometimes tests need to invoke functionality which depends
...
+    monkeypatch.delitem(obj, name, raising=True)
+    monkeypatch.setenv(name, value, prepend=False)
+    monkeypatch.delenv(name, raising=True)
+    monkeypatch.syspath_prepend(path)
+    monkeypatch.chdir(path)
+
+All modifications will be undone after the requesting
+test function or fixture has finished. The ``raising``
+parameter determines if a ``KeyError`` or ``AttributeError``
+will be raised if the set/deletion operation has no target.
```
nicoddemus:
```suggestion
will be raised if the target of the set/deletion operation does not exist.
```

doc/en/monkeypatch.rst
```diff
@@ -8,46 +8,222 @@ Sometimes tests need to invoke functionality which depends
...
+2. Modifying the values of dictionaries e.g. you have a global configuration that
+you want to modify for certain test cases. Use :py:meth:`monkeypatch.setitem` to patch the
+dictionary for the test. :py:meth:`monkeypatch.delitem` can be used to remove items.
+
+3. Modifying environment variables for a test e.g. to test program behavior if an
+environment variable is missing, or to set multiple values to a known variable.
+:py:meth:`monkeypatch.setenv` and :py:meth:`monkeypatch.delenv` can be used for
+these patches.
+
+4. Use :py:meth:`monkeypatch.syspath_prepend` to add to the system ``$PATH`` and
```
nicoddemus:
```suggestion
4. Use :py:meth:`monkeypatch.syspath_prepend` to modify ``sys.path`` safely, and
```

doc/en/monkeypatch.rst
```diff
@@ -8,46 +8,222 @@ Sometimes tests need to invoke functionality which depends
...
     import os.path
-    def getssh(): # pseudo application code
-        return os.path.join(os.path.expanduser("~admin"), '.ssh')
 
-    def test_mytest(monkeypatch):
+
+    def getssh():
+        """Simple function to return expanded homedir ssh path from $USER"""
+        user = os.getenv("$USER")
+        return os.path.join(os.path.expanduser(f"~{user}"), ".ssh")
```
nicoddemus: To simplify this example, you can use:

```python
def getssh():
    """Simple function to return expanded homedir ssh path from $USER"""
    return os.path.expanduser("~/.ssh")
```

doc/en/monkeypatch.rst
```diff
@@ -8,46 +8,222 @@ Sometimes tests need to invoke functionality which depends
...
+The ``mock_get`` function returns an instance of the ``MockResponse`` class, which
+has a ``json()`` method defined to return a known testing dictionary and does not
+require any outside API connection.
+
+You can build the ``MockResponse`` class with the appropriate degree of complexity for
+the scenario you are testing. For instance, it could include an ``ok`` property that
+always returns ``True``, or return different values from the ``json()`` mocked method
+based on input strings.
+
+This mock can be shared across tests using the ``fixture`` syntax:
```
nicoddemus:
```suggestion
This mock can be shared across tests using a ``fixture``:
```

A fixture is not really a new syntax. 😁

doc/en/monkeypatch.rst
```diff
@@ -8,46 +8,222 @@ Sometimes tests need to invoke functionality which depends
...
+
+
+    # Notice our test uses the custom fixture instead of monkeypatch directly.
+    def test_get_json(mock_response):
+        result = app.get_json("https://fakeurl")
+        assert result["mock_key"] == "mock_response"
+
+
+Furthermore, if the mock was designed to be applied to all tests, the ``fixture`` could
+be moved to ``conftest.py`` with ``autouse=True``.
```
nicoddemus:
```suggestion
be moved to a ``conftest.py`` file and use the ``autouse=True`` option.
```

doc/en/monkeypatch.rst
```diff
@@ -90,7 +266,7 @@ to do this using the ``setenv`` and ``delenv`` method. Our example code to test:
 
 .. code-block:: python
 
-    # contents of our original code file e.g. code.py
+    # Contents of our original code file e.g. code.py.
```
nicoddemus: I would leave all those "contents -> Contents" changes lone, just for consistency with the rest of the docs.

doc/en/monkeypatch.rst
```diff
@@ -110,7 +286,7 @@ both paths can be safely tested without impacting the running environment:
 
 .. code-block:: python
 
-    # contents of our test file e.g. test_code.py
+    # Contents of our test file e.g. test_code.py.
```
nicoddemus: Also, if I'm being pedantic, we usually don't use `.` at the end of single line comments

doc/en/monkeypatch.rst
```diff
@@ -8,46 +8,222 @@ Sometimes tests need to invoke functionality which depends
...
+    # os.path is imported for reference in monkeypatch.setattr().
     import os.path
-    def getssh(): # pseudo application code
-        return os.path.join(os.path.expanduser("~admin"), '.ssh')
 
-    def test_mytest(monkeypatch):
+
+    def getssh():
+        """Simple function to return expanded homedir ssh path from $USER"""
+        user = os.getenv("$USER")
```
asottile: this looks up the environment variable named '$USER` -- I think you mean `os.getenv('USER')` (which could return `None`)

The `os.path.expanduser` call for "the current user" would just be:

```python
    return os.path.join(os.path.expanduser("~"), ".ssh")
```

no need for `getenv` or string formatting

doc/en/monkeypatch.rst
```diff
@@ -154,6 +330,112 @@ This behavior can be moved into ``fixture`` structures and shared across tests:
...
+        # Patch the DEFAULT_CONFIG t be missing the 'user' key.
+        monkeypatch.delitem(app.DEFAULT_CONFIG, "user", raising=False)
+
+        # Key error expected because a config is not passed, and the
+        # default is now missing the 'user' entry.
+        with pytest.raises(KeyError):
+            _ = app.create_connection_string()
+
+
+The modularity of the ``fixture`` syntax gives you the flexibility to define
```
nicoddemus:
```suggestion
The modularity of fixtures gives you the flexibility to define
```

doc/en/monkeypatch.rst
```diff
@@ -8,46 +8,222 @@ Sometimes tests need to invoke functionality which depends
...
 
-
 Simple example: monkeypatching functions
 ----------------------------------------
 
-If you want to pretend that ``os.expanduser`` returns a certain
-directory, you can use the :py:meth:`monkeypatch.setattr` method to
-patch this function before calling into a function which uses it::
+Consider a scenario where you are working with user directories. In the context of
+testing, you do not want your test to depend on the running user. ``Monkeypatch``
```
asottile: I think we should consistently match the case of the fixture name, so

```rst
    ``monkeypatch``
```

instead of

```rst
    ``Monkeypatch``
```

EvanKepner:
@nicoddemus @asottile appreciate the fast review! I made the updates, agreed with all suggestions :+1:

doc/en/monkeypatch.rst
EvanKepner: I changed this throughout the file with regard to the single line comments.

doc/en/monkeypatch.rst
EvanKepner: Great catch, also ensured `monkeypatch` is lowercase everywhere except when it is the first word in the headers.

doc/en/monkeypatch.rst
EvanKepner: Applied in https://github.com/pytest-dev/pytest/commit/7fa0e0405d4c051d176ba40ae4205dc3393e27a7

asottile:
test failures here are expected -- until #5063 is merged

doc/en/monkeypatch.rst
EvanKepner: Applied in https://github.com/pytest-dev/pytest/pull/5315/commits/60e4fd448e0ed05bd71436e506fafd4643b0a105

doc/en/monkeypatch.rst
EvanKepner: Applied in https://github.com/pytest-dev/pytest/commit/7fa0e0405d4c051d176ba40ae4205dc3393e27a7 - this is a really good simplification. I was debating using `Path.home()` but the original example used `os` so I wanted to stick with it.

doc/en/monkeypatch.rst
EvanKepner: Applied in https://github.com/pytest-dev/pytest/commit/7fa0e0405d4c051d176ba40ae4205dc3393e27a7

doc/en/monkeypatch.rst
EvanKepner: Applied in https://github.com/pytest-dev/pytest/commit/7fa0e0405d4c051d176ba40ae4205dc3393e27a7

doc/en/monkeypatch.rst
EvanKepner: Applied in https://github.com/pytest-dev/pytest/commit/7fa0e0405d4c051d176ba40ae4205dc3393e27a7 - also realized the second test does not require the `mock_test_database` fixture so I removed it from the signature - https://github.com/pytest-dev/pytest/commit/7fa0e0405d4c051d176ba40ae4205dc3393e27a7#diff-425daef754438812ec1a0e4d3aeaeff8L434

doc/en/monkeypatch.rst
nicoddemus: I still see comments starting with an upper case letter... would you mind changing all comments to start with lowercase? This is the pattern we use through the docs and code. 👍

blueyed: > I still see comments starting with an upper case letter... would you mind changing all comments to start with lowercase? This is the pattern we use through the docs and code. 

That's new to me also - I see that there are comments without punctuation and starting lowercased, but when using punctuation / full sentences they should be capitalized (like above).

nicoddemus: full sentences yes, but in one-liners like this we don't use punctuation and start with a lowercase letter (at least that's what I see browsing the docs quickly now)

blueyed: Well, you will also see no typing hints currently - but maybe soon.. ;)
(what I want to say is that style can be improved, and I prefer it like above also)

nicoddemus: Well I won't get into that discussion right now then, the current style is fine.

@EvanKepner could you just squash and rebase your commits then? Thanks!

EvanKepner: No problem at all, squash/rebase coming soon. Also @nicoddemus sorry I misread your original comment and thought you wanted all comment lines capitalized but without punctuation unless part of a larger sentence :sweat_smile: 

I'm a fan of generally strict type hints and full doc-strings for each function. It's a difficult challenge with documentation because you don't want to add additional syntax that could confuse a reader. These tutorial examples are simplified to show how the patching works so I steered towards the least verbose form. 

Might be worth an update in the guidance to https://github.com/pytest-dev/pytest/blob/master/CONTRIBUTING.rst#write-documentation if you come to a consensus on standardized documentation style beyond the pre-commit checks for `black` in the docs. If you're seeking suggestions I like one sentence per line in RST with a hard-wrap (or just a hard-wrap at something sensible) to make suggested changes and PR reviews easier.

EvanKepner:
Squashed and rebased to `master` at https://github.com/pytest-dev/pytest/commit/ece774f0eb0bb87aa37c8c34df13a18be1500bfb. This includes the prior edits, and I went through and lower-cased all comment lines unless they were part of a sentence.

EvanKepner:
Realized the change to `os.path.expanduser("~/.ssh")` actually caused the test example to fail, since `expanduser` was patched to return `"/abc"`. Added a commit to update to this to use `pathlib` instead which is more clear and confirmed that it runs correctly.

EvanKepner:
@nicoddemus checking to make sure there is no need for additional action from me. If more edits are required please let me know. Thanks!

nicoddemus:
Hi @EvanKepner,

Thanks for the ping, missed that this was ready. 👍

EvanKepner:
Excellent - thanks! No worries at all :+1:

PR #5322, opened 2019-05-28
pdb: move/refactor initialization of PytestPdbWrapper

blueyed:


src/_pytest/debugging.py
```diff
@@ -98,6 +99,11 @@ def _import_pdb_cls(cls):
         if not pdb_cls:
             return pdb.Pdb
 
+        if cls._wrapped_pdb_cls and cls._wrapped_pdb_cls[0] == pdb_cls:
+            return cls._wrapped_pdb_cls[1]
+
+        import _pytest.config
```
blueyed: This is used only for `tw = _pytest.config.create_terminal_writer(cls._config)` below.

src/_pytest/debugging.py
```diff
@@ -89,156 +90,161 @@ def _is_capturing(cls, capman):
...
+                    "--pdbcls: could not import {!r}: {}".format(value, exc)
+                )
 
-        modname, classname = pdb_cls
+        import _pytest.config
 
-        try:
-            __import__(modname)
-            mod = sys.modules[modname]
+        class PytestPdbWrapper(pdb_cls, object):
```
nicoddemus: Can we move this class creation to a module level function? This class in the middle of the code like this makes it harder to understand than it could be.

blueyed: Will move it into a classmethod.

blueyed:
@nicoddemus 
This might be good/better now, please review.

PR #5323, opened 2019-05-28
Correct some tiny typos in changelog

asottile:


nicoddemus:
Those failures are all due to the `importlib` changes in `pluggy` right?

Perhaps we should pin to `pluggy==0.10` on `master` to at least get green builds on `master` until 4.6 gets released?

asottile:
yeah sounds good, we'll need to make sure to revert that when merging master into features -- I can make a branch

PR #5325, opened 2019-05-28
Temporarily pin pluggy on master

asottile:


asottile:
> But we also appear to have new / flaky failures on features..?!

link? I haven't seen any

blueyed:
Only https://travis-ci.org/pytest-dev/pytest/builds/537992776 apparently (via https://travis-ci.org/pytest-dev/pytest/branches).  That is the pypy3-spamming one, so might be unrelated then.

:shipit:

asottile:
we'll need to make sure to revert this on features when we merge the two

nicoddemus:
Thanks @asottile!

PR #5331, opened 2019-05-29
tests: conftest: run slowest tests actually

blueyed:
Regression in 0025e4408.


blueyed:
(With regard to coverage we apparently do not need all the non-pexpect tests using "testdr", which are not explicitly marked as slow.)

blueyed:
re coverage: apparently it was noticed when merging the regression.

Issue #5301, opened 2019-05-23
"Last Failed" mode no longer reruns all tests on successful run


exhuma:
For about a month now (maybe less) our team is experiencing issues with the `--last-failed` mode. Since we regularly update our development dependencies there is a chance that it might be due to an issue in pytest. Unfortunately I don't have an example to reproduce the issue.

I will try downgrading to pytest 4.4 and see if this fixes the issue. If I don't forget it, I will update this issue if that's the case.

When using "last-failed" mode we expect the following behaviour:

* As long as there are failures, only run the failed tests on each run
* The run following the first run without any failures should re-run all tests

What we observe instead:

* As long as there are failures, only run the failed tests on each run
* The run following the first run without any failures *only some tests are re-run*. Only a fraction of tests are executed.

Workaround:

* Running `pytest` once without `--last-failed` solves the issue (but only once).

## pip list output

```
Package                       Version  Location
----------------------------- -------- ------------------------------
alabaster                     0.7.12
apipkg                        1.5
asn1crypto                    0.24.0
astroid                       2.2.5
atomicwrites                  1.3.0
attr                          0.3.1
attrs                         19.1.0
autopep8                      1.4.4
Babel                         2.6.0
bboss                         1.10.1
bcrypt                        3.1.6
certifi                       2019.3.9
cffi                          1.12.3
chardet                       3.0.4
colorama                      0.4.1
commonmark                    0.9.0
config-resolver               4.3.4
coverage                      4.5.3
cryptography                  2.6.1
distribute                    0.7.3
docutils                      0.14
entrypoints                   0.3
execnet                       1.6.0
flake8                        3.7.7
flake8-polyfill               1.0.2
future                        0.17.1
gouge                         1.3.0
idna                          2.8
imagesize                     1.1.0
isort                         4.3.20
jargon                        4.14.0
Jinja2                        2.10.1
lazy-object-proxy             1.4.1
mando                         0.6.4
MarkupSafe                    1.1.1
mccabe                        0.6.1
more-itertools                7.0.0
msgpack-python                0.5.6
mypy                          0.701
mypy-extensions               0.4.1
packaging                     19.0
paramiko                      2.4.2
pathlib2                      2.3.3
pexpect                       3.3
pip                           19.1.1
pkg-resources                 0.0.0
pluggy                        0.11.0
puresnmp                      1.5.1
py                            1.8.0
pyasn1                        0.4.5
pycodestyle                   2.5.0
pycparser                     2.19
[...]

PR #5333, opened 2019-05-29
Fix regression with --lf and non-selected failures

blueyed:
TODO:

- [x] changelog (might be https://github.com/pytest-dev/pytest/issues/5301)


nicoddemus:
This seems to have broken other tests

I agree with using #5301 for the CHANGELOG, if this fixes it

blueyed:
> This seems to have broken other tests

The initial fix was wrong.  Should be good/better now.

It is not clear to me if this is #5301 really, so added a changelog with the PR number instead.

Issue #5317, opened 2019-05-27
pypy and pypy3 Exception warnings upon closing


nicoddemus:
We get a lot of ignored exceptions when the interpreter closes:

```
ResourceWarning: unclosed file <_io.FileIO fd=1114 mode='rb' closefd=True>
Exception ignored in: <_io.FileIO fd=1115 mode='rb' closefd=True>
```

At first glance it seems the capture plugin is not closing files properly.

blueyed:
That are quite a lot - have you checked older logs?  I do not remember seeing that many, i.e. it might be a recent regression.

nicoddemus:
>  have you checked older logs

I haven't... indeed I don't recall seeing so many

blueyed:
Builds on master/features also started to fail: https://travis-ci.org/pytest-dev/pytest/branches

blueyed:
https://travis-ci.org/pytest-dev/pytest/builds/537998957 (master-into-features PR) was OK, let's see how its merge comes out.

blueyed:
As for the CI failures I suspect this is related to the lastest pluggy maybe?!  (at least some failures I've looked at are related to entrypoint-related tests).

asottile:
yes, master failures are expected

releasing 4.6 or merging this will fix `master`: https://github.com/pytest-dev/pytest/pull/5325

pierreglaser:
FYI, I am getting recent failures on `loky` using `pypy` with `pytest >= 4.6`. The root cause seems to be unclosed file descriptors, so it seems related to this issue.

asottile:
I found the root cause of this by the way: 

- issue: https://bitbucket.org/pypy/pypy/issues/3021/ioopen-directory-leaks-a-file-descriptor
- PR: https://bitbucket.org/pypy/pypy/pull-requests/647/fix-leak-of-file-descriptor-with-_iofileio/diff

the same thing is affecting pre-commit's CI as well :S

blueyed:
Awesome.

Can we remove pypy3 until there is some progress / fix in this regard?
Nobody is paying attention to real failures there anyway, and it slows down CI in general, especially now with it being run last due to allowed-failures.

nicoddemus:
> Can we remove pypy3 until there is some progress / fix in this regard?

I agree. We should test it again when a new `pypy3` release comes out.

PR #5334, opened 2019-05-29
WIP: debug pypy3 failures due to "Unroll calls to any #5062 (#5103)"

blueyed:
This reverts commit 2b9ca342809a4e05b4451a074d644c14acda844c, reversing
changes made to 0a57124063a4ed63f85fb8a39dc57f6d6fc5b132.

This is when pypy3 started to fail (https://travis-ci.org/pytest-dev/pytest/builds/537992776).

Ref: https://github.com/pytest-dev/pytest/issues/5317

/cc @Tadaboody


nicoddemus:
How about if instead of reverting this completely, we don't try to unroll calls to `any` in pypy3 instead? It seems to me this would make it easier to fix the problem than trying to reintroduce the entire feature later.

blueyed:
> How about if instead of reverting this completely

This is just for testing / investigating.  I do not want / plan to revert it actually (as-is).

nicoddemus:
Oh OK!

blueyed:
The build passed: https://travis-ci.org/pytest-dev/pytest/jobs/538946951
But IIRC it is rather flaky instead, due to something appearing on stderr.

blueyed:
From the warnings summary:

> .tox/pypy3-xdist/site-packages/_pytest/fixtures.py:1270
  /home/travis/build/pytest-dev/pytest/.tox/pypy3-xdist/site-packages/_pytest/fixtures.py:1270: ResourceWarning: unclosed file <_io.FileIO fd=445 mode='rb' closefd=True>
    marker = getfixturemarker(obj)

nicoddemus:
Thanks for the last commmit.

Argh this makes it harder to debug because I can't get `pypy3` to work on Windows. 😞

blueyed:
Good first shot though I guess.
Might be worth to check if there are warnings also in pypy already, and/or if this is really just maybe flaky due to "output on stderr" (and happened before, but not that much).

nicoddemus:
> Might be worth to check if there are warnings also in pypy already, and/or if this is really just maybe flaky due to "output on stderr" (and happened before, but not that much).

Hmmm interesting, we can try with `-s` next.

blueyed:
> Hmmm interesting, we can try with -s next.

`-s` appears to be pretty broken in general already (started debugging / looking into it in https://github.com/pytest-dev/pytest/pull/5337 - which is against master)

nicoddemus:
`-s` and xdist don't play nice together. We might try `pypy3` without xdist and only a single test file, to see if we can reproduce the issue with a smaller number of tests and shorter feedback. 🤔

blueyed:
I could imagine that rewriting might happen in a place where PyPy does some resource handling itself (https://github.com/pytest-dev/pytest/pull/5103/files#diff-a54d1eaf9d064e9a7d25b1152476bb3dR967).

Bisecting/debugging this quickly is not fun, since there are warnings also on master already.
(Would be easier now with enforced `--lsof` being in place before already)

blueyed:
Still appears to be flaky: https://travis-ci.org/pytest-dev/pytest/builds/539061916 ...
(via https://travis-ci.org/pytest-dev/pytest/branches)

nicoddemus:
*sigh* now travis won't start

nicoddemus:
It was because it had a merge conflict then.

.travis.yml
```diff
@@ -55,7 +55,7 @@ before_script:
       export _PYTEST_TOX_EXTRA_DEP=coverage-enable-subprocess
     fi
 
-script: tox -- testing/acceptance_test.py
+script: tox -- testing/test_capture.py
```
blueyed: In my manual testing it needs more, but `-m "not slow"` should do IIRC.

blueyed: Should be rebased for that though I guess.

nicoddemus: Trying `testing/python` now

But what do you suggest, `testing -m "not slow"`?

blueyed:
btw, just out of interest: could you use pypy3 in Docker maybe?
Also the nt._getfinalpathname should get reported/fixed probably.

.travis.yml
blueyed: Yes.. just `-m "not slow"` basically.

nicoddemus:
> btw, just out of interest: could you use pypy3 in Docker maybe?

I have one or two VMs laying around, I guess there's no other way but try this.

> Also the nt._getfinalpathname should get reported/fixed probably.

Definitely, didn't do that yesterday as it was late and I was about to clock out. Will do it later.

nicoddemus:
Spent the hour or so investigating this. Some observations:

1. I get warnings by just running `tox -e pypy3 -- testing/acceptance_test.py`.
2. If I add `--runpytest=subprocess`, I don't see any warnings. 🤔 

I will clock out now, might give this another try tomorrow. 

But I'm not confident we can find the problem anytime soon. The question is, should this block the 4.6 release? IMHO no; while it is unfortunate that we are getting more of those warnings now, it seems they affect mostly pytest's own test suite, so this is something we can deal later.

blueyed:
> should this block the 4.6 release?

No.
We should either allow failures there, or just remove it temporarily.

nicoddemus:
I agree with allowing failures. I believe you had a PR with that approach right? Could you resurrect it please? 😁

blueyed:
Closing for now.

PR #5336, opened 2019-05-29
Fix pexpect tests on MacOS

blueyed:


blueyed:
Should be good now, assuming that CI is green.

PR #5340, opened 2019-05-30
ci: Travis: add pypy3 to allowed failures temporarily

blueyed:
Ref: https://github.com/pytest-dev/pytest/pull/5334
Ref: https://github.com/pytest-dev/pytest/issues/5317

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -115,6 +115,9 @@ matrix:
   allow_failures:
     - python: '3.8-dev'
       env: TOXENV=py38-xdist
+    # Temporary (https://github.com/pytest-dev/pytest/pull/5334).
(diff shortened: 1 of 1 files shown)


blueyed:
Let's try to fix it instead first.

blueyed:
Re-opening: https://github.com/pytest-dev/pytest/pull/5334#issuecomment-497527688

nicoddemus:
@asottile we're all set for 4.6. Want to make that release? 😁

blueyed:
JFI: this causes Travis builds to become even slower, since it now gets run as the last job: https://travis-ci.org/pytest-dev/pytest/builds/539900223

PR #5343, opened 2019-05-30
pytester: remove unused winpymap

blueyed:
Follow-up to c86d2daf8.

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -153,15 +153,6 @@ class LsofFdLeakChecker(object):
             item.warn(pytest.PytestWarning("\n".join(error)))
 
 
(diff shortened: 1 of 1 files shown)


blueyed:
Also here pypy3-xdist was ok.

nicoddemus:
> Also here pypy3-xdist was ok.

🤔 

I wonder if `features` `pypy3` is fine, only failing on `master`. 

If this is the case, I would just go ahead and release 4.6, because afterwards we will merge `features` into `master` and the problem goes away. 😅

blueyed:
> I wonder if features pypy3 is fine, only failing on master.

No, it started failing on features - but now appears to be OK again?!
Have not investigated (i.e. not checked new commits to features really), but from this morning it looked like something else might have caused it (e.g. load on Travis?!).

Issue #5330, opened 2019-05-29
Collection warning emitted for modules which should not being collected


BeyondEvil:
When setting the `testpaths` key under the `[pytest]` section in `tox.ini`, pytest seems to ignore it.

See the comment thread in this [PR](https://github.com/pytest-dev/pytest-selenium/pull/220#issuecomment-497017048) for more information.

<details>
<summary>Result of `pip list`</summary>

```
appdirs              1.4.3                           
Appium-Python-Client 0.44                            
aspy.yaml            1.3.0                           
atomicwrites         1.3.0                           
attrs                19.1.0                          
black                19.3b0                          
certifi              2019.3.9                        
cfgv                 2.0.0                           
chardet              3.0.4                           
Click                7.0                             
entrypoints          0.3                             
filelock             3.0.12                          
flake8               3.7.7                           
identify             1.4.3                           
idna                 2.8                             
importlib-metadata   0.15                            
mccabe               0.6.1                           
more-itertools       7.0.0                           
nodeenv              1.3.3                           
pip                  19.1.1                          
pluggy               0.12.0                          
pre-commit           1.16.1                          
py                   1.8.0                           
pycodestyle          2.5.0                           
pyflakes             2.1.1                           
pytest               4.5.0                           
pytest-base-url      1.4.1                           
pytest-html          1.20.0                          
pytest-metadata      1.8.0                           
pytest-selenium      1.9.2.dev112+g77fed97.d20190529 /Users/jimbrannlund/dev/pytest-dev/pytest-selenium
pytest-variables     1.7.1                           
PyYAML               5.1                             
requests             2.22.0                          
selenium             3.141.0                         
setuptools           41.0.1                          
six                  1.12.0                          
toml                 0.10.0                          
tox                  3.12.1                          
urllib3              1.25.3                          
virtualenv           16.6.0                          
wcwidth              0.1.7                           
wheel                0.33.4                          
zipp                 0.5.1
```
</details>

python: 3.7.3
macOSX: 10.14.4

I see the same outcome if I run the same command as defined in tox.ini (`pytest -n auto -v -r a`) w/o tox.

pytest header seems to pickup the value correctly at least:
```
rootdir: /Users/<user>/dev/pytest-dev/pytest-selenium, inifile: tox.ini, testpaths: testing
```

@nicoddemus

nicoddemus:
Thanks @BeyondEvil, I will try to take a look when I have some time. 👍

blueyed:
[...]

PR #5346, opened 2019-05-30
Show test module in the PytestCollectionWarning message

nicoddemus:
Related to #5330

### DIFF
--- a/testing/python/collect.py
+++ b/testing/python/collect.py
@@ -146,7 +146,24 @@ class TestClass(object):
         result = testdir.runpytest("-rw")
         result.stdout.fnmatch_lines(
             [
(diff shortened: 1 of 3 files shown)


testing/python/collect.py
```diff
@@ -146,7 +146,24 @@ def __init__(self):
...
+                "*cannot collect test class 'TestClass1' because it has "
+                "a __init__ constructor (from: test_class_with_init_warning.py)"
+            ]
+        )
+
+    def test_class_with_new_warning(self, testdir):
+        testdir.makepyfile(
+            """
+            class TestClass1(object):
+                def __new__(self):
```
nicoddemus: Note: this was being tested, but on a test about `namedtuple` support. Decided to make a specific test for this instead.

PR #5362, opened 2019-06-02
Preparing release version 4.6.1

asottile:


asottile:
Hmm, we'll probably want to after-the-fact merge the CHANGELOG into master 🤔

nicoddemus:
Hmmm indeed. 👍

asottile:
> 
> 
> Hmmm indeed. 👍

I guess it's as easy as cherry-picking this patch -- so not a big deal, I'll add that to the HOWTORELEASE instructions

PR #5363, opened 2019-06-02
Update release instructions for 4.6-maintenance

asottile:
I'm going to run these and make sure they make sense

### DIFF
--- a/HOWTORELEASE.rst
+++ b/HOWTORELEASE.rst
@@ -12,6 +12,8 @@ taking a lot of time to make a new one.
 
 #. Create a branch ``release-X.Y.Z`` with the version for the release.
 
+   * **maintenance releases**: from ``4.6-maintenance``;
+
    * **patch releases**: from the latest ``master``;
 
    * **minor releases**: from the latest ``features``; then merge with the latest ``master``;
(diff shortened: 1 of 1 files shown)


HOWTORELEASE.rst
```diff
@@ -33,7 +36,7 @@ taking a lot of time to make a new one.
 
    Wait for the deploy to complete, then make sure it is `available on PyPI <https://pypi.org/project/pytest>`_.
 
-#. Merge the PR into ``master``.
+#. Merge the PR.
```
nicoddemus:
```suggestion
#. Merge the PR.

#. If this is a maintenance release, cherry-pick the commit that updates the CHANGELOG into `master`, while keeping the `changelog` files intact so they get included in the next release as well.
```

Hmmm this is a little annoying I guess... 🤔

asottile: I think I can put this in command form so it's a little easier 🤔

blueyed: Wouldn't/shouldn't changelog updates happen by merging maintenance into master?

asottile: maintenance is intentionally divergent from `master` / `features`.  We actually want all of the changelogs that happen in 4.6 to additionally be part of the 5.x releases

So what we ~basically want to do is cherry pick the changes to `CHANGELOG.rst` and to `doc/en/announce`

asottile:
updated the instructions and used them to create #5365

PR #5367, opened 2019-06-02
Fix typo about interpreters count in doc/en/example/parametrize.html

apollovy:


blueyed:
Canceled the Travis build - not necessary here.

nicoddemus:
Thanks @apollovy!

I also changed the interpreter list to list python 3 interpreters only. 👍

doc/en/example/multipython.py
```diff
@@ -9,7 +9,7 @@
 
 import pytest
 
-pythonlist = ["python2.7", "python3.4", "python3.5"]
+pythonlist = ["python3.5", "python3.6", "python3.7"]
```
blueyed: :+1:

PR #5368, opened 2019-06-02
Automated / semi-automated python3 upgrades

asottile:
I split this into separate commits to make it easier to review, rebase, and replay.


src/_pytest/assertion/rewrite.py
```diff
@@ -330,16 +328,16 @@ def _write_pyc(state, co, source_stat, pyc):
...
-        state.trace("error writing pyc file at %s: errno=%s" % (pyc, e.errno))
+        state.trace("error writing pyc file at {}: errno={}".format(pyc, e.errno))
         # we ignore any failure to write the cache file
         # there are many reasons, permission-denied, __pycache__ being a
         # file etc.
         return False
     return True
 
 
-RN = "\r\n".encode("utf-8")
```
nicoddemus: Not sure this is correct... doesn't `encode` use the default system locale?

blueyed:
Why is "coding" removed again?

asottile:
> 
> 
> Why is "coding" removed again? ([5e2e6b9](https://github.com/pytest-dev/pytest/commit/5e2e6b9288adeb7d7088d75962a4b8d96fcbf931))

https://www.python.org/dev/peps/pep-3120/

setup.py
```diff
@@ -1,11 +1,9 @@
-# -*- coding: utf-8 -*-
 from setuptools import setup
 
 # TODO: if py gets upgrade to >=1.6,
 #       remove _width_of_current_line in terminal.py
 INSTALL_REQUIRES = [
     "py>=1.5.0",
-    "six>=1.10.0",
```
nicoddemus: Thanks `six` for your service. 🙇

.pre-commit-config.yaml
```diff
@@ -7,17 +7,18 @@ repos:
         args: [--safe, --quiet]
         language_version: python3
```
nicoddemus: Didn't we have a pre-commit hook for the `encoding` comments? 🤔

asottile: yep, and now it is enforcing that they are removed `fix-encoding-pragma` w/ `args: [--remove]`

src/_pytest/assertion/rewrite.py
asottile: in python3+ it uses UTF-8 if no encoding is specified: https://github.com/asottile/pyupgrade/issues/138

setup.py
asottile: o7

/me pours one 🍺 out

testing/code/test_source.py
```diff
@@ -113,7 +113,7 @@ def test_source_strip_multiline():
 def test_syntaxerror_rerepresentation():
     ex = pytest.raises(SyntaxError, _pytest._code.compile, "xyz xyz")
     assert ex.value.lineno == 1
-    assert ex.value.offset == 7
```
asottile: damnit lol

```diff
commit 15669e81310e80018a86ed96678992e575f2f537 (HEAD -> pyupgrade)
Author: Anthony Sottile <asottile@umich.edu>
Date:   Sun Jun 2 18:02:14 2019 -0700

    Allow multiple positions for the SyntaxError in pypy3.6

diff --git a/testing/code/test_source.py b/testing/code/test_source.py
index 72bc628ab..a12a102a0 100644
--- a/testing/code/test_source.py
+++ b/testing/code/test_source.py
@@ -113,7 +113,7 @@ def test_source_strip_multiline():
 def test_syntaxerror_rerepresentation():
     ex = pytest.raises(SyntaxError, _pytest._code.compile, "xyz xyz")
     assert ex.value.lineno == 1
-    assert ex.value.offset == 7
+    assert ex.value.offset in {5, 7}  # cpython: 7, pypy3.6 7.1.1: 5
     assert ex.value.text.strip(), "x x"
 
 
```

asottile: actually, I'm going to push mine since it has more context 🤷‍♂

nicoddemus:
I don't understand why the coverage is reporting uncovered lines... for example, it says here:

https://codecov.io/gh/pytest-dev/pytest/compare/28ac469eaa7206a0bf1439b31ed187beb3598522...15669e81310e80018a86ed96678992e575f2f537/changes

That `test_tmpdir_fixture` was not executed at all, which doesn't really make sense... 🤔

asottile:
I'm always suspicious of any data that `codecov` produces, in this case that test definitely ran: https://travis-ci.org/pytest-dev/pytest/jobs/540538554 🤔

nicoddemus:
Hmmm indeed, but I noticed that `PYTEST_COVERAGE=1` was not set for that build. I will enable it for pure `py37` runs, let's see.

nicoddemus:
Rebased on `master` (hope you don't mind @asottile) and fixed linting.

asottile:
no problem, though I would have replayed the commits myself so that ended up in the right one :man_shrugging:

nicoddemus:
OK I think we are good. Can you go ahead and merge @asottile if you are OK with the PR?

.travis.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         - test $(python -c 'import sys; print("%d%d" % sys.version_info[0:2])') = 37
 
     # Full run of latest supported version, without xdist.
-    - env: TOXENV=py37
+    - env: TOXENV=py37  PYTEST_COVERAGE=1
```
blueyed: We do not need this, do we?
We have py37-xdist with coverage already (at least).

blueyed: I'd guess it was the rebase that fixed it.

We can revert this and this - after things stabilized a bit.. :)

PR #5378, opened 2019-06-03
Revert "ci: Travis: add pypy3 to allowed failures temporarily"

blueyed:
Reverts pytest-dev/pytest#5377

Shouldn't be necessary with the unrolling of all being removed... - and might serve as an indication when trying to get it in again (since this started failing with it).


blueyed:
I had assigned this to me, because I wanted to investigate / verify that this is OK.
Which does not appear to be the case: https://travis-ci.org/pytest-dev/pytest/builds/541022238

PR #5386, opened 2019-06-04
[4.6] Remove --recreate from .travis.yml (#5384)

asottile:
Remove --recreate from .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -130,7 +130,7 @@ before_script:
       export _PYTEST_TOX_EXTRA_DEP=coverage-enable-subprocess
     fi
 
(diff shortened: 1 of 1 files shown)


blueyed:
Isn't it a no-op basically?  (we have `cache: no`)
Why the backport?

asottile:
:man_shrugging: I figure why not - - and potentially reduce conflicts

nicoddemus:
FTR I don't think this has any practical effect on Travis, given that Travis runs on fresh VMs so all environments are created from scratch anyway. 🤷‍♂

PR #5396, opened 2019-06-04
Revert "Enable coverage for 'py37' environment"

blueyed:
This reverts commit 6d393c5dc8bfc9e437f85dae83914004409b1bc2.

It should not be necessary, because we have it via other jobs already.


blueyed:
Looks like subprocess coverage is not working on Azure.

blueyed:
Should be fixed now.

blueyed:
This has still unexpected changes, taken the azure fix out: #5403

Issue #5392, opened 2019-06-04
pkg_resources.NullProvider since pytest 4.6.0+


tobywf:
When using a pytest version 4.6.0 or up, Jinja's `PackageLoader` breaks with "NotImplementedError: Can't perform this operation for unregistered loader type". This is because `pkg_resources` returns a `NullProvider`.

Minimal working example:

```
python3 -m venv env && source env/bin/activate
pip install 'pytest>=4.6.0' 'jinja2==2.10.1'
mkdir -p tests/templates
touch tests/__init__.py
echo "hello, world" > tests/templates/test.txt
cat << EOF > tests/test_jinja.py
from jinja2 import Environment, PackageLoader

def test_package_loader():
    env = Environment(loader=PackageLoader(__name__))
    assert "hello" in env.get_template("test.txt").render()
EOF
pytest tests/
pip install 'pytest==4.5.0'
pytest tests/
echo "done"
```

With `pytest>=4.6.0` (confirmed with `4.6.0`, `4.6.1`, and `4.6.2`), it fails with:
```
self = <pkg_resources.NullProvider object at 0x10c2d69b0>, path = '/workplace/tobflem/pytest/tests/templates/test.txt'

    def _has(self, path):
        raise NotImplementedError(
>           "Can't perform this operation for unregistered loader type"
        )
E       NotImplementedError: Can't perform this operation for unregistered loader type

env/lib/python3.7/site-packages/pkg_resources/__init__.py:1457: NotImplementedError
```
After pytest is pinned to `4.5.0` again, no failure is observed.

There was an issue way back, #366, which seems to be quite similar, where "`__loader__` is being wrapped in an AssertionRewritingHook". Could be a similar issue?

Hope this is enough info, let me know if you need more!

tobywf:
Pretty sure it's to do with commit 13f02af97d676bdf7143aa1834c8898fbf753d87, where `_register_with_pkg_resources` was removed from `AssertionRewritingHook` in `src/_pytest/assertion/rewrite.py`. When I add that back in, the issue goes away.

blueyed:
> Environment(loader=PackageLoader(__name__))

Do you have to use this?

It seems like you are using a different loader then in your real code already probably, and should probably use an `Environment` from your real app?

tobywf:
[...]

PR #5397, opened 2019-06-05
Reinstate AssertionRewritingHook._register_with_pkg_resources

tobywf:
As discussed in #5392, restore `_register_with_pkg_resources` removed in 13f02af97d676bdf7143aa1834c8898fbf753d87 so ``has_resource`` for ``pkg_resources`` providers doesn't fail in tests. Haven't changed the code otherwise, comments and docstring are from the original.

Added a unit test that fails without the fix, and passes with it. Added a changelog entry, might not be required for such a trivial revert?

### DIFF
--- a/changelog/5392.bugfix.rst
+++ b/changelog/5392.bugfix.rst
@@ -1,0 +1,1 @@
+Reinstate ``AssertionRewritingHook._register_with_pkg_resources`` so ``has_resource`` for ``pkg_resources`` providers doesn't fail in tests.
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -51,6 +51,7 @@ class AssertionRewritingHook:
         self.session = None
         self.modules = {}
         self._rewritten_names = set()
+        self._register_with_pkg_resources()
         self._must_rewrite = set()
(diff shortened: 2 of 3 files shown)


blueyed:
I think this removes (at least some of) the performance gain this brought us.

asottile:
> I think this removes (at least some of) the performance gain this brought us.

yeah this would make the performance worse than 4.5.x since we'd gain back all of `pkg_resources` plus this new thing

I didn't know this would break in this way, but I'd suggest switching to `importlib.resources` (`importlib-resources` backport) for your tests

blueyed:
> I'd suggest switching to `importlib.resources` (`importlib-resources` backport) for your tests

Is that something Jinja2 should/could do/use already?

asottile:
> 
> 
> > I'd suggest switching to `importlib.resources` (`importlib-resources` backport) for your tests
> 
> Is that something Jinja2 should/could do/use already?

I think so, I'll see if I can cook one up for them 👍

asottile:
actually, looks like someone is already on it! https://github.com/pallets/jinja/issues/970

blueyed:
Awesome - although it looks like a helping hand is needed: https://github.com/pallets/jinja/pull/976

PR #5393, opened 2019-06-04
item.obj is again a bound method on TestCase function items

nicoddemus:
Fix #5390

PR #5399, opened 2019-06-05
[4.6] Merge pull request #5393 from nicoddemus/unittest-self-5390

asottile:
item.obj is again a bound method on TestCase function items


nicoddemus:
Thanks for picking this up. 👍

PR #5408, opened 2019-06-05
[4.6] tests: restore tracing function

blueyed:
Without this, `testing/test_pdb.py` (already without pexpect) will cause
missing test coverage afterwards (for the same process).

### DIFF
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -1,5 +1,20 @@
 # -*- coding: utf-8 -*-
+import sys
+
 import pytest
+
+if sys.gettrace():
+
+    @pytest.fixture(autouse=True)
+    def restore_tracing():
+        """Restore tracing function (when run with Coverage.py).
+
+        https://bugs.python.org/issue37011
+        """
+        orig_trace = sys.gettrace()
+        yield
+        if sys.gettrace() != orig_trace:
+            sys.settrace(orig_trace)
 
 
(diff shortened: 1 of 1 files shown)


testing/conftest.py
```diff
@@ -1,6 +1,21 @@
 # -*- coding: utf-8 -*-
+import sys
+
 import pytest
 
+if sys.gettrace():
```
asottile: hmmm I would have put this in the function `and orig_trace is not None`

blueyed: Well, this way it avoids using the fixture when not run with coverage (i.e. no initial tracing).
(not sure I understand your suggestion though)

nicoddemus: I agree it seems better to not have this active without need.

nicoddemus:
I will go ahead and merge this as the original PR is already on master. 👍

blueyed:
> I will go ahead and merge this as the original PR is already on master. ￼

Thanks.  We can still improve/change it though (possibly without a backport), if @asottile would like to have it.

nicoddemus:
> We can still improve/change it though (possibly without a backport), if @asottile would like to have it.

Definitely!

I just went ahead with the merge because IMO we should try to make the backports as close to the original as possible. Improvements suggested during the backport PR should happen in another round of PRs if they apply to the original patch.

PR #4849, opened 2019-02-27
ci: remove numpy/trial envs

blueyed:
Just test/install this with all envs.

Ref: https://github.com/pytest-dev/pytest/pull/4848#issuecomment-467909811

PR #5409, opened 2019-06-05
ci: optimize twisted/numpy related jobs

blueyed:
- tox: use twisted as dep only
- Azure: move twisted/numpy to main py37 job
- Travis: move twisted to main py37 build

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,7 +28,7 @@ jobs:
         tox.env: 'py36-xdist'
       py37:
         python.version: '3.7'
-        tox.env: 'py37'
+        tox.env: 'py37-twisted-numpy'
         # Coverage for:
         # - _py36_windowsconsoleio_workaround (with py36+)
         # - test_request_garbage (no xdist)
@@ -36,9 +36,6 @@ jobs:
       py37-linting/docs/doctesting:
         python.version: '3.7'
         tox.env: 'linting,docs,doctesting'
-      py37-twisted/numpy:
-        python.version: '3.7'
-        tox.env: 'py37-twisted,py37-numpy'
       py37-pluggymaster-xdist:
         python.version: '3.7'
         tox.env: 'py37-pluggymaster-xdist'
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,7 @@ commands =
     coverage: coverage report
 passenv = USER USERNAME COVERAGE_* TRAVIS PYTEST_ADDOPTS
 setenv =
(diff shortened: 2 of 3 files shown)


azure-pipelines.yml
```diff
@@ -28,19 +28,16 @@ jobs:
...
-        # - _py36_windowsconsoleio_workaround (with py36+)
-        # - test_request_garbage (no xdist)
-        PYTEST_COVERAGE: '1'
       py37-linting/docs/doctesting:
         python.version: '3.7'
         tox.env: 'linting,docs,doctesting'
       py37-twisted/numpy:
         python.version: '3.7'
-        tox.env: 'py37-twisted,py37-numpy'
+        tox.env: 'py37-twisted-numpy'
```
nicoddemus: This PR looks good, but how about if we go further and just test `numpy` and `twisted` in the main `py37` job? It is after all just a matter of adding some extra dependencies and running a few extra tests, it seems it would be a win overall.

blueyed: That's what this does :) - but I will rename it.

Will also (re)move pexpect on Travis.

nicoddemus: > That's what this does :) - but I will rename it.

Oh indeed. 

But I meant to actually install `numpy` and `twisted` in all test envs as part of the normal build process (just like we install `mock` for all test envs).

blueyed: I think we tried this before, without much luck.

Also I do not like the added overhead of it.

nicoddemus: Which overhead you mean? I believe we will get a slight overhead during installation, but otherwise the extra tests should have a negligible cost.

blueyed: Yes, installation mostly.
In general I still do not like that our CI runs so much, but this PR makes it better, and also of course having dropped py27.. :)
I do not feel like we should slow it down again/more - but IIRC it is not even possible easily (since we have tried something like that).

blueyed: https://github.com/pytest-dev/pytest/pull/4849

nicoddemus: Thanks for the link. 👍

tox.ini
```diff
@@ -39,8 +39,6 @@ setenv =
     pexpect: _PYTEST_TOX_PLATFORM=linux|darwin
     pexpect: _PYTEST_TOX_POSARGS_PEXPECT=-m uses_pexpect
```
blueyed: Should be removed likely.

PR #5391, opened 2019-06-04
Fix verbosity bug in --collect-only

thisch:
Closes: #5383

- [ ] Target the `master` branch for bug fixes, documentation updates and trivial changes.
- [ ] Target the `features` branch for new features and removals/deprecations.
- [ ] Include documentation when adding new features.
- [ ] Include new tests or update existing tests when applicable.

Unless your change is trivial or a small documentation fix (e.g.,  a typo or reword of a small section) please:

- [ ] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>.<TYPE>.rst`. See [changelog/README.rst](https://github.com/pytest-dev/pytest/blob/master/changelog/README.rst) for details.
- [ ] Add yourself to `AUTHORS` in alphabetical order;

PR #5411, opened 2019-06-05
[4.6] Fix verbosity bug in --collect-only (backport of #5391)

nicoddemus:


nicoddemus:
How about if those who merge a backport PR also apply the "backported" label, as it is easier for them. What do you guys think?

asottile:
oh yeah good call, my bad

nicoddemus:
No worries at all, just realized that and thought I would mention it. 👍

PR #5415, opened 2019-06-06
small mypy fixes

asottile:
mypy is very far from passing, but it did point out a few small cleanups that were doable

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -81,19 +81,12 @@ def _format_lines(lines):
     return result
 
 
-# Provide basestring in python3
-try:
-    basestring = basestring
-except NameError:
-    basestring = str
-
-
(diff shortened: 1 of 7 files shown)


blueyed:
Great.
Would be nice to have some mypy config already in setup.cfg maybe, or share it here.

asottile:
agreed, but even with no config there's 1000+ errors right now, I hope to reduce them slowly, eventually

I was using `pre-commit try-repo https://github.com/pre-commit/mirrors-mypy` and then fixing the crashes

blueyed:
`mypy --check-untyped-defs --ignore-missing-imports src/` has 275 issues "only".. ;)

asottile:
I was also looking at `testing` since it's useful to validate those calls as well, I might've also had `--check-untyped-defs` as well I forget

PR #5419, opened 2019-06-07
ci: Travis: remove pypy3 job for now

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5317#issuecomment-499019928


blueyed:
I think it is good enough to have https://github.com/pytest-dev/pytest/issues/5317 for keeping track of it, and I like reverting better than (un)commenting.

PR #5421, opened 2019-06-07
Link deprecation docs pytest.raises 'message' warning

nicoddemus:
As commented in https://github.com/pytest-dev/pytest/issues/3974#issuecomment-499870914


ndevenish:
Looks good - my one (minor) concern was linking to `/latest/` - but it looks like you've accounted for that and old deprecations are kept listed on that page even after removal.

*although* it does look as though the anchor links change (everything before "Removed Features" has a `deprecations.html#named-title-anchor` whereas after has `#id<num>`), this is probably just a manual page-generation issue? And anyway, getting onto the right page is probably good enough.

From unfortunate experience, I know that some people remain stuck to old version for quite some time, especially if v5 is dropping python 2.7.

nicoddemus:
> my one (minor) concern was linking to /latest/ 

Great catch. I've changed the link to point to the `4.6-maintenance` release docs instead. 👍

PR #5427, opened 2019-06-09
Minor:  tox: coverage: use -m with coverage-report

blueyed:


tox.ini
```diff
@@ -20,7 +20,7 @@ envlist =
 commands =
     {env:_PYTEST_TOX_COVERAGE_RUN:} pytest {posargs:{env:_PYTEST_TOX_DEFAULT_POSARGS:}}
     coverage: coverage combine
-    coverage: coverage report
+    coverage: coverage report -m
```
asottile: or we could add `show_missing = true` into `.coveragerc` 🤔

blueyed: Yes, sounds good - although it cannot be disabled/skipped then AFAIK (i.e. if you are using `coverage report` yourself).

blueyed:
Hmm, py37 on Azure failed to upload coverage:
```
…
TOTAL                                      25503   1335   4967    345    94%
+ bash /dev/fd/63 -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml
++ curl -s https://codecov.io/bash
##[section]Finishing: Report and upload coverage
```

Restarted.

PR #5435, opened 2019-06-11
Preparing release version 4.6.3

asottile:
figured we should get these bugfixes out, it costs us basically nothing to release :man_shrugging:


asottile:
50 minutes and counting for the macos build!

PR #5437, opened 2019-06-11
re-enable pypy3

### LINKED ISSUE #5317: pypy and pypy3 Exception warnings upon closing

asottile:
I pushed a workaround into importlib-metadata which avoids the call causing the ResourceWarning

https://gitlab.com/python-devs/importlib_metadata/merge_requests/75

Resolves #5317

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,10 +87,6 @@ matrix:
   allow_failures:
     - python: '3.8-dev'
       env: TOXENV=py38-xdist
-    # failing due to ResourceWarning exceptions:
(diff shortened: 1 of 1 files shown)


nicoddemus:
Sorry for the delay, great work!

PR #5469, opened 2019-06-23
Disallow abbreviated command-line options

### LINKED ISSUE #1149: drop argparse since its broken by design and unfixable

### LINKED ISSUE #3413: add_option no longer works because of ambiguous option

### LINKED ISSUE #4009: If a command line argument is a directory that contains a tox.ini file, that tox file is used

Zac-HD:
This was previously allowed where the ArgumentParser thought it was unambiguous, because this could be incorrect due to delayed parsing of options for plugins.

Fixes #1149, fixes #3413, and fixes #4009 - it's been a known problem for quite a while, but this fix is only available on Python 3.5+... which means we can finally do it!  Replaces #5467, in that it *also* defines `MyOptionParser._parse_optional`, in order to work around [an upstream bug that disables combining short options, such as `-vv`](https://bugs.python.org/issue26967).  

**We should avoid merging this until python/cpython#14316 is merged** and we know that we don't need to do our conditional backport on 3.8.

### DIFF
--- a/src/_pytest/config/argparsing.py
+++ b/src/_pytest/config/argparsing.py
@@ -1,5 +1,7 @@
 import argparse
+import sys
 import warnings
+from gettext import gettext
 
 import py
 
@@ -328,6 +330,7 @@ class MyOptionParser(argparse.ArgumentParser):
             usage=parser._usage,
             add_help=False,
             formatter_class=DropShorterLongHelpFormatter,
+            allow_abbrev=False,
         )
         # extra_info is a dict of (param -> value) to display if there's
         # an usage error to provide more contextual information to the user
@@ -354,6 +357,42 @@ class MyOptionParser(argparse.ArgumentParser):
                     self.error("\n".join(lines))
             getattr(args, FILE_OR_DIR).extend(argv)
         return args
+
+    if sys.version_info[:2] < (3, 8):  # pragma: no cover
+        # Backport of https://github.com/python/cpython/pull/14316 so we can
+        # disable long --argument abbreviations without breaking short flags.
+        def _parse_optional(self, arg_string):
+            if not arg_string:
+                return None
+            if not arg_string[0] in self.prefix_chars:
+                return None
+            if arg_string in self._option_string_actions:
+                action = self._option_string_actions[arg_string]
+                return action, arg_string, None
+            if len(arg_string) == 1:
+                return None
+            if "=" in arg_string:
+                option_string, explicit_arg = arg_string.split("=", 1)
+                if option_string in self._option_string_actions:
+                    action = self._option_string_actions[option_string]
+                    return action, option_string, explicit_arg
+            if self.allow_abbrev or not arg_string.startswith("--"):
+                option_tuples = self._get_option_tuples(arg_string)
+                if len(option_tuples) > 1:
+                    msg = gettext(
+                        "ambiguous option: %(option)s could match %(matches)s"
+                    )
+                    options = ", ".join(option for _, option, _ in option_tuples)
+                    self.error(msg % {"option": arg_string, "matches": options})
+                elif len(option_tuples) == 1:
+                    option_tuple, = option_tuples
+                    return option_tuple
+            if self._negative_number_matcher.match(arg_string):
+                if not self._has_negative_number_optionals:
+                    return None
+            if " " in arg_string:
+                return None
+            return None, arg_string, None
 
 
 class DropShorterLongHelpFormatter(argparse.HelpFormatter):
--- a/changelog/1149.removal.rst
+++ b/changelog/1149.removal.rst
@@ -1,0 +1,7 @@
+Pytest no longer accepts prefixes of command-line arguments, for example
+typing ``pytest --doctest-mod`` inplace of ``--doctest-modules``.
+This was previously allowed where the ``ArgumentParser`` thought it was unambiguous,
+but this could be incorrect due to delayed parsing of options for plugins.
+See for example issues `#1149 <https://github.com/pytest-dev/pytest/issues/1149>`__,
+`#3413 <https://github.com/pytest-dev/pytest/issues/3413>`__, and
+`#4009 <https://github.com/pytest-dev/pytest/issues/4009>`__.
--- a/extra/get_issues.py
+++ b/extra/get_issues.py
@@ -74,7 +74,7 @@ def report(issues):
 if __name__ == "__main__":
     import argparse
 
-    parser = argparse.ArgumentParser("process bitbucket issues")
+    parser = argparse.ArgumentParser("process bitbucket issues", allow_abbrev=False)
     parser.add_argument(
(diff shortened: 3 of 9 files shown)


Zac-HD:
Closing and reopening, in hopes of getting codecov unstuck:

changelog/1149.feature.rst
```diff
@@ -0,0 +1,4 @@
+Pytest no longer accepts prefixes of command-line arguments.
+This was previously allowed where the ``ArgumentParser`` thought it was unambigious,
+because this could be incorrect due to delayed parsing of options for plugins.
+See for example issues #1149, #3413, and #4009.
```
nicoddemus:
```suggestion
See for example issues `#1149 <https://github.com/pytest-dev/pytest/issues/1149>`__, `#3413 <https://github.com/pytest-dev/pytest/issues/3413>`__, and `#4009 <https://github.com/pytest-dev/pytest/issues/4009>`__.
```

changelog/1149.feature.rst
```diff
@@ -0,0 +1,4 @@
+Pytest no longer accepts prefixes of command-line arguments.
```
nicoddemus:
```suggestion
Pytest no longer accepts abbreviation of command-line arguments, for example typing ``pytest --doctest-mod`` inplace of ``--doctest-modules``.
```

changelog/1149.feature.rst
```diff
@@ -0,0 +1,4 @@
...
+This was previously allowed where the ``ArgumentParser`` thought it was unambigious,
```
nicoddemus:
```suggestion

This was previously allowed where the ``ArgumentParser`` thought it was unambigious,

```

src/_pytest/config/argparsing.py
```diff
@@ -355,6 +357,43 @@ def parse_args(self, args=None, namespace=None):
             getattr(args, FILE_OR_DIR).extend(argv)
         return args
 
+    def _parse_optional(self, arg_string):  # pragma: no cover
+        # This method is entirely copied from `argparse.ArgumentParser`, with a
+        # one-line bugfix for https://bugs.python.org/issue26967 so we can disable
```
nicoddemus: It would be nice to contribute this back upstream. 😁 

How can we get some kind of check where this blows up loudly when the problem no longer exists upstream?

How about:

```python
    py_ver = sys.version_info[:2]    
    assert py_ver < (3, 9), "check if this is still required in Python {}.{}".format(*py_ver)
```

This will warn us to check when pytest starts to get tested in 3.9. When the time comes, the outcome might need to change this a more standard version guard like we use in other parts of the code:

```python
if sys.version_info[:2] < (3, 9):
    def _parse_optional(self, arg_string):  # pragma: no cover
```

Or bump the version of the `assert` check.

changelog/1149.feature.rst
nicoddemus: I think this should be a "removal" changelog, as we are dropping a "feature" (as much as it was considered a misfeature).

changelog/1149.feature.rst
```diff
@@ -0,0 +1,4 @@
...
+because this could be incorrect due to delayed parsing of options for plugins.
```
nicoddemus:
```suggestion
but this could be incorrect due to delayed parsing of options for plugins.
```

src/_pytest/config/argparsing.py
Zac-HD: I opened python/cpython#14316 yesterday 😄 

There's a fairly good chance that this will also land in 3.8, so I suggest we just hold off on merging until we know, and then do exactly the conditional definition you suggest.

src/_pytest/config/argparsing.py
```diff
@@ -355,6 +358,42 @@ def parse_args(self, args=None, namespace=None):
             getattr(args, FILE_OR_DIR).extend(argv)
         return args
 
+    if sys.version_info[:2] < (3, 8):  # pragma: no cover
```
nicoddemus: I suppose this is enough because we use `-vv` (for example) in our tests, which would blow up without the patch correct?

Also, I thought this fix was not going into 3.8, am I incorrect?

nicoddemus:
Great work! 👍

src/_pytest/config/argparsing.py
Zac-HD: Yes, along with -ra and a couple of others... all of which failed when I did the naive version in my last PR!

My upstream PR targets master (3.9), but will almost certainly be backported to 3.8. TBH I think we could merge this now and just fix the condition if the backport somehow falls through!

nicoddemus: > TBH I think we could merge this now and just fix the condition if the backport somehow falls through!

Agree, just trying to get the full picture. 😁 👍

PR #5472, opened 2019-06-23
don't use randomly for two tests which assert ordering

asottile:
makes it easier to debug #5468


asottile:
actually a good percentage of the testsuite isn't compatible, abandoning this

blueyed:
> actually a good percentage of the testsuite isn't compatible, abandoning this

Still better than nothing maybe?  (ref: #4351)

Issue #5482, opened 2019-06-24
Parametrize error when passing positional args


kevinjfoley:
Starting with 4.6.0, when passing anything but `argnames` and `argvalues` to `pytest.mark.parametrize` as position arguments, pytest fails during collection with 

```
_pytest/fixtures.py:1145: in _get_direct_parametrize_args
    *marker.args, **marker.kwargs
E TypeError: _parse_parametrize_args() takes exactly 2 arguments (4 given)
```
a minimal example:
```
import pytest
@pytest.mark.parametrize(["a", "b"], [(1, 2), (3, 4)], False, ["1", "2"])
def test_t_pos(a, b):
    assert a + 1 == b
```

I believe the issue stems from this PR - #5254.  The new `_parse_parametrize_args` method only takes `argnames` and `argvalues` as positional arguments and expects all additional arguments to be keywords.

nicoddemus:
Fixed by #5483

PR #5483, opened 2019-06-24
Pickup addition positional args passed to _parse_parametrize_args

kevinjfoley:
Fix for #5482.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -135,6 +135,7 @@ Kale Kundert
 Katarzyna Jachim
 Katerina Koukiou
 Kevin Cox
+Kevin J. Foley
 Kodi B. Arfer
 Kostis Anagnostopoulos
 Kristoffer Nordström
--- a/changelog/5482.bugfix.rst
+++ b/changelog/5482.bugfix.rst
@@ -1,0 +1,2 @@
+Fix bug introduced in 4.6.0 causing collection errors when passing
+more than 2 positional arguments to ``pytest.mark.parametrize``.
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -102,10 +102,7 @@ class ParameterSet(namedtuple("ParameterSet", "values, marks, id")):
             return cls(parameterset, marks=[], id=None)
 
     @staticmethod
-    def _parse_parametrize_args(argnames, argvalues, **_):
-        """It receives an ignored _ (kwargs) argument so this function can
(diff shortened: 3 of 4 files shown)


kevinjfoley:
Will do, will also change the arg names which is causing the tests to fail currently.

Any input on where the test should be added?

nicoddemus:
I think `testing/python/metafunc.py` should do it, thanks.

nicoddemus:
Hmmm I wonder though, since this is already broken, we should make everything except `argnames` and `argvalues` keyword only in 5.0? 🤔 

@asottile @blueyed @Zac-HD what do you think?

nicoddemus:
(@kevinjfoley perhaps hold on to your patch for a bit, wouldn't want you to waste your time)

Zac-HD:
I'm a big fan of keyword-only arguments... but we broke this in 4.6, so I'd prefer to accept the PR (thanks @kevinjfoley!) and then have a follow-up issue to work out what and how to convert to keyword-only arguments.

kevinjfoley:
> thanks @kevinjfoley

No problem!

Pushed a squashed commit with the fix and a test demonstrating the problem, let me know if anything else is needed.

nicoddemus:
>  but we broke this in 4.6, so I'd prefer to accept the PR (thanks @kevinjfoley!) and then have a follow-up issue to work out what and how to convert to keyword-only arguments.

Yeah, I agree, thanks for putting me in check. 👍

kevinjfoley:
> Thanks again @kevinjfoley, appreciate it!

Happy to be able to contribute!

asottile:
Hmmm i don't think this should be merged, it was a bug that parametrize swallowed extra arguments (the test is asserting that a bug exists)

asottile:
Oh already merged :thinking:

nicoddemus:
> Hmmm i don't think this should be merged, it was a bug that parametrize swallowed extra arguments (the test is asserting that a bug exists)

Not really, the extra arguments are valid, here's the signature:

https://github.com/pytest-dev/pytest/blob/37fce6c6e54bc276ca639343344b80b3e3838520/src/_pytest/python.py#L947

So it is in fact a regression, I'm afraid.

asottile:
Yikes, always assumed those were kwargonly, and it's probably too late to do that for 5.0

nicoddemus:
I guess we should open an issue then to discuss how to change that so it emits a warning in 5.X so we can turn it into keyword-only in 6.0 at least.

PR #5484, opened 2019-06-24
Include pytest version in the cached pyc tags

### LINKED ISSUE #1671: assertion rewriting should be sensitive of the pytest version
currently pytest updates do not trigger recreation of the cached bytecode
we should have the pytest version of part of it

nicoddemus:
All credits go to @asottile! Just wanted to make a small contribution before calling it a day.
 
Fix #1671

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -13,6 +13,7 @@ import types
 import atomicwrites
 
 from _pytest._io.saferepr import saferepr
+from _pytest._version import version
 from _pytest.assertion import util
 from _pytest.assertion.util import (  # noqa: F401
     format_explanation as _format_explanation,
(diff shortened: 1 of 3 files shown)


src/_pytest/assertion/rewrite.py
```diff
@@ -21,7 +22,7 @@
 from _pytest.pathlib import PurePath
 
 # pytest caches rewritten pycs in __pycache__.
-PYTEST_TAG = "{}-PYTEST".format(sys.implementation.cache_tag)
+PYTEST_TAG = "{}-pytest-{}".format(sys.implementation.cache_tag, version)
```
nicoddemus: Decided to use lowercase `pytest` here because all the other tag elements are lower case. Decided to follow suit.

nicoddemus:
Btw any concerns about the name of the file containing the dotted pytest version? For example `foo.cpython-37-pytest-5.0.0.pyc`?

I don't think it should be any problem, but thought I would ask.

asottile:
Seems fine to me!

blueyed:
> foo.cpython-37-pytest-5.0.0.pyc

Likely should not really be a problem, but `foo.cpython-37-pytest-50.pyc` would be shorter and looks nicer / more like what Python does.

asottile:
Though dropping dots creates ambiguity between 5.1.10 and 5.11.0 :man_shrugging:

nicoddemus:
> Though dropping dots creates ambiguity between 5.1.10 and 5.11.0

Had the same thought.

Well let's merge this then.

asottile:
pytest: drop the dot... only sometimes ;)

PR #5491, opened 2019-06-25
Merge features into master

nicoddemus:
Didn't realize @Zac-HD's last PR was targeting `features`.

@Zac-HD we didn't really announce it anywhere (sorry), but we will work only on `master` until 5.0 is released, because the py27/py34 support drop happened on `master`.

After 5.0 gets released, it is business as usual regarding `master`/`features`. 👍

### DIFF
--- a/extra/get_issues.py
+++ b/extra/get_issues.py
@@ -74,7 +74,7 @@ def report(issues):
 if __name__ == "__main__":
     import argparse
 
-    parser = argparse.ArgumentParser("process bitbucket issues")
+    parser = argparse.ArgumentParser("process bitbucket issues", allow_abbrev=False)
     parser.add_argument(
         "--refresh", action="store_true", help="invalidate cache, refresh issues"
     )
--- a/extra/get_issues.py
+++ b/extra/get_issues.py
@@ -74,7 +74,7 @@ def report(issues):
 if __name__ == "__main__":
     import argparse
 
-    parser = argparse.ArgumentParser("process bitbucket issues", allow_abbrev=False)
+    parser = argparse.ArgumentParser("process bitbucket issues")
     parser.add_argument(
         "--refresh", action="store_true", help="invalidate cache, refresh issues"
     )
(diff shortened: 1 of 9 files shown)


scripts/release.py
```diff
@@ -105,7 +105,7 @@ def changelog(version, write_out=False):
 
 def main():
     init(autoreset=True)
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(allow_abbrev=False)
```
asottile: idk why we changed this in this file, there's no plugins involved here and prefix matching is actually kind of a nice feature

extra/get_issues.py
```diff
@@ -74,7 +74,7 @@ def report(issues):
 if __name__ == "__main__":
     import argparse
 
-    parser = argparse.ArgumentParser("process bitbucket issues")
+    parser = argparse.ArgumentParser("process bitbucket issues", allow_abbrev=False)
```
asottile: ""

scripts/release.py
nicoddemus: See the mentioned issues for the problems that prefix matching brings, namely that we can't know that a prefix matches a given option during preparsing, because later on plugins might register options that actually match the option given by the user.

nicoddemus: Oh but you got a point about changing it in this file, here doesn't really hurt. 😁

extra/get_issues.py
nicoddemus: Agreed, my bad for not catching this really.

scripts/release.py
asottile: to be clear, I agree with the change in pytest proper, just not in these helper scripts

nicoddemus: Yeah, I understand and agree. 👍

asottile: ok if I add a commit to this PR which undoes those two?

Zac-HD: Fine by me!

nicoddemus:
Merging as the failure is related to tox, which should have been fixed by now with the 1.13.1 release. 👍

PR #5492, opened 2019-06-25
Work around tox-dev/tox#1349

asottile:
tox-dev/tox#1349

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
     - PYTEST_ADDOPTS=-vv
 
 install:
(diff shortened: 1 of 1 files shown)


blueyed:
Is this due to a new tox (pre-)release, or pip?

Could/should we pin pip / virtualenv maybe?  (I've heard we do not want to pin tox)

asottile:
It's a regression between tox 1.12.1 and tox 1.13.0 -- I already have a fix for it in the linked issue from tox's side

asottile:
nevermind, releasing tox 3.13.1 instead :D

PR #5496, opened 2019-06-26
Delete stray comment

asottile:


Zac-HD:
Unrelated to this PR, I just want to say that I'm not enjoying working with codecov... there is simply no way that this PR added 1701 lines of code, and if it made any difference to what code was *covered* then we have a flakiness problem.

I have no solutions short of Hypothesis-style "require 100% coverage of everything and fail CI otherwise" (IMO reasonable but obviously unpopular), but I'm finding it pretty annoying even when it *isn't* hanging forever.

asottile:
> Unrelated to this PR, I just want to say that I'm not enjoying working with codecov... there is simply no way that this PR added 1701 lines of code, and if it made any difference to what code was _covered_ then we have a flakiness problem.
> 
> I have no solutions short of Hypothesis-style "require 100% coverage of everything and fail CI otherwise" (IMO reasonable but obviously unpopular), but I'm finding it pretty annoying even when it _isn't_ hanging forever.

Totally agree, I mostly just ignore codecov

nicoddemus:
> Unrelated to this PR, I just want to say that I'm not enjoying working with codecov... there is simply no way that this PR added 1701 lines of code,

To be fair, those 1701 lines of code are regarding the "coverage diff". Probably it got confused from some other run with less coverage... I'm not sure

PR #5503, opened 2019-06-27
Don't crash with --pyargs and a filename that looks like a module

asottile:


testing/acceptance_test.py
```diff
@@ -646,6 +646,12 @@ def test_pyargs_only_imported_once(self, testdir):
         # should only configure once
         assert result.outlines.count("configuring") == 1
 
+    def test_pyargs_filename_looks_like_module(self, testdir):
+        testdir.tmpdir.join("conftest.py").ensure()
+        testdir.tmpdir.join("t.py").write("def test(): pass")
+        result = testdir.runpytest("--pyargs", "t.py")
```
nicoddemus: Hmm what happens here? Does it run the file? Is that what's expected?

asottile: it runs the file, yes -- it's at least consistent with the 4.6.x behaviour -- I didn't really consider whether this is sane / correct or not :laughing: 

I've never really worked with `--pyargs` so I'm not sure what the expected thing is here

nicoddemus: Well if it works like 4.6.x then that's fine by me! 

Thanks for the quick fix!

nicoddemus:
> Only missing a CHANGELOG entry.

That is, unless that bug was never released in the first place.

asottile:
> > Only missing a CHANGELOG entry.
> 
> That is, unless that bug was never released in the first place.

:this:

asottile:
@nicoddemus approve? :D

nicoddemus:
Oh sorry!

nicoddemus:
Feel free to cut the 5.0.0 release; I won't have time until later today

asottile:
> Feel free to cut the 5.0.0 release; I won't have time until later today

sounds good! I'll take a stab at it and fix up the things in the original PR :+1:

nicoddemus:
Awesome 🎉

PR #5509, opened 2019-06-27
Preparing release version 5.0.0

asottile:


doc/en/example/parametrize.rst
```diff
@@ -434,10 +434,8 @@ Running it results in some skips if we don't have all the python interpreters in
 .. code-block:: pytest
 
    . $ pytest -rs -q multipython.py
-   ...sss...sssssssss...sss...                                          [100%]
-   ========================= short test summary info ==========================
-   SKIPPED [15] $REGENDOC_TMPDIR/CWD/multipython.py:31: 'python3.4' not found
-   12 passed, 15 skipped in 0.12 seconds
+   ...........................                                          [100%]
+   27 passed in 0.12 seconds
```
asottile: oh, should I uninstall some pythons and try this again :laughing:

CHANGELOG.rst
```diff
@@ -18,6 +18,198 @@ with advance notice in the **Deprecations** section of releases.
...
+      filterwarnings =
+          ignore::pytest.PytestDeprecationWarning
+
+  But this will stop working when pytest ``5.1`` is released.
+
+  **If you have concerns** about the removal of a specific feature, please add a
+  comment to `#5402 <https://github.com/pytest-dev/pytest/issues/5402>`__.
+
+
+- `#5412 <https://github.com/pytest-dev/pytest/issues/5412>`_: ``ExceptionInfo`` objects (returned by ``pytest.raises``) now have the same ``str`` representation as ``repr``, which
```
nicoddemus: (not for this release, random thought)

I'm thinking of adding a "Changes" to incorporate changes like this one: it is not a removal, merely a change in behavior which can potentially break existing suites. Thoughts?

CHANGELOG.rst
```diff
@@ -18,6 +18,198 @@ with advance notice in the **Deprecations** section of releases.
...
+  module is now enabled by default to help users diagnose crashes in C modules.
+
+  This functionality was provided by integrating the external
+  `pytest-faulthandler <https://github.com/pytest-dev/pytest-faulthandler>`__ plugin into the core,
+  so users should remove that plugin from their requirements if used.
+
+  For more information see the docs: https://docs.pytest.org/en/latest/usage.html#fault-handler
+
+
+- `#5452 <https://github.com/pytest-dev/pytest/issues/5452>`_: When warnings are configured as errors, pytest warnings now appear as originating from ``pytest.`` instead of the internal ``_pytest.warning_types.`` module.
```
nicoddemus: (not for this release, random thought)

I'm thinking of adding a "Improvements" to incorporate changes like this one: it is not a "new feature", merely an improvement which we don't release in bug-fixes. Thoughts?

CHANGELOG.rst
```diff
@@ -18,6 +18,198 @@ with advance notice in the **Deprecations** section of releases.
...
+
+
+- `#5416 <https://github.com/pytest-dev/pytest/issues/5416>`_: Fix PytestUnknownMarkWarning in run/skip example.
+
+
+
+Trivial/Internal Changes
+------------------------
+
+- `#3457 <https://github.com/pytest-dev/pytest/issues/3457>`_: pytest now also depends on the `astor <https://pypi.org/project/astor/>`__ package.
```
nicoddemus: (not for this release, random thought)

Dependency changes should get their own category as well, perhaps? That's because they should appear higher up on the list of changes too, as this is important for upstream package maintainers (like myself which maintains the conda package).

asottile: oh when did we add `astor` -- I should look at that :D

yeah I think a separate category for new dependencies makes sense

nicoddemus: In https://github.com/pytest-dev/pytest/pull/3479

I had a quick chat with the maintainers: https://github.com/berkerpeksag/astor/issues/149

asottile: hmmm I don't think we should be using `astor` for that -- shouldn't we grab the code out of the original source instead of round-tripping through ast and losing information?

asottile: oh also it's broken :/

```python
def f():
    return 1

def test():
    assert f()
```

```
(Pdb) args
item = <Function test>
lineno = 5
orig = 'f('
expl = '1\n +  where 1 = f()'
```

nicoddemus: Hmmm bummer

@Sup3rGeo what do you think?

CHANGELOG.rst
asottile: :man_shrugging: I think "features" is probably fine for this, an improvement is a feature right :D

CHANGELOG.rst
asottile: yeah seems fair to split it out -- I use a separate "Updating" section for pre-commit's changelogs: https://github.com/pre-commit/pre-commit/blob/master/CHANGELOG.md#updating

CHANGELOG.rst
nicoddemus: My thinking is this: if someone asks you about what are the new features in the release, I doubt this type of small improvement would be something you would mention.

asottile: ok that's fair I guess :man_shrugging: -- could also just call them micro bugs too :laughing:

asottile:
@nicoddemus alright, this should be ready again :D

nicoddemus:
Awesome, ship it! 😁 🎉

nicoddemus:
Don't forget to send the announcement email 😉

asottile:
yep :) just did that!

nicoddemus:
email? Weird still have not gotten it...

asottile:
weird!

PR #5511, opened 2019-06-27
WIP for 5510 [skip ci]

nicoddemus:
cc @asottile

PR #5512, opened 2019-06-28
Remove astor and reproduce the original assertion expression

### LINKED ISSUE #5510: pytest_assertion_pass strange results
(From https://github.com/pytest-dev/pytest/pull/5509#discussion_r298332848)

It seems this patch fixes the problem:

```diff
diff --git a/src/_pytest/assertion/rewrite.py b/src/_pytest/assertion/rewrite.py
index 8810c156c..28091c115 100644
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -750,7 +750,7 @@ class AssertionRewriter(ast.NodeVisitor):

             # Passed
             fmt_pass = self.helper("_format_explanation", msg)
-            orig = astor.to_source(assert_.test).rstrip("\n").lstrip("(").rstrip(")")
+            orig = astor.to_source(assert_.test).strip()
             hook_call_pass = ast.Expr(
                 self.helper(
                     "_call_assertion_pass",
```

```
item = <Function test>
lineno = 5
orig = f()
expl = 1
 +  where 1 = f()
```

The test we have for it fails, but just because it expects the parenthesis to be stripped:

```
E           and: '>       raise Exception("Assertion Passed: {} {} at line {}".format(orig, expl, lineno))'
E           and: 'E       Exception: Assertion Passed: (a + b == c + d) (1 + 2) == (3 + 0) at line 7'
...
E       remains unmatched: '*Assertion Passed: a + b == c + d (1 + 2) == (3 + 0) at line 7*'
```

So I think we should avoid trying to strip the results ourselves. I will open a PR.

cc @Sup3rGeo @asottile

asottile:
Resolves #5510

This is a little different approach than using `astor` in that it reproduces the original assertion source expression directly

I could make this even simpler by including the whole line of the original assertion (then the indentation would be preserved as well???) what do you think of that?

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -1,16 +1,18 @@
 """Rewrite assertion AST to produce nice error messages"""
 import ast
 import errno
+import functools
 import importlib.machinery
 import importlib.util
+import io
 import itertools
 import marshal
 import os
 import struct
 import sys
+import tokenize
 import types
 
-import astor
 import atomicwrites
 
 from _pytest._io.saferepr import saferepr
@@ -285,7 +287,7 @@ def _rewrite_test(fn, config):
     with open(fn, "rb") as f:
         source = f.read()
     tree = ast.parse(source, filename=fn)
-    rewrite_asserts(tree, fn, config)
+    rewrite_asserts(tree, source, fn, config)
     co = compile(tree, fn, "exec", dont_inherit=True)
     return stat, co
 
@@ -327,9 +329,9 @@ def _read_pyc(source, pyc, trace=lambda x: None):
         return co
 
 
-def rewrite_asserts(mod, module_path=None, config=None):
+def rewrite_asserts(mod, source, module_path=None, config=None):
     """Rewrite the assert statements in mod."""
-    AssertionRewriter(module_path, config).run(mod)
+    AssertionRewriter(module_path, config, source).run(mod)
 
 
 def _saferepr(obj):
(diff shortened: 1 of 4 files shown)


src/_pytest/assertion/rewrite.py
```diff
@@ -285,7 +287,7 @@ def _rewrite_test(fn, config):
     with open(fn, "rb") as f:
         source = f.read()
     tree = ast.parse(source, filename=fn)
-    rewrite_asserts(tree, fn, config)
+    rewrite_asserts(tree, source, fn, config)
```
nicoddemus: Not that this is a blocker, but this will end up breaking https://pypi.org/project/pytest-ast-back-to-python.

cc @tomviner

src/_pytest/assertion/rewrite.py
```diff
@@ -457,6 +459,33 @@ def _fix(node, lineno, col_offset):
     return node
 
 
+def _get_assertion_exprs(src: bytes):  # -> Dict[int, str]
```
nicoddemus: A small docstring would be nice here.

testing/test_assertrewrite.py
```diff
@@ -1444,3 +1441,53 @@ def test_simple():
...
+            {2: "1\n"},
+            id="latin1 encoded on first line\n",
+        ),
+        pytest.param(
+            # using the default utf-8 encoding
+            "def ÀÀÀÀÀ(): assert 1\n".encode(),
+            {1: "1\n"},
+            id="utf-8 encoded on first line",
+        ),
+        pytest.param(
```
nicoddemus: Could we add another case where we break the expression itself in multiple lines?

```python
        pytest.param(
            b"def x():\n"
            b"    assert (\n"
            b"        1 + \\"
            b"        2  # comment\n"
            b"    )\n",
            ...
        )
```

Also I commonly do asserts like this:

```python
assert obtained == [
    ...
] 
```

So another one like this might be in order:


```python
        pytest.param(
            b"def x():\n"
            b"    assert values == [\n"
            b"        1,\n"
            b"        2,\n"
            b"    )\n",
            ...
        )
```

Another case that comes to mind is assertions with a message:

```python
        pytest.param(
            b"def x():\n"
            b"    assert 1 + 1, 'some message'\n"
            ...
        )
```

and

```python
        pytest.param(
            b"def x():\n"
            b"    assert 1 + 1, f()\n"
            ...
        )
```

Let me know if those make sense or not.

testing/test_assertrewrite.py
```diff
@@ -1371,10 +1371,21 @@ def test_fails():
         )
         result = testdir.runpytest()
         result.stdout.fnmatch_lines(
-            "*Assertion Passed: a + b == c + d (1 + 2) == (3 + 0) at line 7*"
+            "*Assertion Passed: a+b == c+d (1 + 2) == (3 + 0) at line 7*"
```
nicoddemus: Did you see the approach I took in https://github.com/pytest-dev/pytest/pull/5511 of using `inline_run` and checking the parameters received by the hook directly? Feels more elegant that way as we can check the parameters directly instead of relying on parsed output.

src/_pytest/assertion/rewrite.py
asottile: ah, I can restore the api and just re-read the source in the visitor using `fn`

testing/test_assertrewrite.py
asottile: ooh, I'm definitely not handling messages right now -- should be an easy fix

the multi-line expression ones aren't all that different than the current cases but I can add those as well :+1:

testing/test_assertrewrite.py
asottile: I did yeah, didn't quite feel like expanding the api scope of pytester further but I can take another look at that again :+1:

nicoddemus: I added a two-liner helper function to `ParsedCall`, that's all. I think it is an improvement.

Anyway just a suggestion really. 👍

testing/test_assertrewrite.py
asottile: added a few of these, plus message-based ones and some tests for escaped newlines

asottile:
alrighty, let me replay the release branch :D

src/_pytest/assertion/rewrite.py
tomviner: Just to update here, [pytest-ast-back-to-python](https://pypi.org/project/pytest-ast-back-to-python/) has now been fixed.

nicoddemus: Awesome Tom, thanks for the heads up!

PR #5517, opened 2019-06-28
doc(plugin_hooks): Improve documentation for writing plugin hooks.

NiklasMM:
I wrote my first plugin hooks today and since I found the documentation somewhat lacking, I extended it somewhat, so other people will have an easier time. :)

Would be great if someone with a better understanding of the subject could double check if I didn't make any mistakes.

### DIFF
--- a/doc/en/writing_plugins.rst
+++ b/doc/en/writing_plugins.rst
@@ -621,11 +621,59 @@ the new plugin:
 
 Hooks are usually declared as do-nothing functions that contain only
 documentation describing when the hook will be called and what return values
-are expected.
+are expected. The names of the functions must start with `pytest_` otherwise pytest won't recognize them.
 
-For an example, see `newhooks.py`_ from `xdist <https://github.com/pytest-dev/pytest-xdist>`_.
+Here's an example. Let's assume this code is in the ``hooks.py`` module.
+
+.. code-block:: python
+
+    def pytest_my_hook(config):
+        """
+        Receives the pytest config and does things with it
+        """
+
+To register the hooks with pytest they need to be structured in their own module or class. This
+class or module can then be passed to the ``pluginmanager`` using the ``pytest_addhooks`` function
+(which itself is a hook exposed by pytest).
+
+.. code-block:: python
+
+    def pytest_addhooks(pluginmanager):
+        """ This example assumes the hooks are grouped in the 'hooks' module. """
+        from my_app.tests import hooks
+        pluginmanager.add_hookspecs(hooks)
+
+For a real world example, see `newhooks.py`_ from `xdist <https://github.com/pytest-dev/pytest-xdist>`_.
 
 .. _`newhooks.py`: https://github.com/pytest-dev/pytest-xdist/blob/974bd566c599dc6a9ea291838c6f226197208b46/xdist/newhooks.py
+
+Hooks may be called both from fixtures or from other hooks. In both cases, hooks are called
+through the ``hook`` object, available in the ``config`` object. Most hooks receive a
+``config`` object directly, while fixtures may use the ``pytestconfig`` fixture which provides the same object.
+
+.. code-block:: python
+
+    @pytest.fixture()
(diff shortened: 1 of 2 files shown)


doc/en/writing_plugins.rst
```diff
@@ -621,12 +621,39 @@ the new plugin:
...
 documentation describing when the hook will be called and what return values
-are expected.
+are expected. The names of the functions must start with `pytest_` otherwise pytest won't recognize them.
 
 For an example, see `newhooks.py`_ from `xdist <https://github.com/pytest-dev/pytest-xdist>`_.
 
 .. _`newhooks.py`: https://github.com/pytest-dev/pytest-xdist/blob/974bd566c599dc6a9ea291838c6f226197208b46/xdist/newhooks.py
 
+To register the hooks with pytest they need to be structured in their own module or class. This
+class or module can then be passed to the `pluginmanager` using the `pytest_addhooks` function
```
nicoddemus:
```suggestion
class or module can then be passed to the ``pluginmanager`` using the ``pytest_addhooks`` function
```

doc/en/writing_plugins.rst
```diff
@@ -621,12 +621,39 @@ the new plugin:
...
+(which itself is a hook exposed by pytest):
+
+.. code-block:: python
+
+    def pytest_addhooks(pluginmanager):
+        """ This example assumes the hooks are grouped in the pytest_hooks module. """
+        from my_app.tests import pytest_hooks
```
nicoddemus: It is not a good idea to call a file `pytest_hooks`, because it will appear in the namespace of `my_app.tests` and pytest will complain that it doesn't recognize the hook `pytest_hooks` (because it starts with `pytest_`, pytest assumes it must be a hook).

I suggest using simply `hooks`.

doc/en/writing_plugins.rst
```diff
@@ -621,12 +621,39 @@ the new plugin:
...
+        pluginmanager.add_hookspecs(pytest_hooks)
+
+To execute hooks in one of your fixtures, use the `pytestconfig` fixture. It holds all registered
```
nicoddemus: I think this paragraph can be improved thus:

```rst
Hooks may be called both from fixtures or from other hooks. In both cases, hooks are called 
through the ``hook`` object, available in the ``config`` object. Most hooks receive a 
``config`` object directly, while fixtures may use the ``pytestconfig`` fixture which provides the same object.
```

doc/en/writing_plugins.rst
nicoddemus: It is also probably show the contents of the `hooks` file, and make `pytest_my_hook` receive one or more parameters to showcase argument passing in `my_fixture` down below.

doc/en/writing_plugins.rst
```diff
@@ -621,12 +621,39 @@ the new plugin:
...
+.. code-block:: python
+
+    @pytest.fixture()
+    def my_fixture(pytestconfig):
+        # call the hook called "pytest_my_hook"
+        # `result` will be a list of return values from all registered functions.
+        result = pytestconfig.hook.pytest_my_hook()
+
+Now your hook is ready to be used. To register a function at the hook, other plugins or users must
+now simply define the function `pytest_my_hook` with the correct signature in their `conftest.py`.
```
nicoddemus: Also I believe an example here would be nice.

doc/en/writing_plugins.rst
```diff
@@ -621,12 +621,39 @@ the new plugin:
...
+
+.. note::
+    If a hook uses parameters only keyword arguments can be used to invoke it.
```
nicoddemus:
```suggestion
    Hooks receive parameters using only keyword arguments.
```

NiklasMM:
Thanks @nicoddemus, I tried to implement all of your suggestions.

nicoddemus:
Linting failed, to fix it one can use `tox -e linting`. I've just did it myself and pushed. Thanks!

Issue #5513, opened 2019-06-28
Remove fallback from inspect wrapper


Zac-HD:
Now that the master branch only has to be compatible with Python 3.5+, we can simplify the following code because `inspect.unwrap` was added in 3.4:

https://github.com/pytest-dev/pytest/blob/cf27af734e0b3689944061721c54af0e7d24dd08/src/_pytest/doctest.py#L367-L370

So we can just use `real_unwrap = inspect.unwrap`, and unconditionally run the `else:` block.  This is a good issue for a first-time contributor!

AmirElkess:
I apologize if any mistake happened, as it's my first contribution to open-source community.

I made this change, which passed the tests, then pushed it. 
According to the contribution guide, the next step is to create a new changelog entry, and I am unsure of the type of this issue. 

Can you please check if my push request is correct, and tell me what type is this issue?
Thanks.

nicoddemus:
Hi @AmirElkess,

Thanks for the contribution in https://github.com/pytest-dev/pytest/pull/5518. I will comment about your questions in there. 👍

nicoddemus:
Fixed by #5518

PR #5518, opened 2019-06-28
Refactoring doctests

AmirElkess:
head-fork: AmirElkess/pytest
compare: refactor

base-fork: pytest-dev/pytest
base: master

### DIFF
--- a/src/_pytest/doctest.py
+++ b/src/_pytest/doctest.py
@@ -364,30 +364,27 @@ def _patch_unwrap_mock_aware():
     contextmanager which replaces ``inspect.unwrap`` with a version
     that's aware of mock objects and doesn't recurse on them
     """
-    real_unwrap = getattr(inspect, "unwrap", None)
-    if real_unwrap is None:
-        yield
-    else:
+    real_unwrap = inspect.unwrap
 
-        def _mock_aware_unwrap(obj, stop=None):
-            try:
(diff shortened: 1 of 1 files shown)


nicoddemus:
Hi @AmirElkess,

Thanks a lot for taking this issue and submitting a PR, we really appreciate it. Congratulations on your first open source contribution. 👍 

Regarding the code itself, your PR is fine, thanks! 

About title and description I have a few tips: it helps to use a short title and small description so we maintainers can see what the PR is about at a glance. Including the issue number that the PR is related to is also good (#5513 in this example), as GitHub will automatically create links between issues and PR.

Another tip is to use a [keyword to close issues automatically when merging](https://help.github.com/en/articles/closing-issues-using-keywords), in this case it would be `Fix #5513` (or any of the other keywords known by GitHub).
 
Having said all the above, here is how I would have submitted this PR:

```
title: Remove fallback for legacy inspect.unwrap

description: Fix #5513
```

Note that I wrote a very short description, that's because the issue explains the problem well and the PR is very straightforward, so I just used the "Fix #<issue>" magic comment so GitHub will close the related issue automatically when the PR is merged. 👍 

--- 

About the changelog entry, we usually add one, but given that this is an internal change which doesn't affect users in any way, we don't really need a CHANGELOG for that, so none is necessary! 👍 

I'm merging this as is then. Thanks again!

Issue #5534, opened 2019-07-01
setup.py test is going to be deprecated


jamesmyatt:
There are plans to deprecate `setup.py test` (https://github.com/pypa/setuptools/issues/1684). Is it worth updating the section in "Good Integration Practices" to discourage using the setuptools integration (https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner)?

nicoddemus:
Definitely, thanks for bringing this to attention.

We might even consider removing that section altogether I think, only keeping the `pytest-runner` section.

nicoddemus:
@jaraco what do you think?

graingert:
setting setup_requires in 'setup.py' is not good practice, it should be moved to `pyproject.toml` if used at all

graingert:
https://github.com/pytest-dev/pytest/issues/3814

jaraco:
I agree we should discourage use of `setup_requires` and `tests_require` in favor of extras or requirements.txt or pyproject.toml.

> We might even consider removing that section altogether I think, only keeping the pytest-runner section.

Yes, I'm in favor of removing the section altogether. I don't see any value in keeping a reference to pytest-runner. Instead, I recommend giving users guidance on how best to _orchestrate_ (create environments, declare and install dependencies, etc) their tests (e.g. with `tox`).

jamesmyatt:
I'd be tempted to include a recommendation to not integrate with setuptools, rather than just have the section disappear. Otherwise people won't understand why it's gone.

nicoddemus:
@jamesmyatt that's a good suggestion. Would you like to open a short PR with that? I suppose this can be done even directly from the GH interface. 👍

nicoddemus:
Closed by #5546

PR #5546, opened 2019-07-03
remove documentation about setuptools integration Refs #5534 Fixes #3814

### LINKED ISSUE #3814: recommended pytest-runner in setup_requires means packages fail to install often
The recommendation to add `pytest-runner` to `setup_requires` means that all users of that package end up with an unnecessary pytest-runner package installed. This is bad because it bypasses pip hashes and [`--trusted-host`](https://github.com/pypa/pip/issues/4156)

https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner

https://github.com/MechanicalSoup/MechanicalSoup/pull/224
https://github.com/rxcomm/pyaxo/issues/26
https://github.com/jpadilla/pyjwt/issues/179

graingert:
integrating with `python setup.py test` is not good practice!


graingert:
https://github.com/pytest-dev/pytest/issues/5534

doc/en/goodpractices.rst
```diff
@@ -219,101 +219,4 @@ against your source code checkout, helping to detect packaging
 glitches.
 
 
-Integrating with setuptools / ``python setup.py test`` / ``pytest-runner``
```
nicoddemus: I would keep this section, perhaps renaming it to: 
```rst
Integrating with setuptools using ``pytest-runner``
```

But we should still remove the `Manual Integration` section.

graingert: From @jaraco 

> Yes, I'm in favor of removing the section altogether. I don't see any value in keeping a reference to pytest-runner.

nicoddemus: Oh OK then, I missed that comment. 👍 

In that case let's remove it altogether.

nicoddemus:
Thanks!

PR #5563, opened 2019-07-05
Sort parametrize params to test_external_plugins_integrated

nicoddemus:
This might cause problems during collection with pytest-xdist; we
didn't see any so far mostly by luck I think.

Shame on me for letting that slip in. 😓

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -131,6 +131,7 @@ filterwarnings =
     ignore:yield tests are deprecated, and scheduled to be removed in pytest 4.0:pytest.RemovedInPytest4Warning
     ignore:Metafunc.addcall is deprecated and scheduled to be removed in pytest 4.0:pytest.RemovedInPytest4Warning
     ignore::pytest.RemovedInPytest4Warning
+    default:Using or importing the ABCs:DeprecationWarning:unittest2.*
     ignore:Module already imported so cannot be rewritten:pytest.PytestWarning
     # produced by path.local
     ignore:bad escape.*:DeprecationWarning:re
--- a/testing/deprecated_test.py
+++ b/testing/deprecated_test.py
@@ -70,7 +70,7 @@ def test_terminal_reporter_writer_attr(pytestconfig):
     assert terminal_reporter.writer is terminal_reporter._tw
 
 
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -131,6 +131,7 @@ filterwarnings =
     ignore:yield tests are deprecated, and scheduled to be removed in pytest 4.0:pytest.RemovedInPytest4Warning
     ignore:Metafunc.addcall is deprecated and scheduled to be removed in pytest 4.0:pytest.RemovedInPytest4Warning
     ignore::pytest.RemovedInPytest4Warning
+    default:Using or importing the ABCs:DeprecationWarning
```
nicoddemus: Unrelated to the main purpose of the PR, but Travis is now issuing this warning:

```
/home/travis/build/pytest-dev/pytest/.tox/py37-lsof-numpy-twisted-xdist/lib/python3.7/site-packages/unittest2/compatibility.py:143: in <module>
    class ChainMap(collections.MutableMapping):
name = 'MutableMapping'
    def __getattr__(name):
        # For backwards compatibility, continue to make the collections ABCs
        # through Python 3.6 available through the collections module.
        # Note, no new collections ABCs were added in Python 3.7
        if name in _collections_abc.__all__:
            obj = getattr(_collections_abc, name)
            import warnings
            warnings.warn("Using or importing the ABCs from 'collections' instead "
                          "of from 'collections.abc' is deprecated, "
                          "and in 3.8 it will stop working",
>                         DeprecationWarning, stacklevel=2)
E           DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

Zac-HD: I think this is redundant with #5565, right?  In which case I'd rather wait for that rather than suppress the warning, since it could indicate a real problem somewhere else!

nicoddemus: Changed this to be specific to `unittest2` now, given that we won't be actually dropping unittest2 until 6.0.

nicoddemus: @Zac-HD can you please approve if this looks good to you now? Thanks! 🙇

PR #5565, opened 2019-07-05
Drop unittest2 support

nicoddemus:
[unittest2](https://pypi.org/project/unittest2) is a backport for Python <=3.2, so we no longer need
to support it (last release was in 2015). 

I suspect it has not been used at all by users after 5.0 has been released.

### DIFF
--- a/src/_pytest/nose.py
+++ b/src/_pytest/nose.py
@@ -1,29 +1,7 @@
 """ run test suites written for nose. """
-import sys
-
-import pytest
 from _pytest import python
-from _pytest import runner
 from _pytest import unittest
 from _pytest.config import hookimpl
-
-
-def get_skip_exceptions():
-    skip_classes = set()
-    for module_name in ("unittest", "unittest2", "nose"):
-        mod = sys.modules.get(module_name)
-        if hasattr(mod, "SkipTest"):
-            skip_classes.add(mod.SkipTest)
-    return tuple(skip_classes)
-
-
-def pytest_runtest_makereport(item, call):
-    if call.excinfo and call.excinfo.errisinstance(get_skip_exceptions()):
-        # let's substitute the excinfo with a pytest.skip one
-        call2 = runner.CallInfo.from_call(
-            lambda: pytest.skip(str(call.excinfo.value)), call.when
-        )
-        call.excinfo = call2.excinfo
 
 
 @hookimpl(trylast=True)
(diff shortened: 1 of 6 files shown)


nicoddemus:
> I'd be inclined to detect and warn on use of unittest2, since as noted it's a pointless dependency on Python 3. 

I believe you are correct, we should deprecate it first and remove it only on pytest 6.0. I will change this accordingly.

Zac-HD:
I meant rather that we should retain a warning even *after* removing supoort, to explicitly notify users that they no longer need that dependency (and that Pytest no longer supports it).

Happy to remove in 5.1 personally, but 6.0 would be fine too.

nicoddemus:
Will wait for https://github.com/pytest-dev/pytest/pull/5529 to be merged first before getting back to this.

nicoddemus:
Strangely, even after removing `"unittest2"` from the only place where it is used under `src`:

https://github.com/pytest-dev/pytest/blob/57e5bd0664d23d1d923f2747f93b59595de25237/src/_pytest/nose.py#L11-L17

My simple tests still work. Just pushing a commit removing `unittest2` from the list of supported skip exceptions to see where we break.

nicoddemus:
Ahh I see, `unittest2.TestCase` subclasses `unittest.TestCase`, and it seems all pytest features (like skip) only really depend on attributes obtained from `TestCase` instances. 

For example, skipping works by adding an `__unittest_skip__` attribute to the test function:

```
def skip(reason):
    """
    Unconditionally skip a test.
    """
    def decorator(test_item):
        if not isinstance(test_item, type):
            @functools.wraps(test_item)
            def skip_wrapper(*args, **kwargs):
                raise SkipTest(reason)
            test_item = skip_wrapper

        test_item.__unittest_skip__ = True
        test_item.__unittest_skip_why__ = reason
        return test_item
    return decorator
```

(https://github.com/python/cpython/blob/6aeb2fe606408aae14c246470794f1303b3be812/Lib/unittest/case.py#L111-L125)

Both `unittest.skip` and `unittest2.skip` work in the same way, and that's what we check. 

This means we don't actually need any explicit `unittest2` support since Python 2.7 at all. 😅

Zac-HD:
Right then... I think we can just delete tbat line and we're done!

nicoddemus:
@RonnyPfannschmidt @asottile opinions?

asottile:
Seems fine, especially that it still mysteriously works heh

src/_pytest/nose.py
```diff
@@ -1,31 +1,9 @@
...
-    skip_classes = set()
-    for module_name in ("unittest", "unittest2", "nose"):
-        mod = sys.modules.get(module_name)
-        if hasattr(mod, "SkipTest"):
-            skip_classes.add(mod.SkipTest)
-    return tuple(skip_classes)
-
-
-def pytest_runtest_makereport(item, call):
-    if call.excinfo and call.excinfo.errisinstance(get_skip_exceptions()):
```
nicoddemus: This code was moved to `unittest.py` as I think it makes more sense there.

PR #5568, opened 2019-07-07
Improve docs sidebar layout

timhoffm:
This PR improves the layout of the sidebar:

- Move the search field to the top (should be easily accessible without the need to scroll).
- Reduce the heading font size so that "Table Of Contents" is not wrapped.

Here's a screenshot of the change (left: before, right: after):

grafik grafik


curiousjazz77:
Nice improvement!

nicoddemus:
Thanks! appreciate it! 👍

PR #5571, opened 2019-07-07
Register marks using regex

lhupfeldt:


lhupfeldt:
> interesting approach, please note that we still hope to do away with string matching markers in mid term future

I hope this, or something similar, can be accepted in the mean time, because we are drowning in warnings.

RonnyPfannschmidt:
its already approved, another core dev will merge it, its a really neat improvement

nicoddemus:
I wonder if we shouldn't instead use a special marking instead of using a tuple for regex-matching of marks?

Something like:

```python
config.addinivalue_line(
            "markers", ":regex:AT_[0-9]+:Match marks generated by A-tool")

```

We already have some string-handling magic there in order to support signatures and descriptions:

```python
config.addinivalue_line(
        "markers",
        "pytester_example_path(*path_segments): join the given path "
        "segments to `pytester_example_dir` for this test.",
    )
```

Adding a new one might be more consistent. This also would allow one to use the regex patterns in the `pytest.ini` file:

```ini
[pytest]
markers = 
    :regex:AT_[0-9]+:Match marks generated by A-tool
```

--- 

Another idea is that we can assume the names of the markers are already a regex and always use `re.match` to match them. Marks are normal Python identifiers so they don't contain special characters, so this is even backward compatible:

```python
config.addinivalue_line(
            "markers", "AT_[0-9]+:Match marks generated by A-tool")
```

lhupfeldt:
Hi,

Note that marks are not always identifiers.
We use pytest-bdd which reads marks from cucumber style feature files. We
actually have marks of the form "id:112". I did not investigate which
interface is used to add those to pytest. I propose keeping the tuple based
registration to avoid complicated and hard tp explain string registration
rules.
I will take a look at adding a test for marks with ':'.

I do think that your idea of turning all registrations into regex could
simplify the implementation, while allowing regex in ini file.

Best regards / Venlig Hilsen
Lars Hupfeldt Nielsen
Hupfeldt IT

Phone: +45 2060 3546


Den tir. 9. jul. 2019 kl. 00.20 skrev Bruno Oliveira <
notifications@github.com>:

nicoddemus:
Hi,

> We use pytest-bdd which reads marks from cucumber style feature files. We
actually have marks of the form "id:112". I did not investigate which
interface is used to add those to pytest. 

I guess they might be using something like `getattr(pytest.mark, "id:112")` or such. This can be easily fixed by using `re.escape`: `getattr(pytest.mark, re.escape("id:112"))`.

TBH I have never seen marks which were not identifiers in the wild, seems like `pytest-bdd` is the uncommon case; given that we have an easy workaround, we can contribute it to `pytest-bdd` and it can be applied based on pytest version for backward compatibility:

```
escape = re.escape if Version(pytest.__version__) >= Version('5.1') else lambda v: v
mark = getattr(pytest.mark, escape("id:112"))
```

> I propose keeping the tuple based registration to avoid complicated and hard tp explain string registration rules.

TBH I find "mark names are regexes" much easier to explain than the special case for tuples. 🙃

But before you move this forward either way, I would like other's opinions on this as well.

cc @asottile @blueyed @The-Compiler

lhupfeldt:
Hi

The complexity of the explanation comes from the "... the text following
the last ':' is the description..." which may not be too complicated after
all.

Note that my solution supports any object with a match method. I guess not
important as all markers can probably be handled by regex, and the number
of markers will probably be no more than a few thousand, so performance
should not be a problem either.

Best regards
Lars

RonnyPfannschmidt:
im -1 on regex in the config files

nicoddemus:
> im -1 on regex in the config files

Why? When disagreeing with something @RonnyPfannschmidt, it helps to move the discussion forward to state this in the same post, otherwise people lose a lot of time trying to figure out what you mean and getting back and forth.

RonnyPfannschmidt:
i dont see this being easyly and  sanely implementable in ini, we have certain overlaps of functionality there

guessing whether we have a signature or just a regex in  front of is is plain error prone, i don't want that as a starting point, but rather as a well experimented follow up (regex is that tricky after all)

nicoddemus:
@lhupfeldt,

This just occurred to me, but have you considered ignoring the warnings for the marks you know about? You can use regex matching in the message itself:

```
λ pytest .tmp\test_m.py -q
.
========================= warnings summary ==========================
PytestUnknownMarkWarning: Unknown pytest.mark.AT_12102 - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

1 passed, 1 warnings in 0.01 seconds
```

```
λ pytest .tmp\test_m.py "-Wignore:Unknown pytest.mark.AT_" -q
.
1 passed in 0.01 seconds
```

(of course that setting can be used in the `pytest.ini` file)

nicoddemus:
> guessing whether we have a signature or just a regex in front of is is plain error prone, i don't want that as a starting point, but rather as a well experimented follow up (regex is that tricky after all)

Thanks for following up. Do you mean that for the idea of having a `:regex:` prefix, for just treating mark names as a regex, or both?

TBH I'm definitely not happy with the "tuple" signature, I consider that kind of a hack, which we will regret later down the road. I would prefer to introduce a new function specifically to register "named marks", rather than overloading the generic `addinivalue_line`.

RonnyPfannschmidt:
we already have such a proposal in the open, having the registred marker name be a regex sounds doable

we might also want to  consider that for fixtures, (and in both cases have regex matches from the name be accessible in some way)

asottile:
> TBH I'm definitely not happy with the "tuple" signature, I consider that kind of a hack, which we will regret later down the road. I would prefer to introduce a new function specifically to register "named marks", rather than overloading the generic addinivalue_line.

This was my feeling as well -- also makes it hard to make that function and the related functions "typesafe"

RonnyPfannschmidt:
we could use a representational object then

nicoddemus:
How about introducing functions to register named marks? This goes in the direction of https://github.com/pytest-dev/pytest/issues/5424 anyway.

Something like:

```python
def pytest_configure():
    pytest.mark.register_named_mark(name=re.compile('AT_(\d+)', description="some description")

    pytest.mark.register_named_mark(name="slow", description="slow tests")
```

I really don't think we should insert special mark-related handling in `config.addinivalue_line`, which is generic and doesn't know anything about marks.

RonnyPfannschmidt:
not on the mark object tho

RonnyPfannschmidt:
```

def pytest_configure(config):
   @config.register_marker(re.compile('AT_(\d+)')
   def AT(number):
      "description"
```
could be a precursor to 

full marker spec with object return values

RonnyPfannschmidt:
but in any case im still for merging this pr as experimental and iterating as followup, its a working contribution, and we didnt provide something better in time

lhupfeldt:
Hi @Bruno

Actually it never occured to me that I could suppress warnings. I never
used that feature before.

nicoddemus:
@lhupfeldt, 

> Actually it never occured to me that I could suppress warnings. I never
used that feature before.

I see. Can you please use it and check it solves the original issue you were having? Note you can add the option to your `pytest.ini`:

```ini
[pytest]
filterwarnings = 
    ignore:Unknown pytest.mark.AT_
```

If it does solve the problem for you, I vote -1 on this PR: while we do appreciate your efforts and contribution, it is clear that we don't agree yet on a good API for this yet, and if it does solve your immediate problem, I don't see a reason to rush this in. 👍

lhupfeldt:
@Bruno

Suppressing the warnings works fine.
I'm good with that for now, so take your time to figure out how to register
marks using regex.

Seems that it shold also be aligned with the filterwarnings syntax which
already supports regex.

When configuring things through conftest, a more programming friendly
interface should be used, rather than using addinivalue, e.g. no string
representation of something like regex plus description.

nicoddemus:
Thanks @lhupfeldt for the follow up!

> When configuring things through conftest, a more programming friendly
interface should be used, rather than using addinivalue, e.g. no string
representation of something like regex plus description.

Agreed, that's why I was hesitant and now I'm -1 on the original proposal.

Thanks again for the PR, sorry that in the end there was a simpler solution and you wasted some time on this, we appreciate it nonetheless. 👍 

I think we can close this now then?

asottile:
agreed :) -- closing

lhupfeldt:
Yes go ahead and close.

Thank you for the support.

PR #5576, opened 2019-07-08
doctest: Add +NUMBER option to ignore irrelevant floating-point differences

drothlis:
When enabled, floating-point numbers only need to match as far as the
precision you have written in the expected doctest output. This avoids
false positives caused by limited floating-point precision, like this:

    Expected:
        0.233
    Got:
        0.23300000000000001

This is inspired by Sébastien Boisgérault's [numtest] but the
implementation is a bit different:

* This implementation edits the literals that are in the "got"
  string (the actual output from the expression being tested), and then
  proceeds to compare the strings literally. This is similar to pytest's
  existing ALLOW_UNICODE and ALLOW_BYTES implementation.

* This implementation only compares floats against floats, not ints
  against floats. That is, the following doctest will fail with pytest
  whereas it would pass with numtest:

      >>> math.py  # doctest: +NUMBER
      3

  This behaviour should be less surprising (less false negatives) when
  you enable NUMBER globally in pytest.ini.

Advantages of this implementation compared to numtest:

* Doesn't require `import numtest` at the top level of the file.
* Works with pytest (if you try to use pytest & numtest together, pytest
  raises "TypeError: unbound method check_output() must be called with
  NumTestOutputChecker instance as first argument (got
  LiteralsOutputChecker instance instead)").
* Works with Python 3.

[numtest]: https://github.com/boisgera/numtest

### DIFF
--- a/testing/test_doctest.py
+++ b/testing/test_doctest.py
@@ -3,6 +3,7 @@ import textwrap
 
 import pytest
 from _pytest.compat import MODULE_NOT_FOUND_ERROR
+from _pytest.doctest import _get_checker
 from _pytest.doctest import _is_mocked
 from _pytest.doctest import _patch_unwrap_mock_aware
 from _pytest.doctest import DoctestItem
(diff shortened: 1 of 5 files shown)


thisch:
https://github.com/astropy/pytest-doctestplus already has support for this kind of handling of floating point comparisons in doctests.

drothlis:
Ha, that would have saved me half a day. :-) Thanks for pointing it out!

It might be worth merging this anyway? Here are the advantages of this pull
request compared to pytest-doctestplus, based on a quick inspection of the latter:

- The precision is set by the literal that you put in the expected doctest
  output. For example, if you write this:

      >>> some_function()
      3.14

  ...then it checks that the actual output matches to 2 decimal places.

  If you write this:

      >>> some_function()
      3.1416

  ...then it checks to 4 decimal places. etc.

  In pytest-doctestplus you can only configure atol & rtol globally.

- You can enable NUMBER and ELLIPSIS at the same time. In pytest-doctestplus you
  can't, at least according to https://github.com/astropy/pytest-doctestplus/blob/720266f/tests/test_doctestplus.py#L106

- This implementation is less invasive (it doesn't replicate so much of
  `doctest.OutputChecker.check_output`).
- More tests.

There may be *dis*advantages of this implementation, of course, that I may have missed. :-)

thisch:
Great! Thx for comparing your implementation with the one from doctest-plus! IIRC there were some discussions about merging pytest-doctestplus into pytest-core or merging pytest's doctest plugin into pytest-doctestplus. Maybe @nicoddemus or @RonnyPfannschmidt can help me find the relevant tickets where this was discussed.

nicoddemus:
> Maybe @nicoddemus or @RonnyPfannschmidt can help me find the relevant tickets where this was discussed.

Hmmm what comes to mind was perhaps adding a note to the docs suggesting `pytest-doctestplus` as an alternative to the builtin functionality instead?

nicoddemus:
Hi @drothlis,

Thanks for the patch! It is very well written, both in terms of code clarity, tests and docs, so congratulations! 🎉 

I plan to review this more thoroughly this week. Thanks again.

testing/test_doctest.py
```diff
@@ -838,6 +839,149 @@ def test_bytes_literal(self, testdir):
...
+            ("3e0", "3"),
+            ("1e3", "1000"),
+            ("3", "3.0"),
+            # Rounding:
+            ("3.1", "3.0"),
+            ("3.1", "3.2"),
+            ("3.1", "4.0"),
+            ("8.22e5", "810000.0"),
+            # Only the actual output is rounded up, not the expected output:
+            ("3.0", "2.99"),
```
nicoddemus: Can we also check against strings? For example this passes for me:

```
>>> import math
>>> str(math.pi)[:5]  #doctest: +NUMBER
'3.14'
```

```
λ pytest .tmp\foo.txt -q
.
1 passed in 0.01 seconds
```

When it shouldn't:

```
>>> str(math.pi)[:5]
'3.141'
```

drothlis:
> Can we also check against strings?

The implementation is going to have to be a lot smarter -- currently it just uses a regex to find anything that looks like a number within the output. I'll try to do it using [tokenize] -- it might take me a week or 2.

Thanks for the prompt review!

[tokenize]: https://docs.python.org/3.6/library/tokenize.html

nicoddemus:
> The implementation is going to have to be a lot smarter -- currently it just uses a regex to find anything that looks like a number within the output. I'll try to do it using tokenize -- it might take me a week or 2.

Changing to use tokenize might make it more robust, but if that fails, I'm OK with adding a warning to the docs mentioning this known issue.

Thanks!

nicoddemus:
Alternatively, would you like to add a note to the docs about how the matching is made, and merge this now? Then you can follow up with a refactoring to use tokenize later.

drothlis:
> would you like to add a note to the docs about how the matching is made, and merge this now? Then you can follow up with a refactoring to use tokenize later.

I'm a big fan of this approach of incremental improvement. :-)  I have pushed a commit clarifying the documentation, and with an "xfail" testcase. I also fixed a silly bug in one of my unit tests.

testing/test_doctest.py
drothlis: Addressed by noting this limitation in the documentation (see [comment below](#issuecomment-510009030)).

drothlis:
Thanks very much for the quick review & merge!

Will this be backported to pytest 4.6? Let me know if you want me to raise a pull request with the backport.

nicoddemus:
> Will this be backported to pytest 4.6? 

@drothlis 4.6 is for bug-fixes only, and as this is a new feature it won't require a backport. Thanks for the offer! 👍

PR #5603, opened 2019-07-15
Simplify SafeRepr a bit

bluetech:
Remove some dead code and make it easier to type-check.

The PR is split to commits.

### DIFF
--- a/src/_pytest/_io/saferepr.py
+++ b/src/_pytest/_io/saferepr.py
@@ -2,19 +2,23 @@ import pprint
 import reprlib
 
 
-def _call_and_format_exception(call, x, *args):
+def _format_repr_exception(exc, obj):
+    exc_name = type(exc).__name__
     try:
-        # Try the vanilla repr and make sure that the result is a string
-        return call(x, *args)
-    except Exception as exc:
-        exc_name = type(exc).__name__
-        try:
-            exc_info = str(exc)
-        except Exception:
-            exc_info = "unknown"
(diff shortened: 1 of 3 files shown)


src/_pytest/_io/saferepr.py
```diff
@@ -2,65 +2,59 @@
...
-        if len(s) > self.maxstring:
-            i = max(0, (self.maxstring - 3) // 2)
-            j = max(0, self.maxstring - 3 - i)
-            s = repr(x[:i] + x[len(x) - j :])
-            s = s[:i] + "..." + s[len(s) - j :]
-        return s
+    def repr(self, x):
+        try:
+            s = super().repr(x)
+        except Exception as exc:
```
nicoddemus: This `except` block is [not covered](https://codecov.io/gh/pytest-dev/pytest/compare/499fda234997d65b40e24f6fe3c7341f524b7d97...b47e1679e02029b847a9932e68cdf5db207991b8/diff), can you please add a test for it? Thanks!

nicoddemus:
Even though this is an internal change, I suggest adding a `5603.trivial.rst` changelog entry anyway, with something along this lines:

```rst
Simplified internal ``SafeRepr`` class and remove some dead code.
```

This is useful so if someone has a regression because of the change this can help pinpoint the problem. 👍

bluetech:
Thanks for the review @nicoddemus - updated.

bluetech:
I'm going to take the "better ask for forgiveness" approach and merge this given the approval from @nicoddemus - please let me know if that's not appropriate.

PR #5605, opened 2019-07-15
Validate xunit2 files against the schema

### LINKED ISSUE #5095: Automatically validate junitxml with its schemas
We should automatically validate the output produced by our junitxml plugin during our own testing.

This stack overflow question points to some solutions:

https://stackoverflow.com/questions/299588/validating-with-an-xml-schema-in-python

Seems like `lxml` is the more complete solution, although it might be difficult to install on some systems, but as long as our CI can handle this it should be fine.

nicoddemus:
Fix #5095

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,6 @@ def main():
         use_scm_version={"write_to": "src/_pytest/_version.py"},
         setup_requires=["setuptools-scm", "setuptools>=40.0"],
         package_dir={"": "src"},
-        # fmt: off
         extras_require={
             "testing": [
                 "argcomplete",
@@ -29,9 +28,9 @@ def main():
                 "mock",
                 "nose",
                 "requests",
-            ],
+                "xmlschema",
+            ]
         },
-        # fmt: on
         install_requires=INSTALL_REQUIRES,
     )
 
--- a/testing/test_junitxml.py
+++ b/testing/test_junitxml.py
@@ -1221,7 +1221,6 @@ def test_runs_twice(testdir, run_and_parse):
     assert first == second
 
 
-@pytest.mark.xfail(reason="hangs", run=False)
 def test_runs_twice_xdist(testdir, run_and_parse):
     pytest.importorskip("xdist")
     f = testdir.makepyfile(
--- a/testing/test_junitxml.py
+++ b/testing/test_junitxml.py
@@ -1,20 +1,47 @@
 import os
 import platform
 from datetime import datetime
+from pathlib import Path
 from xml.dom import minidom
 
 import py
+import xmlschema
 
 import pytest
 from _pytest.junitxml import LogXML
 from _pytest.reports import BaseReport
 
 
-def runandparse(testdir, *args):
-    resultpath = testdir.tmpdir.join("junit.xml")
-    result = testdir.runpytest("--junitxml=%s" % resultpath, *args)
-    xmldoc = minidom.parse(str(resultpath))
-    return result, DomNode(xmldoc)
+@pytest.fixture(scope="session")
+def schema():
+    """Returns a xmlschema.XMLSchema object for the junit-10.xsd file"""
+    fn = Path(__file__).parent / "example_scripts/junit-10.xsd"
+    with fn.open() as f:
+        return xmlschema.XMLSchema(f)
+
+
+@pytest.fixture
+def run_and_parse(testdir, schema):
+    """
+    Fixture that returns a function that can be used to execute pytest and return
+    the parsed ``DomNode`` of the root xml node.
+
+    The ``family`` parameter is used to configure the ``junit_family`` of the written report.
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -3,16 +3,16 @@
...
+    "atomicwrites>=1.0",
     "attrs>=17.4.0",
     "more-itertools>=4.0.0",
-    "atomicwrites>=1.0",
-    'pathlib2>=2.2.0;python_version<"3.6"',
-    'colorama;sys_platform=="win32"',
+    "packaging",
     "pluggy>=0.12,<1.0",
-    'importlib-metadata>=0.12;python_version<"3.8"',
+    "py>=1.5.0",
```
Zac-HD: It looks like the diff here is moving many more packages around than we're actually changing - any specific reason for that?

Plus the quotes make me think that `black` is not being applied to `setup.py`, which means we may want to check how autoformatting is configured...

blueyed: See first commit - sorts it.
Not sure if it's worth it - I'd rather leave it to avoid diff churn.

blueyed: Oh, it's not even adding to it?

Zac-HD: I don't have a strong opinion on sorting vs minimal diff, but if we're sorting it I'd go alphabetically *without* dividing by platform-specific requirements vs not.

nicoddemus: This always bothered me a bit because I think lists where the order doesn't matter should be sorted alphabetically in order to avoid conflicts.

I will remove that commit and rebase on `features`.

nicoddemus:
Updated.

Any reviewers? From my POV after this PR gets merged we can release `5.1`.

testing/test_junitxml.py
```diff
@@ -1,20 +1,52 @@
...
+@pytest.fixture(scope="session")
+def schema():
+    """Returns a xmlschema.XMLSchema object for the junit-10.xsd file"""
+    fn = Path(__file__).parent / "example_scripts/junit-10.xsd"
+    with fn.open() as f:
+        return xmlschema.XMLSchema(f)
+
+
+@pytest.fixture(name="run_and_parse")
+def run_and_parse_(testdir, schema):
```
blueyed: Why the indirection via `name`?

testing/test_junitxml.py
```diff
@@ -1,20 +1,52 @@
...
+
+        resultpath = testdir.tmpdir.join("junit.xml")
+        result = testdir.runpytest("--junitxml=%s" % resultpath, *args)
+        if validate:
+            with resultpath.open() as f:
+                schema.validate(f)
+        xmldoc = minidom.parse(str(resultpath))
+        return result, DomNode(xmldoc)
+
+    return run
```
blueyed: What's the benefit of having parametrizing with validate=True/False, if it could just validate it then?  (I see no benefit of having the test without validation, if it gets done anyway in another test)
I.e. I suggest having separate fixtures `run_and_parse` and `run_validate_and_parse` (where the latter could use the former).
This would allow to just use one of them, and remove the `@pytest.mark.parametrize("validate", [True, False])` instances.

testing/test_junitxml.py
nicoddemus: I was replacing `run_and_parse` call by a fixture, this ensures I didn't make any mistakes during development (this is the intention for the `name` parameter to `fixture`).

testing/test_junitxml.py
nicoddemus: > What's the benefit of having parametrizing with validate=True/False, if it could just validate it then? (I see no benefit of having the test without validation, if it gets done anyway in another test)

We only validate xunit2, we don't have validation for xunit1. You will notice that when `validate=True`, I "force" to use xunit2 family.

But thinking about this a bit more, I think it is more explicit to have `run_and_parse(family)` with `family` being "xunit1" or "xunit2", and validation being implicitly done for `xunit2`. What do you thik?

nicoddemus: I went ahead and made the above changes, it is now much clearer IMHO (thanks @blueyed).

@asottile would you like to review this as well?

nicoddemus:
Rebased

testing/test_junitxml.py
blueyed: Great!

FRidh:
Would it be possible to skip tests/files that require `xmlschema` when it is not present in the environment? The reason I am asking is because `pytest` is a test dependency of `xmlschema`. 

In Nixpkgs that means we now have to disable tests of `xmlschema` because we need to build `pytest` first. Clearly we have to improve our situation of circular test dependencies here (there is also `elementpath` to consider), however, that is not trivial. For now we will disable these tests. when running the `pytest` suite.

nicoddemus:
@FRidh sure, I don't see any harm on doing so. Would be glad to review/merge a PR in that direction. 👍

PR #5608, opened 2019-07-15
Remove obsolete "importorskip('unittest.mock')" calls

nicoddemus:
No longer make sense given that `unittest.mock` is part of the stdlib for all Python versions we support now.

### DIFF
--- a/testing/python/integration.py
+++ b/testing/python/integration.py
@@ -104,21 +104,15 @@ class TestMockDecoration:
         values = getfuncargnames(f)
         assert values == ("x",)
 
-    @pytest.mark.xfail(
-        strict=False, reason="getfuncargnames breaks if mock is imported"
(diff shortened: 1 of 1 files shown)


testing/python/integration.py
```diff
@@ -164,7 +162,6 @@ def test_hello(inject_me):
         reprec.assertoutcome(passed=1)
 
     def test_unittest_mock_and_pypi_mock(self, testdir):
-        pytest.importorskip("unittest.mock")
         pytest.importorskip("mock", "1.0.1")
```
bluetech: Is there still a reason for anyone to use `mock` from PyPI?

nicoddemus: Yes: https://github.com/pytest-dev/pytest-mock/pull/54

PR #5609, opened 2019-07-15
CSS changes to the API reference

timhoffm:
### CSS changes to the API reference

- Changed layout of parameters sections
  - Don't indent by the width of "Parameters". Parameter documentation is less cramped.
  - Haning indent for parameter docs: Better overview where the next parameter starts.
  - Draw a vertical line along all parameters: The parameters section is easy to distinguish from other parts of the docstring

  before/after:

  grafik grafik

- Draw a vertical line along class docstrings. Class docstrings can be quite long. The line makes it obvious that the attribute/method definitions therein are not toplevel but belong to a class.

  before/after:

  grafik grafik

- Some spacing adjustments to give the text more space to breathe.

### ReST fixes
Fixed two indentation issues in docstrings, which resulted in incorrect formatting of the rendered API reference.

### Change of section titles
- "Changelog history" -> "Changelog":
  A changelog is by definition a history of changes. No need to double the meaning; and the trem "changelog history" is rather uncommon.
- "Reference" -> "API Reference":
  Reference feels a bit to unspecific. - However this may be caused by my non-native english. Please report back if "Reference" itself is enough.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,6 @@
-=================
-Changelog history
-=================
+=========
+Changelog
+=========
 
 Versions follow `Semantic Versioning <https://semver.org/>`_ (``<major>.<minor>.<patch>``).
 
--- a/doc/en/_templates/globaltoc.html
+++ b/doc/en/_templates/globaltoc.html
@@ -4,7 +4,7 @@
   <li><a href="{{ pathto('index') }}">Home</a></li>
   <li><a href="{{ pathto('getting-started') }}">Install</a></li>
   <li><a href="{{ pathto('contents') }}">Contents</a></li>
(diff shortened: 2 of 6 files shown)


nicoddemus:
> Reference feels a bit to unspecific. - However this may be caused by my non-native english. Please report back if "Reference" itself is enough.

Sounds good to me! :+1:

timhoffm:
For some reason codecov hangs. However, codecov should not be relevant to this PR. The only change in a .py file is within a docstring.

timhoffm:
I've noticed that some style changes do not take effect on the website. - Likely because I developed with sphinx 2.1.2 but the website is built with 1.8.5.

Anyway, it's already an improvement as is - and it will get even better when switching to sphinx 2 at some point :smile:. Not going to do more about this for now.

PR #5612, opened 2019-07-16
Remove Europython training

The-Compiler:
Europython is over! The training went surprisingly well, especially considering that it was only @obestwalter and me with ~60 participants. :tada:


nicoddemus:
👍

PR #5613, opened 2019-07-16
Fix/Remove broken links

steffenschroeder:
https://github.com/pytest-dev/pytest/issues/1722 reported broken links in CHANGELOG.
This PR fixes or removes many broken links.
This is prerequisite for a follow-up PR which adds [linkcheck](https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-the-linkcheck-builder)

PR #5620, opened 2019-07-17
Fix broken venv link

jmsanders:
There's a small typo in the `venv` link that returns a 404. This change removes the trailing `/` to fix the link.


jmsanders:
Disregard - I'm closing because I see this is already addressed in:

https://github.com/pytest-dev/pytest/pull/5613/files#diff-8508591be63a9ca0fbc482e8edcbc02fL17

Issue #5624, opened 2019-07-18
cache: last_failed_paths might crash with ELOOP / does not honor/use args


blueyed:
```
% COLUMNS=80 p --lf testing
============================= test session starts ==============================
platform linux -- Python 3.7.3, pytest-4.6.1.dev186+gc9923a3a5, py-1.8.0, pluggy-0.12.0
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('…/Vcs/pytest/.hypothesis/examples')
rootdir: …/Vcs/pytest, inifile: tox.ini, testpaths: testing
plugins: hypothesis-4.23.8
collected 0 items / 1 errors
run-last-failure: 335 known failures not in selected tests

==================================== ERRORS ====================================
________________________ ERROR collecting test session _________________________
src/_pytest/cacheprovider.py:161: in last_failed_paths
    return self._last_failed_paths
E   AttributeError: 'LFPlugin' object has no attribute '_last_failed_paths'

During handling of the above exception, another exception occurred:
.venv/lib/python3.7/site-packages/pluggy/hooks.py:291: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
.venv/lib/python3.7/site-packages/pluggy/manager.py:87: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
.venv/lib/python3.7/site-packages/pluggy/manager.py:81: in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
src/_pytest/cacheprovider.py:175: in pytest_ignore_collect
    last_failed_paths = self.last_failed_paths()
src/_pytest/cacheprovider.py:165: in last_failed_paths
    result = {x for x in result if x.exists()}
src/_pytest/cacheprovider.py:165: in <setcomp>
    result = {x for x in result if x.exists()}
/usr/lib/python3.7/pathlib.py:1339: in exists
    self.stat()
/usr/lib/python3.7/pathlib.py:1161: in stat
    return self._accessor.stat(self)
E   OSError: [Errno 40] Too many levels of symbolic links: '…/Vcs/pytest/issue-4782/pkg/test_symlink.py'
=========================== short test summary info ============================
FAILED  - OSError: [Errno 40] Too many levels of symbolic links: '/home/danie...
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
[...]

PR #5625, opened 2019-07-18
[RFC] last-failed: honor args

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5624

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -166,8 +166,17 @@ class LFPlugin:
             return self._last_failed_paths
         except AttributeError:
             rootpath = Path(self.config.rootdir)
-            result = {rootpath / nodeid.split("::")[0] for nodeid in self.lastfailed}
-            result = {x for x in result if x.exists()}
+            result = []
+            abs_args = {
+                x if Path(x).is_absolute() else rootpath / x for x in self.config.args
+            }
+            for nodeid in self.lastfailed:
+                path = nodeid.split("::")[0]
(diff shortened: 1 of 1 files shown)


blueyed:
Any idea for a good test?
Reproducing https://github.com/pytest-dev/pytest/issues/5624 there might change in the future (i.e. not necessary test this anymore).
Let's see what coverage has to say.

blueyed:
A good test might come via https://github.com/pytest-dev/pytest/issues/5301.

src/_pytest/cacheprovider.py
```diff
@@ -161,8 +161,17 @@ def last_failed_paths(self):
...
+            abs_args = {
+                x if Path(x).is_absolute() else rootpath / x for x in self.config.args
+            }
+            for nodeid in self.lastfailed:
+                path = nodeid.split("::")[0]
+                abs_path = rootpath / path
+                if not any(str(abs_path).startswith(x) for x in abs_args):
+                    continue
+                if abs_path.exists():
+                    result.append(abs_path)
```
blueyed: @nicoddemus 
This was / feels very ad-hoc, but I fear there is not a better mechanism?

blueyed:
Will add a test using the plugin directly.  After https://github.com/pytest-dev/pytest/pull/6016 for likely some more "config" refactoring.

src/_pytest/cacheprovider.py
blueyed: @nicoddemus ping

PR #5630, opened 2019-07-19
doctest: handle BdbQuit

blueyed:
Map `BdbQuit` exception to `outcomes.Exit`.

This is necessary since we are not wrapping `pdb.set_trace` there, and
therefore our `do_quit` is not called.


blueyed:
Not too sure about this - just came across this when looking at fixing it for unittest.

blueyed:
Any feedback?

blueyed:
Thanks! :heart:

PR #5632, opened 2019-07-19
warnings: use _setoption without escaping for cmdline options

blueyed:
It is nice that pytest supports regular expressions (as documented by
Python itself already), but should do so also for cmdline options.

It is likely to be fixed in Python eventually.
Ref: https://github.com/python/cpython/pull/9358

Targetting features since it is a change in behavior.

TODO:
- [ ] test
- [ ] changelog

### DIFF
--- a/src/_pytest/warnings.py
+++ b/src/_pytest/warnings.py
@@ -78,7 +78,7 @@ def catch_warnings_for_item(config, ihook, when, item):
             _setoption(warnings, arg)
 
         for arg in cmdline_filters:
-            warnings._setoption(arg)
+            _setoption(warnings, arg)
 
         if item is not None:
             for mark in item.iter_markers(name="filterwarnings"):


blueyed:
Would/could need some clarification in the help then, which currently says:

> set which warnings to report, see -W option of python itself.

I've found that you can use `-o filterwarnings=…` to get the behavior from there, but still would appreciate feedback, and maybe we should have test and/or clarify the help/doc in this regard.

nicoddemus:
> I've found that you can use -o filterwarnings=… to get the behavior from there, but still would appreciate feedback, and maybe we should have test and/or clarify the help/doc in this regard.

Given that we have an alternative for the command-line (`-o`), my preference is we don't change this behavior, and instead follow what Python itself does to avoid confusion. Agree that a help/doc change about this would be highly welcome.

blueyed:
Cool, will follow up with a doc.

But note also that passing newlines in env vars (for the settings value) is not trivial, and not supported by tox yet (https://github.com/tox-dev/tox/pull/1373).

nicoddemus:
> Cool, will follow up with a doc.

Sounds good.

> But note also that passing newlines in env vars (for the settings value) is not trivial, and not supported by tox yet (tox-dev/tox#1373).

Thanks for the note, although thinking about it, `-o filterwarnings` is probably used as a quick test to pass a single option and see if a particular warning is still triggered, which is then added to `pytest.ini` or not for future uses. But I understand your point that `-o filterwarnings` and `-W` are not identical.

blueyed:
My use case is having a strict mode, where (own) DeprecationWarnings become errors, and for this to work you need to filter based on the path/filename, and for this the pattern must not be escaped.
I was to have a "strict" factor in tox (instead of just `$PYTEST_ADDOPTS` on CI) to being able to run this locally in the same way.  As for envs, you would use something like `PYTEST_ADDOPTS=$'-o filterwarnings="foo\nbar"'` (requires bash/zsh, and is maybe even not supported on Travis via its yaml (but have not checked)).

PR #5633, opened 2019-07-19
tox.ini: clean up filterwarnings

blueyed:
- path.local/path.readlines is not used anymore
- enhance doc for "invalid escape sequence" filter


blueyed:
Hmm.. this had a fixup commit.
(so now it's not only a IMHO unnecessary merge commit, but also that one)

nicoddemus:
Sorry, missed that one. 😕

blueyed:
Yeah, not your fault really (assuming that you're not against squash-merging.. ;))

PR #5634, opened 2019-07-20
unittest: handle outcomes.Exit

blueyed:
This is required for pytest to stop when using "quit" in pdb, and should
get honored/handled in general.

### DIFF
--- a/changelog/5634.bugfix.rst
+++ b/changelog/5634.bugfix.rst
@@ -1,0 +1,1 @@
+``pytest.exit`` and ``bdb.BdbQuit`` are now correctly handled in ``unittest`` cases.
--- a/src/_pytest/unittest.py
+++ b/src/_pytest/unittest.py
@@ -6,6 +6,7 @@ import _pytest._code
 import pytest
 from _pytest.compat import getimfunc
 from _pytest.config import hookimpl
+from _pytest.outcomes import exit
 from _pytest.outcomes import fail
 from _pytest.outcomes import skip
 from _pytest.outcomes import xfail
@@ -153,6 +154,11 @@ class TestCaseFunction(Function):
         self.__dict__.setdefault("_excinfo", []).append(excinfo)
 
     def addError(self, testcase, rawexcinfo):
+        try:
+            if isinstance(rawexcinfo[1], exit.Exception):
+                exit(rawexcinfo[1].msg)
+        except TypeError:
+            pass
         self._addexcinfo(rawexcinfo)
 
     def addFailure(self, testcase, rawexcinfo):
--- a/testing/test_unittest.py
+++ b/testing/test_unittest.py
@@ -1078,11 +1078,11 @@ def test_exit_outcome(testdir):
 
         class MyTestCase(unittest.TestCase):
             def test_exit_outcome(self):
-                pytest.exit("pytest_exit")
+                pytest.exit("pytest_exit called")
 
             def test_should_not_run(self):
                 pass
     """
     )
-    reprec = testdir.inline_run()
-    reprec.assertoutcome()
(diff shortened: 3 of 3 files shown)


blueyed:
Initial test, but testing for outcomes.Exit is faster / more concise:

```diff
diff --git i/testing/test_pdb.py w/testing/test_pdb.py
index 8d327cbb3..602db0aa7 100644
--- i/testing/test_pdb.py
+++ w/testing/test_pdb.py
@@ -189,6 +189,30 @@ def test_one(self):
         child.sendeof()
         self.flush(child)

+    def test_pdb_unittest_quit(self, testdir):
+        p1 = testdir.makepyfile(
+            """
+            import unittest
+
+            class MyTestCase(unittest.TestCase):
+            # class TestClass:
+                def test_one(self):
+                    __import__('pdb').set_trace()
+
+                def test_two(self):
+                    pass
+        """
+        )
+        child = testdir.spawn_pytest("%s" % p1)
+        child.expect("Pdb")
+        child.sendline("q")
+        child.expect_exact("Exit: Quitting debugger")
+        rest = child.read().decode("utf8")
+        assert "no tests ran in" in rest
+        assert "reading from stdin while output" not in rest
+        # Only printed once - not on stderr.
+        assert "Exit: Quitting debugger" not in child.before.decode("utf8")
+
     def test_pdb_print_captured_stdout_and_stderr(self, testdir):
         p1 = testdir.makepyfile(
             """
```

src/_pytest/unittest.py
```diff
@@ -153,6 +154,11 @@ def _addexcinfo(self, rawexcinfo):
         self.__dict__.setdefault("_excinfo", []).append(excinfo)
 
     def addError(self, testcase, rawexcinfo):
+        try:
+            if isinstance(rawexcinfo[1], exit.Exception):
+                exit(rawexcinfo[1].msg)
+        except TypeError:
+            pass
```
blueyed: This is ugly, but some test passes in 42 here, and I did not feel like having it in `self._addexcinfo`.

blueyed:
I think in general it might be better for unittests to use `pytest_pyfunc_call` instead of `runtest`, which would also make e.g. `--trace` work with them, but this here appears to be a more minimal fix.

testing/test_unittest.py
```diff
@@ -1050,3 +1050,39 @@ def test_setup_inheritance_skipping(testdir, test_name, expected_outcome):
...
+        class MyTestCase(unittest.TestCase):
+            def test_exit_outcome(self):
+                pytest.exit("pytest_exit")
+
+            def test_should_not_run(self):
+                pass
+    """
+    )
+    reprec = testdir.inline_run()
+    reprec.assertoutcome()
```
blueyed: Should assert more here..

changelog/5634.bugfix.rst
```diff
@@ -0,0 +1 @@
+``pytest.exit`` and ``bdb.BdbQuit`` are now correctly handled in ``unittest`` cases.
```
blueyed: Behavior with `bdb.BdbQuit` is not changed.

testing/test_unittest.py
```diff
@@ -1078,11 +1078,11 @@ def test_exit_outcome(testdir):
...
+                pytest.exit("pytest_exit called")
 
             def test_should_not_run(self):
                 pass
     """
     )
-    reprec = testdir.inline_run()
-    reprec.assertoutcome()
+    result = testdir.runpytest()
+    result.stdout.fnmatch_lines("*Exit: pytest_exit called*")
```
blueyed: This changed it to have less assertions - previously it asserted that no test was run, no it does not anymore.

PR #5643, opened 2019-07-22
typo

wjwwood:
Just a small typo I found while reading.


The-Compiler:
You'll need to change the link label a few lines down as well - currently the build fails with `ERROR README.rst:114 Unknown target name: "tidelift subscription".`.

blueyed:
~Please squash / force-push then also~

Amended/pushed in 13d8577.

Thanks!

Issue #5579, opened 2019-07-08
No longer can extract message from exceptions in 5+ versions


cscetbon:
I can't extract the message of an exception using str(my_exception) when using pytest 5+. Here is an example :

```
E       AssertionError: assert 'blabla' in '<ExceptionInfo Exception tblen=1>'
E        +  where '<ExceptionInfo Exception tblen=1>' = str(<ExceptionInfo Exception tblen=1>)
```

That test succeeds in previous versions (4.6.4 for instance)

I've been able to test it using the following example :
```
import pytest

def test_ex():
    with pytest.raises(Exception) as e:
        raise Exception('blabla')

    assert 'blabla' in str(e)
```

banderson5586:
assert 'blabla' in str(e.value)

Haven't looked at release notes yet to see why but accessing the exception text changed slightly.

cscetbon:
@banderson5586 but why accessing the error message using pytest is different from the way it's done in python 3 ? it does not make any sense to me.

nicoddemus:
Hi,

Here's the related CHANGELOG entry:

> #5412: ExceptionInfo objects (returned by pytest.raises) now have the same str representation as repr, which avoids some confusion when users use print(e) to inspect the object.

So it was mostly to avoid confusion. Note that `pytest.raises` returns a `ExceptionInfo` object when used in a `with` context, *not* the exception object itself. 

Here are the docs for `ExceptionInfo`: https://docs.pytest.org/en/latest/reference.html#exceptioninfo

As can be seen in the docs, `ExceptionInfo` objects contain a lot of other attributes, so what happened before is that it was a little confusing when someone printed an `ExceptionInfo` object and it behave as if it was the exception *value*.

cscetbon:
Understood. This change is probably gonna break a ton of unit tests ...

nicoddemus:
Let's see, we are hopíng not, as this was not really documented behavior.

I'm closing this for now, feel free to follow up with further questions though.

stadelmanma:
@nicoddemus I understand the change but it did smash a good bit of my test suite out of the blue when my CI suite pulled v5 instead of v4.X. I didn't realize I was using undocumented behavior which in the end is on me but I suspect several others will get hit in the face by this as well. 

[...]

PR #5648, opened 2019-07-23
Make ExceptionInfo.repr change more prominent

nicoddemus:
Related to #5579


stadelmanma:
Looks good! That is all I would have needed to see to realize the problem.

nicoddemus:
> Looks good! That is all I would have needed to see to realize the problem.

Sorry, that's on us for 1) don't realize it would be such a problem, 2) don't realize it wouldn't be abvious.

Issue #5647, opened 2019-07-23
--fulltrace is not mentioned in --help


anntzer:
`pytest --help` mentions `--full-trace`, but not `--fulltrace`; compare with the message on KeyboardInterrupt: "(to show a full traceback on KeyboardInterrupt use --fulltrace)".  Having the same spelling for the option on both sides (even if keeping the other spelling as an alias) may save a bit of head-scratching.

Python 3.7 on Arch Linux, clean venv with just pytest pip-installed.

```
$ pip list
Package            Version
------------------ -------
atomicwrites       1.3.0  
attrs              19.1.0 
importlib-metadata 0.18   
more-itertools     7.2.0  
packaging          19.0   
pip                19.0.3 
pluggy             0.12.0 
py                 1.8.0  
pyparsing          2.4.1  
pytest             5.0.1  
setuptools         40.8.0 
six                1.12.0 
wcwidth            0.1.7  
zipp               0.5.2
```

- [X] a detailed description of the bug or suggestion
- [X] output of `pip list` from the virtual environment you are using
- [X] pytest and operating system versions
- [X] minimal example if possible

blueyed:
Makes sense.  Please consider creating a PR.

noviluni:
I think that this issue is already resolved. :)

nicoddemus:
Yes, thanks!

PR #5668, opened 2019-07-27
Fix the confusing different names of `fulltrace`

charlesdong1991:
Fix the confusing different names of `fulltrace` showed in `pytest --help` (#5647)

Since this is a small fix, and I don't think this needs to create a changelog, please let me know if needed.


nicoddemus:
Thanks @charlesdong1991 for the PR, but the fix for #5647 should be a docs update, not removing the `--full-trace` option.

nicoddemus:
Thanks @charlesdong1991 for the follow up.

Issue #5516, opened 2019-06-28
Tests collection is slow on high number of tests


Nepherhotep:
Hi,

We have over 15k tests in the repository, and recently switched from nosetests to pytest. Apparently, tests collection in pytest is way slower than in nosetests - 130s vs 30s. Also, I need to mention that `norecursedirs` and `testpaths` parameters are both set, just to exclude the case of walking in .git or other unrelated directories.
It's also interesting, that the number of collected tests is asymptotic to time:
![image](https://user-images.githubusercontent.com/2122618/60345588-5eb3ff00-9987-11e9-8ef1-59da7c43c9a3.png)

I tried to profile it using pyflame and cProfile, it seems like the slowest part is calling node.py:FixtureManager._matchfactories method call:
![image](https://user-images.githubusercontent.com/2122618/60345970-2cef6800-9988-11e9-9b75-eb99ed216a23.png)

Here is a callgraph:
![image](https://user-images.githubusercontent.com/2122618/60346065-61632400-9988-11e9-86ce-b09eb99c949c.png)

When I tried to log, what's is happening, it seems like, the more tests, the more method is called with polynomial dependency.

Going deeper into `_matchfactories` method, I was able to reduce collection time from 130s to 85s just by caching the results of `_splitnode` method:
```diff
diff --git a/src/_pytest/nodes.py b/src/_pytest/nodes.py
index 491cf9d2c..62eb66086 100644
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -13,6 +13,9 @@ SEP = "/"
 tracebackcutdir = py.path.local(_pytest.__file__).dirpath()
 
 
+_node_cache = {}
+
+
 def _splitnode(nodeid):
     """Split a nodeid into constituent 'parts'.
 
@@ -31,9 +34,13 @@ def _splitnode(nodeid):
     if nodeid == "":
         # If there is no root node at all, return an empty list so the caller's logic can remain sane
         return []
-    parts = nodeid.split(SEP)
-    # Replace single last element 'test_foo.py::Bar' with multiple elements 'test_foo.py', 'Bar'
-    parts[-1:] = parts[-1].split("::")
+
+    parts = _node_cache.get(nodeid)
+    if not parts:
+        parts = nodeid.split(SEP)
+        # Replace single last element 'test_foo.py::Bar' with multiple elements 'test_foo.py', 'Bar'
+        parts[-1:] = parts[-1].split("::")
[...]

PR #5681, opened 2019-08-01
Cache splitnode results to improve tests collection time

Nepherhotep:
https://github.com/pytest-dev/pytest/issues/5516

On a large number of tests, collection time can take over 2 minutes for 20k of tests. 60% of the time is spent on splitting nodes, so caching nodesplit results will improve performance considerably.

### DIFF
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -1,5 +1,6 @@
 import os
 import warnings
+from functools import lru_cache
 
 import py
 
@@ -13,6 +14,7 @@ SEP = "/"
 tracebackcutdir = py.path.local(_pytest.__file__).dirpath()
 
 
+@lru_cache(maxsize=None)
 def _splitnode(nodeid):
     """Split a nodeid into constituent 'parts'.
 
@@ -30,11 +32,12 @@ def _splitnode(nodeid):
     """
     if nodeid == "":
         # If there is no root node at all, return an empty list so the caller's logic can remain sane
-        return []
+        return ()
     parts = nodeid.split(SEP)
     # Replace single last element 'test_foo.py::Bar' with multiple elements 'test_foo.py', 'Bar'
     parts[-1:] = parts[-1].split("::")
-    return parts
+    # Convert parts into a tuple to avoid possible errors with caching of a mutable type
+    return tuple(parts)
 
 
 def ischildnode(baseid, nodeid):
--- a/changelog/5516.trivial.rst
+++ b/changelog/5516.trivial.rst
@@ -1,0 +1,1 @@
+Cache node splitting function which can improve collection performance in very large test suites.


src/_pytest/nodes.py
```diff
@@ -13,6 +14,7 @@
 tracebackcutdir = py.path.local(_pytest.__file__).dirpath()
 
 
+@lru_cache(maxsize=None)
```
asottile: a cache of mutable return values is asking for trouble -- if this returned `tuple`s I'd be more comfortable

nicoddemus: Yeah, it would be safer. @Nepherhotep would you mind changing that so we can see how it fares?

asottile:
@Nepherhotep can you show the difference in cpu time / memory consumption? `lru_cache(maxsize=None)` is ~usually a memory leak -- want to make sure the trade offs are correct here or if we should be looking to refactor the code such that it doesn't do as many repeated lookups instead

Nepherhotep:
> @Nepherhotep can you show the difference in cpu time / memory consumption? `lru_cache(maxsize=None)` is ~usually a memory leak -- want to make sure the trade offs are correct here or if we should be looking to refactor the code such that it doesn't do as many repeated lookups instead
For ~15k tests, collection time decreases from 130 to 90 seconds (i7 8-gen CPU), as for memory, I don't have this information right away. I'll try to collect it.
I don't think memory increase is substantial - the problem is in splitting the same node path multiple times, but the number of nodes is relatively small.

src/_pytest/nodes.py
Nepherhotep: @nicoddemus, a simple converting response format from list to tuple caused 1500 errors. Most of them failed with the error "fixture testdir not found", some of them failed with repr error.
I need to dig into "testdir" fixture issue a bit deepr.

nicoddemus: I'm surprised... strange.

nicoddemus:
> I don't think memory increase is substantial - the problem is in splitting the same node path multiple times

I agree, the cache for this function should get equal to the number of nodeids in the section, and that number is fixed and relatively small.

src/_pytest/nodes.py
nicoddemus: Oh you missed the early return:

```python
    if nodeid == "":
        # If there is no root node at all, return an empty list so the caller's logic can remain sane
        return []
```

So sometimes it would return a tuple, others a list.

Nepherhotep: Ah, my bad, thanks for pointing to that!

src/_pytest/nodes.py
```diff
@@ -30,11 +32,12 @@ def _splitnode(nodeid):
     """
     if nodeid == "":
         # If there is no root node at all, return an empty list so the caller's logic can remain sane
-        return []
+        return tuple()
```
asottile:
```suggestion
        return ()
```

nicoddemus:
Great, thanks! Would you mind squashing the commits before merging? Or I would be happy to do it myself if you let me. 👍

Nepherhotep:
@nicoddemus done.
I wonder, though, why don't you configure GitHub project to do "Squash and Merge" pull request by default - it's a very convenient setup :)

nicoddemus:
@asottile doesn't like it, IIRC.

PR #5684, opened 2019-08-01
Use OSError subclasses instead of handling errno

nicoddemus:
img

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -116,24 +116,11 @@ class AssertionRewritingHook:
         write = not sys.dont_write_bytecode
         cache_dir = os.path.join(os.path.dirname(fn), "__pycache__")
         if write:
-            try:
-                os.mkdir(cache_dir)
-            except OSError:
-                e = sys.exc_info()[1].errno
-                if e == errno.EEXIST:
-                    # Either the __pycache__ directory already exists (the
-                    # common case) or it's blocked by a non-dir node. In the
-                    # latter case, we'll ignore it in _write_pyc.
-                    pass
-                elif e in {errno.ENOENT, errno.ENOTDIR}:
-                    # One of the path components was not a directory, likely
-                    # because we're in a zip file.
-                    write = False
-                elif e in {errno.EACCES, errno.EROFS, errno.EPERM}:
-                    state.trace("read only directory: %r" % os.path.dirname(fn))
-                    write = False
-                else:
-                    raise
+            ok = try_mkdir(cache_dir)
+            if not ok:
+                write = False
+                state.trace("read only directory: {}".format(os.path.dirname(fn)))
+
         cache_name = os.path.basename(fn)[:-3] + PYC_TAIL
         pyc = os.path.join(cache_dir, cache_name)
         # Notice that even if we're in a read-only directory, I'm going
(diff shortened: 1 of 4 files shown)


src/_pytest/assertion/rewrite.py
```diff
@@ -118,22 +117,19 @@ def exec_module(self, module):
...
+            except FileExistsError:
+                # Either the __pycache__ directory already exists (the
+                # common case) or it's blocked by a non-dir node. In the
+                # latter case, we'll ignore it in _write_pyc.
+                pass
+            except (FileNotFoundError, NotADirectoryError):
+                # One of the path components was not a directory, likely
+                # because we're in a zip file.
+                write = False
+            except PermissionError:
```
nicoddemus: > exception PermissionError
Raised when trying to run an operation without the adequate access rights - for example filesystem permissions. Corresponds to errno EACCES and EPERM.

I couldn't find an exception corresponding to `errno.EROFS` (Read-only file system), but I suppose it will fall under the umbrella of `PermissionError`.

src/_pytest/assertion/rewrite.py
```diff
@@ -118,22 +117,19 @@ def exec_module(self, module):
...
-                if e == errno.EEXIST:
-                    # Either the __pycache__ directory already exists (the
-                    # common case) or it's blocked by a non-dir node. In the
-                    # latter case, we'll ignore it in _write_pyc.
-                    pass
-                elif e in {errno.ENOENT, errno.ENOTDIR}:
-                    # One of the path components was not a directory, likely
-                    # because we're in a zip file.
-                    write = False
-                elif e in {errno.EACCES, errno.EROFS, errno.EPERM}:
```
nicoddemus: > exception PermissionError
Raised when trying to run an operation without the adequate access rights - for example filesystem permissions. Corresponds to errno EACCES and EPERM.

I couldn't find an exception corresponding to `errno.EROFS` (Read-only file system), but I suppose it will fall under the umbrella of `PermissionError`.

The-Compiler: I can't find it handled [in the CPython source](https://github.com/python/cpython/blob/v3.8.0b3/Objects/exceptions.c#L2603-L2623) so I'm guessing it'll be a generic `OSError`.

nicoddemus: Ahhh thanks for checking. I will update the code to handle that then. 👍

nicoddemus: Done, also refactored that into a separate function.

src/_pytest/assertion/rewrite.py
```diff
@@ -113,27 +113,14 @@ def exec_module(self, module):
         # tricky race conditions, we maintain the following invariant: The
         # cached pyc is always a complete, valid pyc. Operations on it must be
         # atomic. POSIX's atomic rename comes in handy.
-        write = not sys.dont_write_bytecode
+        write = not sys.dont_write_bytecode  # love me a double negative
```
blueyed: The comment is not necessary, is it?

nicoddemus: I was feeling snarky at the time and added it. Want me to remove it?

blueyed: I can see that.. :)
I'd rather not have it, since it adds no value - and gets read unnecessarily.

nicoddemus: Removed

blueyed: Ok, should not land in the Git history though also - i.e. please squash after approval.

nicoddemus:
Will do

PR #5691, opened 2019-08-03
[4.6] Handle only known functions in rm_rf (#5627)

nicoddemus:
Backport of #5627

Conflicts:
- 	src/_pytest/pathlib.py

Also had to adapt:

- PermissionError into OSError with appropriate errno
- Change keyword-only argument to `**kwargs`
- Remove type annotations

### DIFF
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -11,6 +11,7 @@ import shutil
 import sys
 import uuid
 import warnings
+from functools import partial
 from functools import reduce
 from os.path import expanduser
 from os.path import expandvars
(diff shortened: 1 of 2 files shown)


src/_pytest/pathlib.py
```diff
@@ -46,38 +47,53 @@ def ensure_reset_dir(path):
     path.mkdir()
 
 
-def rm_rf(path):
-    """Remove the path contents recursively, even if some elements
-    are read-only.
-    """
+def on_rm_rf_error(func, path, exc, **kwargs):
+    """Handles known read-only errors during rmtree."""
+    start_path = kwargs["start_path"]
```
asottile: should this be `kwargs.pop(...)` and then `if not kwargs: raise TypeError(...)` such that we retain the api from 5.x?

nicoddemus: Good point, but given that this is called by Python's stdlib I don'think it actually matters.

asottile: ok, as long as it's not an api we could ever expose to users 👍

PR #5694, opened 2019-08-04
Added info to access exitcodes directly

yashtodi94:
Relates to:

1. [pytest#5125](https://github.com/pytest-dev/pytest/issues/5125)
2. [pytest#5420](https://github.com/pytest-dev/pytest/pull/5420)

Better visibility on usage of exitcodes.


asottile:
oof, bad rng you got the creepy not-cat

PR #5697, opened 2019-08-05
Fix RuntimeError when trying to collect package with "__init__.py" only

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/4344.


blueyed:
Will do the backport.

blueyed:
Meant to create a PR, but pushed it accidentally already: b71f87318.

PR #5705, opened 2019-08-06
pyupgrade docs to Python 3

verhovsky:
ran [pyupgrade-docs](https://github.com/verhovsky/pyupgrade-docs/blob/master/pyupgrade_docs.py) (it's just a copy of [asottile/blacken-docs](https://github.com/asottile/blacken-docs/blob/master/blacken_docs.py) but with pyupgrade instead) with `--py3-only` 

then I labeled all the unlabeled code blocks I could find (except the ones in `/doc/en/announce`) as python/text/bash because pyupgrade-docs and blacken-docs don't pick up unlabeled `code-block`s

then I re-ran pyupgrade and also blacken-docs

### DIFF
--- a/doc/en/builtin.rst
+++ b/doc/en/builtin.rst
@@ -162,7 +162,7 @@ For information about fixtures, see :ref:`fixtures`. To see a complete list of a
 
     no tests ran in 0.12 seconds
 
-You can also interactively ask for help, e.g. by typing on the Python interactive prompt something like
+You can also interactively ask for help, e.g. by typing on the Python interactive prompt something like::
 
 .. code-block:: python
 
--- a/doc/en/builtin.rst
+++ b/doc/en/builtin.rst
@@ -164,5 +164,7 @@ For information about fixtures, see :ref:`fixtures`. To see a complete list of a
 
 You can also interactively ask for help, e.g. by typing on the Python interactive prompt something like::
 
+.. code-block:: python
+
     import pytest
     help(pytest)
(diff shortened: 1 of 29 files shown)


doc/en/example/parametrize.rst
```diff
@@ -173,22 +187,27 @@ an add-on from Robert Collins for the standard unittest framework. We
...
 
+
     def pytest_generate_tests(metafunc):
         idlist = []
         argvalues = []
         for scenario in metafunc.cls.scenarios:
             idlist.append(scenario[0])
             items = scenario[1].items()
             argnames = [x[0] for x in items]
-            argvalues.append(([x[1] for x in items]))
```
verhovsky: this might look like a mistake, but it's not.

doc/en/cache.rst
```diff
@@ -35,13 +35,16 @@ Rerunning only failures or failures first
 
 First, let's create 50 test invocation of which only 2 fail::
```
nicoddemus: If one adds a code-block, we need to remove the implicit code block (`::`) from the previous line, that's why the build is failing.

doc/en/builtin.rst
```diff
@@ -162,7 +162,7 @@ For information about fixtures, see :ref:`fixtures`. To see a complete list of a
 
     no tests ran in 0.12 seconds
 
-You can also interactively ask for help, e.g. by typing on the Python interactive prompt something like::
+You can also interactively ask for help, e.g. by typing on the Python interactive prompt something like
```
nicoddemus: Oh sorry I should have been more clear, the `::` should be replaced by `:`, as the `:` makes gramatical sense.

blueyed:
@verhovsky 
Thanks for the contribution.
For the future it would be better to not have reverted commits, but rather remove them by force-pushing.  Also @nicoddemus of course, who could have asked for this / squashing.

PR #5711, opened 2019-08-08
Add a code of conduct

The-Compiler:
See #5709 and the mails linked from there. I added a list of people included in coc@pytest.org at the bottom.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,84 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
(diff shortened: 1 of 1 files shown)


nicoddemus:
Looks good!

I believe we should add this to the docs (so it needs to be converted to `.rst`), and link it from the README too to gain more visibility.

besides being on the docs, I think it deserves to appear on the side bar too:

image

Perhaps after "Contributing"?

The-Compiler:
Sounds good as long as GitHub still picks it up as well. @nicoddemus if you have the time, could you maybe pick this up? I'm quite busy with exams so I can't commit to continuing working on this until next month.

nicoddemus:
Hmmm oh right: https://help.github.com/en/articles/adding-a-code-of-conduct-to-your-project

It seems it only supports `.md`. Well in that case I suppose we can close this, as it will get enough visibility by GitHub's own means (issues and PRs).

PR #5713, opened 2019-08-09
Add `enabled()` context manager to output capture fixtures, to force capture even if it is disabled globally

fjarri:
Use case: testing a block of code whose output must be captured (and checked). Since the test suite might be executed with or without `-s` flag, there should be a way to force output capture in a block if the output is used in some assertions. For example:

    def test_enabling_capturing(capsys):
        print("this output may be captured")
        with capsys.enabled():
            print("this output is always captured")
        captured = capsys.readouterr()
        assert "this output is always captured\n" in captured.out

The current PR is made by analogy with `CaptureFixture.disabled()`, and has a drawback: if `-s` is not given, there is no way to separate the output in the `enabled()` block from the output of the rest of the testcase (that is, in the example above `captured.out == "this output may be captured\nthis output is always captured\n"`). Ideally, it would be more convenient to capture only the output of the block, but I am not sure how to approach this without breaking the existing capturing mechanic.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -41,6 +41,7 @@ Ben Webb
 Benjamin Peterson
 Bernard Pratz
 Bob Ippolito
+Bogdan Opanchuk
 Brian Dorsey
 Brian Maissy
 Brian Okken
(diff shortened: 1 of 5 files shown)


fjarri:
Apparently I misunderstood the output capturing logic in `py.test`. The output is always captured; if `-s` is given, it is shown when the testcase is finished, otherwise it is shown in the end report (if the testcase failed). Also, `capsys.readouterr()` consumes all previously captured output. With this approach an `.enabled()` method makes no sense.

PR #5731, opened 2019-08-12
assertrepr_compare: prefer same maxsize

blueyed:
Previously it would say:

    > assert '123456789012...901234567890A' == '1234567890123...901234567890B'"

This makes it look like the "3" might be different already.

This is clearer, and it is OK to have potentially one less char in the
right one:

    > assert '123456789012...901234567890A' == '123456789012...901234567890B'"

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -119,9 +119,9 @@ def isiterable(obj):
 
 def assertrepr_compare(config, op, left, right):
     """Return specialised explanations for some operators/operands"""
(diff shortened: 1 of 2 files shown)


nicoddemus:
We could use a `trivial` changelog entry I think

blueyed:
@nicoddemus 
> We could use a `trivial` changelog entry I think

I don't think so.

But in general, what do you think about having a way to also look at commit messages when generating the changelog?
Then it could be added to the commit message when merging - without extra pushing etc.

nicoddemus:
Not sure, we would need to see if it would be worth it

PR #5732, opened 2019-08-12
assertion.rewrite: use ast.NameConstant(None) directly

blueyed:
`ast.parse("None")` was added/used in 3e6f1fa2d for differences between
Python 2/3, but we do not support py2 anymore.


blueyed:
Thanks for the review!

> I think using constants for this is a bit much, but OK either way.

Not sure either, but thought that this gets called/used a lot after all.

bluetech:
This seems trivial enough for me to merge, so I did. Also to avoid conflict when rebasing the above mentioned PR.

PR #5736, opened 2019-08-13
pytester: spawn: skip without pexpect.spawn

blueyed:
"pexpect" can be imported on Windows, but does not have "pexpect.spawn"
then.

Ref: https://github.com/pexpect/pexpect/blob/a803933ed53/pexpect/__init__.py#L73-L76


nicoddemus:
Thanks!

PR #5740, opened 2019-08-14
Use TRAVIS_REPO_SLUG instead of hard-coding pytest-dev/pytest

nicoddemus:
I was doing final tests on the script today, and forgot to change
the hardecoded "pytest-dev/pytest", which ended up publishing a
`4.99.10` release to the main repository by mistake, as my token
has access to both my fork and main repository. 🤕 

I deleted the tag immeditely just a few seconds later, so hopefully
this won't cause major problems.

This change makes it safer to test this in the future, never publishing
to the main repository by mistake (as long as the tags are pushed to the
right repositories of course).

### DIFF
--- a/scripts/publish_gh_release_notes.py
+++ b/scripts/publish_gh_release_notes.py
@@ -25,9 +25,10 @@ import github3
 import pypandoc
 
 
-def publish_github_release(token, tag_name, body):
+def publish_github_release(slug, token, tag_name, body):
     github = github3.login(token=token)
-    repo = github.repository("pytest-dev", "pytest")
+    owner, repo = slug.split("/")
+    repo = github.repository(owner, repo)
     return repo.create_release(tag_name=tag_name, body=body)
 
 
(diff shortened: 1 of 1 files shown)


scripts/publish_gh_release_notes.py
```diff
@@ -71,14 +72,19 @@ def main(argv):
         print("GH_RELEASE_NOTES_TOKEN not set", file=sys.stderr)
         return 1
 
+    slug = os.environ.get("TRAVIS_REPO_SLUG")
+    if not slug:
+        print("TRAVIS_REPO_SLUG not set", file=sys.stderr)
+        return 1
```
blueyed: Just use `slug = os.environ["TRAVIS_REPO_SLUG"]` maybe to get a traceback also?  (i.e. more info)
But it's OK like that also.

scripts/publish_gh_release_notes.py
```diff
@@ -71,14 +72,19 @@ def main(argv):
...
     rst_body = parse_changelog(tag_name)
     md_body = convert_rst_to_md(rst_body)
-    if not publish_github_release(token, tag_name, md_body):
+    if not publish_github_release(slug, token, tag_name, md_body):
         print("Could not publish release notes:", file=sys.stderr)
         print(md_body, file=sys.stderr)
         return 5
 
-    print(f"Release notes for {tag_name} published successfully")
+    print(f"Release notes for {tag_name} published successfully to {slug}")
```
blueyed: Might be cool to have a link to it there, assuming it's in the response somehow.
(the next time you debug/change it ;))

scripts/publish_gh_release_notes.py
nicoddemus: I think the message is pretty clear TBH 😁

scripts/publish_gh_release_notes.py
nicoddemus: Great idea, done. 👍

scripts/publish_gh_release_notes.py
blueyed: Sure.
I've just meant that just asserting it to be there is enough/shorter, and provides validation (all in a single statement).

PR #5742, opened 2019-08-15
Async result warn

graingert:

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -151,10 +151,7 @@ def pytest_configure(config):
 
 @hookimpl(trylast=True)
 def pytest_pyfunc_call(pyfuncitem):
-    testfunction = pyfuncitem.obj
-    if iscoroutinefunction(testfunction) or (
-        sys.version_info >= (3, 6) and inspect.isasyncgenfunction(testfunction)
-    ):
+    def async_warn():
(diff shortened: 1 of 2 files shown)


nicoddemus:
Follow up to https://github.com/pytest-dev/pytest/pull/5734 I assume? Can you please rebase given that https://github.com/pytest-dev/pytest/pull/5734 was merged already? Thanks

nicoddemus:
Do we really need this? I mean, Python itself already issues a warning which is clear enough IMHO:

```python
async def foo():
    pass

def test_foo():
    return foo()
```

```
====================== warnings summary ======================
test_async.py::test_foo
  c:\users\bruno.esss.co\pytest\src\_pytest\python.py:165: RuntimeWarning: coroutine 'foo' was never awaited
    testfunction(**testargs)

```

nicoddemus:
Oh for the future: new warnings should target `features`, because they might break test suites. No need to change this now because we are about to release 5.1.0. 👍

graingert:
I'm experiencing the problem with twisted Deferreds, however. Where they are awaitable but no error is thrown if you do not await them and there is no way to determine if a function will return one statically like inspect.iscoro...

nicoddemus:
Oh I see, thanks!

PR #5746, opened 2019-08-15
ci: add codecov.yml to turn comments off

blueyed:
The only benefit for me is to get notified about finished builds, but
that might happen to early anyway.  Apart from that they are rather big
and distract from actual comments.


blueyed:
@asottile 
We could have it with a dot in front also.

Issue #5754, opened 2019-08-17
Reported location for assertion error is off with coverage


blueyed:
Locations for assertion errors are off when using pytest with coverage (pytest-cov):

```
    def test_config():
        assert False
>       assert False
E       assert False
```

Bisected to 39ba99613.

More information (original issue):

In https://travis-ci.com/pdbpp/fancycompleter/jobs/226053243#L153 it reports this error:
```
================================== FAILURES ===================================
_________________________________ test_config _________________________________
tmphome = local('C:\\Users\\travis\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_config0')
capsys = <_pytest.capture.CaptureFixture object at 0x0000026CE8AD3508>
    def test_config(tmphome, capsys):
        class DefaultCfg:
            default = 42
    
        class MyCfg(ConfigurableClass):
            DefaultConfig = DefaultCfg
            config_filename = ".mycfg"
    
        cfg = MyCfg()
    
        # Calls passed in Config.
        assert cfg.get_config(Config=lambda: "42") == "42"
    
        # Uses DefaultConfig instance.
        assert isinstance(cfg.get_config(None), DefaultCfg)
    
        cfgfile = tmphome.join(MyCfg.config_filename)
    
        # Handles empty config file (missing "Config").
        cfgfile.ensure()
        assert isinstance(cfg.get_config(None), DefaultCfg)
        out, err = capsys.readouterr()
        assert out == ""
        assert err == ""
    
        # Exception when instantiating Config.
        p = tmphome.ensure(MyCfg.config_filename)
        print(p, file=capsys._capture.out._old)
        cfgfile.write("def Config(): raise Exception('my_exc')")
        assert isinstance(cfg.get_config(None), DefaultCfg)
        out, err = capsys.readouterr()
        assert out == ""
        assert err == (
            "** error when setting up Config from ~/.mycfg: my_exc (%s:1) **\n" % p
        )
        print("passed first", file=capsys._capture.out._old)
    
        # Error during execfile.
        config_filename = tmphome.ensure(MyCfg.config_filename)
        print("config_filename:", config_filename, file=capsys._capture.out._old)
        cfgfile.write("raise Exception('my_execfile_exc')")
        assert isinstance(cfg.get_config(None), DefaultCfg)
        out, err = capsys.readouterr()
        assert out == ""
>       assert err == "** error when importing ~/.mycfg: my_execfile_exc **\n"
[...]

PR #5755, opened 2019-08-17
test_assertion_location

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5754

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -33,9 +33,6 @@ PYTEST_TAG = "{}-pytest-{}".format(sys.implementation.cache_tag, version)
 PYC_EXT = ".py" + (__debug__ and "c" or "o")
 PYC_TAIL = "." + PYTEST_TAG + PYC_EXT
 
-AST_IS = ast.Is()
-AST_NONE = ast.NameConstant(None)
-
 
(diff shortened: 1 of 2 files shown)


blueyed:
Failed as expected under coverage: https://travis-ci.org/pytest-dev/pytest/jobs/573192533#L786

/cc @nedbat @asottile 
Any idea what is happening here?

testing/test_assertrewrite.py
```diff
@@ -1580,3 +1580,21 @@ def fake_mkdir(p, *, exc):
     with pytest.raises(OSError) as exc_info:
         try_mkdir(str(p))
     assert exc_info.value.errno == errno.ECHILD
+
+
+def test_assertion_location(testdir):
```
blueyed: Should/could be included with an existing basic test probably.

blueyed: Could not find a better place.. :/

asottile:
since it's global, all `assert`s will share those ast instances (with last-one-wins positioning information) -- not sure why coverage changes it though 🤔 

patch looks good though 👍

blueyed:
@asottile 
Thanks!

@nedbat 
Still interested in your input - might also indicate some bug / issue with Coverage.py after all.

PR #5760, opened 2019-08-17
minor: make noqa comment more specific

blueyed:


src/_pytest/compat.py
```diff
@@ -27,9 +27,9 @@
 
 
 if sys.version_info >= (3, 8):
-    from importlib import metadata as importlib_metadata  # noqa
+    from importlib import metadata as importlib_metadata  # noqa: F401
```
graingert: If you defined an __all__ you wouldn't need this noqa

PR #5764, opened 2019-08-18
Gracefully handle HTTP errors from pastebin

goerz:
We find that the `--pastebin` option to `pytest` sometimes fails with "HTTP Error 400: Bad Request". We're still investigating the exact cause of these errors, but in the meantime, a failure to upload to the pastebin service should probably not crash pytest and cause a test failure in the continuous-integration.

This patch catches exceptions like `HTTPError` that may be thrown while trying to communicate with the pastebin service, and reports them as a "bad response", without crashing with a backtrace or failing the entire test suite.

### DIFF
--- a/src/_pytest/pastebin.py
+++ b/src/_pytest/pastebin.py
@@ -59,7 +59,7 @@ def create_new_paste(contents):
     Creates a new paste using bpaste.net service.
 
     :contents: paste contents as utf-8 encoded bytes
-    :returns: url to the pasted contents
+    :returns: url to the pasted contents or error message
     """
     import re
     from urllib.request import urlopen
@@ -67,12 +67,17 @@ def create_new_paste(contents):
 
     params = {"code": contents, "lexer": "python3", "expiry": "1week"}
     url = "https://bpaste.net"
-    response = urlopen(url, data=urlencode(params).encode("ascii")).read()
-    m = re.search(r'href="/raw/(\w+)"', response.decode("utf-8"))
+    try:
+        response = (
+            urlopen(url, data=urlencode(params).encode("ascii")).read().decode("utf-8")
+        )
+    except OSError as exc_info:  # urllib errors
+        return "bad response: %s" % exc_info
+    m = re.search(r'href="/raw/(\w+)"', response)
     if m:
         return "{}/show/{}".format(url, m.group(1))
     else:
-        return "bad response: " + response.decode("utf-8")
+        return "bad response: invalid format ('" + response + "')"
 
 
 def pytest_terminal_summary(terminalreporter):
--- a/AUTHORS
+++ b/AUTHORS
@@ -173,6 +173,7 @@ mbyt
 Michael Aquilina
 Michael Birtwell
 Michael Droettboom
+Michael Goerz
 Michael Seifert
 Michal Wajszczuk
 Mihai Capotă
--- a/changelog/5764.feature.rst
+++ b/changelog/5764.feature.rst
@@ -1,0 +1,1 @@
+New behavior of the ``--pastebin`` option: failures to connect to the pastebin server are reported, without failing the pytest run
--- a/testing/test_pastebin.py
+++ b/testing/test_pastebin.py
@@ -83,6 +83,47 @@ class TestPaste:
         return request.config.pluginmanager.getplugin("pastebin")
 
     @pytest.fixture
+    def mocked_urlopen_fail(self, monkeypatch):
+        """
+        monkeypatch the actual urlopen call to emulate a HTTP Error 400
+        """
+        calls = []
+
+        import urllib.error
+        import urllib.request
+
+        def mocked(url, data):
+            calls.append((url, data))
+            raise urllib.error.HTTPError(url, 400, "Bad request", None, None)
+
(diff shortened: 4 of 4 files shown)


goerz:
I don't understand what pytest-CI is complaining about.

src/_pytest/pastebin.py
```diff
@@ -67,8 +67,13 @@ def create_new_paste(contents):
 
     params = {"code": contents, "lexer": "python3", "expiry": "1week"}
     url = "https://bpaste.net"
-    response = urlopen(url, data=urlencode(params).encode("ascii")).read()
-    m = re.search(r'href="/raw/(\w+)"', response.decode("utf-8"))
+    try:
+        response = (
+            urlopen(url, data=urlencode(params).encode("ascii")).read().decode("utf-8")
+        )
+    except OSError as exc_info:  # urllib errors
```
asottile: I don't think we should silently gobble the errors here, if someone requests `--pastebin` and pytest fails to pastebin then the command should probably exit nonzero

goerz: Our use-case is a home-spun continuous integration script for https://www.qdyn-library.net. This project is hosted on a gitolite installation on a university workstation running commercial Fortran compilers, which also acts as the CI server via a simple post-commit hook kicking off a [SLURM](https://slurm.schedmd.com) job to run the test matrix. Everything is totally headless and not web-based, so the CI-server sends developers an email with the results of the CI. These only include a summary, the detailed test output tends to be a bit too large to send around by email, so the CI-server uses `--pastebin` to upload the output, and puts a link in the email.

Recently, this started failing for some test runs with an HTTP 400 error. We're not quite sure what exactly the cause is. In any case, having the full test output available is very much secondary to the actual tests passing or failing. We don't want to get a "test failure" just because the upload to pastebin failed for some reason. After all, it might just be a temporary network failure. It would be enough to have the appropriate error message in the email (as this PR implements). We can then still investigate what the problem is, but a successful upload to pastebin is no longer a necessary condition for accepting patches. So to "silently gobble the errors" is explicitly the point here. The error will still be logged to the output, so it's not *totally* silent. Of course, this behavior is debatable. It's just that with the current behavior of pytest failing if pastebin fails, we'd probably have to stop using that feature and re-implement it as a pytest-plugin that matches our use case; that would obviously be more work than this small patch.

goerz: A possible compromise *might* be to exit with a specific status code that indicates "Everything OK except for the `--pastebin`". The CI-script might be able to catch that as a special case.

blueyed:
While at it: what about retrying it 3 times after e.g. 10, 30, 60 seconds?

src/_pytest/pastebin.py
nicoddemus: I'm OK with catching and displaying the error instead of returning non-zero. I think most (all?) use cases for `--pastebin` follow the pattern of "posting to pastebin for a nice-to-have report" but that is not really required.

Either way, this should target `features` as it changes the behavior. 👍

goerz:
> While at it: what about retrying it 3 times after e.g. 10, 30, 60 seconds?

That sounds like a good idea, but definitely a separate pull request

src/_pytest/pastebin.py
```diff
@@ -67,8 +67,13 @@ def create_new_paste(contents):
...
+        response = str(exc_info)
```
blueyed: Would the exception info ever contain the URL?

Errors should get logged in any case probably. (not sure about non-zero exit code myself)

goerz: *Probably* not, but I can't give a guarantee. So it's probably best to revise this.

goerz: New commit 25dba82171dc93ee9d81b41e21609eda1f82846a addresses this

src/_pytest/pastebin.py
goerz: So should I rename `changelog/5764.improvement.rst` to `changelog/5764.feature.rst` and reword the changelog text?

goerz: I changed the base branch from `master` to `features` just now.

goerz:
I'll see if I can change the test a little bit to increase patch coverage (if you care)

blueyed:
..appears to no have been rebased (but only the branch switched in the UI).
You need to rebase it locally and force-push (then you can also remove the mention :))

goerz:
I've force-pushed a new PR that is rebased directly on `features`. Sorry about the confusion. The revised code contains all of @blueyed's suggestions (no `else` statement for the `try` block).

This is still three commits (the patch itself, the changelog entry, and the tests), but I can squash it into a single commit if you prefer.

nicoddemus:
@blueyed @asottile would you like to do a 2nd round of reviews?

asottile:
I mean, the code looks fine -- retries would be nice, and I still think exiting 0 in this case is incorrect as `pytest` has failed to accomplish a thing that's explicitly asked for on the commandline

but I also would never use the pastebin feature so I don't really care about the outcome here :man_shrugging: 

good enough for me :+0:

nicoddemus:
Thanks guys!

goerz:
Thanks!

PR #5768, opened 2019-08-19
Fix self reference in function scoped fixtures

robholt:
This fixes an old issue #2270 that was closed at the time without a fix. That issue has more info, but basically function scope fixtures in a class (not a test class) are given a reference to the test class, not their own class, as `self`. This causes problems when that fixture tries to access attributes on its class.


nicoddemus:
The failure is unrelated (https://github.com/pytest-dev/pytest/issues/5795).

Thanks again @robholt!

PR #5773, opened 2019-08-20
Preparing release version 5.1.1

asottile:
@nicoddemus how do we fund you a better computer, mine's from 2015 and twice as fast :'(


nicoddemus:
> @nicoddemus how do we fund you a better computer, mine's from 2015 and twice as fast :'(

🤔 I didn't get it. 😅

nicoddemus:
Hah now I see. Don't know, perhaps because I'm on a VM? 😛

nicoddemus:
Thanks!

Issue #5795, opened 2019-08-28
Flaky tests in Python 3.5.0 with xdist 😬


nicoddemus:
```
__________________________ testing/test_collection.py __________________________
[gw1] linux -- Python 3.5.0 /home/travis/build/pytest-dev/pytest/.tox/py35-xdist/bin/python
worker 'gw1' crashed while running 'testing/test_collection.py::test_exit_on_collection_with_maxfail_bigger_than_n_errors'
```

https://travis-ci.org/pytest-dev/pytest/jobs/577770604

😬 

---

In https://github.com/pytest-dev/pytest/pull/5803#issuecomment-526557021, we suspect this is related to 3.5.0, as we have never seen this failure before started testing on that Python version on Travis.

nicoddemus:
As mentioned in https://github.com/pytest-dev/pytest/pull/5803#issuecomment-526557021, it might have started with `3.5.0`, but I'm not sure myself.

nicoddemus:
Unfortunately py35, without xdist and coverage, still is crashing randomly.

Unless someone has other ideas, I think we should consider not testing `Python 3.5.0` anymore. Having it failing randomly is too disruptive. 

Does flake8 now cover most gotchas related to poor typing in 3.5.0?

PR #5772, opened 2019-08-20
Coverage for 3.5.0

blueyed:
Ref: https://github.com/pytest-dev/pytest/pull/5752#issuecomment-523064809

PR #5803, opened 2019-08-30
Mark flaky test as xfail on Python 3.5.0

nicoddemus:
I've tried to reproduce the issue, but it seems to only happen sporadically on Linux.

Mark it is flaky now to avoid people wasting time on PRs when they encounter this problem.

Related to #5795

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
     - env: TOXENV=pypy3-xdist
       python: 'pypy3'
 
-    - env: TOXENV=py35-xdist
+    - env: TOXENV=py35
       dist: trusty
       python: '3.5.0'
 


blueyed:
Have you seen it without xdist?
It is also related to 3.5.0 likely, not 3.5.x?

nicoddemus:
> Have you seen it without xdist?

That's what I've tried, yes.

> It is also related to 3.5.0 likely, not 3.5.x?

Good point, have not tried that.

Either way, I think it is fine to merge this; if/when someone champions this, we just remove that line of code. It is worst to have this failing randomly on PRs and generating friction, IMHO.

blueyed:
Sure.  I just got the impression that it started to happen after explicitly testing 3.5.0.

nicoddemus:
> Sure. I just got the impression that it started to happen after explicitly testing 3.5.0.

You got a point, I will mention it in the issue. If you are OK with the PR, could you approve please? Thanks

blueyed:
> You got a point, I will mention it in the issue.

Why not just amend it here instead? :)

blueyed:
> > Have you seen it without xdist?
> 
> That's what I've tried, yes.

So you've not seen it yourself?  "it seems to only happen sporadically on Linux." refers to the CI job then?

nicoddemus:
> So you've not seen it yourself? "it seems to only happen sporadically on Linux." refers to the CI job then?

Yep, and on Linux only, which we use `3.5.0` (on Windows we use `3.5.4` I think). So you are probably right that is 3.5.0 related.

I have changed the check to use `3.5.0` then.

nicoddemus:
Wait, another one failed now: test_continue_on_collection_errors

blueyed:
```
__________________________ testing/test_collection.py __________________________
[gw1] linux -- Python 3.5.0 /home/travis/build/pytest-dev/pytest/.tox/py35-xdist/bin/python
worker 'gw1' crashed while running 'testing/test_collection.py::test_continue_on_collection_errors'
```
It is a bit unfortunate that xdist provides not more information there, isn't it?

blueyed:
Maybe the job should be changed to not use xdist?

nicoddemus:
Hmmm that would probably help, let me try it.

nicoddemus:
I will remove the `xfail` marker, let's see how it fares without xdist for awhile then.

nicoddemus:
🤞

blueyed:
Cool, for reference it might still break when using coverage with it (https://github.com/pytest-dev/pytest/pull/5772).

nicoddemus:
😓

PR #5828, opened 2019-09-07
Fix pypy3.6 on windows

asottile:
pypy3.6 on windows doesn't have an `_io.WinConsoleIO`


nicoddemus:
Cool, thanks!

(Btw, I'm probably won't be participating much other than reviewing/merging for the next few days, my HD died on me so I'm waiting for a new one).

PR #5834, opened 2019-09-09
Update doc regarding pytest.raises

gene1wood:
Remove reference to the `message` argument in the docs as it was deprecated in #4539


nicoddemus:
Great, thank you!

PR #5864, opened 2019-09-19
Support sys.pycache_prefix on py38

### LINKED ISSUE #4730: assert rewriting: support sys.pycache_prefix (py38)
A new feature will be added in Python 3.8 allowing to specify an alternate prefix/location for the pycache files: https://github.com/python/cpython/blame/dcfcd146f8e6fc5c2fc16a4c192a0c5f5ca8c53c/Doc/whatsnew/3.8.rst#L70-L83

Pytest should support/use this with its assertion rewriting.

nicoddemus:
Fix #4730

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -13,6 +13,7 @@ import struct
 import sys
 import tokenize
 import types
+from pathlib import Path
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -30,7 +31,7 @@ from _pytest.assertion.util import (  # noqa: F401
 from _pytest.pathlib import fnmatch_ex
 from _pytest.pathlib import PurePath
 
-# pytest caches rewritten pycs in __pycache__.
+# pytest caches rewritten pycs in pycache dirs
 PYTEST_TAG = "{}-pytest-{}".format(sys.implementation.cache_tag, version)
 PYC_EXT = ".py" + (__debug__ and "c" or "o")
 PYC_TAIL = "." + PYTEST_TAG + PYC_EXT
@@ -103,7 +104,7 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder):
         return None  # default behaviour is fine
 
     def exec_module(self, module):
-        fn = module.__spec__.origin
+        fn = Path(module.__spec__.origin)
         state = self.config._assertstate
 
         self._rewritten_names.add(module.__name__)
@@ -117,15 +118,15 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder):
         # cached pyc is always a complete, valid pyc. Operations on it must be
         # atomic. POSIX's atomic rename comes in handy.
         write = not sys.dont_write_bytecode
-        cache_dir = os.path.join(os.path.dirname(fn), "__pycache__")
+        cache_dir = get_cache_dir(fn)
         if write:
             ok = try_mkdir(cache_dir)
             if not ok:
                 write = False
-                state.trace("read only directory: {}".format(os.path.dirname(fn)))
+                state.trace("read only directory: {}".format(cache_dir))
 
-        cache_name = os.path.basename(fn)[:-3] + PYC_TAIL
-        pyc = os.path.join(cache_dir, cache_name)
+        cache_name = fn.name[:-3] + PYC_TAIL
+        pyc = cache_dir / cache_name
         # Notice that even if we're in a read-only directory, I'm going
         # to check for a cached pyc. This may not be optimal...
         co = _read_pyc(fn, pyc, state.trace)
@@ -139,7 +140,7 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder):
                 finally:
                     self._writing_pyc = False
         else:
-            state.trace("found cached rewritten pyc for {!r}".format(fn))
+            state.trace("found cached rewritten pyc for {}".format(fn))
         exec(co, module.__dict__)
 
     def _early_rewrite_bailout(self, name, state):
@@ -258,7 +259,7 @@ def _write_pyc(state, co, source_stat, pyc):
     # (C)Python, since these "pycs" should never be seen by builtin
     # import. However, there's little reason deviate.
     try:
-        with atomicwrites.atomic_write(pyc, mode="wb", overwrite=True) as fp:
+        with atomicwrites.atomic_write(str(pyc), mode="wb", overwrite=True) as fp:
             fp.write(importlib.util.MAGIC_NUMBER)
             # as of now, bytecode header expects 32-bit numbers for size and mtime (#4903)
             mtime = int(source_stat.st_mtime) & 0xFFFFFFFF
@@ -269,7 +270,7 @@ def _write_pyc(state, co, source_stat, pyc):
     except EnvironmentError as e:
         state.trace("error writing pyc file at {}: errno={}".format(pyc, e.errno))
         # we ignore any failure to write the cache file
-        # there are many reasons, permission-denied, __pycache__ being a
+        # there are many reasons, permission-denied, pycache dir being a
         # file etc.
         return False
     return True
@@ -277,6 +278,7 @@ def _write_pyc(state, co, source_stat, pyc):
 
 def _rewrite_test(fn, config):
     """read and rewrite *fn* and return the code object."""
+    fn = str(fn)
     stat = os.stat(fn)
     with open(fn, "rb") as f:
         source = f.read()
@@ -292,12 +294,12 @@ def _read_pyc(source, pyc, trace=lambda x: None):
     Return rewritten code if successful or None if not.
     """
     try:
-        fp = open(pyc, "rb")
+        fp = open(str(pyc), "rb")
     except IOError:
         return None
     with fp:
         try:
-            stat_result = os.stat(source)
+            stat_result = os.stat(str(source))
             mtime = int(stat_result.st_mtime)
             size = stat_result.st_size
             data = fp.read(12)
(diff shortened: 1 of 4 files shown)


src/_pytest/assertion/rewrite.py
```diff
@@ -291,12 +293,12 @@ def _read_pyc(source, pyc, trace=lambda x: None):
     Return rewritten code if successful or None if not.
     """
     try:
-        fp = open(pyc, "rb")
+        fp = open(str(pyc), "rb")
```
nicoddemus: Ideally we should use `os.fspath`, but py36+ only. 😬

blueyed: Can be done for it then already, no?

nicoddemus: What do you mean? We still need to support py35.

nicoddemus: Hmmm probably it pays of to create a `fspath`-like function in `compat`... 🤔

src/_pytest/assertion/rewrite.py
```diff
@@ -102,7 +104,7 @@ def create_module(self, spec):
         return None  # default behaviour is fine
 
     def exec_module(self, module):
-        fn = module.__spec__.origin
+        fn = Path(module.__spec__.origin)
```
asottile: lol, in my last patch I moved this away from pathlib due to needing `fspath` hax -- this is fine though I guess 🤷‍♂

src/_pytest/assertion/rewrite.py
```diff
@@ -291,12 +294,12 @@ def _read_pyc(source, pyc, trace=lambda x: None):
     Return rewritten code if successful or None if not.
     """
     try:
-        fp = open(pyc, "rb")
+        fp = open(fspath(pyc), "rb")
```
asottile: this one shouldn't need `fspath` should it?

src/_pytest/assertion/rewrite.py
```diff
@@ -1021,9 +1024,9 @@ def visit_Compare(self, comp: ast.Compare):
 def try_mkdir(cache_dir):
     """Attempts to create the given directory, returns True if successful"""
     try:
-        os.mkdir(cache_dir)
+        os.makedirs(fspath(cache_dir))
```
asottile: `exist_ok=True` and then I think you can avoid the `except:` block maybe -- or just use `suppress`

src/_pytest/assertion/rewrite.py
```diff
@@ -1039,3 +1042,17 @@ def try_mkdir(cache_dir):
...
+
+def get_cache_dir(file_path: Path) -> Path:
+    """Returns the cache directory to write .pyc files for the given .py file path"""
+    if sys.version_info >= (3, 8) and sys.pycache_prefix:
+        # given:
+        #   prefix = '/tmp/pycs'
+        #   path = '/home/user/proj/test_app.py'
+        # we want:
+        #   '/tmp/pycs/home/user/proj'
+        return Path(sys.pycache_prefix) / Path(*file_path.parts[1:-1])
```
asottile: this seems broken for windows if I had to guess

src/_pytest/compat.py
```diff
@@ -41,6 +42,19 @@ def _format_args(func):
 REGEX_TYPE = type(re.compile(""))
 
 
+if sys.version_info < (3, 6):
+
+    def fspath(p):
+        """os.fspath replacement, useful to point out when we should replace it by the
+        real function once we drop py35.
+        """
+        return str(p)
```
asottile: this is not ~quite right -- but maybe good enough for our uses? here's the implementation from python3.6:

```python
def _fspath(path):
    """Return the path representation of a path-like object.

    If str or bytes is passed in, it is returned unchanged. Otherwise the
    os.PathLike interface is used to get the path representation. If the
    path representation is not str or bytes, TypeError is raised. If the
    provided path is not str, bytes, or os.PathLike, TypeError is raised.
    """
    if isinstance(path, (str, bytes)):
        return path

    # Work from the object's type to match method resolution of other magic
    # methods.
    path_type = type(path)
    try:
        path_repr = path_type.__fspath__(path)
    except AttributeError:
        if hasattr(path_type, '__fspath__'):
            raise
        else:
            raise TypeError("expected str, bytes or os.PathLike object, "
                            "not " + path_type.__name__)
    if isinstance(path_repr, (str, bytes)):
        return path_repr
    else:
        raise TypeError("expected {}.__fspath__() to return str or bytes, "
                        "not {}".format(path_type.__name__,
                                        type(path_repr).__name__))
```

nicoddemus: > but maybe good enough for our uses?

I think so. We will be using esporadically to convert `Path` objects only, and `__fspath__` of `Path` objects implement a call to `str()`:

```python
    def __fspath__(self):
        return str(self)
```

And it is also cached:

```python
    def __str__(self):
        """Return the string representation of the path, suitable for
        passing to system calls."""
        try:
            return self._str
        except AttributeError:
            self._str = self._format_parsed_parts(self._drv, self._root,
                                                  self._parts) or '.'
            return self._str
```

So I think it is fine as is. Also we will be dropping py35 soon I think, so we can just call `os.fspath` directly. 👍

testing/test_assertrewrite.py
```diff
@@ -1580,3 +1583,59 @@ def fake_mkdir(p, *, exc):
     with pytest.raises(OSError) as exc_info:
         try_mkdir(str(p))
     assert exc_info.value.errno == errno.ECHILD
+
+
+class TestPyCacheDir:
+    @pytest.mark.parametrize(
+        "prefix, source, expected",
+        [
+            ("c:/tmp/pycs", "d:/projects/src/foo.py", "c:/tmp/pycs/projects/src"),
```
nicoddemus: @asottile here are the tests for Windows, which I obtained by testing how `sys.pycache_prefix` worked on my system.

src/_pytest/assertion/rewrite.py
nicoddemus: I changed to use `cache_dir.mkdir` as you suggest, unfortunately I could not get rid of `FileExistsError` handling (got some test failures), plus `test_try_mkdir` seems to be fine tuned already and I would have to change it now that `cache_dir` would be a `Path`, so I would prefer to not really change this if possible. 😬 

I also tried to use `suppress`, but it loses its appeal because we still need try/except for the other errors (so we can return `False` on some of them):

```diff
diff --git a/src/_pytest/assertion/rewrite.py b/src/_pytest/assertion/rewrite.py
index 415676aaa..0527a3639 100644
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -1,5 +1,6 @@
 """Rewrite assertion AST to produce nice error messages"""
 import ast
+import contextlib
 import errno
 import functools
 import importlib.abc
@@ -1024,8 +1025,8 @@ warn_explicit(
 def try_mkdir(cache_dir):
     """Attempts to create the given directory, returns True if successful"""
     try:
-        os.makedirs(fspath(cache_dir))
-    except FileExistsError:
+        with contextlib.suppress(FileExistsError):
+            os.makedirs(fspath(cache_dir))
         # Either the pycache directory already exists (the
         # common case) or it's blocked by a non-dir node. In the
         # latter case, we'll ignore it in _write_pyc.
```

asottile: I mean to use `makedirs(..., exist_ok=True)` ~essentially `mkdir -p` but only available in python3 :)

src/_pytest/assertion/rewrite.py
asottile: oh I see, it's using `.parts` -- thought this was slicing strings -- I'll have to look into this more :/

src/_pytest/assertion/rewrite.py
asottile: if we need hax for py35 to use pathlib I'd prefer we don't use pathlib until we drop python3.5

nicoddemus: I don't think it is hack to be honest... it is a compatibility function, like we have many others in `compat`. It is literally a if/else with an implementation specific to py35.

This request would need me to review the entire patch, and I really would like to avoid that unless we have good reason.

asottile: I agree that this particular case is just a compat function, but pathlib in general feels like a liability if it's going to blow up mysteriously in python3.5

nicoddemus: > if it's going to blow up mysteriously in python3.5

Definitely agree for external APIs, but this is an internal change. And it doesn't blow mysteriously: remove that call and it does fail on python3.5 immediately, with a good error message. Or perhaps you mean something else by mysteriously?

asottile: this particular usage doesn't, but I'm not confident of test coverage about all of our uses of such functions

there's also potentially more surface area than expected here since it's an import hook -- but maybe it's fine

I just have my reservations about pathlib being a good idea when there's some pretty easy footguns :) (and I don't like pathlib myself)

nicoddemus: Well, let's hold this PR until we drop py35 drops then. 👍

asottile: I'm fine with this PR btw, just _concerned_ is all ;) -- I think if you use `exist_ok=True` below I'll press approve :D

nicoddemus:
> So I think it is fine as is. Also we will be dropping py35 soon I think, so we can just call os.fspath directly. 

Actually someone brought to my attention that py35 support ends in late 2020, not late 2019 as I thought initially.

Regarding this PR, should we:

1. Go ahead with it the way it is, using the `fspath` in compat for py35. 
2. Revert all the `Path` changes.
3. Wait until we drop py35.

I myself vote for 1), `str(p)` is how we have been dealing with `Path` on Python versions that don't have `os.fspath` since we introduced `pathlib`.

asottile:
> I'm fine with this PR btw, just concerned is all ;) -- I think if you use exist_ok=True below I'll press approve :D

yeah I'm ok to go ahead with this -- just want to see `os.makedirs(..., exist_ok=True)`

nicoddemus:
> just want to see os.makedirs(..., exist_ok=True)

Done in 870752187.

PR #5869, opened 2019-09-21
Change report-coverage.sh in attempt to fix Azure

nicoddemus:
Recently sometimes Azure has failed with:

```
++ curl -s https://codecov.io/bash
bash: /dev/fd/63: No such file or directory
```

This attempts to fix this by modifying report-coverage.sh slightly.

### DIFF
--- a/scripts/report-coverage.sh
+++ b/scripts/report-coverage.sh
@@ -13,4 +13,5 @@ fi
 python -m coverage combine
 python -m coverage xml
 python -m coverage report -m
-bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml
+curl -S -L --retry 3 -s https://codecov.io/bash -o codecov-upload.sh
+bash codecov-upload.sh -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml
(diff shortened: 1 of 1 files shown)


scripts/report-coverage.sh
```diff
@@ -13,4 +13,5 @@ fi
 python -m coverage combine
 python -m coverage xml
 python -m coverage report -m
-bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X gcovout -X fix -f coverage.xml
+curl -s https://codecov.io/bash -o codecov-upload.sh
```
bluetech: I'd suggest adding `-S` (show errors), `-L` (follow redirects). Maybe also `--retry 2`.

nicoddemus: Good idea, done (used retry 3 thought, because why not :grin:)

blueyed:
Isn't it "just" a timeout issue on their end?

nicoddemus:
Doesn't seem like a timeout issue, more some flaky system settings or something.

blueyed:
Yes - the point being that `curl` returns nothing here etc.

But I see that you've added `--retry 3`.. :+1: 

Note that `-X gcov -X coveragepy -X search -X xcode -X gcovout` could be dropped (not necessary with `-f` IIRC).

nicoddemus:
Err not sure what you are suggesting... please feel free to push into this branch with any recomentadtions!

nicoddemus:
Thanks @blueyed and @bluetech!

blueyed:
Just for the record, it still might fail (of course):
```
+ curl -S -L --retry 3 -s https://codecov.io/bash -o codecov-upload.sh
curl: (7) Failed to connect to codecov.io port 443: Timed out
Bash exited with code '7'.
```
(https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=2260)

Issue #5655, opened 2019-07-24
stepwise doesn't stop the suite on first failure with xdist


graingert:
```sh
graingert@onomastic:~/projects/foo$ cat tests/test_foo.py 
import pytest

def test_1():
    assert True

def test_2():
    assert True

def test_3():
    assert False

def test_4():
    assert True

def test_5():
    assert True

def test_6():
    assert True
```

Running it the first time, the test suite doesn't stop: (no -x behavior)
```sh
graingert@onomastic:~/projects/foo$ pytest -vvv --sw -n 1 tests/
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.7.3, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /home/graingert/.virtualenvs/pytest/bin/python3.7
cachedir: .pytest_cache
rootdir: /home/graingert/projects/foo/tests, inifile: pytest.ini
plugins: xdist-1.29.0, forked-1.0.2
[gw0] linux Python 3.7.3 cwd: /home/graingert/projects/foo
[gw0] Python 3.7.3 (default, Apr  3 2019, 05:39:12)  -- [GCC 8.3.0]
gw0 [6]
scheduling tests via LoadScheduling

tests/test_foo.py::test_1 
[gw0] [ 16%] PASSED tests/test_foo.py::test_1 
tests/test_foo.py::test_2 
[gw0] [ 33%] PASSED tests/test_foo.py::test_2 
tests/test_foo.py::test_3 
[gw0] [ 50%] FAILED tests/test_foo.py::test_3 
tests/test_foo.py::test_4 
[gw0] [ 66%] PASSED tests/test_foo.py::test_4 
tests/test_foo.py::test_5 
[gw0] [ 83%] PASSED tests/test_foo.py::test_5 
tests/test_foo.py::test_6 
[gw0] [100%] PASSED tests/test_foo.py::test_6 

================================================================================================= FAILURES =================================================================================================
__________________________________________________________________________________________________ test_3 __________________________________________________________________________________________________
[gw0] linux -- Python 3.7.3 /home/graingert/.virtualenvs/pytest/bin/python3.7

    def test_3():
>       assert False
E       assert False

tests/test_foo.py:10: AssertionError
==================================================================================== 1 failed, 5 passed in 0.23 seconds ====================================================================================
```

Running it the second time, the test suite starts from the first failing test
but doesn't stop: (no -x behavior)
```sh
[...]

PR #5871, opened 2019-09-23
remember all stepwise lastfailed pytest-dev/pytest#5655

graingert:


src/_pytest/stepwise.py
```diff
@@ -78,13 +81,15 @@ def pytest_runtest_logreport(self, report):
             if self.skip:
                 # Remove test from the failed ones (if it exists) and unset the skip option
                 # to make sure the following tests will not be skipped.
-                if report.nodeid == self.lastfailed:
```
graingert: so it turns out pytest_runtest_logreport can get called multiple times when running under xdist

graingert:
@nicoddemus hmm this still doesn't work because xdist doesn't run log report for passing tests so it won't remove the test from lastfailed...

graingert:
also you can end up with a situation with batched tests:
```
[1, 2, 3, 4, 5] gw0
[6, 7, 8, 9, 10] gw1
```

if 5 and 9 are both failing tests, and only tests 1, 2, 3, 6, 7, 8, 9 run (because gw0 is shutdown after 9 fails) then 5 will be marked as passing

blueyed:
@nicoddemus why was this closed?

blueyed:
Moved to https://github.com/pytest-dev/pytest-xdist/issues/458.

@nicoddemus please adhere to your own suggestions though (https://github.com/pytest-dev/pytest/pull/6533#issuecomment-577179396).

nicoddemus:
Definitely, I should have left a note that this was moved to pytest-xdist. 👍

PR #5873, opened 2019-09-23
Do not call python directly but use sys.executable. Fixes #5872

### LINKED ISSUE #5872: test_argcomplete should not call python directly
on the most distros it points to py2 or it is not really deterministic.

The call itself is following:
`fp.write('COMP_WORDBREAKS="$COMP_WORDBREAKS" python -m pytest 8>&1 9>&2')`

I suppose it should use `sys.executable` or at least `python3 `as the py2 is not compatible anyway.

scarabeusiv:
Fixes #5872.

### DIFF
--- a/testing/test_parseopt.py
+++ b/testing/test_parseopt.py
@@ -298,7 +298,11 @@ def test_argcomplete(testdir, monkeypatch):
         # redirect output from argcomplete to stdin and stderr is not trivial
         # http://stackoverflow.com/q/12589419/1307905
         # so we use bash
-        fp.write('COMP_WORDBREAKS="$COMP_WORDBREAKS" python -m pytest 8>&1 9>&2')
+        fp.write(
+            'COMP_WORDBREAKS="$COMP_WORDBREAKS" {} -m pytest 8>&1 9>&2'.format(
+                sys.executable
+            )
+        )
     # alternative would be exteneded Testdir.{run(),_run(),popen()} to be able
     # to handle a keyword argument env that replaces os.environ in popen or
     # extends the copy, advantage: could not forget to restore


blueyed:
Thanks!

Just for reference: it is best to use branches for PRs.
Apart from that it would be nice to have a better commit message - it could eg mention the test name etc, and should not be too long.

Normally not really a problem, when it could be squash-merged, but this is disabled here (#4361).

testing/test_parseopt.py
```diff
@@ -298,7 +298,11 @@ def test_argcomplete(testdir, monkeypatch):
         # redirect output from argcomplete to stdin and stderr is not trivial
         # http://stackoverflow.com/q/12589419/1307905
         # so we use bash
-        fp.write('COMP_WORDBREAKS="$COMP_WORDBREAKS" python -m pytest 8>&1 9>&2')
+        fp.write(
+            'COMP_WORDBREAKS="$COMP_WORDBREAKS" {} -m pytest 8>&1 9>&2'.format(
+                sys.executable
```
asottile: should ~probably be `shlex.quote(sys.executable)` since we're sticking it into a shell script

asottile: yep!

blueyed: `{!r}` might be enough though.

asottile: repr is not a substitute for proper shell quoting, that's asking for shell injection ;)

blueyed: Sure.. but this is something we have control over here more or less.
It's good to use the proper way though, of course.. :)

asottile: we don't control this at all though, it's a user's directory -- they could do all sorts of wild things with the directory they clone/install pytest into

blueyed: I've meant the user / test runner - who is setting/using `sys.executable`.

asottile: of course, that doesn't stop people from having spaces in their development dirs, or non-ascii usernames, or having quotes in their names

blueyed: Spaces would be handled by the `repr()` then.
Anyway, your fix is better.

PR #5910, opened 2019-10-03
Fix dynamic scoping changelog link

asottile:


nicoddemus:
Thanks, failures are unrelated (I hope to investigate those later)

asottile:
yeah the failures are #5901 / #5902

PR #5913, opened 2019-10-04
fix deprecation warning of attrs on NodeMarkers

isidentical:
```
/home/batuhan/setuptools/.venv/lib/python3.8/site-packages/_pytest/mark/structures.py:370: DeprecationWarning: The usage of `cmp` is deprecated and will be removed on or after 2021-06-01.  Please use `eq` and `order` instead.
```


asottile:
already handling this in #5902

PR #5916, opened 2019-10-06
updated sidebarintro

readthroughmyglasses:
Added more information in the sidebar for clarification

### DIFF
--- a/doc/en/_templates/sidebarintro.html
+++ b/doc/en/_templates/sidebarintro.html
@@ -1,5 +1,8 @@
 <h3>About pytest</h3>
 <p>
-  pytest is a mature full-featured Python testing tool that helps
+  PyTest is a mature full-featured Python testing tool that helps
   you write better programs.
+</p>
+<p>
+  you can write code to test anything like database,API,even UI if you want.
 </p>


doc/en/_templates/sidebarintro.html
```diff
@@ -1,5 +1,8 @@
 <h3>About pytest</h3>
 <p>
-  pytest is a mature full-featured Python testing tool that helps
+  PyTest is a mature full-featured Python testing tool that helps
   you write better programs.
 </p>
+<p>
+  you can write code to test anything like database,API,even UI if you want.
```
nicoddemus: Hi @readthroughmyglasses,

Unfortunately I don't see how this improves the blurb on the side bar. Ideally it should stay short (1 paragraph), and concise. Plus you have a few grammar mistakes: `you` instead of `You`, and missing spaces between commas.

doc/en/_templates/sidebarintro.html
```diff
@@ -1,5 +1,8 @@
 <h3>About pytest</h3>
 <p>
-  pytest is a mature full-featured Python testing tool that helps
+  PyTest is a mature full-featured Python testing tool that helps
```
nicoddemus: It really is `pytest`, not `PyTest`

blueyed:
Looks like Hacktoberfest spam.. :(

The-Compiler:
@blueyed Assume good intentions. This was someone's first contribution (at least via GitHub PR) to an opensource project ever - it ended up not being an improvement, but probably the author felt that it was, so calling it spam seems unnecessarily hostile to me.

blueyed:
@The-Compiler sure, but we've seen some of those in neovim recently also, and there was no answer to the question/closing from @nicoddemus - I've just left a comment then about what I suspect this was.
Looking at the profile shows that it's not a totally new account, and should know a bit therefore.
It's also meant to make maintainers aware of this - AFAIK those should get an "invalid" label then.

PR #5920, opened 2019-10-06
Allow for "pdb" module to be rewritten

blueyed:
Noticed this with tests for pdb++.

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -1,8 +1,6 @@
 """ interactive debugging with PDB, the Python Debugger. """
 import argparse
-import pdb
 import sys
-from doctest import UnexpectedException
 
 from _pytest import outcomes
 from _pytest.config import hookimpl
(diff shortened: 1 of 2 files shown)


blueyed:
Note that it still tries to rewrite it (https://github.com/pytest-dev/pytest/blob/c1361b48f83911aa721b21a4515a5446515642e2/src/_pytest/assertion/rewrite.py#L233-L234), and might succeed (likely due to the "pdb" module being re-imported explicitly in the tests).

nicoddemus:
Needs rebase once https://github.com/pytest-dev/pytest/pull/5923 is merged

blueyed:
> Missing changelog, otherwise LGTM. ￼

Isn't that just noise?  (quite internal isn't it?).

nicoddemus:
> Isn't that just noise? (quite internal isn't it?)

Fair enough. 👍

PR #5922, opened 2019-10-06
ci: test oldest supported attrs

blueyed:


tox.ini
```diff
@@ -41,6 +41,8 @@ setenv =
     xdist: _PYTEST_TOX_POSARGS_XDIST=-n auto
 extras = testing
 deps =
+    oldattrs: attrs==17.4.0
+    oldattrs: hypothesis<=4.38.1
```
blueyed: hypothesis 4.38.2 bumped the attrs requirement itself.

blueyed:
Covers the new compat code now: https://codecov.io/gh/pytest-dev/pytest/compare/8c9ea5e055e7ef7a7d33db65c781adee16b55bc0...b0fd8742da28c2df2bd20cab9d67b432ce542951/src/src/_pytest/compat.py

asottile:
I think we should instead `>=19.2` on features

nicoddemus:
> I think we should instead >=19.2 on features

That's arguable, but neverthless it doesn't discard this PR, we don't know when `features` will get released.

Back to the point, not sure we should require attrs>=19.2: we are an upstream library which is loaded into the user's runtime, so we should be careful when to bump our minimal requirements, specially for a widely used library such as `attrs`. 

Given that so far this is the only issue that has bitten us, I would leave it alone. If more issues start to creep in, we should definitely reconsider this though.

Issue #5515, opened 2019-06-28
Option to disable cli log indentation


mitzkia:
Hi, 

I have noticed that, the console log output has changed in #5312.

I understand the benefit: the output will be easier to read with eyes.
But in my use-case I have often print generated configurations to the console, and if I wanted to copy-and-paste it, it 
will be harder to work with it. I will got unnecessary spaces before my printed configuration (because of the indentation).

Sorry, but I think this change is a little bit strict. I am not against with the change, but it would be good if my "constructed" text could be printed without any change. 
Maybe this formatted indentation could be optional (turn on/off with an arg or option)?

If needed I can provide problematic console log.

Thanks

nicoddemus:
Hi @mitzkia,

An option seems like a good compromise (logging has tons of options already, another one wouldn't hurt I think). I've changed the title to reflect that. 👍

earthman1:
This option is fairly important for me, as some of my log lines are long, and I use a newline to shift the output to the left in order to make the output visible without scrolling an output window, or breaking the line at an artificial location based on terminal size.

I had hoped that `\r` could be a workaround, as this is defined as moving the output cursor to the beginning of the line. Unfortunately, the call to `record.message.splitlines()` removes each `\r` by treating `\r` as `\n`, treating `\r\n` as `\n`, and treating `\n\r` as `\n\n` when splitting. If the method of splitting the lines were adjusted to ignore the `\r` in `\n\r`, it would split on the `\n` and leave the `\r` at the front of the second line, which would send the cursor to column 0 after printing the indentation, but before printing the text. This would allow another option for formatting the output for readability, as needed, without the requirement to turn the feature on or off everywhere.

This would not address the issue experienced by @mitzkia, so a configuration option would still be advantageous, and may open questions of how it would behave on different platforms such as windows, which may treat `\r` differently. On the other hand, a character that would trigger a carriage return without a newline (which would visually suppress the automatic indentation) would allow for the default formatting to occur at all times, and still allow for multi-line log entries to (appear to) begin at column 0 of the output on an ad-hoc basis.

nicoddemus:
@earthman1 thanks for describing your use case.

Please consider writing a PR with the option; we would love to guide you if there are questions.

earthman1:
Thank you for the invitation. I am working on the PR now.

earthman1:
I submitted PR https://github.com/pytest-dev/pytest/pull/5926 to implement this feature

[...]

Issue #4283, opened 2018-10-31
pre-commit: rst: allow pytest directives in changelog entries


blueyed:
Via: https://github.com/pytest-dev/pytest/pull/4271#issue-227093306

It is useful to link options etc in changelog entries, but linting fails on them.

asottile:
rst-lint is quite inflexible, there isn't really anything that can be done here

blueyed:
Well, not using it in the first place would be an option.
We're checking docs already for every PR in total, and should/could then instead check them with a generated CHANGELOG.rst then.

asottile:
:man_shrugging: `rst-lint` has definitely saved me in changelogs before, I expect it to be even more helpful to those not familiar with the intricacies of rst

nicoddemus:
👍 on keeping `rst-lint`: it also checks the changelog entries, while the generated changelog is only materialized on releases. It would make the release task much more frustating if it meant having to fix rst syntax problems in the CHANGELOG.

blueyed:
> should/could then instead check them with a generated CHANGELOG.rst then.

blueyed:
I would not sacrifice better documentation because of it.

(Also for linting rst in general I recommend using a linter / syntax highlighting in your editor already.)

nicoddemus:
You mean generating the CHANGELOG during `docs` and checking that?

That would bring the advantage of using full rst syntax and directives (say `:ref:fixtures`)... it is an interesting idea indeed.

blueyed:
Exactly.

PR #5926, opened 2019-10-07
Add log-auto-indent option to control multiline formatting

earthman1:
Allow selective auto-indentation of multiline log messages.

Supports command line option ``--log-auto-indent [value]``
and config option ``log_auto_indent`` to set the default for all logging.

Supports passing kwarg ``extra={"auto_indent": [value]}`` to
calls to ``log.log()`` to specify auto-indentation behavior for
a specific entry in the log.

``[value]`` can be:
    * True or On - Dynamically auto-indent multiline log messages
    * False or Off or 0 - Do not auto-indent multiline log messages
    * [integer] - auto-indent multiline log messages by [integer] spaces

``extra`` kwarg overrides the value specified on the command line or in the config.

Implements feature request https://github.com/pytest-dev/pytest/issues/5515

### DIFF
--- a/changelog/5515.feature.rst
+++ b/changelog/5515.feature.rst
@@ -1,0 +1,11 @@
+Allow selective auto-indentation of multiline log messages.
+
+Adds command line option ``--log-auto-indent``, config option
+``log_auto_indent`` and support for per-entry configuration of
+indentation behavior on calls to ``logging.log()``.
+
+Alters the default for auto-indention from ``on`` to ``off``. This
+restores the older behavior that existed prior to v4.6.0. This
+reversion to earlier behavior was done because it is better to
+activate new features that may lead to broken tests explicitly
+rather than implicitly.
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1192,6 +1192,29 @@ passed multiple times. The expected format is ``name=value``. For example::
         [pytest]
         junit_suite_name = my_suite
 
+.. confval:: log_auto_indent
+
+    Allow selective auto-indentation of multiline log messages.
+
+    Supports command line option ``--log-auto-indent [value]``
+    and config option ``log_auto_indent = [value]`` to set the
+    auto-indentation behavior for all logging.
+
+    ``[value]`` can be:
+        * True or "On" - Dynamically auto-indent multiline log messages
+        * False or "Off" or 0 - Do not auto-indent multiline log messages (the default behavior)
+        * [positive integer] - auto-indent multiline log messages by [value] spaces
+
+    .. code-block:: ini
+
+        [pytest]
+        log_auto_indent = False
+
+    Supports passing kwarg ``extra={"auto_indent": [value]}`` to
+    calls to ``logging.log()`` to specify auto-indentation behavior for
+    a specific entry in the log. ``extra`` kwarg overrides the value specified
+    on the command line or in the config.
+
 
 .. confval:: log_cli_date_format
 
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -7,6 +7,7 @@ import py
 
 import pytest
 from _pytest.compat import nullcontext
+from _pytest.config import _strtobool
 from _pytest.config import create_terminal_writer
 from _pytest.pathlib import Path
 
@@ -72,24 +73,87 @@ class PercentStyleMultiline(logging.PercentStyle):
     formats the message as if each line were logged separately.
     """
 
+    def __init__(self, fmt, auto_indent):
+        super().__init__(fmt)
+        self._auto_indent = self._get_auto_indent(auto_indent)
+
     @staticmethod
     def _update_message(record_dict, message):
         tmp = record_dict.copy()
         tmp["message"] = message
         return tmp
 
+    @staticmethod
+    def _get_auto_indent(auto_indent_option) -> int:
+        """Determines the current auto indentation setting
+
+        Specify auto indent behavior (on/off/fixed) by passing in
+        extra={"auto_indent": [value]} to the call to logging.log() or
+        using a --log-auto-indent [value] command line or the
+        log_auto_indent [value] config option.
+
+        Default behavior is auto-indent off.
+
+        Using the string "True" or "on" or the boolean True as the value
+        turns auto indent on, using the string "False" or "off" or the
+        boolean False or the int 0 turns it off, and specifying a
+        positive integer fixes the indentation position to the value
+        specified.
+
+        Any other values for the option are invalid, and will silently be
+        converted to the default.
+
+        :param any auto_indent_option: User specified option for indentation
+            from command line, config or extra kwarg. Accepts int, bool or str.
+            str option accepts the same range of values as boolean config options,
+            as well as positive integers represented in str form.
+
+        :returns: indentation value, which can be
+            -1 (automatically determine indentation) or
+            0 (auto-indent turned off) or
+            >0 (explicitly set indentation position).
+        """
+
+        if type(auto_indent_option) is int:
+            return int(auto_indent_option)
+        elif type(auto_indent_option) is str:
+            try:
+                return int(auto_indent_option)
+            except ValueError:
+                pass
+            try:
+                if _strtobool(auto_indent_option):
+                    return -1
+            except ValueError:
+                return 0
+        elif type(auto_indent_option) is bool:
+            if auto_indent_option:
+                return -1
+
+        return 0
+
     def format(self, record):
         if "\n" in record.message:
-            lines = record.message.splitlines()
-            formatted = self._fmt % self._update_message(record.__dict__, lines[0])
-            # TODO optimize this by introducing an option that tells the
-            # logging framework that the indentation doesn't
-            # change. This allows to compute the indentation only once.
-            indentation = _remove_ansi_escape_sequences(formatted).find(lines[0])
-            lines[0] = formatted
-            return ("\n" + " " * indentation).join(lines)
-        else:
-            return self._fmt % record.__dict__
+            if hasattr(record, "auto_indent"):
+                # passed in from the "extra={}" kwarg on the call to logging.log()
+                auto_indent = self._get_auto_indent(record.auto_indent)
+            else:
+                auto_indent = self._auto_indent
+
+            if auto_indent:
+                lines = record.message.splitlines()
+                formatted = self._fmt % self._update_message(record.__dict__, lines[0])
+
+                if auto_indent < 0:
+                    indentation = _remove_ansi_escape_sequences(formatted).find(
+                        lines[0]
+                    )
+                else:
+                    # optimizes logging by allowing a fixed indentation
+                    indentation = auto_indent
+                lines[0] = formatted
+                return ("\n" + " " * indentation).join(lines)
+        return self._fmt % record.__dict__
 
 
 def get_option_ini(config, *names):
@@ -182,6 +246,12 @@ def pytest_addoption(parser):
         dest="log_file_date_format",
         default=DEFAULT_LOG_DATE_FORMAT,
         help="log date format as used by the logging module.",
+    )
+    add_option_ini(
+        "--log-auto-indent",
+        dest="log_auto_indent",
+        default=None,
+        help="Auto-indent multiline messages passed to the logging module. Accepts true|on, false|off or an integer.",
     )
 
 
@@ -413,6 +483,7 @@ class LoggingPlugin:
         self.formatter = self._create_formatter(
             get_option_ini(config, "log_format"),
             get_option_ini(config, "log_date_format"),
+            get_option_ini(config, "log_auto_indent"),
         )
         self.log_level = get_actual_log_level(config, "log_level")
 
@@ -444,7 +515,7 @@ class LoggingPlugin:
         if self._log_cli_enabled():
             self._setup_cli_logging()
 
-    def _create_formatter(self, log_format, log_date_format):
+    def _create_formatter(self, log_format, log_date_format, auto_indent):
         # color option doesn't exist if terminal plugin is disabled
         color = getattr(self._config.option, "color", "no")
         if color != "no" and ColoredLevelFormatter.LEVELNAME_FMT_REGEX.search(
@@ -456,7 +527,10 @@ class LoggingPlugin:
         else:
             formatter = logging.Formatter(log_format, log_date_format)
 
-        formatter._style = PercentStyleMultiline(formatter._style._fmt)
+        formatter._style = PercentStyleMultiline(
+            formatter._style._fmt, auto_indent=auto_indent
+        )
+
         return formatter
 
     def _setup_cli_logging(self):
@@ -473,6 +547,7 @@ class LoggingPlugin:
         log_cli_formatter = self._create_formatter(
             get_option_ini(config, "log_cli_format", "log_format"),
             get_option_ini(config, "log_cli_date_format", "log_date_format"),
+            get_option_ini(config, "log_auto_indent"),
         )
 
         log_cli_level = get_actual_log_level(config, "log_cli_level", "log_level")
--- a/testing/logging/test_formatter.py
+++ b/testing/logging/test_formatter.py
@@ -53,11 +53,75 @@ def test_multiline_message():
     # this is called by logging.Formatter.format
     record.message = record.getMessage()
 
-    style = PercentStyleMultiline(logfmt)
-    output = style.format(record)
+    ai_on_style = PercentStyleMultiline(logfmt, True)
+    output = ai_on_style.format(record)
     assert output == (
         "dummypath                   10 INFO     Test Message line1\n"
         "                                        line2"
+    )
+
+    ai_off_style = PercentStyleMultiline(logfmt, False)
+    output = ai_off_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\nline2"
+    )
+
+    ai_none_style = PercentStyleMultiline(logfmt, None)
+    output = ai_none_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\nline2"
+    )
+
+    record.auto_indent = False
+    output = ai_on_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\nline2"
+    )
+
+    record.auto_indent = True
+    output = ai_off_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\n"
+        "                                        line2"
+    )
+
+    record.auto_indent = "False"
+    output = ai_on_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\nline2"
+    )
+
+    record.auto_indent = "True"
+    output = ai_off_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\n"
+        "                                        line2"
+    )
+
+    # bad string values default to False
+    record.auto_indent = "junk"
+    output = ai_off_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\nline2"
+    )
+
+    # anything other than string or int will default to False
+    record.auto_indent = dict()
+    output = ai_off_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\nline2"
+    )
+
+    record.auto_indent = "5"
+    output = ai_off_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\n     line2"
+    )
+
+    record.auto_indent = 5
+    output = ai_off_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\n     line2"
     )
 
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -177,6 +177,7 @@ Michael Aquilina
 Michael Birtwell
 Michael Droettboom
 Michael Goerz
+Michael Krebs
 Michael Seifert
 Michal Wajszczuk
 Mihai Capotă


changelog/5515.feature.rst
```diff
@@ -0,0 +1,15 @@
+Allow selective auto-indentation of multiline log messages.
```
nicoddemus: That's a great changelog entry! 😍

src/_pytest/logging.py
```diff
@@ -72,24 +72,62 @@ class PercentStyleMultiline(logging.PercentStyle):
...
+        super().__init__(fmt)
+        self._auto_indent = auto_indent
+
     @staticmethod
     def _update_message(record_dict, message):
         tmp = record_dict.copy()
         tmp["message"] = message
         return tmp
 
+    def _get_auto_indent(self, record):
```
nicoddemus: Can we add a docstring here please? It is not entirely obvious all that's being done here.

I also suggest turning this into a function:

```python
def get_auto_indent(record, default_auto_indent) -> int:
```

Where `default_auto_indent` is `self._auto_indent` at the call site.

Then we can unittest this function in isolation and account for all the branches and cases it handles. 👍

src/_pytest/logging.py
```diff
@@ -72,24 +72,62 @@ class PercentStyleMultiline(logging.PercentStyle):
...
-            formatted = self._fmt % self._update_message(record.__dict__, lines[0])
-            # TODO optimize this by introducing an option that tells the
-            # logging framework that the indentation doesn't
-            # change. This allows to compute the indentation only once.
-            indentation = _remove_ansi_escape_sequences(formatted).find(lines[0])
-            lines[0] = formatted
-            return ("\n" + " " * indentation).join(lines)
-        else:
-            return self._fmt % record.__dict__
+            auto_indent = self._get_auto_indent(record)
```
nicoddemus: Just to be perfectly clear, per my suggestion above this becomes:

```python
auto_indent = get_auto_indent(record, self._auto_indent)
```

src/_pytest/logging.py
```diff
@@ -183,6 +221,12 @@ def add_option_ini(option, dest, default=None, type=None, **kwargs):
         default=DEFAULT_LOG_DATE_FORMAT,
         help="log date format as used by the logging module.",
     )
+    add_option_ini(
+        "--log-auto-indent",
+        dest="log_auto_indent",
+        default=None,
+        help="auto indent multiline messages passed to the logging module. accepts true|on, false|off or an integer",
```
nicoddemus: I'm not sure we need to support `on/off`... any reason for it?

changelog/5515.feature.rst
earthman1: Thanks!

src/_pytest/logging.py
earthman1: Docstring: yes, I'll create that now.
Altering the method definition: Yes, I can make that static. Do you feel it is important to make the method public? To me it feels like it is private, but I don't know your views on calling private methods from the unittests.

src/_pytest/logging.py
earthman1: Understood

src/_pytest/logging.py
earthman1: To confirm, you mean only support true/false and remove support for on/off?

src/_pytest/logging.py
earthman1: With respect to the docstring, you indicated that the purpose for it is that "It is not entirely obvious all that's being done here." As I began to write it, it started to seem wordy, because I'm trying to address your concern. Do you really want something detailed?

src/_pytest/logging.py
earthman1: No specific need to support on/off. If you'd like me to remove it, let me know.

src/_pytest/logging.py
nicoddemus: More about the overall purpose of the method

> Altering the method definition:

I meant to make it a function; the fact that it is public is not really relevant, because the module itself is not public (but it is OK to import the private module on our own test suite).

src/_pytest/logging.py
nicoddemus: > To confirm, you mean only support true/false and remove support for on/off?

Exactly, sorry I wasn't clear. Let's remove it, no need to have support for two set of options. 👍 

Thanks!

changelog/5515.feature.rst
```diff
@@ -0,0 +1,15 @@
...
+and config option ``log_auto_indent`` to set the default for all logging.
+
+Supports passing kwarg ``extra={"auto_indent": [value]}`` to
+calls to ``log.log()`` to specify auto-indentation behavior for
+a specific entry in the log.
+
+``[value]`` can be:
+    * True or On - Dynamically auto-indent multiline log messages
+    * False or Off or 0 - Do not auto-indent multiline log messages
+    * [integer] - auto-indent multiline log messages by [integer] spaces
```
thisch: Great that you've added support for configuring the the indentation level explicitly!

changelog/5515.feature.rst
```diff
@@ -0,0 +1,15 @@
+Allow selective auto-indentation of multiline log messages.
+
+Supports command line option ``--log-auto-indent [value]``
+and config option ``log_auto_indent`` to set the default for all logging.
+
+Supports passing kwarg ``extra={"auto_indent": [value]}`` to
+calls to ``log.log()`` to specify auto-indentation behavior for
```
thisch: What do you mean by `log.log`? Probably you mean logging.log from the stdlib, right? Please try to make this a bit clearer in the changelog entry.

changelog/5515.feature.rst
```diff
@@ -0,0 +1,15 @@
+Allow selective auto-indentation of multiline log messages.
+
+Supports command line option ``--log-auto-indent [value]``
+and config option ``log_auto_indent`` to set the default for all logging.
+
+Supports passing kwarg ``extra={"auto_indent": [value]}`` to
```
thisch: What is the motiviation for adding an auto_indent parameter to `log.log`? Do you want to configure auto_indent per log record?

src/_pytest/logging.py
```diff
@@ -72,24 +72,80 @@ class PercentStyleMultiline(logging.PercentStyle):
...
         tmp = record_dict.copy()
         tmp["message"] = message
         return tmp
 
+    @staticmethod
+    def _get_auto_indent(record, default_auto_indent) -> int:
+        """Determines the current auto indentation setting
+
+        Specify auto indent behavior (on/off/fixed) by passing in
+        extra={"auto_indent": [value]} to the call to log.log() or
```
thisch: "s/log.log/logging.log/" ?

src/_pytest/logging.py
```diff
@@ -72,24 +72,80 @@ class PercentStyleMultiline(logging.PercentStyle):
...
+        Default behavior is auto-indent on.
+
+        Using the string "True" or "on" or the boolean True as the value
+        turns auto indent on, using the string "False" or "off" or the
+        boolean False or the int 0 turns it off, and specifying a
+        positive integer fixes the indentation position to the value
+        specified.
+
+        Any other values for the option are invalid, and will silently be
+        converted to "on".
```
thisch: Please document the meaning of the return value using the `:returns:` directive
sth like 
```
:returns: indenation value, which can be
-1 (automatically determine indentation) or
0 (auto-indent turned off) or
>0 (explicitly set indentation position).
```

src/_pytest/logging.py
```diff
@@ -183,6 +239,12 @@ def add_option_ini(option, dest, default=None, type=None, **kwargs):
         default=DEFAULT_LOG_DATE_FORMAT,
         help="log date format as used by the logging module.",
     )
+    add_option_ini(
+        "--log-auto-indent",
+        dest="log_auto_indent",
+        default=None,
```
thisch: Why do you set auto-indent by default to None and not to "true"?

src/_pytest/logging.py
```diff
@@ -72,24 +72,80 @@ class PercentStyleMultiline(logging.PercentStyle):
...
+        converted to "on".
+        """
+        out_auto_indent = -1
+        if hasattr(record, "auto_indent"):
+            # passed in from the "extra={}" kwarg on the call to log.log()
+            in_auto_indent = record.auto_indent
+        elif default_auto_indent is not None:
+            in_auto_indent = default_auto_indent
+        else:
+            in_auto_indent = -1
```
thisch: This branch is only reachable if the user didn't specify a parameter to log-cli-level. Maybe we don't need distinguish between default_auto_indent being None and not-None.

src/_pytest/logging.py
```diff
@@ -72,24 +72,80 @@ class PercentStyleMultiline(logging.PercentStyle):
...
+            else:
+                try:
+                    out_auto_indent = int(in_auto_indent)
+                except ValueError:
+                    out_auto_indent = -1
+        elif type(in_auto_indent) is bool:
+            if in_auto_indent:
+                out_auto_indent = -1
+            else:
+                out_auto_indent = 0
```
thisch: Please move this code to a new method and call it in `__init__`, where you assign the return value to self._auto_indent and in `_get_auto_indent`. You can then change the implemenation of _get_auto_indent to 

```
def _get_auto_indent(self, .....):
        if hasattr(record, "auto_indent"):
            # passed in from the "extra={}" kwarg on the call to log.log()
            out_auto_indent = self.TODO_NEW_METHOD(record.auto_indent)
        else:
            out_auto_indent = self._auto_indent
        return out_auto_indent
```
Since the above method is very simple you can also move it directly to `def format`.

changelog/5515.feature.rst
earonesty: yes, this adds per log line support, so that you can strategically turn it off when printing a preformatted table of trace info, for example.

changelog/5515.feature.rst
earthman1: Thanks

changelog/5515.feature.rst
earthman1: You're right, of course. I'll change that

changelog/5515.feature.rst
earthman1: @earonesty is correct. Sometimes it is good to leave it default on, and then strategically turn it off on a particularly long log entry that doesn't want to be pushed so far to the right, or to set the indentation position to a position that is suited to that entry

src/_pytest/logging.py
earthman1: correct, I'll change it

src/_pytest/logging.py
earthman1: Will add that

src/_pytest/logging.py
earthman1: My preference is to set the variables that represent options to `None` if the option is not specified, rather than to a valid value, so that the default condition can be distinguished from the user explicitly choosing the default. This can make a difference when there are multiple ways to specify the option, and have it fall back if the option was not explicitly set.

This may not specifically relate to the setting I'm working on, but I find it to be a good practice overall because it may not be clear up front whether or not distinguishing _the default_ from _an explicit choice equal to the default_ may be necessary later.

nicoddemus: FTR I like this approach too.

src/_pytest/logging.py
earthman1: I ended up taking a different route with respect to this discussion. I discovered that there is a ``strtobool`` method in pytest's config module, which I simply imported and reused for this purpose, which means I don't need to hard-code the accepted values for the strings, and it will simply accept the same strings as all the config options. 

It also happens to be that the standard for all pytest config options is to accept y/n, yes/no, t/f, true/false, on/off and 1/0, so now this option will follow that standard as well.

src/_pytest/logging.py
earthman1: I followed this advice in the latest iteration, including calling this code for the config option in the `__init__`, but I did not need to create a new method. Having moved the code to `def format` as you suggested, this left all the code that would have been moved to a new method already isolated in `def _get_auto_indent`, so this accomplished what I believe you had intended.

src/_pytest/logging.py
thisch: Thx for the explaination.

src/_pytest/logging.py
thisch: Thx!

earthman1:
Any suggestions how to resolve the codecov/changes failing check? I don't know what "1 file has unexpected coverage changes not visible in diff" means.

earthman1:
After checking, my diff has 100% code coverage, and I don't see how I might have reduced coverage elsewhere

earthman1:
I rebased from the features branch and force pushed in order to fix the base that codecov was comparing against. It still does not pass codecov, so a little assistance getting this approved PR over the finish line would be greatly appreciated

blueyed:
The changes to junitxml in the coverage report are due to flakiness (see #5959).  You can ignore it here.

changelog/5515.feature.rst
blueyed: I'd rather see a link from the changelog to docs, instead of having it expanded/explained that much there.
(people apparently (according to #5952 at least) will find it then in the changelog, and might see an out-of-date documenation)

testing/logging/test_formatter.py
```diff
@@ -53,13 +53,80 @@ def test_multiline_message():
...
+    output = ai_none_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\n"
+        "                                        line2"
+    )
+
+    record.auto_indent = False
+    output = ai_on_style.format(record)
+    assert output == (
+        "dummypath                   10 INFO     Test Message line1\n" "line2"
```
blueyed: This is hard to read / confusing.
Please consider using `.splitlines()` instead and comparing lists instead.
Also below.

blueyed: If black gets in your way here you can also use `# fmt: off` … `# fmt: on`..

earthman1: This is the way the test was written before I introduced my feature change. I merely duplicated the check within the existing test, and adjusted it to check for the new behavior

blueyed: This is all new code, and I do not see `"…line1\n" "line2"` being used anywhere before?!

earthman1: lines 58-61

blueyed: They are not on the same line though.  That's what I mean.
So in your case `"…line1\nline2"` should be used at least.

blueyed: Also please squash and force-push it then already, since we cannot do this from the UI (#4361).

earthman1: I did not place them on the same line, the auto formatter did that. I'm fine with getting rid of the '" "', but once we've gotten to the point where we are code reviewing the auto formatter's contribution to test code, it starts to feel like code review for code review's sake? I've worked on this for over a week, and it would be really nice to get this merged

earthman1: I did not place them on the same line, the auto formatter did that. I'm fine with getting rid of the '" "', but once we've gotten to the point where we are code reviewing the auto formatter's contribution to test code, it starts to feel like code review for code review's sake? I've worked on this for over a week, and it would be really nice to get this merge

blueyed: > I did not place them on the same line, the auto formatter did that.

You ran `black` however (either manually or via pre-commit).
It is not smart enough in this case, so it makes sense to manually review it.

> I've worked on this for over a week, and it would be really nice to get this merged

Sure, but you are not the one maintaining it then from there likely.

I've only skimmed this in general.  As for me I do not like the added complexity, but can see that it is necessary.

Please squash / force-push - assuming that this is now ready from your side, and was approved already.

earthman1: Do you have the ability to merge the branch when codecov is still failing?

blueyed: Yes.

earthman1: squash/force-push complete

earthman1: With respect to your comment that you do not like the added complexity, I am in full agreement. It is my view that, while automatic indentation of multiline log entries may be desirable, it does not belong in the pytest project, it belongs in the logging project instead. That fact that it is here instead of there requires the addition of this extra logic to at least permit controlling/disabling the feature when desired.

earthman1: If this is ready to merge, please proceed, thx

blueyed: > That fact that it is here instead of there requires the addition of this extra logic to at least permit controlling/disabling the feature when desired.

Yes, unfortunately.
It might have been better to not enable it by default, but rather only add it via settings similar to what you've done now.
I've not looked into details, but could imagine that the "sane" behavior would be a good default then?

earthman1: I'd say the behavior is arguably sane. It just needed a bit of extra configuration to globally or selectively revert logging to the original behavior, plus threw in a little extra power because it was easy

earthman1: Is this ready to merge, then?

src/_pytest/logging.py
nicoddemus: Great, good call

testing/logging/test_formatter.py
blueyed: > I'd say the behavior is arguably sane. It just needed a bit of extra configuration to globally or selectively revert logging to the original behavior

What I've meant is that it maybe good to change the defaults here, so that you would not have to configure anything - but people wanting the extra indent would have to instead.

earonesty: Personally, I agree the current behavior might not be a desirable default for some ... but I'd be wary of changing it out from under existing users.    For example: if people have done curious things with their ci logs, like parsing them and loading them into splunk, changing the default would cause those same people may have to modify their parsers.  I'm all for keeping the current default ... not because it's particularly more or less sane, but because leaving defaults the same is often a good choice.

blueyed: @earonesty
Thanks for the feedback.  I've thought we only changed this recently, but it is in 4.6.0 already (https://github.com/pytest-dev/pytest/commit/c8d23c206bd25d869acac2ee87d01289b04c3cda).

blueyed: But according to https://github.com/pytest-dev/pytest/pull/5312#issuecomment-498660973 it does not only affect the output, but also existing tests where `.text` is different then..! (not sure if that is addressed here when changing the output, and/or was fixed inbetween already).

Therefore (and in general) I'd be still +1 to restore the behavior before 4.6.0 by default (i.e. change the defaults in this PR).  It is unfortunate that it would change back and forth then, but seems to be better / more sane in the long term.
I assume @asottile agrees here (using pytest as a tool that should not do too much magic)?

@thisch 
What do you think?

changelog/5515.feature.rst
earthman1: > s/integer/positive integer/ ?

Agreed

earthman1: > I'd rather see a link from the changelog to docs, instead of having it expanded/explained that much there.

Can you provide more detail what you'd like to see here, or perhaps the exact text you prefer? I'm not sure what you're looking for.

blueyed: @earthman1
Not sure what's unclear.. the changelog basically documents the feature/change, while that should be in the docs themselves instead (where it gets updated in the future).

I.e. https://github.com/pytest-dev/pytest/blob/817a42d48fb00cf50553facecdbbc24dadabf36c/changelog/5515.feature.rst should go into https://github.com/pytest-dev/pytest/blame/914a9465aba501e776282961fb3317837772435d/doc/en/reference.rst#L1196, and then the changelog should link to that (IIRC that's not possible though due to pre-commit failing on it then.. https://github.com/pytest-dev/pytest/issues/4283).  So assuming #4283 is still a problem, it should just mention/describe the setting briefly, but not document it there.

src/_pytest/logging.py
```diff
@@ -183,6 +244,12 @@ def add_option_ini(option, dest, default=None, type=None, **kwargs):
         default=DEFAULT_LOG_DATE_FORMAT,
         help="log date format as used by the logging module.",
     )
+    add_option_ini(
+        "--log-auto-indent",
+        dest="log_auto_indent",
+        default=None,
+        help="auto indent multiline messages passed to the logging module. accepts true|on, false|off or an integer",
```
blueyed:
```suggestion
        help="Auto-indent multiline messages passed to the logging module. Accepts true|on, false|off or an integer.",
```

The style with help messages is not really consistent, but that's more in line.. see `pytest --help`.

earthman1: done

blueyed: You missed 3/4 of my suggestions?!

src/_pytest/logging.py
```diff
@@ -183,6 +244,12 @@ def add_option_ini(option, dest, default=None, type=None, **kwargs):
         default=DEFAULT_LOG_DATE_FORMAT,
         help="log date format as used by the logging module.",
     )
+    add_option_ini(
+        "--log-auto-indent",
+        dest="log_auto_indent",
+        default=None,
+        help="Auto indent multiline messages passed to the logging module. accepts true|on, false|off or an integer",
```
earonesty: `Auto indent multiline messages passed to the logging module. Accepts true|on, false|off or an integer.`

src/_pytest/logging.py
earthman1: I didn't see all of the changes because the line was cut off and some of the changes were hidden behind the sideways scroll. I will do it again

earthman1:
Are there any remaining tasks with respect to the PR?

blueyed:
From my side: https://github.com/pytest-dev/pytest/pull/5926#discussion_r335631033 and https://github.com/pytest-dev/pytest/pull/5926#discussion_r335489237.

earthman1:
I made the change to the changelog/reference.rst

earthman1:
Let me know if there are any remaining tasks for me

blueyed:
Thanks for the doc/changelog update.
I am still in the camp to have the old behavior by default (from before you were forced to create these options for it in the first place - https://github.com/pytest-dev/pytest/pull/5926#discussion_r335489237).  Therefore I'll at least not approve/merge it myself - JFI.

earthman1:
Any suggestion how to proceed?

earthman1:
Permit to propose a solution to the current obstacle. I will create a second PR that changes the default from true to false, and then once that PR looks good to you, you can feel better about merging this improvement, and still debate that issue, isolated to the PR that effectuates that specific change.

blueyed:
Why not change it here already?
As far as I can see it cannot be changed before this is merged, since it won't be there already - or do I misunderstand?

testing/logging/test_formatter.py
thisch: And as a justification of the restoration of the old default value, I would only say in the changelog that
"It is better to activate new features that may lead to broken tests explicitly instead of implicitly."

earthman1:
I can change it here, but I am under the impression that changing the default is a matter of further discussion. If we permit the configurability of auto-indentation to be isolated from the changing of the default, that allows this merge to go in without waiting for the resolution of that discussion.

Having a separate PR up that changes the default, that is a branch off of ``optional-multiline`` allows it to successfully merge after this PR merges. The hope is that this will give you the confidence that the discussion is still alive, but without the necessity of holding up a PR that is arguably not related to what the default behavior is (or what it should be)

earthman1:
It looks like there is a consensus building around reverting the default from True to False. I anticipate to submit this change tomorrow.

earthman1:
The default has been changed to log_auto_indent off. Let me know if you have any further comments, questions or requests.

doc/en/reference.rst
```diff
@@ -1192,6 +1192,45 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+    Allow selective auto-indentation of multiline log messages.
+
+    Supports command line option ``--log-auto-indent [value]``
+    and config option ``log_auto_indent = [value]`` to set the
+    auto-indentation behavior for all logging.
+
+    ``[value]`` can be:
+        * True or On - Dynamically auto-indent multiline log messages
+        * False or Off or 0 - Do not auto-indent multiline log messages
+        * [positive integer] - auto-indent multiline log messages by [value] spaces
```
thisch: Please mention that auto-indentation is turned off by default.

doc/en/reference.rst
```diff
@@ -1192,6 +1192,45 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+.. confval:: log_auto_indent
+
+    Allow selective auto-indentation of multiline log messages.
+
+    Supports command line option ``--log-auto-indent [value]``
+    and config option ``log_auto_indent = [value]`` to set the
+    auto-indentation behavior for all logging.
+
+    ``[value]`` can be:
+        * True or On - Dynamically auto-indent multiline log messages
```
thisch: Maybe add quotes around On.

doc/en/reference.rst
```diff
@@ -1192,6 +1192,45 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+        * False or Off or 0 - Do not auto-indent multiline log messages
```
thisch: Same for Off

doc/en/reference.rst
```diff
@@ -1192,6 +1192,45 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+
+        log = logging.getLogger(__name__)
+
+
+        def test_multiline():
+            log.debug("line 1\nline 2 indented", extra={"auto_indent": True})
+            log.warning("line 1\nline 2 not indented", extra={"auto_indent": "off"})
+            log.log(
+                logging.ERROR, "line 1\nline 2 indented by 5 spaces", extra={"auto_indent": 5}
+            )
```
thisch: I'm not sure if we need an usage example for the `extra` kwarg. I think the paragraph starting at L1213 explains the feature pretty well and doesn't need an additional code block.

doc/en/reference.rst
earthman1: done

doc/en/reference.rst
earthman1: done

doc/en/reference.rst
earthman1: done

doc/en/reference.rst
earthman1: done

earthman1:
As before, please let me know if there are any remaining tasks or questions. Thx

changelog/5515.feature.rst
```diff
@@ -0,0 +1,10 @@
+Allow selective auto-indentation of multiline log messages.
+
+Adds command line option ``--log-auto-indent``, config option
+``log_auto_indent`` and support for per-entry configuration of
+indentation behavior on calls to ``logging.log()``.
+
+Alters the default for auto-indention from ``on`` to ``off``,
+reverting to the behavior prior to v4.6.0. This reversion to earlier
+behavior was done because it is better to activate new features that
+may lead to broken tests explicitly rather than implicitly.
```
blueyed: Sorry to be annoying, but I think the changelog should clearly mention that this restores the (older) behavior.

earthman1: Can you provide the language you need? I felt that "...reverting to the behavior prior to v4.6.0. This reversion to earlier behavior was done because..." did clearly indicate that we are restoring old behavior.

earthman1: > Sorry to be annoying, but I think the changelog should clearly mention that this restores the (older) behavior.

You're not annoying, you're committed to quality. 

I changed the wording of the changelog to be closer to the wording in your request. Please let me know if I can improve this further, or if there are any other requests you may have.

earonesty:
Just pointing out and cementing the fact that at 100 mentions, this is the most commented pull req of all 200+ reqs in this repo's history.

thisch:
@nicoddemus I would like to merge this PR. Any objections/suggestions?

earthman1:
> Just pointing out and cementing the fact that at 100 mentions, this is the most commented pull req of all 200+ reqs in this repo's history.

Wow, I hadn't noticed that. Funny

link: [PRs sorted by most commented](https://github.com/pytest-dev/pytest/pulls?q=is%3Apr+is%3Aclosed+sort%3Acomments-desc)

I think it's more like 2,600 or so PRs at this point. [The original PR](https://github.com/pytest-dev/pytest/pull/5312) had 8 comments :-)

earthman1:
Haven't seen any activity on the PR in a week, just checking in

blueyed:
Yeah...
I was waiting for others to (re-)approve, but apparenently they are busy with other things.

Since @thisch approved a week ago already, and I am very happy for you to have understood my commitment to quality I'm going to merge it now.

Thanks!  (also to everyone having contributed to the constructive part of the 100+ comments)

Issue #5192, opened 2019-05-01
Failed test output is less verbose with -v than not using -v


pfctdayelise:
I don't know if this has changed recently but it seems in some cases using -v makes the output *less* verbose...

```
def test_verbosity():
    s = [{'SKU': '12345678', 'Name': 'Linda', 'Description': 'Fancy dress', 'Net Price': '9.16', 'VAT rate': '20', 'VAT': '1.83', 'Gross Price': '10.99', 'Colour': 'Grey', 'Image': 'FancyDressImage'}, {'SKU': '12345679', 'Name': 'Jane', 'Description': 'Skirt', 'Net Price': '10.00', 'VAT rate': '0', 'VAT': '0.00', 'Gross Price': '10.00', 'Colour': 'Pink', 'Image': 'ChildsSkirt'}]
    assert len(s) == 1
```

Normal output
```
$ pytest -k verbosity --tb=short
====================================== test session starts ======================================
platform linux -- Python 3.6.7, pytest-4.4.1, py-1.8.0, pluggy-0.9.0
rootdir: /mnt/c/Users/bjl/wsl/bdd-examples-project/tests, inifile: pytest.ini
plugins: bdd-3.1.0
collected 27 items / 26 deselected / 1 selected

steps/test_example.py F                                                                   [100%]

=========================================== FAILURES ============================================
________________________________________ test_verbosity _________________________________________
steps/test_example.py:55: in test_verbosity
    assert len(s) == 1
E   AssertionError: assert 2 == 1
E    +  where 2 = len([{'Colour': 'Grey', 'Description': 'Fancy dress', 'Gross Price': '10.99', 'Image': 'FancyDressImage', ...}, {'Colour': 'Pink', 'Description': 'Skirt', 'Gross Price': '10.00', 'Image': 'ChildsSkirt', ...}])
============================ 1 failed, 26 deselected in 0.12 seconds ============================
```

Output with -v
```
$ pytest -k verbosity --tb=short -v
====================================== test session starts ======================================
platform linux -- Python 3.6.7, pytest-4.4.1, py-1.8.0, pluggy-0.9.0 -- /home/bjl/.local/share/virtualenvs/bdd-examples-project-dSboamh_/bin/python3
cachedir: .pytest_cache
rootdir: /mnt/c/Users/bjl/wsl/bdd-examples-project/tests, inifile: pytest.ini
plugins: bdd-3.1.0
collected 27 items / 26 deselected / 1 selected

[...]

Issue #5932, opened 2019-10-09
Increasing verbosity turns off the assertion rewriting


JBKahn:
Given a file:
```python
a = [1, 2, 3]
b = [1, 2]

def test_a():
    assert len(a) == len(b)

def test_bigger_assertion():
    first_dict = {a: a for a in range(100)}
    second_dict = {b: b for b in range(1, 101)}
    assert first_dict == second_dict
```

pytest will output
```
 ✔  pytest test_file.py
============================================================================================== test session starts ==============================================================================================
platform darwin -- Python 3.7.4, pytest-5.2.1, py-1.8.0, pluggy-0.13.0
rootdir: /Users/jkahn/wave/src/external/pytest-clarity
collected 2 items

test_file.py FF                                                                                                                                                                                           [100%]

=================================================================================================== FAILURES ====================================================================================================
____________________________________________________________________________________________________ test_a _____________________________________________________________________________________________________

    def test_a():
>       assert len(a) == len(b)
E       assert 3 == 2
E        +  where 3 = len([1, 2, 3])
E        +  and   2 = len([1, 2])

test_file.py:5: AssertionError
_____________________________________________________________________________________________ test_bigger_assertion _____________________________________________________________________________________________

    def test_bigger_assertion():
        first_dict = {a: a for a in range(100)}
        second_dict = {b: b for b in range(1, 101)}
>       assert first_dict == second_dict
E       assert {0: 0, 1: 1, 2: 2, 3: 3, ...} == {1: 1, 2: 2, 3: 3, 4: 4, ...}
E         Omitting 99 identical items, use -vv to show
E         Left contains 1 more item:
E         {0: 0}
E         Right contains 1 more item:
E         {100: 100}
E         Use -v to get the full diff

test_file.py:10: AssertionError
=============================================================================================== 2 failed in 0.04s ===============================================================================================
```

However if I increase the verbosity, so that I can see more of the diff (either `-v` or `-vv`) like so, then it removed the helpful assertion rewriting from the first test

```
 ↵ 1  pytest test_file.py -vv
============================================================================================== test session starts ==============================================================================================
platform darwin -- Python 3.7.4, pytest-5.2.1, py-1.8.0, pluggy-0.13.0 -- /Users/jkahn/.local/share/virtualenvs/pytest-clarity-JGV2Xd-w/bin/python3.7
cachedir: .pytest_cache
rootdir: /Users/jkahn/wave/src/external/pytest-clarity
collected 2 items

[...]

Issue #5931, opened 2019-10-08
Help with plugin assertion rewriting


JBKahn:
Hey, I was looking at the [pytest-clarity package](https://github.com/darrenburns/pytest-clarity) and it does something really nice, imo, which is it turns this:

![pytest](https://raw.githubusercontent.com/darrenburns/pytest-clarity/master/without-clarity.png)

into:

![pytest](https://raw.githubusercontent.com/darrenburns/pytest-clarity/master/pytest-clarity.png)

The downside is that the rewriting results in a loss of functionality:

```python
a = [1, 2, 3]
b = [1, 2]

def test_a():
    assert len(a) == len(b)
```

Goes from
```
    def test_a():
>       assert len(a) == len(b)
E       assert 3 == 2
E        +  where 3 = len([1, 2, 3])
E        +  and   2 = len([1, 2])
```

to

```
    def test_a():
>       assert len(a) == len(b)
E       assert left == right failed.
E         Showing split diff:
E
E         left:  3
E         right: 2
```

I've been digging through the source and I can't quite find the right part to use to do this. It stems from use of `pytest_assertrepr_compare` but I can't find anything in pytest I can use to construct that `3 = len([1, 2, 3])`. I'm looking for some pointers if anyone can help put me on the right track. I was barking down  the `AssertionRewriter` which I think is what I'm looking for but I can't quite put my finger on how to modify these assertions in way that still utilizes that code.

blueyed:
Should we also close this as duplicate?  (#5932)

blueyed:
Might be good to get feedback in https://github.com/pytest-dev/pytest/pull/5933 how the new method / alternatives would be useful for the plugin, since the plugin would benefit here from having the original explanation (or a way to say where it should be inserted (via special placeholder)).

JBKahn:
I'll take a look tomorrow myself

JBKahn:
Thanks for getting back so quickly!

PR #5933, opened 2019-10-09
[RFC] assert: keep "where … and …" output in verbose mode

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/5932
Fixes: https://github.com/pytest-dev/pytest/issues/5192

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -14,7 +14,9 @@ from _pytest.compat import ATTRS_EQ_FIELD
 # interpretation code and assertion rewriter to detect this plugin was
 # loaded and in turn call the hooks defined here as part of the
 # DebugInterpreter.
-_reprcompare = None  # type: Optional[Callable[[str, object, object], Optional[str]]]
+_reprcompare = (
+    None
+)  # type: Optional[Callable[[str, object, object, str], Optional[str]]]
 
 # Works similarly as _reprcompare attribute. Is populated with the hook call
 # when pytest_runtest_setup is called.
@@ -121,7 +123,7 @@ def isiterable(obj):
         return False
 
 
-def assertrepr_compare(config, op, left, right):
+def assertrepr_compare(config, op, left, right, expl):
     """Return specialised explanations for some operators/operands"""
     maxsize = (80 - 15 - len(op) - 2) // 2  # 15 chars indentation, 1 space around op
     left_repr = saferepr(left, maxsize=maxsize)
@@ -146,7 +148,7 @@ def assertrepr_compare(config, op, left, right):
                     type_fn = (isdatacls, isattrs)
                     explanation = _compare_eq_cls(left, right, verbose, type_fn)
                 elif verbose > 0:
-                    explanation = _compare_eq_verbose(left, right)
+                    return expl + "\n~" + "\n~".join(_compare_eq_verbose(left, right))
                 if isiterable(left) and isiterable(right):
                     expl = _compare_eq_iterable(left, right, verbose)
                     if explanation is not None:
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -472,7 +472,7 @@ def pytest_unconfigure(config):
 # -------------------------------------------------------------------------
 
 
-def pytest_assertrepr_compare(config, op, left, right):
+def pytest_assertrepr_compare(config, op, left, right, expl):
     """return explanation for comparisons in failing assert expressions.
 
     Return None for no custom explanation, otherwise return a list
--- a/src/_pytest/assertion/__init__.py
+++ b/src/_pytest/assertion/__init__.py
@@ -110,8 +110,8 @@ def pytest_runtest_setup(item):
     comparison for the test.
     """
 
-    def callbinrepr(op, left, right):
-        # type: (str, object, object) -> Optional[str]
+    def callbinrepr(op, left, right, expl):
+        # type: (str, object, object, str) -> Optional[str]
         """Call the pytest_assertrepr_compare hook and prepare the result
 
         This uses the first result from the hook and then ensures the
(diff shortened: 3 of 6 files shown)


blueyed:
This is quite hacky (but might still pass existing tests..).

I think it might be better to replace something like "~%(default)" instead, which would avoid having to pass through the original `repl`.
OTOH it might be useful in custom hooks.

blueyed:
/cc @darrenburns via https://github.com/darrenburns/pytest-clarity (https://github.com/pytest-dev/pytest/issues/5931)
I guess your feedback would be helpful here.. :)

darrenburns:
This would be really cool! I personally prefer this approach over the templating suggestion, but I guess it may break backwards compatibility since the hook interface will change?

src/_pytest/hookspec.py
```diff
@@ -472,7 +472,7 @@ def pytest_unconfigure(config):
 # -------------------------------------------------------------------------
 
 
-def pytest_assertrepr_compare(config, op, left, right):
+def pytest_assertrepr_compare(config, op, left, right, expl):
```
nicoddemus: Please update the docstring with `expl` as well. 😁

src/_pytest/assertion/util.py
```diff
@@ -146,7 +148,7 @@ def assertrepr_compare(config, op, left, right):
                     type_fn = (isdatacls, isattrs)
                     explanation = _compare_eq_cls(left, right, verbose, type_fn)
                 elif verbose > 0:
-                    explanation = _compare_eq_verbose(left, right)
+                    return expl + "\n~" + "\n~".join(_compare_eq_verbose(left, right))
```
nicoddemus: Sorry, I don't get the `"\n~"` part...? 🤔

nicoddemus:
> but I guess it may break backwards compatibility since the hook interface will change?

No, `pluggy` is smart and discards parameters that an implementation doesn't declare. 😉

src/_pytest/hookspec.py
blueyed: Would need more updating in general then, especially for when we go with the return-string-to-skip special postprocessing approach.

src/_pytest/assertion/util.py
blueyed: It mimics https://github.com/blueyed/pytest/blob/1543a16330a5be10022505429b66608e7ee0ed19/src/_pytest/assertion/__init__.py#L139 (which is skipped for strings).
It's the "mini language" used with https://github.com/blueyed/pytest/blob/1543a16330a5be10022505429b66608e7ee0ed19/src/_pytest/assertion/util.py#L59-L91.

blueyed:
Why do you both prefer this over a special template var (keep in mind that you have to use the "mini template language" here already).
It requires you to be aware of internals more, including that you have to prefix things with "~"/">".
It also clearly is a hack to handle strings in a special way (to skip post-processing).

blueyed:
@nicoddemus 
Could you make up your mind?

(it is a long-standing regression, so it would be great to make some progress here)

blueyed:
Who can we ping about this - being more into the assertion rewriting part of pytest?

nicoddemus:
> Could you make up your mind?

Sorry about the delay, I've been busy with other things.

The problem that I have with the `expl` parameter is that the user can't really do anything with it in practice:

```python
def y():
    return [0, 1]

def test():
    assert y() == [1, 2, 3, 4]
```

For this test failure, the `expl` parameter is:

```
"%(py2)s\n{%(py2)s = %(py0)s()\n} == %(py5)s"
```

So the `expl` parameter of `pytest_assertrepr_compare` is only really usefl for pytest's own internal implementation, `assertrepr_compare`.

I see two options:

1. We have to rethink how `pytest_assertrepr_compare` actually works so users can reuse the variables saved by pytest's rewriting mechanism.
2. We need to somehow pass `expl` from the rewriting mechanism directly to `assertrepr_compare` somehow.

I wouldn't even know where to start with 1), and 2) would probably require some hacking, like setting `expl` to a global. 

At first look 2) is ugly, but we might evaluate it under the light that we already are doing something similar by installing `util._reprcompare` during `pytest_runtest_setup`. What if we set `util._reprcompare_expls` during `_call_reprcompare`?

```python
    if util._reprcompare is not None:
        util._reprcompare_expl = expl
        try:
            custom = util._reprcompare(ops[i], each_obj[i], each_obj[i + 1])
        finally:
            util._reprcompare_expl = None
        if custom is not None:
            return custom
```

Again, while this is ugly, at least it is kept internal and won't introduce a parameter which we might regret later.

What do you think?

blueyed:
> The problem that I have with the `expl` parameter is that the user can't really do anything with it in practice:

Yes.

> What if we set `util._reprcompare_expls` during `_call_reprcompare`?

How does that help plugins with using the default?

nicoddemus:
> How does that help plugins with using the default?

Unless I'm misunderstanding, I thought the patch was about to fix "keep "where … and …" output in verbose mode" in the pytest core, not necessarily expose to plugins necessarily.

In my previous post, `assertrepr_compare` would access the global `util._reprcompare_expl` instead of receiving it by argument... again a hack, but we can avoid exposing a `expl` parameter to users, which would require us to carry this over many releases if we ever find a better way to represent the explanation obtained by the rewriter.

blueyed:
@nicoddemus 
The idea here is to also make this available for plugins / do it "properly".  See previous comment: https://github.com/pytest-dev/pytest/pull/6104

nicoddemus:
> The idea here is to also make this available for plugins / do it "properly". See previous comment: #6104

Is that the right issue number? 🤔

blueyed:
> > The idea here is to also make this available for plugins / do it "properly". See previous comment: #6104
> 
> Is that the right issue number?

No, sorry.  More likely https://github.com/pytest-dev/pytest/issues/5932.

JBKahn:
@blueyed does closing this mean no charge coming?

blueyed:
@JBKahn I've closed all my open PRs for now.  Feel free to pick it up, I might come back to it later myself, but just had to clean up things for myself.

JBKahn:
@nicoddemus any more thoughts on this one? We'd really like to be able to use the explanation in the rewrite

Issue #3962, opened 2018-09-10
Add distinct option to show full (assertion) diffs always


blueyed:
Currently you have to use `-vv` to show the whole diff for dicts, and not to truncate in general.

Code:

 - https://github.com/pytest-dev/pytest/blob/c00d934b213606fac60c2502e5bc8fe5b3e2aecb/src/_pytest/assertion/util.py#L295-L296
 - https://github.com/pytest-dev/pytest/blob/c00d934b213606fac60c2502e5bc8fe5b3e2aecb/src/_pytest/assertion/truncate.py#L27-L32

1. there should be a distinct option additionally, because I usually have to re-run the test with `-vv` then, but would rather have it by default - without using very verbose mode by default

2. `pytest.assertion.truncate` checks if it is running on CI and enables this then always.  This should also be done for comparing dicts anyway.  Should I file a separate issue / create a PR for this already?

I would like to add this to the config file for projects where it makes sense, e.g. https://github.com/Vimjas/covimerage, where most of the tests are comparing program output (by lines), which usually gets truncated.
`-vv` would still have the same effect of course.

cheezman34:
A potential workaround that might work for some:

Add to your conftest.py
```
from _pytest.assertion.util import assertrepr_compare
def pytest_assertrepr_compare(config, op, left, right):
    # hack the verbosity so we always show full diffs on assertion failures,
    # even if we're otherwise not fully verbose
    config.option.verbose = 2
    return assertrepr_compare(config, op, left, right)
```

blueyed:
The hack does not for truncation via `saferepr` unfortunately (for a overlong assertion message itself).

See also https://github.com/pytest-dev/pytest/issues/2382 and https://github.com/pytest-dev/pytest/issues/3693.

This is really frustrating, but at least `--assert=plain` can be used to get the full assertion message.

blueyed:
@pytest-dev/contributors 
Any input on this (given that there are already some WIP PRs to fix issues in this regard), please?

blueyed:
btw/JFI: for skipping truncation (just one part of it), you can use `CI=true` as a workaround:
https://github.com/pytest-dev/pytest/blob/cbe3e3ebd3ca941118254025764aefd392383a8a/src/_pytest/assertion/truncate.py#L23-L34

nicoddemus:
> Any input on this (given that there are already some WIP PRs to fix issues in this regard), please?

[...]

PR #5936, opened 2019-10-09
assertrepr_compare: use safeformat with -vv

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/3962

TODO:

- [x] changelog

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -7,6 +7,7 @@ from typing import Optional
 
 import _pytest._code
 from _pytest import outcomes
+from _pytest._io.saferepr import safeformat
 from _pytest._io.saferepr import saferepr
 from _pytest.compat import ATTRS_EQ_FIELD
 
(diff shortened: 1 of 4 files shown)


src/_pytest/assertion/util.py
```diff
@@ -123,13 +124,21 @@ def isiterable(obj):
 
 def assertrepr_compare(config, op, left, right):
     """Return specialised explanations for some operators/operands"""
-    maxsize = (80 - 15 - len(op) - 2) // 2  # 15 chars indentation, 1 space around op
-    left_repr = saferepr(left, maxsize=maxsize)
-    right_repr = saferepr(right, maxsize=maxsize)
+    verbose = config.getoption("verbose")
+    if verbose > 1:
+        left_repr = safeformat(left)
```
nicoddemus: So the idea here is to don't use `maxsize` with `-vv`? Seems reasonable.

Please add a CHANGELOG entry as well.

blueyed:
Uncovered a bug in pytest-xdist: https://github.com/pytest-dev/pytest-xdist/pull/478

blueyed:
Only added changelog, should be good.

blueyed:
Should I pull out the CI improvement into a separate PR, or can we merge this already?

PR #5950, opened 2019-10-12
Casting fixture parameter to list at the beginning of parameter parsing.

### LINKED ISSUE #5946: Fixture parametrization breaks if parameter is a numpy array
**Description**

On latest release and also on master branch, windows 10

The following test:

```python
from numpy import linspace
from pytest import fixture

@fixture(params=linspace(1, 10, 10))
def value(request):
    return request.param

def test_bug(value):
    assert value == value
```

breaks with:

```python-traceback
  File "D:/WORK/pytest-nparray-params-bug/test_bug.py", line 4, in <module>
    @fixture(params=linspace(1, 10, 10))
  File "D:\WORK\pytest-nparray-params-bug\venv\lib\site-packages\_pytest\fixtures.py", line 1123, in fixture
    name=name
  File "D:\WORK\pytest-nparray-params-bug\venv\lib\site-packages\_pytest\fixtures.py", line 1032, in _parse_fixture_args
    key: value for key, value in kwargs.items() if arguments.get(key) != value
  File "D:\WORK\pytest-nparray-params-bug\venv\lib\site-packages\_pytest\fixtures.py", line 1032, in <dictcomp>
    key: value for key, value in kwargs.items() if arguments.get(key) != value
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

**Cause**

It seems that this line:

https://github.com/pytest-dev/pytest/blob/b162ab6a45d5432744b193bb2efbd80a6fe2cc6c/src/_pytest/fixtures.py#L1031-L1033

Tries to get a boolean result from comparing `None != <array>` which is ambiguous.

**Possible solution**

This might be solved by moving the cast to list (line 1137) at the very beginning (1116) of this function:

https://github.com/pytest-dev/pytest/blob/b162ab6a45d5432744b193bb2efbd80a6fe2cc6c/src/_pytest/fixtures.py#L1116-L1139

I am actually running tests with this change but not 100% certain about side-effects, although it seems to be that casting to list that soon should not cause problems (equivalent of casting to list in user test code).

If you think this is okay I can create the PR.

Sup3rGeo:
Closes #5946 

- [x] (NON APPLICABLE) Target the `features` branch for new features, improvements, and removals/deprecations.
- [x] (NON APPLICABLE) Include documentation when adding new features.


src/_pytest/fixtures.py
```diff
@@ -1113,6 +1113,9 @@ def fixture(
                 ``fixture_<fixturename>`` and then use
                 ``@pytest.fixture(name='<fixturename>')``.
     """
+    if params is not None and not isinstance(params, (list, tuple)):
```
nicoddemus: I think we can always create a list, for consistency:

```suggestion
    if params is not None:
```

It also ensures our params list is not affected by changes from the outside. I don't expect this to make any difference performance-wise.

testing/python/fixtures.py
```diff
@@ -4187,3 +4187,21 @@ def test_check_fixture_instantiations():
     )
     result = testdir.runpytest()
     result.assert_outcomes(passed=7)
+
+
+def test_fixture_parametrization_nparray(testdir):
+    testdir.makepyfile(
```
nicoddemus: I'm not sure if `numpy` is installed in all test environments; if not, I suggest using `importorskip` here:

```suggestion
    pytest.importorskip('numpy')
    testdir.makepyfile(
```

nicoddemus:
Looks great, thanks @Sup3rGeo!

testing/python/fixtures.py
Sup3rGeo: Well, I also cannot tell. In this case what should we do?

Sup3rGeo: Adding `numpy` to `extras_require` in `setup.py` don't make sure the environments have numpy installed?

nicoddemus: I'm just commenting in case we get a failure on CI because of this. If everything passes, please disregard this. 👍

Sup3rGeo: I see. All tests seem to have passed, the only failed check being something related to coverage - but I didn't understand the reason why.

Sup3rGeo:
It seems that because we now always have `numpy` in the testing environments, the part that return false in this function never gets executed/tested:

https://github.com/pytest-dev/pytest/blob/4f2abd7ae08585d9695f78da9c4cc24dcf918947/src/_pytest/python_api.py#L529-L539

What to do about it?

nicoddemus:
TBH I don't think we need to bother testing that, adding `# pragma: no cover` to the `return False` should be enough.

nicoddemus:
Thanks for the PR and reference! 👍

setup.py
```diff
@@ -28,6 +28,7 @@ def main():
                 "mock",
                 "nose",
                 "requests",
+                "numpy",
```
blueyed: @nicoddemus 
Adding numpy to all builds make the numpy tox env obsolete, doesn't it?
I'd rather not have it installed always, but in case this is kept it should clean up the env, jobs, and conditional skips we have for it.

nicoddemus: Oh right, thanks for catching that.

Practically I think it makes sense to install an additional dependency in order to have less jobs, as an extra dependency is usually much quicker to install than starting up an entire new job, which potentionally might run all tests again.

I will follow up with a fix.

blueyed: Just started with it already myself.

blueyed: I wanted to keep numpy optional, but what you've said sounds different?
Keep in mind that numpy might be slow to install without a wheel.

nicoddemus: You are right.

Let's remove `numpy` from `extras` and use `importorskip` on the test.

I have the diff ready:

```diff
diff --git a/setup.py b/setup.py
index be38857b0..dcf63f6fd 100644
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,6 @@ def main():
                 "mock",
                 "nose",
                 "requests",
-                "numpy",
                 "xmlschema",
             ]
         },
diff --git a/testing/python/fixtures.py b/testing/python/fixtures.py
index 4fcab0245..9ef7f0245 100644
--- a/testing/python/fixtures.py
+++ b/testing/python/fixtures.py
@@ -4190,6 +4190,7 @@ def test_indirect_fixture_does_not_break_scope(testdir):


 def test_fixture_parametrization_nparray(testdir):
+    pytest.importorskip("numpy")
     testdir.makepyfile(
         """
         from numpy import linspace
```

PR #5951, opened 2019-10-12
Workaround curl bug which makes retries of fetching codecov.io/bash not work

bluetech:
Fetching this URL is very flaky, even after adding retries. I found a possible explanation here: https://github.com/python-trio/trio/issues/1231, let's try it!

### DIFF
--- a/scripts/report-coverage.sh
+++ b/scripts/report-coverage.sh
@@ -13,5 +13,6 @@ fi
 python -m coverage combine
 python -m coverage xml
 python -m coverage report -m
-curl -S -L --retry 6 -s https://codecov.io/bash -o codecov-upload.sh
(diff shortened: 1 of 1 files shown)


blueyed:
JFI: might still fail:
```
+ curl -S -L --connect-timeout 5 --retry 6 -s https://codecov.io/bash -o codecov-upload.sh
curl: (52) Empty reply from server
```
https://pytest-dev.visualstudio.com/pytest/_build/results?buildId=2556&view=logs&jobId=b93ccf05-0a7e-58c4-578e-33859c8603c3&taskId=5025fdd7-3029-57fb-e408-894fff7790eb&lineStart=119&lineEnd=121&colStart=1&colEnd=1

Issue #5523, opened 2019-06-29
Command line parsing error with Python 3.8


thorstenkampe:
- [X] a detailed description of the bug or suggestion
Command line parsing error with Python 3.8 when concatenating short options
- [X] pytest and operating system versions
Pytest 5.0.0, Python 3.8.0b1, Windows 10
- [X] minimal example if possible
```
> pytest -vx
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -vx
  inifile: None
  rootdir: F:\Documents\Scripts\Runscript
```
`pytest -v -x` works

nicoddemus:
Hmmm indeed we only apply a workaround for that Python bug in Python < 3.8:

https://github.com/pytest-dev/pytest/blob/4bc041572022624bbda81e46efb419fabd8d782e/src/_pytest/config/argparsing.py#L361

@Zac-HD any thoughts? Perhaps we should use `< 3.9` until the patch lands and a new 3.8 gets released?

Zac-HD:
Sure - we should probably drop in an assertion that we're running on Python < 3.9 too, with a message that will remind us what to follow up on if it's slipped until then :smile:

nicoddemus:
Do you have some time to get to this later @Zac-HD? If you don't I can pick it up later in the weekend.

Zac-HD:
Probably not sorry - family commitments all Sunday and a lot of writing I'm already behind on...

nicoddemus:
No worries, thanks!

asottile:
going to unpin since 5.0.1 is out 👍

encukou:
The Python fix should get into the upcoming 3.8 beta 3.

PR #5952, opened 2019-10-13
Port CHANGELOG from 4.6.6 release

nicoddemus:


blueyed:
Shouldn't we merge the tag instead?
#5523 shows up 3 times in the changelog now.

blueyed:
I think the changelog for 4.6 should not go into master / 5.x at all, no?

blueyed:
As for #5523 in particular: it shows up with 5.0.1 and 5.1.0, probably because master was not merged into features before doing the release?

asottile:
yes we should include the 4.6 changelog, otherwise it does not appear in the docs

blueyed:
It appears with https://docs.pytest.org/en/4.6-maintenance/changelog.html.

asottile:
yes but nobody is going to land on that

blueyed:
If you are using 4.x you should look at the docs for it, no?

asottile:
should? perhaps -- but I don't expect anyone to be able to find anything except latest from google, etc.

blueyed:
Well, it might actually help Google etc if we would mention 4.6.x not within 5.x docs.

Apart from that, is googling for pytest changes/changelogs a common / valid use case after all?
(IMHO I think that is rather distracting to land on the changelog when you're interested in actual docs, but that's more an issue of (too verbose) changelogs in general - but it does not help to duplicate items there then additionally)

What about a clear note at the top that changes for 4.6.x are somewhere else? (with the above link)

blueyed:
Also people using 5.x (the common use case) have to skip 4.6 entries when looking at the changelog.

asottile:
the 4.6 changelog entries are already below 5.x ones, not sure how anyone would mistake the two

blueyed:
I've not said that they could be mistaken.

The-Compiler:
FWIW pylint does this (only having changelog in branches) and (at least when I followed its development a while back) it was a frequent source of confusion. IMHO, the changelog in the "main" docs should contain all versions.

blueyed:
@The-Compiler 
Any references?  (I've tried to find something via https://github.com/PyCQA/pylint/issues?q=is%3Aissue+changelog+is%3Aclosed only quickly, but could not)

There's no such thing as "main" docs really.
I can see that the docs default to "latest", but then it should be made more clear (apart from the RTD version selector) maybe, that you are looking at docs for 5.x.

It makes sense to merge the current release branch into the development branch before doing a new major release (as we are doing with master => features before a new release from there in general), but here we have two separate release branches really (5.x (master), and 4.6.x (4.6-maintenance)).
Having to copy/add changelogs manually (instead of through a merge) is clearly an indicator that this might be wrong, isn't it?

PR #5955, opened 2019-10-14
ci: codecov: disable comments really

blueyed:
Apparently it has to be "false" instead of "off".

Follow-up to d50198a3f.


blueyed:
Was missing in 4.6-maintenance only: https://github.com/pytest-dev/pytest/pull/5957

PR #5957, opened 2019-10-14
[4.6] ci: add codecov.yml to turn comments off

blueyed:
The only benefit for me is to get notified about finished builds, but
that might happen to early anyway.  Apart from that they are rather big
and distract from actual comments.

(cherry picked from commit d50198a3ff1458b6b175cb9ae274a5a98be965ab)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,7 @@
+coverage:
+  status:
+    project: true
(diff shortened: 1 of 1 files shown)


blueyed:
What's up with Azure / pytest-CI?

PR #5959, opened 2019-10-14
[WIP] TEST: junitxml: write_captured_output: assert for flakiness

blueyed:


src/_pytest/junitxml.py
```diff
@@ -207,6 +207,8 @@ def write_captured_output(self, report):
             else:
                 content = content_err
 
+            # XXX: flaky in test suite, trying to figure out why.
+            assert content, repr([content, content_log, content_err])
             if content:
```
blueyed: @nicoddemus 
Any idea why this might be Falsy during tests?
Does this happen during failures of outer tests then only?

Does it make sense to cover this branch, or should we either assert it here then always (probably a bad idea since it is covered sometimes), or just ignore it for coverage (`# pragma: no branch`)?

nicoddemus:
(@blueyed please use `[WIP]` instead of `[W I P]` so the WIP bot can hold back the PR)

src/_pytest/junitxml.py
nicoddemus: Not sure, do you have an example failure?

blueyed: No.

blueyed:
@nicoddemus 
JFI: the flakiness there marks a lot of PRs with an red X due to the flaky coverage.

nicoddemus:
Yeah I've noticed. I did not notice though it coming from junitxml, but I've often seen flaky coverage in capture.py:

https://codecov.io/gh/pytest-dev/pytest/compare/afabbb6346e1b76cf84f367c3198e3313fbef959...c481df2edec82f4557058129e9f7b40d4f3973d8/changes

blueyed:
```
if (
        not sys.platform.startswith("win32")
        or sys.version_info[:2] < (3, 6)
        or hasattr(sys, "pypy_version_info")
    ):
        return
```
I assume that is only the case before all jobs have finished then?

Issue #5906, opened 2019-10-02
Crash when using --setup-show and KeyboardInterrupt


iwanb:
When interrupting a test using Ctrl-C (KeyboardInterrupt/SIGINT), the "setuponly" plugin crashes due to some interaction with the capture manager.

Using pytest 5.2.0 (fresh install on Python 3.7 in a virtual env), given a dummy test:
```
import time
import pytest


@pytest.fixture()
def setup():
    print("Setup")
    yield
    print("Teardown")


def test_something(setup):
    time.sleep(60)
```

Run it with `pytest --setup-show test_teardown.py` and abort it while the test is running.

Stacktrace:
```
Traceback (most recent call last):
  File "venv/bin/pytest", line 10, in <module>
    sys.exit(main())
  File "venv/lib/python3.7/site-packages/_pytest/config/__init__.py", line 90, in main
    return config.hook.pytest_cmdline_main(config=config)
  File "venv/lib/python3.7/site-packages/pluggy/hooks.py", line 289, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "venv/lib/python3.7/site-packages/pluggy/manager.py", line 87, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "venv/lib/python3.7/site-packages/pluggy/manager.py", line 81, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "venv/lib/python3.7/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "venv/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "venv/lib/python3.7/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "venv/lib/python3.7/site-packages/_pytest/main.py", line 228, in pytest_cmdline_main
    return wrap_session(config, _main)
  File "venv/lib/python3.7/site-packages/_pytest/main.py", line 221, in wrap_session
    session=session, exitstatus=session.exitstatus
  File "venv/lib/python3.7/site-packages/pluggy/hooks.py", line 289, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "venv/lib/python3.7/site-packages/pluggy/manager.py", line 87, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
[...]

PR #5960, opened 2019-10-14
[WIP/RFC] setuponly: remove printing out/err from capman

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5906


blueyed:
`stop_global_capturing` pops it already: https://github.com/pytest-dev/pytest/blob/a20880cca268c4f7fea393145f47c93b5d9c1dac/src/_pytest/capture.py#L116

asottile:
:S I was helping @yoavcaspi make their first contribution

blueyed:
@asottile 
https://github.com/pytest-dev/pytest/issues/5906#issuecomment-541933918 ?

That was the wrong direction though.

I've just created a PR already while looking into it.

Can you approve it then maybe?

asottile:
next time help them in the right direction then!  it's valuable to get people excited about contributing

I don't think this is mergeable yet, there's no test for the bug or change

blueyed:
> next time help them in the right direction then! it's valuable to get people excited about contributing

Sure, but should I have then just posted the diff instead to the issue?
From my experience things here often get stalled, and then it's better to have a fix for things right away, isn't it?

> I don't think this is mergeable yet, there's no test for the bug or change

Ok, you can work with @yoavcaspi on it then maybe?
Closing this here for now.

PR #5962, opened 2019-10-15
[WIP] ci: codecov: require_ci_to_pass: yes / notify.after_n_builds

blueyed:
Delays notification (and hopefully status) until CI has passed.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,8 +1,9 @@
+codecov:
+  notify:
+    after_n_builds: 6
 coverage:
   status:
     project: true
     patch: true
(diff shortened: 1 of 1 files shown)


blueyed:
Should be the default already.. not sure if it is a problem with the pytest repo after all..

nicoddemus:
Did you use `[W I P]` here on purpose so CI would finish?

blueyed:
> Did you use [W I P] here on purpose so CI would finish?

More to get a final status (it is a bit annoying that WIP makes it orange).

Issue #4445, opened 2018-11-22
Update/add acceptance tests for "config warnings" stacklevel


nicoddemus:
Follow up to: https://github.com/pytest-dev/pytest/pull/4440#issuecomment-441106932

SaiHarshaK:
Hi, can i work on this issue? i'm a newbie so some help is appreciated.Could you please explain the issue in detail.

blueyed:
@SaiHarshaK 
It is about writing tests for the fix(es) in #4440.

davidfyu:
I would like to try working on this issue. Is there any update as to whether any acceptance tests have already been written for the fix(es) in #4440 at this time? Thanks!

mandarvaze:
@davidfyu  @SaiHarshaK  Either of you working on this ?

xibalba01:
I am also interested in working on this, but i don't really get what the fixes in #4440 are ?! Does it mean the introduction of the `stacklevel=somenumber` in `_issue_config_warning`? I need some clarification and help for starting :)

xibalba01:
@nicoddemus Thanks for the explanation! Are there already some acceptance tests available that i could have a look at to know in which direction this should go?

xibalba01:
@nicoddemus @RonnyPfannschmidt @blueyed  I am still trying to understand how this will work... A bit more details would be really nice :) In the end these specific warnings in mentioned in #4440 need to be triggered, right? But exactly this is where i am struggeling... How to trigger these warnings?

blueyed:
@xibalba01 
They are hopefully covered with tests already (you could check the coverage report on codecov for the master branch to verify this).
Then just put an `assert 0, 1`, `assert 0, 2` etc in front of them (using different numbers for them), and run the whole test suite.  Then you should see what tests are executing the code, and could harden them then.

blueyed:
@xibalba01 
Also take a look at https://github.com/pytest-dev/pytest/pull/5929 maybe, which is only a single one.

xibalba01:
@blueyed Thanks, i will have look!

xibalba01:
@blueyed in the end shall there be entirely new tests that are put in the acceptance_test.py module or shall the existing test that make use of _issue_capture_warning be enhanced to also check for a correct warning message?

blueyed:
@xibalba01 
I'd enhance/harden the existing tests.

blueyed:
@xibalba01 
[...]

PR #5929, opened 2019-10-08
MarkGenerator: Report unknown mark warning at caller level

allanlewis:
Currently, `PytestUnknownMarkWarning` is reported at the level of `MarkGenerator`, which is unhelpful to callers as the warning is shown as being "caused" by pytest. This commit therefore increases the stack level passed to the warning call to 2, which will make the warning log as being "caused" by the caller, which is much more helpful.

- [x] Target the `master` branch for bug fixes, documentation updates and trivial changes.
- [ ] Target the `features` branch for new features, improvements, and removals/deprecations.
- [ ] Include documentation when adding new features.
- [ ] Include new tests or update existing tests when applicable.

Unless your change is trivial or a small documentation fix (e.g.,  a typo or reword of a small section) please:

- [x] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>.<TYPE>.rst`. See [changelog/README.rst](https://github.com/pytest-dev/pytest/blob/master/changelog/README.rst) for details.
- [ ] Add yourself to `AUTHORS` in alphabetical order;

Closes #5928.

PR #5984, opened 2019-10-17
Add acceptance tests for "config warnings" stacklevel (#4445)

phloose:
This pull requests aims to add the acceptance tests mentioned in #4445. The main intention is to check whether a specific warning emitted by `_issue_warning_captured` points to a reasonable location and, depending on the context, not a location deep inside pytest's internals that may confuse users.

### DIFF
--- a/changelog/5928.bugfix.rst
+++ b/changelog/5928.bugfix.rst
@@ -1,0 +1,1 @@
+Report ``PytestUnknownMarkWarning`` at the level of the user's code, not ``pytest``'s.
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -562,7 +562,7 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
 
 
 @hookspec(historic=True)
-def pytest_warning_captured(warning_message, when, item):
+def pytest_warning_captured(warning_message, when, item, location):
     """
     Process a warning captured by the internal pytest warnings plugin.
 
@@ -582,6 +582,10 @@ def pytest_warning_captured(warning_message, when, item):
         in a future release.
 
         The item being executed if ``when`` is ``"runtest"``, otherwise ``None``.
+
+    :param tuple location:
+        Holds information about the execution context of the captured warning (filename, linenumber, function).
+        ``function`` evaluates to <module> when the execution context is at the module level.
     """
 
 
--- a/src/_pytest/warnings.py
+++ b/src/_pytest/warnings.py
@@ -149,6 +149,10 @@ def _issue_warning_captured(warning, hook, stacklevel):
         warnings.warn(warning, stacklevel=stacklevel)
     # Mypy can't infer that record=True means records is not None; help it.
     assert records is not None
+    frame = sys._getframe(stacklevel - 1)
+    location = frame.f_code.co_filename, frame.f_lineno, frame.f_code.co_name
     hook.pytest_warning_captured.call_historic(
-        kwargs=dict(warning_message=records[0], when="config", item=None)
+        kwargs=dict(
+            warning_message=records[0], when="config", item=None, location=location
+        )
     )
--- a/testing/test_warnings.py
+++ b/testing/test_warnings.py
@@ -1,3 +1,4 @@
+import os
 import warnings
 
 import pytest
@@ -641,3 +642,160 @@ def test_pytest_configure_warning(testdir, recwarn):
     assert "INTERNALERROR" not in result.stderr.str()
     warning = recwarn.pop()
     assert str(warning.message) == "from pytest_configure"
+
+
+class TestStackLevel:
+    @pytest.fixture
+    def capwarn(self, testdir):
+        class CapturedWarnings:
+            captured = []
+
+            @classmethod
+            def pytest_warning_captured(cls, warning_message, when, item, location):
+                cls.captured.append((warning_message, location))
+
+        testdir.plugins = [CapturedWarnings()]
+
+        return CapturedWarnings
+
+    def test_issue4445_rewrite(self, testdir, capwarn):
+        """#4445: Make sure the warning points to a reasonable location
+        See origin of _issue_warning_captured at: _pytest.assertion.rewrite.py:241
+        """
+        testdir.makepyfile(some_mod="")
+        conftest = testdir.makeconftest(
+            """
+                import some_mod
+                import pytest
+
+                pytest.register_assert_rewrite("some_mod")
+            """
+        )
+        testdir.parseconfig()
+
+        # with stacklevel=5 the warning originates from register_assert_rewrite
+        # function in the created conftest.py
+        assert len(capwarn.captured) == 1
+        warning, location = capwarn.captured.pop()
+        file, lineno, func = location
+
+        assert "Module already imported" in str(warning.message)
+        assert file == str(conftest)
+        assert func == "<module>"  # the above conftest.py
+        assert lineno == 4
+
+    def test_issue4445_preparse(self, testdir, capwarn):
+        """#4445: Make sure the warning points to a reasonable location
+        See origin of _issue_warning_captured at: _pytest.config.__init__.py:910
+        """
+        testdir.makeconftest(
+            """
+            import nothing
+            """
+        )
+        testdir.parseconfig("--help")
+
+        # with stacklevel=2 the warning should originate from config._preparse and is
+        # thrown by an errorneous conftest.py
+        assert len(capwarn.captured) == 1
+        warning, location = capwarn.captured.pop()
+        file, _, func = location
+
+        assert "could not load initial conftests" in str(warning.message)
+        assert "config{sep}__init__.py".format(sep=os.sep) in file
+        assert func == "_preparse"
+
+    def test_issue4445_import_plugin(self, testdir, capwarn):
+        """#4445: Make sure the warning points to a reasonable location
+        See origin of _issue_warning_captured at: _pytest.config.__init__.py:585
+        """
+        testdir.makepyfile(
+            some_plugin="""
+            import pytest
+            pytest.skip("thing", allow_module_level=True)
+            """
+        )
+        testdir.syspathinsert()
+        testdir.parseconfig("-p", "some_plugin")
+
+        # with stacklevel=2 the warning should originate from
+        # config.PytestPluginManager.import_plugin is thrown by a skipped plugin
+
+        # During config parsing the the pluginargs are checked in a while loop
+        # that as a result of the argument count runs import_plugin twice, hence
+        # two identical warnings are captured (is this intentional?).
+        assert len(capwarn.captured) == 2
+        warning, location = capwarn.captured.pop()
+        file, _, func = location
+
+        assert "skipped plugin 'some_plugin': thing" in str(warning.message)
+        assert "config{sep}__init__.py".format(sep=os.sep) in file
+        assert func == "import_plugin"
+
+    def test_issue4445_resultlog(self, testdir, capwarn):
+        """#4445: Make sure the warning points to a reasonable location
+        See origin of _issue_warning_captured at: _pytest.resultlog.py:35
+        """
+        testdir.makepyfile(
+            """
+            def test_dummy():
+                pass
+        """
+        )
+        # Use parseconfigure() because the warning in resultlog.py is triggered in
+        # the pytest_configure hook
+        testdir.parseconfigure(
+            "--result-log={dir}".format(dir=testdir.tmpdir.join("result.log"))
+        )
+
+        # with stacklevel=2 the warning originates from resultlog.pytest_configure
+        # and is thrown when --result-log is used
+        warning, location = capwarn.captured.pop()
+        file, _, func = location
+
+        assert "--result-log is deprecated" in str(warning.message)
+        assert "resultlog.py" in file
+        assert func == "pytest_configure"
+
+    def test_issue4445_cacheprovider_set(self, testdir, capwarn):
+        """#4445: Make sure the warning points to a reasonable location
+        See origin of _issue_warning_captured at: _pytest.cacheprovider.py:59
+        """
+        testdir.tmpdir.join(".pytest_cache").write("something wrong")
+        testdir.runpytest(plugins=[capwarn()])
+
+        # with stacklevel=3 the warning originates from one stacklevel above
+        # _issue_warning_captured in cacheprovider.Cache.set and is thrown
+        # when there are errors during cache folder creation
+
+        # set is called twice (in module stepwise and in cacheprovider) so emits
+        # two warnings when there are errors during cache folder creation. (is this intentional?)
+        assert len(capwarn.captured) == 2
+        warning, location = capwarn.captured.pop()
+        file, lineno, func = location
+
+        assert "could not create cache path" in str(warning.message)
+        assert "cacheprovider.py" in file
+        assert func == "set"
+
+    def test_issue4445_issue5928_mark_generator(self, testdir):
+        """#4445 and #5928: Make sure the warning from an unknown mark points to
+        the test file where this mark is used.
+        """
+        testfile = testdir.makepyfile(
+            """
+            import pytest
+
+            @pytest.mark.unknown
+            def test_it():
+                pass
+            """
+        )
+        result = testdir.runpytest_subprocess()
+        # with stacklevel=2 the warning should originate from the above created test file
+        result.stdout.fnmatch_lines_random(
+            [
+                "*{testfile}:3*".format(testfile=str(testfile)),
+                "*Unknown pytest.mark.unknown*",
+            ]
+        )
--- a/AUTHORS
+++ b/AUTHORS
@@ -207,6 +207,7 @@ Oscar Benjamin
 Patrick Hayes
 Paweł Adamczak
 Pedro Algarvio
+Philipp Loose
 Pieter Mulder
 Piotr Banaszkiewicz
 Pulkit Goyal
--- a/changelog/4445.bugfix.rst
+++ b/changelog/4445.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed some warning reports produced by pytest to point to the correct location of the warning in the user's code.
--- a/changelog/5984.improvement.rst
+++ b/changelog/5984.improvement.rst
@@ -1,0 +1,1 @@
+The ``pytest_warning_captured`` hook now receives a ``location`` parameter with the code location that generated the warning.
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -133,13 +133,7 @@ def directory_arg(path, optname):
 
 
 # Plugins that cannot be disabled via "-p no:X" currently.
-essential_plugins = (
-    "mark",
-    "main",
-    "runner",
-    "fixtures",
-    "helpconfig",  # Provides -p.
-)
+essential_plugins = ("mark", "main", "runner", "fixtures", "helpconfig")  # Provides -p.
 
 default_plugins = essential_plugins + (
     "python",
@@ -589,7 +583,7 @@ class PytestPluginManager(PluginManager):
             _issue_warning_captured(
                 PytestConfigWarning("skipped plugin {!r}: {}".format(modname, e.msg)),
                 self.hook,
-                stacklevel=1,
+                stacklevel=2,
             )
         else:
             mod = sys.modules[importspec]
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -347,6 +347,7 @@ class MarkGenerator:
                     "custom marks to avoid this warning - for details, see "
                     "https://docs.pytest.org/en/latest/mark.html" % name,
                     PytestUnknownMarkWarning,
+                    2,
                 )
 
         return MarkDecorator(Mark(name, (), {}))


testing/acceptance_test.py
```diff
@@ -1238,3 +1238,24 @@ def test_3():
...
+
+            pytest.register_assert_rewrite("some_mod")
+        """
+    )
+    result = testdir.runpytest()
+    # with stacklevel=5 the warning originates from register_assert_rewrite
+    # function in the created conftest.py
+    result.stdout.fnmatch_lines(
+        ["*conftest.py:4*", "*Module already imported*", "*some_mod*"]
+    )
```
blueyed: I'm not sure `acceptance_test` is the best place.
From what I know it is meant as a top-level test module, and this appears to be something more detaled.
What about `test_warnings.py`?
(just a minor nit, not clear to me either)

xibalba01: Yeah i am also not really sure. I just thought about putting it in there, because the issue #4445 mentioned acceptance tests.... But as your suggestion sounds reasonable i will put it in `test_warnings.py` then because the other tests are already written and also living in the `acceptance_test.py` but are not committed yet.

xibalba01:
i dont get why the coverage is reduced with this commit: cf8e261 ?!?

blueyed:
> i dont get why the coverage is reduced with this commit: [cf8e261](https://github.com/pytest-dev/pytest/commit/cf8e26116e4c47384b9686168c13b1889f06f0a0) ?!?

CI did not finish - does it say it dropped?  (for me it says that CI failed)

testing/acceptance_test.py
RonnyPfannschmidt: it should go to the assertion rewriting tests

we need a better name for the historic acceptance tests file , but thats a followup

nicoddemus: Well those are testing warnings, so perhaps we can have a `TestStackLevel` class in `test_warnings` to group all those?

xibalba01:
> 
> 
> > i dont get why the coverage is reduced with this commit: [cf8e261](https://github.com/pytest-dev/pytest/commit/cf8e26116e4c47384b9686168c13b1889f06f0a0) ?!?
> 
> CI did not finish - does it say it dropped? (for me it says that CI failed)

CI failed for all commits prior to ab17a00. There i changed the stacklevel in `config.PytestPluginManager` to 2, so that the warning points to the correct location. One of the new tests was written against this location...

testing/acceptance_test.py
xibalba01: Sounds reasonable, i will prepare that!

xibalba01:
I was questioning myself: At the moment the new tests check for a specific file and **also** a line number. I find this very strict with regard to refactorings which that change the amount of code in a file and thus moving the respective part up or down. Then everytime this happends the tests would fail - also when the refactoring has nothing to do with the stacklevel -  and need to be adjusted to the new position of the warning. I think i should take out the line number... what do you think @nicoddemus @blueyed @RonnyPfannschmidt ?

blueyed:
Yeah, no line numbers IMHO.
(also please do not notify people all the time - the ones interested to help (e.g. me now) are subscribed and will notice)

xibalba01:
> 
> 
> Yeah, no line numbers IMHO.
> (also please do not notify people all the time - the ones interested to help (e.g. me now) are subscribed and will notice)

Alright, sorry :)

xibalba01:
what does codecov/patch mean? why did the last commit fail with that?

blueyed:
It shows that some code is not covered/executed: https://codecov.io/gh/pytest-dev/pytest/compare/813ef9e88f0e6f74a54e803d672cd65f358cabb4...3b654a702fe5b1656553fb590ab537c6b06bdfcb/diff#D2-505
Seems to be ok, since you only moved this I assume?

blueyed:
..but it should not have been moved in the first place.
Please try to keep the diff small to avoid churn with git-blame etc.

xibalba01:
> 
> 
> ..but it should not have been moved in the first place.
> Please try to keep the diff small to avoid churn with git-blame etc.

Yeah i am really wondering why the diff is that large... i only added the new TestStackLevel class at the end and since some of the tests use `os.sep` i added the import os statement at the top... weird

blueyed:
Line endings maybe?

xibalba01:
yes i think so, already investigating...

testing/test_warnings.py
```diff
@@ -678,7 +678,7 @@ def test_issue4445_preparse(self, testdir):
         # thrown by an errorneous conftest.py
         result.stdout.fnmatch_lines_random(
             [
-                "*config{sep}__init__.py:915*".format(sep=os.sep),
+                "*config{sep}__init__.py*".format(sep=os.sep),
```
blueyed: `:*` ?

xibalba01: what exactly do you mean with that? Should the colon be again part of the string to check?

xibalba01: For that specific file (config/__init__.py) is was wondering if checking again against line numbers is maybe not such a bad idea. For more unique files to check for a filename is ok i think, but __init__.py is so generic and has lots of code inside that stacklevel changes that point to another location in that file would be eaten up by this check...

blueyed: Yeah, true.
I think we should find a way to match the function name - maybe through reconfiguring display of warnings, if necessary?

blueyed: > How to do that? Using `inspect.stack` and choose the second `FrameInfo` object of the returned list?

Give it a try I guess.  Otherwise I've thought about changing the format of warnings to include the function name.
Maybe @nicoddemus can help?

xibalba01: I added a new commit that uses inspect.stack

src/_pytest/hookspec.py
```diff
@@ -577,7 +577,7 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
 
 
 @hookspec(historic=True)
-def pytest_warning_captured(warning_message, when, item):
+def pytest_warning_captured(warning_message, when, item, frame):
```
blueyed: Should be added to the doc, and can use typing (`types.FrameType`).

src/_pytest/warnings.py
```diff
@@ -142,9 +142,12 @@ def _issue_warning_captured(warning, hook, stacklevel):
     :param hook: the hook caller
     :param stacklevel: stacklevel forwarded to warnings.warn
     """
+    import inspect
+
     with warnings.catch_warnings(record=True) as records:
         warnings.simplefilter("always", type(warning))
         warnings.warn(warning, stacklevel=stacklevel)
+    frame = inspect.stack()[stacklevel - 1]
```
blueyed: I think it would be ok / faster to use `sys._getframe().f_back` directly.

testing/test_warnings.py
```diff
@@ -657,51 +669,57 @@ def test_issue4445_rewrite(self, testdir):
...
         # with stacklevel=5 the warning originates from register_assert_rewrite
         # function in the created conftest.py
-        result.stdout.fnmatch_lines(
-            ["*conftest.py:4*", "*Module already imported*", "*some_mod*"]
-        )
+        warning, frame = capwarn.captured
+
+        assert "Module already imported" in str(warning.message)
+        assert "conftest.py" in str(warning.filename)
+        assert frame.function == "<module>"  # this is the above created conftest module
```
blueyed: Assert the filename or something else to ensure this?

xibalba01:
Should i edit the changelog also for the tests? In 11d897e i changed the stacklevel to 2 but that was a finding during the test writing and not related to any issue...

blueyed:
Include the changelog from https://github.com/pytest-dev/pytest/pull/5929?
(I would have cherry-picked the commit from the actually, giving credit etc, with hub: `hub am https://github.com/pytest-dev/pytest/pull/5929`)

xibalba01:
ok i will have a look! At the moment i merged simply the branch from @allanlewis into this branch...

xibalba01:
I added a test for the MarkGenerator warning. @allanlewis or were you working on a test for this also?

xibalba01:
@blueyed @nicoddemus i was wondering about the behaviour of the warnings in general. The changes from #5929 changed the stacklevel in a `warnings.warn` function call. I can test it the way i did now, just by matching the returned warning with an expected location and other infos. 

Most of the other stacklevel changes though were all made inside `_issue_warning_captured` that uses the `pytest_warning_captured` hook to collect calls and meta information to `_issue_warning_captured`. This way inside a test case warnings can (mostly) be captured via a plugin that intercepts the warning collection. This makes tests more robust as one can now test against the function name obtained from the stackframe the warning was emitted in, rather than to use the stdout from a test and matching it against e.g. an expected filename or line numbers, that are prone to change. (Thereby i don't understand why i can catch some of the warnings with `pytest_warning_captured` but not all (e.g. [resultlog](https://github.com/xibalba01/pytest/blob/ac0df83685cd1a2db482b50e216907675db4ca53/testing/test_warnings.py#L723) or [cacheprovider](https://github.com/xibalba01/pytest/blob/ac0df83685cd1a2db482b50e216907675db4ca53/testing/test_warnings.py#L740)))

Shouldn't this warning "thing" be a bit more uniform or is there a specific reason why some of the warnings are captured while others are not?

allanlewis:
No, I thought that had been addressed already? Sorry, I haven't looked
through this PR in detail.

phloose:
I would like to know how i can improve the recent changes made by me or if they are accepted in general...

src/_pytest/hookspec.py
```diff
@@ -597,6 +597,10 @@ def pytest_warning_captured(warning_message, when, item):
         in a future release.
 
         The item being executed if ``when`` is ``"runtest"``, otherwise ``None``.
+
+    :param types.FrameType frame:
```
nicoddemus: Hi @phloose, sorry for the delay!

That's definitely an interesting approach! 

`call_historic` here means that this hook will be called even for plugins registered after this call was made, if they implement the hook (see [https://pluggy.readthedocs.io/en/latest/index.html#historic-hooks](https://pluggy.readthedocs.io/en/latest/index.html#historic-hooks)).

Problem is that, to implement that, `call_historic` needs to keep the parameters passed to it around *forever*, which means that all locals on that frame will also be kept around, meaning a huge memory leak. 😕

changelog/5928.bugfix.rst
```diff
@@ -0,0 +1 @@
+Report ``PytestUnknownMarkWarning`` at the level of the user's code, not ``pytest``'s.
```
nicoddemus: This PR is about all warnings, so this should probably be:

```rst
Fixed some warning reports produced by pytest to point to the correct location of the warning in the user's code.
```

src/_pytest/hookspec.py
phloose: Ok sounds reasonable... Didn't have such a deep insight into pluggy and didnt thought of a memory leak regarding the locals from the frame and `call_historic`. I just thought of a way to not only match against strings/testoutput/linenumbers when the exact warning location is important (described here: https://github.com/pytest-dev/pytest/pull/5984#issuecomment-543776819). So should i revert the recent changes regarding the frame thing to maybe 3237e05 ?

changelog/5928.bugfix.rst
phloose: I will adapt this, was merged in from https://github.com/pytest-dev/pytest/pull/5929...

src/_pytest/hookspec.py
nicoddemus: Sounds reasonable, except the last commit is "Add test for MarkGenerator warning location" which I guess should be kept. But I think you can squash all the commits together anyway for the final merge. 👍

phloose: Did an interactive rebase and removed the frame related commits. I am still questioning how to check for the exact location of a warning, if linenumbers of a warning are too strict as outlined in https://github.com/pytest-dev/pytest/pull/5984#issuecomment-543776819...

nicoddemus: Yeah it is not clear to me as well, but having the test failures at least alerts us to the problem, which is better than letting the warning point to an unrelated location after a refactoring. 😕

changelog/5928.bugfix.rst
nicoddemus: Just needs this update in order to merge, IMO.

src/_pytest/hookspec.py
phloose: Hmm ok. I am wondering how for a large module like `_pytest.config.__init__.py` (`config._preparse` and `config.PytestPluginManager.import_plugin`) this can be helpful at all if you only check that the error comes from this module and not in which context, because then the stacklevel changes cannot be precisely checked in the test - it could be anywhere in that module... But yeah since checking against frame is no option for obvious reasons and linenumbers are too strict with respect to refactorings this is maybe the only way until now.

changelog/5928.bugfix.rst
phloose: I will update this when i get back home... But i will create another changelog entry for #4445 so that both are addressed

src/_pytest/hookspec.py
blueyed: Maybe it could pass the frame only when under test / via some config?
Otherwise it might also be not a big problem after all keeping the frame(s) around?  (there should not be many warning after all)

blueyed: Also not handling the frame for historic calls might be an option, i.e. call it (passing a reference), but set it to `None` afterwards.
Also considering if `pytest_warning_captured` really needs to be historic might be worth it (IIRC there some issue about leaks with warnings already, which might be related).

blueyed: Also we could of course not pass the frame as is, but copies of values we can use for matching then (e.g. `co_name`).

phloose: > Also we could of course not pass the frame as is, but copies of values we can use for matching then (e.g. `co_name`).

This sounds feasible to me 🤔 worth rethinking the whole thing ... Unfortunatly i threw away the commits with the frame related stuff. Maybe i am lucky looking in the reflog. Then we would need to find a way to distinguish clearly between names for modules and functions (`co_name` evaluates to the string `<module>` instead of the module's name when the context is a module)

testing/test_warnings.py
```diff
@@ -640,3 +641,117 @@ def pytest_configure():
...
+            def test_it():
+                pass
+            """
+        )
+        result = testdir.runpytest_subprocess()
+        # with stacklevel=2 the warning should originate from the above created test file
+        # referenced by request.node.name
+        result.stdout.fnmatch_lines(
+            [
+                "*{testfile}.py:3*".format(testfile=request.node.name),
```
blueyed: You could just use `str(p1)` here with `p1 = testdir.makepyfile…` above (I guess).

blueyed: The test should be more specific (hardened) though, checking for the warning header above, and could match the whole line etc.
Like it is now it would still pass if theres an INTERNALERROR where the file is mentioned.
Also `runpytest_subprocess` should not be necessary, is it?

blueyed: Also matching the "*= 1 passed, 1 warning in *" line is recommended (ensures also the exitcode).

src/_pytest/hookspec.py
blueyed: @phloose 
Yeah, another case of having to force-push here before merging.. :/
But Github keeps track of this, see the comments:

> phloose force-pushed the phloose:acceptance-tests_pytest-dev/pytest/#4445  branch from 591328c to c489af3 10 days ago
> phloose force-pushed the phloose:acceptance-tests_pytest-dev/pytest/#4445  branch from ac0df83 to 46663af 18 hours ago

So you can just `git reset --hard …` with it (create a backup if you're not experienced with resetting.. ;))

I wonder if you could just configure the warnings format with the (hopefully single) outer run to include more/necessary information then in the output?

testing/test_warnings.py
phloose: > You could just use `str(p1)` here with `p1 = testdir.makepyfile…` above (I guess).

I will have a look at this, as this surely can be simpler

>The test should be more specific (hardened) though, checking for the warning header above, > and could match the whole line etc.

I agree totally that they should be more strict, but also thought the most essential parts are the locations, because of the changed stacklevel. And for determining the exact location we are limited now, because of the implications regarding the usage of frame. For me i doesn't make sense to match against more text. I will think about it.

> Also `runpytest_subprocess` should not be necessary, is it?

There was a reason for using `runpytest_subprocess`: Using only `runpytest` resulted in no warning at all for that specific test.

> Also matching the "*= 1 passed, 1 warning in *" line is recommended (ensures also the exitcode). 

Good point!

phloose:
> I do not like that there are many different tests for this, especially with `runpytest_subprocess`, which is slow.
> It would be great if all those could be done in a single test.
> While it is usually good to have separate tests, those go all together, and are kind of not expected to change/fail, except for refactorings, and then it is ok to not have 5 failures, instead of just one that you have to iterate over.

For me it seemed best to have a single test case for every warning that was affected by the stacklevel change, especially as they are triggered somewhat different. I think this way it is more clean and more readable. Also they are grouped in a test class called `TestStackLevel`

phloose:
I came up with some thoughts about the whole stacklevel thing: 

The main concern is that a warning points to a correct location, that is related to an action done by the user (typos, errors, misconfiguration etc.) and in turn does not point to a location inside pytests internals that is confusing and not helpful for the user.

What i've noticed during my work on the test cases is that there is no uniform mechanism to catch the warning information (e.g. warnings in `resultlog` and `cacheprovider` cannot be catched by the `pytest_warning_captured` hook... do they use a different hook object?). Furthermore there are warnings that are raised directly from the users code in created modules (`testdir.makepyfile`) and warnings that originate from pytests internals (like `_pytest.config.__init__.py`).  While the former is easy to test with exact line numbers to be strict and sure that the warning does not point to a deeper origin, the latter is hard to test because, as already outlined above, being strict here with line numbers may produce unrelated test failures during refactorings. 

In general if find there should be a more uniform mechanism how warnings in general can be catched (be it during config, collection, call etc.) and tracked to be sure they point to the correct origin.

Regarding @blueyed's suggestion to only use stackframe information and not the whole frame object i came up with a new idea:

```python
def _issue_warning_captured(warning, hook, stacklevel):
    with warnings.catch_warnings(record=True) as records:
        warnings.simplefilter("always", type(warning))
        warnings.warn(warning, stacklevel=stacklevel)
    # get information about the current execution context
    frame = sys._getframe(stacklevel - 1)
    context = frame.f_code.co_filename, frame.f_code.co_name, frame.f_lineno
    hook.pytest_warning_captured.call_historic(
        kwargs=dict(
            warning_message=records[0], when="config", item=None, context=context
        )
    )
```

This way you get information about the module, a possible function (co_name evaluates to `<module>` when the current frame is not a function) and the current frames line number. This would work for all stacklevel changes affected by #4445 except the ones in `resultlog` and `cacheprovider`

At the moment i am not happy with the tests at all, because they CANNOT be strict for the correct locations (no linenumbers, no stackinformation) and may also not fail when there are new stacklevel changes... Therefore i would also call them useless until there is a mechanism that catches the execution context...

blueyed:
Yeah, I think having `context` being provided via the hook makes sense in general (what you've done is what I had in mind).  It's also useful outside of pytest's own tests, and might make sense to stick on the `WarningMsg` then even (so that tests could use it in general).

phloose:
friendly reminder for a review @nicoddemus ... I just saw there are conflicts (because of the new release?) should i resolve them?

nicoddemus:
Hi @phloose, sorry about the delay, and thanks for the ping!

I will take another look later (the context seems like a good solution), meanwhile if you want to fix the conflicts it would be nice. 👍

phloose:
> meanwhile if you want to fix the conflicts it would be nice.

Done!

src/_pytest/hookspec.py
```diff
@@ -582,6 +582,10 @@ def pytest_warning_captured(warning_message, when, item):
         in a future release.
 
         The item being executed if ``when`` is ``"runtest"``, otherwise ``None``.
+
+    :param tuple context:
```
nicoddemus: This is nice, but pytest kind already has this concept and it is called `location`, for example `pytest_runtest_logstart` receives a `location` parameter:

```python
def pytest_runtest_logstart(nodeid, location):
    """ signal the start of running a single test item.

    This hook will be called **before** :func:`pytest_runtest_setup`, :func:`pytest_runtest_call` and
    :func:`pytest_runtest_teardown` hooks.

    :param str nodeid: full id of the item
    :param location: a triple of ``(filename, linenum, testname)``
    """
```

I suggest we follow the same approach here, except you might call the last parameter `name` and use your description. 👍

testing/test_warnings.py
```diff
@@ -641,3 +642,151 @@ def pytest_configure():
     assert "INTERNALERROR" not in result.stderr.str()
     warning = recwarn.pop()
     assert str(warning.message) == "from pytest_configure"
+
+
+class TestStackLevel:
+    @pytest.fixture
+    def capwarn(self):
```
nicoddemus: This is a good approach, but I think we can improve it a bit:

1. Instead of only holding the last warning, it would be better to accumulate them on a list. This way we ensure the tests will fail with better messages in case more than one warning is raised in the future.

2. Following 1, you can return the list object directly, and automatically register it in `testdir`:

   ```python
       
       @pytest.fixture
       def capwarn(self, testdir):
           class CapturedWarnings:
               captured = []
   
               @classmethod
               def pytest_warning_captured(cls, warning_message, when, item, context):
                   cls.captured.append((warning_message, context))
   
           testdir.plugins = [CapturedWarnings()]
   
           return CapturedWarnings.captured
  ```

testing/test_warnings.py
```diff
@@ -641,3 +642,151 @@ def pytest_configure():
...
+                import pytest
+
+                pytest.register_assert_rewrite("some_mod")
+            """
+        )
+        testdir.parseconfig()
+
+        # with stacklevel=5 the warning originates from register_assert_rewrite
+        # function in the created conftest.py
+        warning, context = capwarn.captured
```
nicoddemus: Following the suggestion of `capwarn` returning a list, this would then be:

```python
assert len(capwarn) == 1
warning, context = capwarn[0]
```

testing/test_warnings.py
phloose: To accumulate them makes sense to me but i am concerned about setting testdir.plugins in that fixture. During writing the tests i discovered that it is (strangely) not uniform to apply a plugin to pytests internal tests. While it works most of the time it **does not** work the above way in the [cacheprovider test](https://github.com/phloose/pytest/blob/acceptance-tests_pytest-dev/pytest/%234445/testing/test_warnings.py#L760) (if i am missing something please point me in the right direction). Therefore i would propose to still use the class as return value, so that you can do both ways.

testing/test_warnings.py
phloose: I find this another good step in making the tests stronger! 

However i found out strange things while enhancing the tests with your suggestions regarding using `len(capwarn)`. Depending on the order of calls it can happen, that some of the warning emitting parts are called twice (like cacheprovider.set and import_plugin) so that two identical warnings are captured. I only know few to nothing about the respective parts that account for the call order but for me it is worth to mention that. I will dig deeper with the debugger and enhance the respective tests with comments about my findings. 

But maybe this has to do with how pytester works and i am lacking necessary understanding of that procedure ;)

src/_pytest/hookspec.py
phloose: Ok makes sense to stay with that naming. I will change context to location. At the moment `context` consists of `(filename, function, linenumber)`. I would prefer to not name `function` to `name` as it can be confusing. I can change the order in the way of the above example!

phloose:
While refactoring the changes i stumbled over another occurence of `_issue_warning_captured` in `junitxml.py`. This should also be tested right? Or will this be removed once pytest 6.0 is released?

src/_pytest/hookspec.py
phloose: Done!

testing/test_warnings.py
phloose: Done!

phloose:
@nicoddemus i changed the code according to your suggestions and will squash all commits after your approval!

nicoddemus:
> While refactoring the changes i stumbled over another occurence of _issue_warning_captured in junitxml.py . This should also be tested right? Or will this be removed once pytest 6.0 is released?

There's a test already here:

https://github.com/pytest-dev/pytest/blob/master/testing/deprecated_test.py#L50

But I don't think we need to harden it though. 👍

nicoddemus:
Also please fix linting. 👍 

(You can run `pre-commit install` locally to install the pre-commit hooks which automatically fix linting issues).

nicoddemus:
Thanks @phloose!

Right before merging I realized that by adding the 'location' parameter we need to target `features` for this. I took the liberty of adding an extra changelog for that and rebasing on `features` myself, hope that's ok.

phloose:
> Right before merging I realized that by adding the 'location' parameter we need to target `features` for this. I took the liberty of adding an extra changelog for that and rebasing on `features` myself, hope that's ok.

Of course! No problem!

nicoddemus:
Thanks again @phloose for the patience on getting this through! 👍

blueyed:
Thanks indeed, @phloose!

phloose:
Thanks to you guys for your help!

PR #5990, opened 2019-10-18
Plurality matching

### LINKED ISSUE #5989: Plurality mismatch in warnings

Admittedly this is a very minor issue.

If I write a file `pytest_example.py` as

```
import warnings

warnings.warn("one warning")
```
and then run
```
pytest pytest_example.py
```
from the command line, the output reads
> 1 warnings in 0.02s

I believe this should read "1 warning in 0.02s".

Would a PR to address this be welcome?

<details>

```
$ pip list
Package            Version
------------------ -------
atomicwrites       1.3.0  
attrs              19.3.0 
importlib-metadata 0.23   
more-itertools     7.2.0  
packaging          19.2   
pip                19.0.3 
pluggy             0.13.0 
py                 1.8.0  
pyparsing          2.4.2  
pytest             5.2.1  
setuptools         40.8.0 
six                1.12.0 
wcwidth            0.1.7  
zipp               0.6.0  

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.3 LTS
Release:        18.04
Codename:       bionic
```

</details>

MarcoGorelli:
Fixes #5989 .

I presume this is small enough that I shouldn't add it to the changelog (or myself to the authors), but please do let me know if I should.

### DIFF
--- a/changelog/5990.improvement.rst
+++ b/changelog/5990.improvement.rst
@@ -1,0 +1,1 @@
+Fix plurality mismatch in test summary (e.g. display "1 error" instead of "1 errors").
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -539,7 +539,7 @@ class TerminalReporter:
             str(self._numcollected) + " item" + ("" if self._numcollected == 1 else "s")
         )
         if errors:
-            line += " / %d errors" % errors
+            line += " / %d error%s" % (errors, "s" if errors != 1 else "")
         if deselected:
             line += " / %d deselected" % deselected
         if skipped:
@@ -1056,6 +1056,19 @@ _color_for_type = {
 _color_for_type_default = "yellow"
 
 
+def _make_plural(count, noun):
+    # No need to pluralize words such as `failed` or `passed`.
+    if noun not in ["error", "warnings"]:
+        return count, noun
+
+    # The `warnings` key is plural. To avoid API breakage, we keep it that way but
+    # set it to singular here so we can determine plurality in the same way as we do
+    # for `error`.
+    noun = noun.replace("warnings", "warning")
+
+    return count, noun + "s" if count != 1 else noun
+
+
 def build_summary_stats_line(stats):
     known_types = (
         "failed passed skipped deselected xfailed xpassed warnings error".split()
@@ -1086,7 +1099,7 @@ def build_summary_stats_line(stats):
             )
             color = _color_for_type.get(key, _color_for_type_default)
             markup = {color: True, "bold": color == main_color}
-            parts.append(("%d %s" % (count, key), markup))
+            parts.append(("%d %s" % _make_plural(count, key), markup))
 
     if not parts:
         parts = [("no tests ran", {_color_for_type_default: True})]
--- a/AUTHORS
+++ b/AUTHORS
@@ -160,6 +160,7 @@ Manuel Krebber
 Marc Schlaich
 Marcelo Duarte Trevisani
 Marcin Bachry
+Marco Gorelli
 Mark Abramowitz
 Markus Unterwaditzer
 Martijn Faassen
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -248,7 +248,10 @@ def pytest_collection(session):
 
 def pytest_runtestloop(session):
     if session.testsfailed and not session.config.option.continue_on_collection_errors:
-        raise session.Interrupted("%d errors during collection" % session.testsfailed)
+        raise session.Interrupted(
+            "%d error%s during collection"
+            % (session.testsfailed, "s" if session.testsfailed != 1 else "")
+        )
 
     if session.config.option.collectonly:
         return True
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -628,7 +628,7 @@ class TestInvocationVariants:
         result = testdir.runpytest("--pyargs", "tpkg.test_hello", syspathinsert=True)
         assert result.ret != 0
 
-        result.stdout.fnmatch_lines(["collected*0*items*/*1*errors"])
+        result.stdout.fnmatch_lines(["collected*0*items*/*1*error"])
 
     def test_pyargs_only_imported_once(self, testdir):
         pkg = testdir.mkpydir("foo")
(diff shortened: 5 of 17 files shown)


The-Compiler:
Nice work! IMHO this is big enough for a changelog/author entry! However, the tests are currently failing - it seems like there are other places which you adjusted in the tests (e.g. "1 errors during collection" -> "1 error during collection") but not in the code. Could you take a look at the failing tests?

MarcoGorelli:
Thanks! Yes, just working on it, I'll ping you when it's ready.

src/_pytest/terminal.py
```diff
@@ -1007,6 +1007,14 @@ def _folded_skips(skipped):
     return values
 
 
+def _match_plurality(count, key):
```
nicoddemus: I think it makes more sense to always pass the key as singular, and add `s` as necessary:

```python
def _make_plural(noun, count):
   return noun + "s" if count != 1 else noun
```

This of course requires a change in the key themselves, but I believe this to be more elegant. What do you think?

changelog/5990.improvement.rst
```diff
@@ -0,0 +1 @@
+Fix plurality mismatch in test summary (e.g. display "2 errors" instead of "2 error").
```
nicoddemus: This is actually the other way around:

```
Fix plurality mismatch in test summary (e.g. display "1 error" instead of "1 errors").
```

MarcoGorelli: Good catch, thanks!

src/_pytest/terminal.py
MarcoGorelli: Yes, `_make_plural` is an easier-to-understand name, thanks. I've kept the guard in though as we wouldn't want to add an 's' to the end of 'skipped', for example.

The-Compiler: > This of course requires a change in the key themselves, but I believe this to be more elegant. What do you think?

How backwards-compatible would that be (warnings -> warning) for e.g. plugins?

kalekundert:
I may be a little late to the game here, but I wanted to share a snippet of code I use to get plurality right in my scripts.  I really like how succinct and flexible it is:

```python
class plural:
    """
    The format string has three sections, separated by '/'.  The first section 
    is always printed, the second (optional) section is printed if the number 
    is singular, and the third section is printed if the number is plural.  Any 
    '?' in the format string are replaced with the actual number.

    >>> "I have {:? thing/s}.".format(plural(1))
    I have 1 thing.
    >>> "I have {:? thing/s}.".format(plural(2))
    I have 2 things.

    >>> "I want {:/a cactus/? cacti}.".format(plural(1))
    I want a cactus.
    >>> "I want {:/a cactus/? cacti}.".format(plural(2))
    I want 2 cacti.

    If the provided object supports len() (e.g. lists, dicts, etc.), then that 
    value is automatically used as the number:

    >>> "I have {:? thing/s}.".format(plural([3]))
    I have 1 thing.
    >>> "I have {:? thing/s}.".format(plural([3,4]))
    I have 2 things.

    From Veedrac on Stack Overflow (with some modifications): 
    http://stackoverflow.com/questions/21872366/plural-string-formatting
    """
    def __init__(self, value):
        self.value = value

    def __format__(self, formatter):
        from collections.abc import Sized

        x = self.value
        number = len(x) if isinstance(x, Sized) else self.value
        formatter = formatter.replace("?", str(number))
        always, _, suffixes = formatter.partition("/")
        singular, _, plural = suffixes.rpartition("/")

        return "{}{}".format(always, singular if number == 1 else plural)
```
I have this snippet on PyPI (`nonstdlib.plural`), but if you want to use it in `pytest`, I think it would be better to avoid a dependency for something so trivial.

src/_pytest/terminal.py
nicoddemus: It might cause minor breakages for plugins testing things with `testdir`, depending on the test, good point.

To be on the safe side we should target `features` instead and get this out in `5.3`. @MarcoGorelli could you rebase on `features` and change the target branch of the PR? Thanks!

blueyed: > changed the base branch from master  to features

@marcogorelli
You also need to rebase it on features then (and force-push).

blueyed: Do not add a commit that removes the changes (https://github.com/pytest-dev/pytest/pull/5990/commits/31af9692127612786adcec01f1379be3cca16e35), but rebase it so that only your commit(s) are there.

MarcoGorelli: Ok, I think I've got it now (sorry, bit of a git-rebase noob here), thanks to you all for your time!

src/_pytest/terminal.py
```diff
@@ -1007,9 +1007,15 @@ def _folded_skips(skipped):
...
+def _make_plural(count, key):
+    if key in ["error", "warning"] and count != 1:
+        key += "s"
+    return count, key
+
+
 def build_summary_stats_line(stats):
     known_types = (
-        "failed passed skipped deselected xfailed xpassed warnings error".split()
+        "failed passed skipped deselected xfailed xpassed warning error".split()
```
blueyed: That's an unnecessary API breakage, isn't it?  I.e. plugins might use `stats.get("warnings")`.
I think we should handle/set both then still, but can use "warning" only internally then.

src/_pytest/terminal.py
blueyed: Better.. but the commit message needs adjustments.. ;)

MarcoGorelli:
Hi all - just FYI, I'm away until the 26th of October, I'll finish this after that. Thanks for your patience in helping me get it right!

MarcoGorelli:
I've updated the PR so it's a single commit which targets the `features` branch.

I've kept the API the same (so `stats.get('warnings')` continues to work), only changing it to `warning` within the scope of `_make_plural`.

blueyed:
Went ahead with merging when noticing the issue with #6094 .. :)
Thanks!

aragilar:
This appears to break the testdir fixture, as the fixture assumes that the outcomes will be one of (with no variations) passed, skipped, failed, error, xpassed or xfailed. Can we revert this, as the errors I'm getting are really unhelpful, and make this a pain to debug?

PR #5992, opened 2019-10-18
tests: debugging: disable pdb++ within inner tests

blueyed:
Ref: https://github.com/pytest-dev/pytest/pull/5306#issuecomment-495690643


blueyed:
@nicoddemus 
I assume you improve it implicitly via https://github.com/pytest-dev/pytest/pull/5306#issuecomment-496032660 ?

PR #5996, opened 2019-10-18
unittest: do not use TestCase.debug() with `--pdb`

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5991.

/cc @mbyt 

TODO:

- [x] changelog
- [x] doc update (https://github.com/pytest-dev/pytest/pull/1890/files#diff-c4141365b63ea92331fc88e5f7cc7abf)
- [x] rebase on features: while being a bugfix, it certainly changes behavior (e.g. with pytest-django)


blueyed:
Ready.

blueyed:
> Are you positive this is the right approach?

Quite.
It's not the first time I've came across https://github.com/pytest-dev/pytest/pull/1890 as being problematic, and IIRC @mbyt never reacted to me pinging them.
pytest-django had to work around that also then (monkeypatching `.debug` itself to make it work with Django), while `.debug()` should never have been used in the first place.
AFAICS it is meant to debug your tests manually, and not to be used in general (what #1890) does.

The situation in pytest might have improved since #1890, but the test still works, and I've hardened it.

> the original change was meant for people to be able to debug teardown methods in unittest.TestCase subclasses.

From #1890:

> These tests cannot be easily post mortem debugged (pytest --pdb), as unittests tearDown and cleanup (which close the GUI) are called on an exception before the post mortem debugger jumps into place. Therefore, the GUI is closed and no direct interaction is possible any more.

This is asserted explicitly in https://github.com/pytest-dev/pytest/blob/c7eeb05b598a997e3709e66afecf32b2f23526d9/testing/test_pdb.py#L176

nicoddemus:
Thanks for the explanation! Let's go with your take then, you are the team's debugger expert. 😁 👍

blueyed:
I've noticed that this is actually wrong (and I've hidden the regression by weakening the test actually).
Will look into this later - I think it is finally necessary to clean up / improve the unittest integration, so that e.g. `--trace` also works.

> you are the team's debugger expert.

Thanks! :) - but this needs a unittest expert.. ;)

blueyed:
=> 6014 for now.

blueyed:
Turns out that this is quite ready/good already, but that `test_trial_error` changes - I cannot tell if it is for good or better.

blueyed:
pypy3 stops in functools with `--trace`:
```
test_trace.py 
>>>>>>>>>>>>>>>>>>>> PDB runcall (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>
> /home/travis/build/pytest-dev/pytest/.tox/pypy3-xdist/lib_pypy/_functools.py(78)__call__()
-> if self._keywords:
(Pdb) F                                                          [100%]
```

blueyed:
I've also tried invoking the pdbinvoke plugin directly from within the wrapped function (simpler in general), but it missing properties that come via `from_call` etc normally.
This would basically make it more behave like "just put your pdb.set_trace" where you need it.
So given that it might be better to have a more specific method that could be used here then (from within the wrapped call), when `--pdb` is used - or give using the existing `post_mortem` a try.

nicoddemus:
Thanks @blueyed, I'm convinced that we should *not* use `TestCase.debug()` at all.

I think we should pdbinvoke directly in the `addError` and `addFailure` if `--pdb` is passed in the command line. I will try to make this in a separate PR, if I'm successful, I believe this will supersed this one and #6014.

nicoddemus:
> I think we should pdbinvoke directly in the addError and addFailure if --pdb is passed in the command line.

Just tried that, but unfortunately `unittest` calls `tearDown` before reporting errors and failures to the underlying "result" object (comments are mine):

```python
            if outcome.success:
                outcome.expecting_failure = expecting_failure
                with outcome.testPartExecutor(self, isTest=True):
                    self._callTestMethod(testMethod)  # <-- call test method
                outcome.expecting_failure = False
                with outcome.testPartExecutor(self):
                    self._callTearDown()  # <-- call teardown

            self.doCleanups()
            for test, reason in outcome.skipped:
                self._addSkip(result, test, reason)
            self._feedErrorsToResult(result, outcome.errors)  # <-- call addError, addFailure, etc.
```

So `unittest` doesn't really gives us a chance to "hook" into the result of the test method call for pytest to jump into the debugger, before `tearDown` is called.

You approach (wrapping the actual method call) seems like the only one that can actually work, thanks for looking at it into such detail.

src/_pytest/unittest.py
```diff
@@ -187,29 +191,45 @@ def addSuccess(self, testcase):
...
+                )
+                expecting_failure = expecting_failure_class or expecting_failure_method
+                self._need_tearDown = True
+
+                if expecting_failure:
+                    raise
+
+                raise _GetOutOf_testPartExecutor(exc)
+
+        self._testcase._wrapped_testMethod = wrapped_testMethod
```
nicoddemus: I think we should override the original method, otherwise the internal name will show up to the user when entering PDB:

```
self = <foo.T testMethod=_wrapped_testMethod>

    def test_foo(self):
        self.filename = 'hello'
>       assert 0
```

src/_pytest/unittest.py
```diff
@@ -187,29 +191,45 @@ def addSuccess(self, testcase):
...
-            self._testcase._addSkip(self, self._testcase, skip_why)
-            return True
-        return False
 
-    def runtest(self):
-        if self.config.pluginmanager.get_plugin("pdbinvoke") is None:
+        class _GetOutOf_testPartExecutor(KeyboardInterrupt):
+            """Helper exception to get out of unittests's testPartExecutor."""
+
+        unittest = sys.modules.get("unittest")
```
nicoddemus: At this point `unittest` has definitely been imported, no need to guard against that. 👍

src/_pytest/unittest.py
```diff
@@ -115,6 +117,8 @@ def setup(self):
             self._request._fillfixtures()
 
     def teardown(self):
+        if self._need_tearDown:
```
nicoddemus: Are you sure we need this? Even if I comment these two lines, tear down still gets called for this example:

```python
from unittest import TestCase, expectedFailure

class T(TestCase):

    def tearDown(self):
        print('*********** tearDown')
        self.filename = None

    @expectedFailure
    def test_foo(self):
        self.filename = 'hello'
        assert 0

```

```
λ pytest  .tmp\foo.py -sq
*********** tearDown
x
1 xfailed in 0.05s
```

Uncommenting them actually shows tearDown being called twice:

```
 λ pytest  .tmp\foo.py -sq
*********** tearDown
x*********** tearDown

```

doc/en/unittest.rst
```diff
@@ -238,17 +238,6 @@ was executed ahead of the ``test_method``.
 
 .. _pdb-unittest-note:
 
-.. note::
-
-    Running tests from ``unittest.TestCase`` subclasses with ``--pdb`` will
```
nicoddemus: 👍

src/_pytest/unittest.py
nicoddemus: Nevermind, we actually need it, but only if a test is not an expected failure

src/_pytest/unittest.py
```diff
@@ -217,6 +220,7 @@ def wrapped_testMethod(*args, **kwargs):
                 expecting_failure = self._expecting_failure(testMethod)
                 if expecting_failure:
                     raise
+                self._needs_explicit_tearDown = True
                 raise _GetOutOf_testPartExecutor(exc)
 
         setattr(self._testcase, self._testcase._testMethodName, wrapped_testMethod)
```
blueyed: Does it pass https://github.com/pytest-dev/pytest/pull/5996#discussion_r345282784 then?

blueyed:
Thanks for looking into this!
I think it is worth revisiting completely though.
After all using `post_mortem(sys.exc_info()[2])` in `except Exception:` comes a long way already, and it should be possible to invoke the normal hooks from there, so that e.g. capturing etc is handled.
It likely means to decouple / change the code around `CallInfo.from_call` etc though.
I will look into that later.

src/_pytest/unittest.py
nicoddemus: Yep! 👍

nicoddemus:
> I will look into that later.

Sounds good, but we can consider merging it already (as it solves a bunch of issues and doesn't introduce regressions or external changes), and follow up later with any other changes that improve it. What do you think?

blueyed:
I'd like to avoid unnecessary diff churn, and think it might be a good opportunity to improve internals, since it shows that it is currently not flexible enough.
See https://github.com/blueyed/pytest/pull/107 for a dirty POC.

nicoddemus:
@blueyed I figured you won't be working on this anymore, so decided to re-open it, fix conflicts and merge it.

Thanks again!

PR #6003, opened 2019-10-19
summary/OutcomeException: add support for short_msg

blueyed:
This is meant to improve the short msg with `-r`, where you want to see
the non-matched line, and not the first line (that might have matched),
from pytester's `_match_lines`.

This also improves `__str__` / `__repr__`.

TODO:

- [ ] use it for doctests (ReprFailDoctest, https://github.com/pytest-dev/pytest/blob/5e7b2ae704da84db5e9d4a786ebe1efe1997f9ff/src/_pytest/doctest.py#L229-L280) - not really sure what though - but might be just as simple as "doctest failed" - currently it is empty, see `test_doctest_id`.

### DIFF
--- a/src/_pytest/outcomes.py
+++ b/src/_pytest/outcomes.py
@@ -17,7 +17,13 @@ class OutcomeException(BaseException):
         contain info about test and collection outcomes.
     """
 
-    def __init__(self, msg: Optional[str] = None, pytrace: bool = True) -> None:
+    def __init__(
+        self,
+        msg: Optional[str] = None,
+        pytrace: bool = True,
+        *,
+        short_msg: Optional[str] = None
+    ) -> None:
         if msg is not None and not isinstance(msg, str):
             error_msg = (
                 "{} expected string as 'msg' parameter, got '{}' instead.\n"
@@ -27,13 +33,24 @@ class OutcomeException(BaseException):
         BaseException.__init__(self, msg)
         self.msg = msg
         self.pytrace = pytrace
+        self.short_msg = short_msg
 
     def __repr__(self) -> str:
+        if self.short_msg:
+            return "<{} short_msg={!r}>".format(self.__class__.__name__, self.short_msg)
+        msg = self.msg
+        if msg:
+            lines = msg.split("\n", maxsplit=1)
+            if len(lines) > 1:
+                msg = lines[0] + "..."
+            else:
+                msg = lines[0]
+        return "<{} msg={!r}>".format(self.__class__.__name__, msg)
+
+    def __str__(self) -> str:
         if self.msg:
             return self.msg
-        return "<{} instance>".format(self.__class__.__name__)
-
-    __str__ = __repr__
+        return repr(self)
 
 
 TEST_OUTCOME = (OutcomeException, Exception)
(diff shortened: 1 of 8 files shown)


changelog/6003.feature.rst
```diff
@@ -0,0 +1 @@
+Improve short excinfo with LineMatcher failures, via new ``OutcomeException.short_msg``.
```
blueyed:
```suggestion
Improve short excinfo with LineMatcher failures in short test summaries, via new ``OutcomeException.short_msg``.
```
?

nicoddemus: The change is good, but I also suggest to rename it to `6003.improvement.rst`. 👍

src/_pytest/outcomes.py
```diff
@@ -17,7 +17,12 @@ class OutcomeException(BaseException):
         contain info about test and collection outcomes.
     """
 
-    def __init__(self, msg: Optional[str] = None, pytrace: bool = True) -> None:
+    def __init__(
+        self,
+        msg: Optional[str] = None,
+        pytrace: bool = True,
+        short_msg: Optional[str] = None,
```
nicoddemus: Perhaps change `short_message` into a keyword only argument?

src/_pytest/outcomes.py
```diff
@@ -27,13 +32,24 @@ def __init__(self, msg: Optional[str] = None, pytrace: bool = True) -> None:
...
+                msg = lines[0]
+        return "<{} msg={!r}>".format(self.__class__.__name__, msg)
+
+    def __str__(self) -> str:
         if self.msg:
             return self.msg
-        return "<{} instance>".format(self.__class__.__name__)
-
-    __str__ = __repr__
+        return self.__repr__()
```
nicoddemus:
```suggestion
        return repr(self)
```

src/_pytest/outcomes.py
```diff
@@ -116,7 +132,9 @@ def skip(msg: str = "", *, allow_module_level: bool = False) -> "NoReturn":
 skip.Exception = Skipped  # type: ignore
 
 
-def fail(msg: str = "", pytrace: bool = True) -> "NoReturn":
+def fail(
+    msg: str = "", pytrace: bool = True, short_msg: Optional[str] = None
```
nicoddemus: We should make this a keyword only argument because it is a public API.

```suggestion
    msg: str = "", pytrace: bool = True, *, short_msg: Optional[str] = None
```

testing/test_terminal.py
```diff
@@ -757,14 +757,26 @@ def test(i):
...
+    """
+    )
     result = testdir.runpytest()
     result.stdout.no_fnmatch_line("*short test summary*")
     result = testdir.runpytest("-rf")
     result.stdout.fnmatch_lines(
         [
             "*test summary*",
             "FAILED test_fail_extra_reporting.py::test_this - AssertionError: this_failedt...",
+            "FAILED test_fail_extra_reporting.py::test_linematcher - remains unmatched: 'l...",
```
nicoddemus: 👍

src/_pytest/outcomes.py
blueyed: I assume you mean `short_msg`, makes sense then.

blueyed:
Incorporated feedback.

blueyed:
Updated in https://github.com/blueyed/pytest/pull/44.
Closing for now, might come back to it later.

Issue #5206, opened 2019-05-03
--newest-first forgets about existing nodeids when collection is filtered


blueyed:
`--newest-first` (added via https://github.com/pytest-dev/pytest/issues/3034) forgets about cached nodeids when e.g. run with some file as argument.

Code: https://github.com/pytest-dev/pytest/blob/3f5622c577df6d8706b46f3bfe62a3d16e19556e/src/_pytest/cacheprovider.py#L269-L290

I think it should not forget about those - this makes it likely consider all others as new when running without args then again.

It should probably just add to the known items, and have a mechanism to garbage-collect them from time to time, but that is not really critical I suppose.

/cc @feuillemorte

blueyed:
While at it `nodeids` could be renamed to `known_nodeids` maybe also, and a timestamp for its discovery and last collection (or something similar) could be added to the data then.

PR #6004, opened 2019-10-19
cache: NFPlugin: keep known nodeids

blueyed:
Caveat: does not forget about old nodeids

Fixes https://github.com/pytest-dev/pytest/issues/5206

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -264,8 +264,8 @@ class NFPlugin:
         self.cached_nodeids = config.cache.get("cache/nodeids", [])
 
     def pytest_collection_modifyitems(self, session, config, items):
+        new_items = OrderedDict()
         if self.active:
-            new_items = OrderedDict()
             other_items = OrderedDict()
             for item in items:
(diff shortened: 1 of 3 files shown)


src/_pytest/cacheprovider.py
```diff
@@ -276,7 +276,11 @@ def pytest_collection_modifyitems(self, session, config, items):
             items[:] = self._get_increasing_order(
                 new_items.values()
             ) + self._get_increasing_order(other_items.values())
-        self.cached_nodeids = [x.nodeid for x in items if isinstance(x, pytest.Item)]
+        else:
+            for item in items:
+                if item.nodeid not in self.cached_nodeids:
+                    new_items[item.nodeid] = item
+        self.cached_nodeids.extend(new_items)
```
blueyed: With regard to dropping `isinstance` here: see https://github.com/pytest-dev/pytest/pull/3218#discussion_r168623666
Should either be kept if we have a test for this (`testing/acceptance_test.py` worked for me), or have a test to cover it if necessary.

nicoddemus: Fair enough!

PR #6005, opened 2019-10-19
tests: harden/fix test_trial_error

blueyed:


blueyed:
I want to use it with fixing unittests (in features).

PR #6014, opened 2019-10-20
unittest: do not use `TestCase.debug` with `--pdb`

blueyed:
Reverts https://github.com/pytest-dev/pytest/pull/1890, which needs to
be fixed/addressed in another way
(https://github.com/pytest-dev/pytest/pull/5996).

Fixes https://github.com/pytest-dev/pytest/issues/5991.

### DIFF
--- a/changelog/5991.bugfix.rst
+++ b/changelog/5991.bugfix.rst
@@ -1,0 +1,1 @@
+Do not use ``TestCase.debug()`` for unittests with ``--pdb``.
--- a/src/_pytest/unittest.py
+++ b/src/_pytest/unittest.py
@@ -187,29 +187,8 @@ class TestCaseFunction(Function):
     def stopTest(self, testcase):
         pass
 
-    def _handle_skip(self):
-        # implements the skipping machinery (see #2137)
-        # analog to pythons Lib/unittest/case.py:run
-        testMethod = getattr(self._testcase, self._testcase._testMethodName)
-        if getattr(self._testcase.__class__, "__unittest_skip__", False) or getattr(
-            testMethod, "__unittest_skip__", False
-        ):
-            # If the class or method was skipped.
-            skip_why = getattr(
-                self._testcase.__class__, "__unittest_skip_why__", ""
-            ) or getattr(testMethod, "__unittest_skip_why__", "")
-            self._testcase._addSkip(self, self._testcase, skip_why)
-            return True
-        return False
-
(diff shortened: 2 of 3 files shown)


blueyed:
Decided to revert to sane behavior, until this is addressed properly (https://github.com/pytest-dev/pytest/pull/5996 is a start).

blueyed:
Sure - but it was a regression / bug from the beginning.
Therefore I'd rather revert it already, and fix / re-try it later.

blueyed:
Should I do this against master maybe instead? ;)

Seriously, I am aware that it reverts a previous change, but that one is just wrong, and caused multiple (and more serious) regressions to start with.

I've came across this several times, and it had to be adjusted / fixed multiple times.
The original author never reacted to my pings, and might not even be using pytest anymore.
We should not protect this kind of "feature" (which is really a hack after all).
The proper / better way might just be to add/support "--unittest-debug" - using "--pdb" for that was just a misunderstanding to begin with (actually people are using that by default, and it causes all teardowns to be skipped then (IIRC)).

blueyed:
But since this conflicts anyway already, and me being tired of pointing it out (and people not caring), I'll just close it for now.
(I'm not using unittests myself btw)

blueyed:
@nicoddemus 
/cc @pytest-dev/contributors 
To summary: unittests use `.debug()`, which is only meant to be used manually to debug unittests.
Somebody came along and found that tests are being teared down already when using `--pdb` for post-portem debugging, which closes the GUI.
This was in 2016 already (#1890), and a constant source of issues, e.g. reported via pytest-django, but also here (the latest https://github.com/pytest-dev/pytest/issues/5991).

After investigating, and giving it a try myself to fix it I've concluded that it is the best to revert this, especially since the issue mentions a specific workaround already, and the more obvious would be to just put a `pdb.set_trace()` where you need it (before the failure).

Anyway, what I really do not get is that "reverting this mis-feature" is being rejected as being a regression itself..!
Like I've mentioned above the original author never reacted to my concerns about this, and might not even be using pytest anymore - but still this bug is being protected from being reverted?!

nicoddemus:
You got a point there. I'm on vacation at the moment so I can't look at the code now, but I will pick it up again when I get back.

nicoddemus:
I agree, see https://github.com/pytest-dev/pytest/pull/5996#issuecomment-552639319.

PR #6016, opened 2019-10-20
pytester: parseconfigure: remove duplicate config._ensure_unconfigure

blueyed:
This gets done in `parseconfig` already.


blueyed:
Shouldn't make a difference, because they are added just after each other.

PR #6018, opened 2019-10-21
[RFC] pytester. support stdin with runpytest_inprocess

blueyed:
This allows for speeding up / replacing pexpect with some pdb tests.

Ref: https://github.com/pytest-dev/pytest/pull/4996 (less controversial).


blueyed:
Taken into https://github.com/blueyed/pytest/pull/29 - from past experience this is not likely to fly here anyway.

PR #6021, opened 2019-10-21
ci: Travis: cover verbose=1

blueyed:
Taken out of https://github.com/pytest-dev/pytest/pull/5936.

This should be less controversial / time-taking to merge, and is useful for other PRs with regard to verbosity.


nicoddemus:
Thanks, but #5936 has incorporated this changes already. 👍 (Only saw this after I've merged it).

PR #6025, opened 2019-10-21
pytester: typing

blueyed:


blueyed:
Backed out to use `Union[int, ExitCode]`.

src/_pytest/pytester.py
```diff
@@ -879,16 +881,19 @@ class reprec:
...
 
-    def runpytest(self, *args, **kwargs):
+    def runpytest(self, *args, **kwargs) -> RunResult:
         """Run pytest inline or in a subprocess, depending on the command line
         option "--runpytest" and return a :py:class:`RunResult`.
 
         """
         args = self._ensure_basetemp(args)
-        return self._runpytest_method(*args, **kwargs)
+        if self._method == "inprocess":
```
bluetech: I find it more symmetric to write as

```
if self._method == "inprocess": ...
elif self._method == "subprocess": ...
else:
     assert False, "Unrecognized runpytest option: {}".format(self._method)
```

Also, if we sometime decide to change this `_method` variable to be a `Literal["inprocess", "subprocess"]`, then we can add exhaustiveness checking in the `else` branch, so mypy warns if we forget to handle a case.

src/_pytest/pytester.py
```diff
@@ -361,7 +362,7 @@ class RunResult:
     :ivar duration: duration in seconds
     """
 
-    def __init__(self, ret, outlines, errlines, duration):
+    def __init__(self, ret: Union[int, ExitCode], outlines, errlines, duration):
```
bluetech: Add `-> None`?

testing/test_pytester.py
```diff
@@ -406,6 +406,27 @@ def test_testdir_subprocess(testdir):
     assert testdir.runpytest_subprocess(testfile).ret == 0
 
 
+def test_testdir_subprocess_via_runpytest_arg(testdir):
```
bluetech: Add `-> None`?

src/_pytest/pytester.py
blueyed: > Literal["inprocess", "subprocess"]

Cool, could do this here already then.

blueyed: But of course it would not prevent runtime error really, for that we should maybe check it in `__init__` then already?

blueyed: Changed it to RuntimeError and your style.
It is using "choices" however already: https://github.com/blueyed/pytest/blob/bf4c45df96346684330943cd9db71af013a2b93a/src/_pytest/pytester.py#L46

src/_pytest/pytester.py
```diff
@@ -361,9 +362,9 @@ class RunResult:
     :ivar duration: duration in seconds
     """
 
-    def __init__(self, ret, outlines, errlines, duration):
+    def __init__(self, ret: Union[int, ExitCode], outlines, errlines, duration) -> None:
         try:
-            self.ret = pytest.ExitCode(ret)
+            self.ret = pytest.ExitCode(ret)  # type: Union[int, ExitCode]
```
blueyed: Required to fix `mypy: Incompatible types in assignment (expression has type "Union[int, ExitCode]", variable has type "ExitCode")` below on line 369 - makes sense?

bluetech: Yes, makes perfect sense.

PR #6029, opened 2019-10-22
ReprEntry: enhance coloring of "E" lines

blueyed:
- only make the first line red + bold
- only make "E" prefix of following lines red.  Could also be just not
  bold there, but would still result in a wall-of-red with LineMatcher
  failures.

Slightly conflicts with https://github.com/pytest-dev/pytest/pull/5064.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -971,18 +971,30 @@ class ReprEntry(TerminalRepr):
         self.reprfileloc = filelocrepr
         self.style = style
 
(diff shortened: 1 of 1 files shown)


nicoddemus:
`master`:

image

image

I'm +0 TBH... often the interesting part is not the first `E` line, but some other of the following lines and it really depends on the code being tested...

blueyed:
I'm happy with that in my fork for now (https://github.com/blueyed/pytest/pull/13).

PR #6030, opened 2019-10-22
ci: use tox -vv

blueyed:
This will display durations, and is useful in logs in general.


blueyed:
Yes, it would be nice if `tox` had it with `-v` already, or even just `--durations` or something similar.

PR #6032, opened 2019-10-22
ci: Travis: configure/restrict branch builds

blueyed:


.travis.yml
```diff
@@ -130,3 +130,10 @@ notifications:
     skip_join: true
   email:
     - pytest-commit@python.org
+
+branches:
+  only:
+    - master
+    - features
+    - 4.6-maintenance
+    - /^\d+(\.\d)+$/
```
bluetech: Confusing that tags are filtered along with branches, but so it is!

Did you mean to write `/^\d+(\.\d+)+$/` (add `+` to the second `\d`)?

blueyed: Oh, sure.  Amended - thanks for catching!

nicoddemus:
What's the reason for this change? We rarely have branches on the main repository, so I'm not sure this is an issue.

nicoddemus:
Oh! This avoids doing builds twice in case a PR is opened with a branch in the main repo? In that case I'm OK with the change. 👍

blueyed:
@nicoddemus 
Yes.  See also https://github.com/pytest-dev/pytest/pull/6033#pullrequestreview-306153160.

PR #6033, opened 2019-10-22
[4.6] ci: Travis: configure/restrict branch builds

blueyed:


.travis.yml
```diff
@@ -166,3 +166,10 @@ notifications:
     skip_join: true
   email:
     - pytest-commit@python.org
+
+branches:
+  only:
+    - master
+    - features
+    - 4.6-maintenance
+    - /^\d+(\.\d+)+$/
```
asottile: I usually add another explicit pattern in case people need to actually play with CI (especially useful in forks)

blueyed: Care to share?

asottile: https://github.com/pre-commit/pre-commit/blob/7728162d7a47f387187e379b8011b224b168555d/azure-pipelines.yml#L3

blueyed: So you mean to include a prefix "test-me-*"?

I cannot really see why that would be useful - forks then have to use branches "test-me-foo" to have Travis run for their branch pushes, right?
The idea is that push branches for a "foo" branch are not run with this config, right?

I think real forks should also override that config then.

Regarding your use case we should just exclude "/^.*-patch-.*/" only then?

btw: it would make more sense to discuss this on the master PR - I've only created this already, because I've thought it would be easy/non-controversial.

PR #5847, opened 2019-09-13
2/X Fix check_untyped_defs = True mypy errors

bluetech:
Fixes for some more `check_untyped_defs` errors from mypy.

The first commit required a small refactoring (see commit message).

Continuation of #5673.

PR #6035, opened 2019-10-22
minor: fix mypy error

blueyed:
> Incompatible types in assignment
> Value of type "Set[Any]" is not indexable


bluetech:
I fixed it like this: https://github.com/pytest-dev/pytest/pull/5847/files#diff-e4269384a671a462a4824e8b06d4644bR245

This is also fine though!

blueyed:
I like yours better, and it avoidss conflicts - thanks!

PR #6036, opened 2019-10-22
mypy: no_implicit_optional = False

blueyed:
This allows for easier signatures.

doc: https://mypy.readthedocs.io/en/latest/command_line.html#none-and-optional-handling

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,7 +64,7 @@ formats = sdist.tgz,bdist_wheel
 
 [mypy]
 ignore_missing_imports = True
-no_implicit_optional = True
+no_implicit_optional = False
 strict_equality = True
 warn_redundant_casts = True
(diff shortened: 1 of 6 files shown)


bluetech:
I personally prefer it to be True, I find a signature like `exprinfo: str = None` to be inconsistent and surprising. But I don't feel strongly about it.

blueyed:
Me neither, but I can see no benefit in the boilerplate overhead.
It is a nice shortcut IMHO, and mypy default to it IIRC.

bluetech:
> mypy default to it IIRC.

The consensus seems to be that this was a bad default:

- https://github.com/python/typing/issues/275
- https://github.com/python/peps/pull/689

asottile:
:-1: strongly prefer being explicit here

PR #6040, opened 2019-10-22
tests: mark test_meta as slow

blueyed:
This moves it to the end of tests during collection.  Takes ~7s for me.

### DIFF
--- a/testing/test_meta.py
+++ b/testing/test_meta.py
@@ -15,6 +15,7 @@ def _modules():
     )
 
 
+@pytest.mark.slow
 @pytest.mark.parametrize("module", _modules())
 def test_no_warnings(module):
     # fmt: off


asottile:
for optimal packing, shouldn't slow tests be run first?

also the individual parametrized tests here aren't slow -- just the collective group of them

blueyed:
> for optimal packing, shouldn't slow tests be run first?

The idea is to run `pytest -x` to see where relevant tests are, for example.

> also the individual parametrized tests here aren't slow -- just the collective group of them

Yeah.  Could mark the module instead.

It is also a bit about "those are tests that should be run on CI, but are not really required / expected to fail usually".  But I do not feel like we should have an extra mark for them.

blueyed:
Can we just merge this for now?

nicoddemus:
I'm +0 on this, so unless @asottile prefers otherwise, I'm OK with merging it.

PR #6041, opened 2019-10-22
pytester: runpytest_inprocess: use splitlines()

blueyed:
This avoids having a trailing empty lines always.


blueyed:
Thanks!  I still do not like your gifs FWIW.. ;)  (smiley to prove I have humor in general)

PR #6042, opened 2019-10-22
conftest: allow skipping of test re-ordering

blueyed:
I've wanted this when checking why a certain test is failing when only
being run directly, and thought it would be due to the re-ordering.
Therefore it is useful to being able to skip it via an env.

Adds it to a Travis job, so that we are running tests in two different
orders at least.


blueyed:
Appears to cause "testing/test_pdb.py::TestPDB::test_pdb_unittest_postmortem" to hang.

blueyed:
Shows that the suite is broken, no interest in fixing it myself.

PR #6045, opened 2019-10-23
minor: typing for ReprFailDoctest

blueyed:


blueyed:
Job canceled manually.

src/_pytest/doctest.py
```diff
@@ -113,11 +115,10 @@ def _is_doctest(config, path, parent):
 
 
 class ReprFailDoctest(TerminalRepr):
-    def __init__(self, reprlocation_lines):
-        # List of (reprlocation, lines) tuples
+    def __init__(self, reprlocation_lines: List[Tuple[ReprFileLocation, List[str]]]):
```
bluetech: Can the Lists be Sequences?

blueyed: Probably.. just based it on the comment.. :D - this was just from a stash when working at that code.
Should I change it?

bluetech: Yes, it doesn't modify the input so Sequence is better all around.

LGTM otherwise.

blueyed: Changed both `List`s now.  Hope that's ok? :)

Issue #6057, opened 2019-10-25
Show a tolerance on complex numbers


AnjoMan:
**Suggestion**

Pytest currently doesn't show tolerance when asserting approximate equality on complex numbers. This can make it a little hard to understand just how `approx` compares complex numbers.

for example, on pytest 5.2.2

```
>       assert 3.0002 + 4.0003 * 1j == approx(3 + 4j)
E       assert (3.0002+4.0003j) == (3+4j)
E         -(3.0002+4.0003j)
E         +(3+4j)
```

When pytest does comparisons involving complex numbers, it computes the complex different between the two numbers and then asserts that the absolute value of the difference falls below the tolerance (which, for relative tolerance is calculated as a fraction of the absolute value of the value passed to `approx`).

This means that if you draw a circle around the target point in the complex plane, with a radius equal to the tolerance, the compared value should fall within that circle
![image](https://user-images.githubusercontent.com/2414360/67540876-0c74b780-f6b5-11e9-8562-88c6dd6b8d4a.png)

given this, I propose adding a tolerance indicator to complex numbers using a polar notation:

```
>       assert 3.0002 + 4.0003 * 1j == approx(3 + 4j)
E       assert (3.0002+4.0003j) == (3+4j) ± 5e-06 ∠ ±180°
E         -(3.0002+4.0003j)
E         +(3+4j) ± 5e-06 ∠ ±180°
```

in this notation, `5e-06` is the calculated relative tolerance, derived as `abs(3+4j) * 1e-06`. the `∠ ±180°` component is fixed to indicate that the delta between actual  and expected is a complex value as well, and that it can have any angle as long as the magnitude is within the tolerance.

**Version**

    * pytest 5.2.2
    * python 3.7.3
    * ubuntu 19.04

Zac-HD:
Closed by #6058 - thanks for the writeup and your pull request!

FYI you can [set PRs to automatically close issues when merged](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) in future 😄

AnjoMan:
oo, thanks Zac!

PR #6058, opened 2019-10-25
6057 tolerance on complex approx

AnjoMan:
Adds a tolerance indicator to `pytest.approx` when the expected value is a complex number. This should make it easier to reason about how complex number approximations are handled by pytest.

Relates to #6057.

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -223,26 +223,24 @@ class ApproxScalar(ApproxBase):
     def __repr__(self):
         """
         Return a string communicating both the expected value and the tolerance
-        for the comparison being made, e.g. '1.0 +- 1e-6'.  Use the unicode
-        plus/minus symbol if this is python3 (it's too hard to get right for
-        python2).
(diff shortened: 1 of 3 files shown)


src/_pytest/python_api.py
```diff
@@ -223,26 +223,24 @@ class ApproxScalar(ApproxBase):
...
-        for the comparison being made, e.g. '1.0 +- 1e-6'.  Use the unicode
-        plus/minus symbol if this is python3 (it's too hard to get right for
-        python2).
+        for the comparison being made, e.g. '1.0 ± 1e-6', '(3+4j) ± 5e-6 ∠ ±180°'.
         """
-        if isinstance(self.expected, complex):
-            return str(self.expected)
 
         # Infinities aren't compared using tolerances, so don't show a
         # tolerance.
```
kalekundert:
```suggestion
        # tolerance.  The `abs()` call is required to get complex numbers right.
```

AnjoMan: Yes, good idea!

AnjoMan:
@kalekundert I made the change you requested. Thanks for your review!

PR #6064, opened 2019-10-26
Replace py.io.TextIO with io.StringIO

bluetech:
In Python3, `py.io.TextIO` is just an alias to `io.StringIO`. Remove the indirection.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -4,6 +4,7 @@ import sys
 import traceback
 from inspect import CO_VARARGS
 from inspect import CO_VARKEYWORDS
+from io import StringIO
 from traceback import format_exception_only
 from types import CodeType
 from types import TracebackType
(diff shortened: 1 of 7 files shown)


blueyed:
Thanks for this!
Just wanted to note that this is not really a bugfix, and would have been better off in features.

asottile:
> Thanks for this!
> Just wanted to note that this is not really a bugfix, and would have been better off in features.

Seems to perfectly fit the bill for `trivial` to me -- `master` seems appropriate

blueyed:
Also "trivial" is more like fixing some typo, not changing imports in multiple files.

asottile:
this change literally does nothing, they are referrentially equal -- if this isn't trivial I don't know what is

blueyed:
https://github.com/pytest-dev/pytest/pull/6104

PR #6066, opened 2019-10-26
Add type annotations to _pytest.pathlib and _pytest.tmpdir

bluetech:
These files are mostly straightforward to annotate.

### DIFF
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -1,7 +1,6 @@
 import atexit
 import fnmatch
 import itertools
-import operator
 import os
 import shutil
 import sys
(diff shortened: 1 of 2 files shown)


src/_pytest/pathlib.py
```diff
@@ -311,7 +326,7 @@ def resolve_from_str(input, root):
         return root.joinpath(input)
 
 
-def fnmatch_ex(pattern, path):
+def fnmatch_ex(pattern: str, path) -> bool:
     """FNMatcher port from py.path.common which works with PurePath() instances.
```
blueyed: Also `path` here?

bluetech: There is a `path = PurePath(path)` below which means it can have various types. This feels like something to clean up, so I didn't touch it for now.

PR #6068, opened 2019-10-26
[RDY] tox: remove platform restriction, only used for pexpect

blueyed:
This would prevent TOXENV=py37-pexpect-… from running on e.g. FreeBSD.
And even on Windows it is pytest's job of skipping the tests then.  This
was probably still from when the pexpect env was only running
pexpect-based tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -36,8 +36,6 @@ setenv =
 
     lsof: _PYTEST_TOX_POSARGS_LSOF=--lsof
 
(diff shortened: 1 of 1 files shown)


blueyed:
Azure failure is unrelated (codecov).

blueyed:
> Thanks, this should go into `master` though.

Then I'd rather wait until master is merged from features with the next release then.

nicoddemus:
No need to wait, better have this in now than dragging it further.

PR #6078, opened 2019-10-27
[WIP/RFC] mark: filter keywords/marks before _genfunctions

blueyed:
Ref: https://github.com/pytest-dev/pytest-repeat/issues/15


blueyed:
While this works better it is still not cool.
I would prefer if items could be easily added in `pytest_collection_modifyitems`.

PR #6099, opened 2019-10-29
Fix --trace for parametrized tests

davidszotten:
I'm not sure if this is the correct / best way to fix so please check
critically when reviewing

Without this, the second time it tries to stop in a parametrized
function it raises instead:

`ValueError: --trace can't be used with a fixture named func!`

Also lift the restriction that was disallowing `--trace` on tests with a parameter named `func`

### DIFF
--- a/src/_pytest/debugging.py
+++ b/src/_pytest/debugging.py
@@ -1,5 +1,6 @@
 """ interactive debugging with PDB, the Python Debugger. """
 import argparse
+import functools
 import pdb
 import sys
 from doctest import UnexpectedException
@@ -274,13 +275,16 @@ class PdbTrace:
 def _test_pytest_function(pyfuncitem):
     _pdb = pytestPDB._init_pdb("runcall")
     testfunction = pyfuncitem.obj
-    pyfuncitem.obj = _pdb.runcall
-    if "func" in pyfuncitem._fixtureinfo.argnames:  # pragma: no branch
-        raise ValueError("--trace can't be used with a fixture named func!")
-    pyfuncitem.funcargs["func"] = testfunction
-    new_list = list(pyfuncitem._fixtureinfo.argnames)
-    new_list.append("func")
-    pyfuncitem._fixtureinfo.argnames = tuple(new_list)
+
+    # we can't just return `partial(pdb.runcall, testfunction)` because (on
+    # python < 3.7.4) runcall's first param is `func`, which means we'd get
+    # an exception if one of the kwargs to testfunction was called `func`
+    @functools.wraps(testfunction)
+    def wrapper(*args, **kwargs):
+        func = functools.partial(testfunction, *args, **kwargs)
+        _pdb.runcall(func)
+
+    pyfuncitem.obj = wrapper
 
 
 def _enter_pdb(node, excinfo, rep):
--- a/testing/test_pdb.py
+++ b/testing/test_pdb.py
@@ -1025,6 +1025,51 @@ class TestTraceOption:
         assert "Exit: Quitting debugger" not in child.before.decode("utf8")
         TestPDB.flush(child)
 
+    def test_trace_with_parametrize_handles_shared_fixtureinfo(self, testdir):
+        p1 = testdir.makepyfile(
+            """
+            import pytest
+            @pytest.mark.parametrize('myparam', [1,2])
+            def test_1(myparam, request):
+                assert myparam in (1, 2)
+                assert request.function.__name__ == "test_1"
+            @pytest.mark.parametrize('func', [1,2])
+            def test_func(func, request):
+                assert func in (1, 2)
+                assert request.function.__name__ == "test_func"
+            @pytest.mark.parametrize('myparam', [1,2])
+            def test_func_kw(myparam, request, func="func_kw"):
+                assert myparam in (1, 2)
+                assert func == "func_kw"
+                assert request.function.__name__ == "test_func_kw"
+            """
+        )
+        child = testdir.spawn_pytest("--trace " + str(p1))
+        for func, argname in [
+            ("test_1", "myparam"),
+            ("test_func", "func"),
+            ("test_func_kw", "myparam"),
+        ]:
+            child.expect_exact("> PDB runcall (IO-capturing turned off) >")
+            child.expect_exact(func)
+            child.expect_exact("Pdb")
+            child.sendline("args")
+            child.expect_exact("{} = 1\r\n".format(argname))
+            child.expect_exact("Pdb")
+            child.sendline("c")
+            child.expect_exact("Pdb")
+            child.sendline("args")
+            child.expect_exact("{} = 2\r\n".format(argname))
+            child.expect_exact("Pdb")
+            child.sendline("c")
+            child.expect_exact("> PDB continue (IO-capturing resumed) >")
+        rest = child.read().decode("utf8")
+        assert "6 passed in" in rest
+        assert "reading from stdin while output" not in rest
+        # Only printed once - not on stderr.
+        assert "Exit: Quitting debugger" not in child.before.decode("utf8")
+        TestPDB.flush(child)
+
 
 def test_trace_after_runpytest(testdir):
     """Test that debugging's pytest_configure is re-entrant."""
@@ -1150,7 +1195,6 @@ def test_pdbcls_via_local_module(testdir):
 
                 def runcall(self, *args, **kwds):
                     print("runcall_called", args, kwds)
-                    assert "func" in kwds
         """,
     )
     result = testdir.runpytest(
--- a/changelog/6099.bugfix.rst
+++ b/changelog/6099.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ``--trace`` when used with parametrized functions.


src/_pytest/debugging.py
```diff
@@ -275,12 +275,14 @@ def _test_pytest_function(pyfuncitem):
...
     testfunction = pyfuncitem.obj
     pyfuncitem.obj = _pdb.runcall
-    if "func" in pyfuncitem._fixtureinfo.argnames:  # pragma: no branch
+    if "func" in pyfuncitem.funcargs:  # pragma: no branch
         raise ValueError("--trace can't be used with a fixture named func!")
     pyfuncitem.funcargs["func"] = testfunction
-    new_list = list(pyfuncitem._fixtureinfo.argnames)
-    new_list.append("func")
-    pyfuncitem._fixtureinfo.argnames = tuple(new_list)
+    if "func" not in pyfuncitem._fixtureinfo.argnames:
```
RonnyPfannschmidt: indeed this workaround requires a comment

practically this belongs to the FunctionDefininition (and should only happen once at that)
unfortunately with the current design that's not exposed and we have a hack instead

the ""correct" fix is not to have more than one place where this is added, but that's for after FunctionDefinition lands

testing/test_pdb.py
```diff
@@ -1025,6 +1025,30 @@ def test_3():
         assert "Exit: Quitting debugger" not in child.before.decode("utf8")
         TestPDB.flush(child)
 
+    def test_trace_with_parametrize(self, testdir):
```
RonnyPfannschmidt:
```suggestion
    def test_trace_with_parametrize_handles_shared_fixtureinfo(self, testdir):
```

src/_pytest/debugging.py
```diff
@@ -275,12 +275,14 @@ def _test_pytest_function(pyfuncitem):
...
+        # if using e.g. parametrize the _fixtureinfo is shared
```
RonnyPfannschmidt:
```suggestion
        # TODO: when using parameterized tests, the _fixtureinfo is shared
        # that needs to move to FunctionDefinition
```

davidszotten:
ah, apologies i misread the instructions. while updating i also squashed the 3 commits, hope that's ok

davidszotten:
Thank you for the quick reviews!

blueyed:
Also amended the changelog (punctuation).

davidszotten:
Thanks @blueyed

src/_pytest/debugging.py
```diff
@@ -275,12 +275,15 @@ def _test_pytest_function(pyfuncitem):
...
     pyfuncitem.funcargs["func"] = testfunction
-    new_list = list(pyfuncitem._fixtureinfo.argnames)
-    new_list.append("func")
-    pyfuncitem._fixtureinfo.argnames = tuple(new_list)
+    if "func" not in pyfuncitem._fixtureinfo.argnames:
+        # TODO: when using parameterized tests, the _fixtureinfo is shared
+        # that needs to move to FunctionDefinition
+        new_list = list(pyfuncitem._fixtureinfo.argnames)
+        new_list.append("func")
+        pyfuncitem._fixtureinfo.argnames = tuple(new_list)
```
blueyed: The false branch for "if "func" not in pyfuncitem._fixtureinfo.argnames:" is not covered.
@RonnyPfannschmidt 
Is that something that can happen?
Can/should we cover it then?

blueyed:
I've pushed a fixup to use functools.partial there.

src/_pytest/debugging.py
```diff
@@ -274,13 +275,8 @@ def pytest_pyfunc_call(self, pyfuncitem):
...
     testfunction = pyfuncitem.obj
-    pyfuncitem.obj = _pdb.runcall
-    if "func" in pyfuncitem._fixtureinfo.argnames:  # pragma: no branch
-        raise ValueError("--trace can't be used with a fixture named func!")
-    pyfuncitem.funcargs["func"] = testfunction
-    new_list = list(pyfuncitem._fixtureinfo.argnames)
-    new_list.append("func")
-    pyfuncitem._fixtureinfo.argnames = tuple(new_list)
+    pyfuncitem.obj = functools.partial(_pdb.runcall, testfunction)
+    pyfuncitem.obj.__name__ = testfunction.__name__
```
RonnyPfannschmidt: that solution checks out proper,
nonetheless this feels like a hijacking of the pr to me

blueyed: > nonetheless this feels like a hijacking of the pr to me

￼why?
If it wasn't wanted then you could uncheck the box that allows it.
It can always be reset also (I've explicitly mentioned the old head).
I'm just trying to help fixing the issue (and limitations that were in place due to unknown reasons).

Why shouldn't I push hardening/improvements when I have them locally already, so that they can be reviewed/tested on CI?

@davidszotten 
What do you think?

blueyed: Reset it to  b421aef (from c45761c).  The test was timing out on CI, and I do not want to leave the impression to hijack / being hostile.
Will try to fix this in my fork, and then leave a comment here.

davidszotten:
i don't feel that strongly and can see both sides of this argument. i'm not sure i know enough about the internals to tell how "controversial" your new implementation is. but if that needs more discussion then that does feel like maybe it could be split out into a separate pr and leaving this pr to just tweak the existing workaround to work better

on the other hand, if this is just obviously better _and_ fixes the problem, maybe it's a small enough change that it's better to just bundle this together


on the contributor management side, i'm relatively seasoned in open source and experienced enough to not mind, but someone more new to open source or this project might find it daunting and/or off-putting to have their pr "hijacked" (have someone else force-push a whole new / different implementation without talking it over first)

src/_pytest/debugging.py
RonnyPfannschmidt: the human interaction perception of such actions of merit can widely vary,
whiles some may be accepting of that, for me personally it would be the last time that i'd contribute to the project 

we should perhaps clarify on what we hope to accomplish by enabling maintainers to push commits to other prs

for me the main reason is to help new  contributors that struggle with the tooling,
for better solutions/code change suggestions outside of pre-commit,
i would always  strongly suggest to use something like proposed edits for guidance

for me there is a very important difference between guiding/mentoring and  accidentally taking over

and *i* perceive your actions as "taking over" instead of guiding/mentoring

blueyed:
Thanks for clarifying - it wasn't my intention to step on toes here.
It is reset, so we're back to where I can suggest to using functools here maybe.. :)

> have their pr "hijacked" (have someone else force-push a whole new / different implementation without talking it over first)

FWIW, it was a separate commit on top.  The previous force-push was for the test hardening and typo/changelog fix (I do not think that merits extra commits, and we cannot squash-merge (#4361)).

RonnyPfannschmidt:
i also agree that the proposed/new solution by @blueyed is a better solution

however with the "maintainer/project hat on" i had to make a clear mention about the maintainer interaction behaviour here is potentially problematic

i'm happy that @davidszotten is on the accepting side and would like to see the proposed solution go into the final thing in a collaborative manner

src/_pytest/debugging.py
blueyed: > for me personally it would be the last time that i'd contribute to the project

why?
It's not like anything has been merged here already, it's just still being worked on.  References are/were provided to revert/reset it, and IMHO it is far better to have CI results and an actual commit rather than a diff in a comment.  But I will stick to the latter then in the future here.

> strongly suggest to use something like proposed edits for guidance

I'm doing that also, but here it was not fitting.  Posting a diff then maybe instead.

davidszotten:
ah, i missed that the impl change was a separate commit. however i think the comments about people feel like their prs are taken over would almost equally apply with new commits added to their pr. 

as for multiple commits, i think it's easier to have multiple commits while in wip, and then (e.g.) ask the pr owner to squash once it's done


re: the actual impl (which i'm happy to force-push onto this myself). did you figure out why the tests hang?

davidszotten:
@blueyed hm from what i can see, even with the impl based on `partial`, `--trace` is unhappy with a param named `func` (easier to see by running manually instead of inside pexpect)

RonnyPfannschmidt:
ah, it needs a different wrapper, the partial has the positional argument named `func` already used up,

so it cant have that as fixture anymore

the solution is to create a wrapper with a complete own signature and invoke the running inside

src/_pytest/debugging.py
RonnyPfannschmidt: actions demonstrate values and projects that brush over contributors/new  contributors don't foster/respect them, and i don't want that kind of interaction

davidszotten: agreed. though as feedback to you @RonnyPfannschmidt , of course @blueyed is also a contributor and maybe we can think of a better way to explain the behaviours we want to promote and why than something as brief as "this feels like a hijacking of the pr"

davidszotten:
actually, the issue is that the implementation of p(well, b)db.runcall "uses" a param named func for itself. i think the solution is to lift out its implementation. pr update coming soon

davidszotten:
Updated, please re-review. Thanks @blueyed for the implementation idea and the test. Some of the lines codecov claim to be uncovered seem wrong but i may be misunderstanding something

src/_pytest/debugging.py
```diff
@@ -274,13 +276,24 @@ def pytest_pyfunc_call(self, pyfuncitem):
...
-    pyfuncitem.obj = _pdb.runcall
-    if "func" in pyfuncitem._fixtureinfo.argnames:  # pragma: no branch
-        raise ValueError("--trace can't be used with a fixture named func!")
-    pyfuncitem.funcargs["func"] = testfunction
-    new_list = list(pyfuncitem._fixtureinfo.argnames)
-    new_list.append("func")
-    pyfuncitem._fixtureinfo.argnames = tuple(new_list)
+
+    # we need a copy of bdb.runcall because it takes `func` as a parameter
+    # which means we can't have a kwarg called func
```
blueyed: Ah.. in 3.8 it is positional-only (https://github.com/python/cpython/blob/9e87d5d964f4f49c24dc3dd4bb7e84ec8db1876e/Lib/bdb.py#L621).

I'm not sure why we need to support kwargs named "func" for the param, when pytest does not allow for them to be kwargs in the first place?  ("function already takes an argument 'param' with a default value")
```python
import pytest


@pytest.mark.parametrize("param", [1, 2])
def test_func_kw(param="foo"):
    pass
```

blueyed:
> @blueyed hm from what i can see, even with the impl based on `partial`, `--trace` is unhappy with a param named `func` (easier to see by running manually instead of inside pexpect)

This passes?
```python
@pytest.mark.parametrize('func', [1,2])
def test_func(func, request):
    assert func in (1, 2)
    assert request.function.__name__ == "test_func"
```

davidszotten:
> Also, btw, please do not mention me in commit messages via @blueyed (triggers notifications, also later when being "merged around").

ah, apologies. would you like me to edit that out? if so, can i keep your name without the `@` for credit?

(not sure i understand what you mean by "merged around")

davidszotten:
maybe the `partial` solution needs 3.8+ to work (sorry a little hard to tell the order of your two comments so maybe this is already what you were saying)

blueyed:
> ah, apologies. would you like me to edit that out? if so, can i keep your name without the `@` for credit?

Use co-authored-by?

> (not sure i understand what you mean by "merged around")

Every time the commit is used, also here already:
2019-10-30-224836_745x136_escrotum

blueyed:
> maybe the `partial` solution needs 3.8+ to work (sorry a little hard to tell the order of your two comments so maybe this is already what you were saying)

No, I've tested that with 3.7.4.  But might be different with older versions even more.

blueyed:
Does https://github.com/pytest-dev/pytest/pull/6099#issuecomment-548126638 not work for you?  Which Python then?

davidszotten:
```
python -V
Python 3.7.0
(pytest)david@odie:pytest (trace_parametrize)(2)$ cat p.py
import pdb

def foo(func):
    pass

pdb.runcall(foo, func=None)
(pytest)david@odie:pytest (trace_parametrize)(2)$ python p.py
  File "p.py", line 6, in <module>
    pdb.runcall(foo, func=None)
  File "/Users/david/.pyenv/versions/3.7.0/lib/python3.7/pdb.py", line 1600, in runcall
    return Pdb().runcall(*args, **kwds)
TypeError: runcall() got multiple values for argument 'func'
```

davidszotten:
> Use co-authored-by?
ah, hadn't heard of that before

blueyed:
> > Use co-authored-by?
> 
> ah, hadn't heard of that before

You have it for ronny.. ;)
(GitHub might have added it)

> pdb.runcall(foo, func=None)

Sure.
But the functools.partial solution does not use that, but `pdb.runcall(func)`, doesn't it?
(`pyfuncitem.obj = functools.partial(_pdb.runcall, testfunction)`, not `func=testfunction`)

It does not work when the original function as a kwarg "func", but like I've said pytest dues not support this for fixtures anyway.
All of these pass for me:
```python
import pytest


@pytest.mark.parametrize('myparam', [1, 2])
def test_1(myparam, request):
    assert myparam in (1, 2)
    assert request.function.__name__ == "test_1"


@pytest.mark.parametrize('func', [1, 2])
def test_func(func, request):
    assert func in (1, 2)
    assert request.function.__name__ == "test_func"


@pytest.mark.parametrize('myparam', [1, 2])
def test_func_kw(request, myparam, func="foo"):
    assert myparam in (1, 2)
    assert request.function.__name__ == "test_func_kw"


# pytest fails here with:
# > function already takes an argument 'func' with a default value
@pytest.mark.parametrize('func', [1, 2])
def test_func_kw2(request, func="foo"):
    pass
```
What combination am I missing?

davidszotten:
for me,  the `partial` implementation fails on your second test case (with `--trace`, raising `TypeError: runcall() got multiple values for argument 'func'`)   (it passes without `--trace`)

davidszotten:
the signature of the original `runcall` is `def runcall(self, func, *args, **kwds):` which means it can't handle `kwds` containing `func` (as my `pdb.runcall` example shows)

blueyed:
It's fixed here: https://github.com/python/cpython/commit/a37f356de19828241bf19129f804369794c72ed3#diff-5c870dcafc8a80ef16e789bb8e901de0.
But yeah, needs wrapping then.
You could use the same method though, instead of copying the internals / function body.

blueyed:
But might be ok to just bail out for "func" on TypeErrors (like befored).
I.e. it will be lean, and will support "func" with 3.7.4+ automatically.

davidszotten:
not sure i understand what you mean by 

> You could use the same method though, instead of copying the internals / function body

could you explain please?

davidszotten:
@RonnyPfannschmidt do you have a preference between a more complicated implementation to support params named `func` on py < 3.7.4 vs just bailing out as we do now?

RonnyPfannschmidt:
I'm on mobile, the trick is to make a new function that shares the original signature and invokes run all with a partial of the original function and it's arguments as the sole parameters

RonnyPfannschmidt:
This is why I am talking about creating a real wrapper that passes only a func into runcall

davidszotten:
Oh I see what you mean. But won’t that mean pdb will start in the wrapper? That might be confusing

RonnyPfannschmidt:
No, runcall would be passed a partial as func on test call

davidszotten:
> No, runcall would be passed a partial as func on test call

ok now i'm confused again. could you send some (pseudo/)code?

davidszotten:
ok think i finally understand. pushed an update. way better, thanks!

blueyed:
Thanks! :)

Issue #6097, opened 2019-10-28
terminal: indicate current outcome/status with color of percentage indicator


blueyed:
It would be nice to use different colors for the "[XXX%]" indicator, according to if there are failed tests already especially.
Basically it could have the color that would be used with the final (new) multi-colored line's main color.

MarcoGorelli:
This would be a nice feature! I'd be happy to work on it if nobody else has taken it on

blueyed:
Fixed in features.

As a follow-up I think this would be nice:

- [ ] bold for when the status / number of errors changed in the current line
- [ ] a count for new (and updated totals) in/from the current line, maybe only with a new "console_output_style" option, but it seems to be useful with both existing modes (percent and counts)

nicoddemus:
Closing this for now, given that the original issue has been fixed.

PR #6107, opened 2019-10-30
Color percentage indicator

MarcoGorelli:
Fixes https://github.com/pytest-dev/pytest/issues/6097.

### DIFF
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -21,6 +21,10 @@ from _pytest.terminal import getreportopt
 from _pytest.terminal import TerminalReporter
 
 DistInfo = collections.namedtuple("DistInfo", ["project_name", "version"])
+RED = r"\x1b\[31m"
+GREEN = r"\x1b\[32m"
+YELLOW = r"\x1b\[33m"
+RESET = r"\x1b\[0m"
 
 
 class Option:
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
I'm aware I'm supposed to write tests before writing code, I just haven't (yet) figured out how to test this, sorry

blueyed:
No worries, if you're lucky you can adjust an existing one.
(I've edited the original comment, to add a reference, and remove the boilerplate)
Thanks for working on this! :)

blueyed:
While working on this we should keep in mind to use it also for the "SKIPPED", "FAILED", "XFAIL", etc prefixes in the short test summary.

blueyed:
CI passes (except for the flaky hypothesis issue), so you need a new test.
I suggest using one based on `monkeypatch.setenv("PY_COLORS", "1")`, so that color output is forced.

MarcoGorelli:
Thanks, I was wondering about the hypothesis issue.

OK, I'll give that a go and will let you know

testing/test_terminal.py
```diff
@@ -1835,13 +1835,13 @@ def test_collecterror(testdir):
...
             "*= ERRORS =*",
             "*_ ERROR collecting test_collecterror.py _*",
             "E   SyntaxError: *",
             "*= short test summary info =*",
             "ERROR test_collecterror.py",
-            "*! Interrupted: 1 errors during collection !*",
+            "*! Interrupted: 1 error during collection !*",
             "*= 1 error in *",
         ]
     )
```
blueyed: Yes, bad/outdated merge.
Cherry-picked into features in 9303de8.

MarcoGorelli:
I've added tests with `monkeypatch.setenv("PY_COLORS", "1")`, is this what you had in mind?

blueyed:
> I've added tests with `monkeypatch.setenv("PY_COLORS", "1")`, is this what you had in mind?

Yes.  Although I think we should them more readable (via format strings for color escapes), so that it would use `"{red}"` there then.
Also I am not sure if the base/default test is really needed, and if it would be faster to have them all in a single run (have not looked at what the "many files" fixture really does).
Looking forward to testing this out later..!

MarcoGorelli:
Ok, I've removed two of the tests and have kept the one that has all three colors in it (green, red, yellow). I've also added constants to make the tests more readable.

Just realised that pytest-CI checks for Python 3.5 to work too, so no f-strings....I'll change this later today

blueyed:
Nice!  You also need a changelog then (see `changelog/README.rst`).

MarcoGorelli:
Thanks! Changelog added

blueyed:
Thanks, will review it properly later - you could however squash and force-push it already (sorry, I'd like we could just do it here (#4361)).

blueyed:
Awesome!

PR #6108, opened 2019-10-31
doctest: lazily import bdb

blueyed:
Follow-up to 63d517645.

### DIFF
--- a/src/_pytest/doctest.py
+++ b/src/_pytest/doctest.py
@@ -1,5 +1,4 @@
 """ discover and run doctests in modules and test files."""
-import bdb
 import inspect
 import platform
 import sys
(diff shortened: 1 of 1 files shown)


The-Compiler:
Why? Is there a downside to importing it globally?

blueyed:
A bit faster I assume?

The-Compiler:
The same could be said about all the imports in all files - I don't really see the point of deviating from what's being done everywhere else.

blueyed:
Sure, but that showed in the diff between master and features (recently added, I've linked the commit).  Therefore I've thought it would be better to be faster.
Feel free to closeit, I have no strong opinion about it.

blueyed:
Closing - please speak up if you care for quality.

PR #6118, opened 2019-11-01
addini: allow for default=None, use notset sentinel

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/6117.


blueyed:
As per https://github.com/pytest-dev/pytest/issues/6117 this might change behavior, closing for now.

nicoddemus:
TBH I think we should reopen and go for this in the next minor release... I don't think it will cause too much breakage, and the new behavior is much more sane and seems to me is fixing an annoying wart that was included as an oversight.

If people are really concerned, I still would like to see this in 6.0 at least, but I'm not sure how to properly warn on this given that the problem is the return value.

cc @RonnyPfannschmidt?

PR #6119, opened 2019-11-02
FSCollector: keep/use given fspath

blueyed:
Via https://github.com/blueyed/pytest/pull/42.

### DIFF
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -364,8 +364,9 @@ def _check_initialpaths_for_relpath(session, fspath):
 
 
 class FSCollector(Collector):
-    def __init__(self, fspath, parent=None, config=None, session=None, nodeid=None):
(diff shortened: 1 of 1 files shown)


src/_pytest/nodes.py
```diff
@@ -364,8 +364,9 @@ def _check_initialpaths_for_relpath(session, fspath):
 
 
 class FSCollector(Collector):
-    def __init__(self, fspath, parent=None, config=None, session=None, nodeid=None):
-        fspath = py.path.local(fspath)  # xxx only for test_resultlog.py?
+    def __init__(
+        self, fspath: py.path.local, parent=None, config=None, session=None, nodeid=None
```
blueyed: @bluetech 
Here the `fspath: py.path.local` slipped in - JFI (via #6129 - where I've removed it somewhere else after your review).
However, it might be good for documentation, and/or if we should ever add typing info for it.

PR #6120, opened 2019-11-02
Session: collect: keep/use already parsed initialpart

blueyed:
Via https://github.com/blueyed/pytest/pull/42.


src/_pytest/main.py
```diff
@@ -483,22 +483,23 @@ def _perform_collect(self, args, genitems):
 
     def collect(self):
         for initialpart in self._initialparts:
-            arg = "::".join(map(str, initialpart))
+            arg = initialpart
```
nicoddemus: I think you can get rid of `arg` as well. 👍

PR #6122, opened 2019-11-03
Add type annotations to _pytest._io.saferepr and _pytest.assertion.util._diff_text

bluetech:
Annotate a couple more leaf modules.

While annotating, I found some dead code in the `_diff_text` function (see second commit). I included it in this PR since the next commit depends on it.


blueyed:
`features` is fixes, in case you want to rebase it.

bluetech:
Thanks @blueyed - rebased.

PR #6129, opened 2019-11-04
typing around Node.location, reportinfo, repr_excinfo etc

blueyed:


src/_pytest/_code/code.py
```diff
@@ -818,19 +818,19 @@ def _truncate_recursive_traceback(self, traceback):
 
         return traceback, extraline
 
-    def repr_excinfo(self, excinfo):
-
+    def repr_excinfo(self, excinfo: Optional[ExceptionInfo]) -> "ExceptionChainRepr":
```
blueyed: `Optional` only since it gets re-assigned in the function - did not want to rename it everywhere.. should I do that instead, or is there a better pattern?

src/_pytest/_code/code.py
```diff
@@ -549,7 +549,7 @@ def getrepr(
         funcargs: bool = False,
         truncate_locals: bool = True,
         chain: bool = True,
-    ):
+    ) -> Union["ReprExceptionInfo", "ExceptionChainRepr"]:
```
bluetech: Functions returning a Union are not great, because the caller must disambiguate which one they got.

In this case, both types share a common base class `ExceptionRepr`. If all the caller cares about is getting an `ExceptionRepr` from this function, I suggest using that instead. This will also be more future proof.

If however the distinction is important (not in this case AFAIU), I think it will be better to either use two overloads on `style`, `Literal["long"]` which only returns `ExceptionChainRepr` and  `Literal["native"]` which only returns `ReprExceptionInfo`. Or just split the function to two.

src/_pytest/_code/code.py
bluetech: Yeah that's an annoying situation.

I'd prefer not to change the function type to something incorrect because of an internal detail. So I'd do the rename, but you said you don't like it. The alternative is to set `allow_redefinition = True` mypy option and then you're allowed to add `excinfo = excinfo  # type: Optional[ExceptionInfo]`. It's nice to be able to do that but risk is that it can mask mistakes.

src/_pytest/_code/code.py
```diff
@@ -876,6 +876,9 @@ def __str__(self):
     def __repr__(self):
         return "<{} instance at {:0x}>".format(self.__class__, id(self))
 
+    def toterminal(self, tw):
```
bluetech: `tw` can't be typed because it's in `py`, but can add `-> None` here (and in all the subclasses if you feel like it).

src/_pytest/main.py
```diff
@@ -357,7 +359,7 @@ class Failed(Exception):
 
 @attr.s
 class _bestrelpath_cache(dict):
-    path = attr.ib()
+    path = attr.ib(type=py.path.local)
```
bluetech: Hmm is this accepted? Does mypy understand `py.path.local` to be a proper type? (I'm asking because I remember I had some issue with that).

src/_pytest/main.py
```diff
@@ -368,6 +370,7 @@ def __missing__(self, path):
 class Session(nodes.FSCollector):
     Interrupted = Interrupted
     Failed = Failed
+    _setupstate = None  # type: SetupState
```
bluetech: If it's type is `SetupState` how can it be initialized to `None`? I guess what we want is `_setupstate: SetupState` but that's Python>=3.6 only. So it's good for now.

src/_pytest/main.py
```diff
@@ -383,7 +386,9 @@ def __init__(self, config):
         self._initialpaths = frozenset()
         # Keep track of any collected nodes in here, so we don't duplicate fixtures
         self._node_cache = {}
-        self._bestrelpathcache = _bestrelpath_cache(config.rootdir)
+        self._bestrelpathcache = _bestrelpath_cache(
```
bluetech: Not sure if that's even possible, but does changing `_bestrelpath_cache` to inherit from `Dict[str, str]` instead of `dict` work? If so, then this could be avoided I think.

src/_pytest/main.py
```diff
@@ -398,7 +403,7 @@ def __repr__(self):
             self.testscollected,
         )
 
-    def _node_location_to_relpath(self, node_path):
+    def _node_location_to_relpath(self, node_path) -> str:
```
bluetech: `node_path: str` ?

src/_pytest/nodes.py
```diff
@@ -81,7 +88,7 @@ def __init__(
         self.config = config or parent.config
 
         #: the session this node is part of
-        self.session = session or parent.session
+        self.session = session or parent.session  # type: Session
```
bluetech: Would be better to type the `__init__` argument instead. Since it's `Optional[Session]`, the following might be needed, not sure:

```py
if session is None:
    assert parent.session is not None
    self.session = parent.session
else:
   self.session = session
```

src/_pytest/nodes.py
```diff
@@ -254,13 +261,13 @@ def getparent(self, cls):
     def _prunetraceback(self, excinfo):
         pass
 
-    def _repr_failure_py(self, excinfo, style=None):
-        # Type ignored: see comment where fail.Exception is defined.
-        if excinfo.errisinstance(fail.Exception):  # type: ignore
+    def _repr_failure_py(
+        self, excinfo: ExceptionInfo[Union[Failed, "FixtureLookupError"]], style=None
+    ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, "FixtureLookupErrorRepr"]:
+        if isinstance(excinfo.value, Failed):
```
bluetech: :+1:

src/_pytest/nodes.py
```diff
@@ -254,13 +261,13 @@ def getparent(self, cls):
     def _prunetraceback(self, excinfo):
         pass
 
-    def _repr_failure_py(self, excinfo, style=None):
-        # Type ignored: see comment where fail.Exception is defined.
-        if excinfo.errisinstance(fail.Exception):  # type: ignore
+    def _repr_failure_py(
+        self, excinfo: ExceptionInfo[Union[Failed, "FixtureLookupError"]], style=None
```
bluetech: Are we sure only `Failed` and `FixtureLookupError` are possible here?

src/_pytest/nodes.py
```diff
@@ -254,13 +261,13 @@ def getparent(self, cls):
...
+    ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, "FixtureLookupErrorRepr"]:
```
bluetech: Same comment here re. returning unions -- if there's a way to simplify, would be better. Haven't checked this one, though.

src/_pytest/nodes.py
```diff
@@ -425,16 +434,19 @@ def add_report_section(self, when: str, key: str, content: str) -> None:
...
-    def reportinfo(self):
+    def reportinfo(self) -> Tuple[str, Optional[int], str]:
         return self.fspath, None, ""
 
     @property
-    def location(self):
+    def location(self) -> Tuple[str, Optional[int], str]:
         try:
             return self._location
         except AttributeError:
```
bluetech: [`@cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) would be nice for this idiom. Maybe we should start using it (with a compat implementation).

src/_pytest/nodes.py
```diff
@@ -425,16 +434,19 @@ def add_report_section(self, when: str, key: str, content: str) -> None:
...
             location = self.reportinfo()
             fspath = self.session._node_location_to_relpath(location[0])
-            location = (fspath, location[1], str(location[2]))
-            self._location = location
-            return location
+            self._location = (
+                fspath,
+                location[1],
+                str(location[2]),
```
bluetech: `self.reportinfo` type says `location[2]` is an `str`. So either its inaccurate, or the `str()` here is redundant.

src/_pytest/python.py
```diff
@@ -288,7 +289,7 @@ def getmodpath(self, stopatmodule=True, includemodule=False):
         s = ".".join(parts)
         return s.replace(".[", "[")
 
-    def reportinfo(self):
+    def reportinfo(self) -> Tuple[str, int, str]:
```
bluetech: Not sure if this `reportinfo` and `Item`'s `reportinfo` are related, but their types are slightly different (`int` vs. `Optional[int]`).

src/_pytest/main.py
blueyed: Yeah..

src/_pytest/main.py
blueyed: Appears to be "Any" - will remove it.

src/_pytest/nodes.py
blueyed: That are the ones I've found - can be adjusted if needed?

src/_pytest/nodes.py
blueyed: Thought the same multiple times already.

src/_pytest/nodes.py
blueyed: will harden it with an assert.

src/_pytest/python.py
blueyed: They are related.. the base allows for None, but here it is always int.
Should be ok, no?

src/_pytest/_code/code.py
blueyed: ok, will rename it internally then to `excinfo_` or similar (for the loop).

src/_pytest/_code/code.py
blueyed: Thanks for the input.
I've tried to be as strict / verbose as possible, given the code around all this.
Will consider using `ExceptionRepr` then.

src/_pytest/nodes.py
blueyed: Seems to work with just "session: Session".

blueyed: Done.

src/_pytest/nodes.py
blueyed: => https://github.com/pytest-dev/pytest/issues/6131

src/_pytest/_code/code.py
blueyed: Done.

src/_pytest/main.py
blueyed: Does not work for https://github.com/blueyed/pytest/blob/1b3c0e6d5c078dadd11dfb2a109dc330742ac75c/src/_pytest/main.py#L408.

blueyed: I think _bestrelpath_cache should be dropped, instead of being fixed further..
(using generic caching around pathlib likely)

bluetech: It does not work because `__init__` is untyped. Adding `-> None` to `__init__` would be sufficient to get it working, I think. Although you would then need to type the couple empty collections `self._initialpaths` and `self._node_cache`, if you know what their types are supposed to be.

src/_pytest/nodes.py
bluetech: It depends on what the intention of the function is - to accept only these two types, or to accept any `ExceptionInfo[E]`.

src/_pytest/python.py
bluetech: Yes, a subclass can be more specific in a return type.

src/_pytest/fixtures.py
```diff
@@ -33,6 +33,7 @@
 from _pytest.outcomes import TEST_OUTCOME
 
 if False:  # TYPE_CHECKING
+    from _pytest import nodes
```
bluetech: Type annotations reveal many "semantic" import cycles. I think we should try to improve the situation some time!

BTW, for good measure, would move this import below the `from typing import Type` one.

src/_pytest/nodes.py
```diff
@@ -69,7 +76,13 @@ class Node:
...
     def __init__(
-        self, name, parent=None, config=None, session=None, fspath=None, nodeid=None
+        self,
+        name,
+        parent=None,
+        config=None,
+        session: Optional["Session"] = None,
+        fspath=None,
+        nodeid=None,
     ):
```
bluetech: Didn't mention before, would be good to add a `-> None` here.

src/_pytest/reports.py
```diff
@@ -304,7 +309,7 @@ def __init__(self, nodeid, outcome, longrepr, result, sections=(), **extra):
         self.nodeid = nodeid
         self.outcome = outcome
         self.longrepr = longrepr
-        self.result = result or []
+        self.result = result or []  # type: List
```
bluetech: Do you what type of list it is?

src/_pytest/main.py
blueyed: I'd rather leave it as-is then.

src/_pytest/nodes.py
blueyed: My intention is to being made aware if other types are used.
This should aid with refactoring / cleanup - my main motivation to add types here.

src/_pytest/reports.py
blueyed: Via CallInfo.from_call usually: https://github.com/blueyed/pytest/blob/1b3c0e6d5c078dadd11dfb2a109dc330742ac75c/src/_pytest/runner.py#L210-L236

bluetech: So can this be `List[CallInfo]`?

(btw, it's perfectly fine if you left it out because it's too much work to check, I'm just commenting in case you forgot to fill it in).

src/_pytest/_code/code.py
bluetech: Not sure if you decided to keep it this way.

Other than this, LGTM.

blueyed:
Updated (987f912), then squashed (741f0fe).
Please approve it if it loooks ok, so I can merge it then.

PR #6136, opened 2019-11-05
Fix rm_rf tests

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5974.
Via https://github.com/pytest-dev/pytest/issues/5974#issuecomment-549822509.


blueyed:
> we would still fail for other code

We won't fail really, but only emit a warning - I think that's fair/good.

> please add a changelog entry

I do not think we need a changelog that state we fixed our own tests really, do we?

PR #5924, opened 2019-10-07
Improve full diff output for lists

blueyed:
Massage text input for difflib when comparing pformat output of
different line lengths.

Also do not strip ndiff output on the left, which currently already
removes indenting for lines with no differences.

Before:

    E   AssertionError: assert ['version', '...version_info'] == ['version', '...version', ...]
    E     Right contains 3 more items, first extra item: ' '
    E     Full diff:
    E     - ['version', 'version_info', 'sys.version', 'sys.version_info']
    E     + ['version',
    E     +  'version_info',
    E     +  'sys.version',
    E     +  'sys.version_info',
    E     +  ' ',
    E     +  'sys.version',
    E     +  'sys.version_info']

After:

    E   AssertionError: assert ['version', '...version_info'] == ['version', '...version', ...]
    E     Right contains 3 more items, first extra item: ' '
    E     Full diff:
    E       ['version',
    E        'version_info',
    E        'sys.version',
    E     +  'sys.version_info',
    E     +  ' ',
    E     +  'sys.version',
    E        'sys.version_info'
    E       ]

TODO:

- [x] changelog
- [x] "improve" / remove `asserts`
- [x] more tests (?)

PR #6137, opened 2019-11-06
assert: fix _compare_eq_iterable: re-format both sides

blueyed:
Follow-up to 946434c61 (#5924).

Before this patch the test would look like this:

    {'env': {'sub...s wrapped'}}}} == {'env': {'sub...}}}, 'new': 1}
    Omitting 1 identical items, use -vv to show
    Right contains 1 more item:
    {'new': 1}
    Full diff:
      {
       'env': {'sub': {'long_a': 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
    -                  'sub1': {'long_a': 'substring '
    +                  'sub1': {'long_a': 'substring that gets wrapped'}}},
    ?                                                +++++++++++++++++ ++++
    +  'new': 1,
    -                                     'that '
    -                                     'gets '
    -                                     'wrapped'}}},
      }

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -273,12 +273,13 @@ def _compare_eq_iterable(
     if lines_left != lines_right:
         if lines_left > lines_right:
             max_width = min(len(x) for x in left_formatting)
-            right_formatting = pprint.pformat(right, width=max_width).splitlines()
-            lines_right = len(right_formatting)
         else:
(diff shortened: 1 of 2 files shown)


blueyed:
No changelog, since it has not been released.

PR #6140, opened 2019-11-06
Explicitly implement pytest_assertrepr_compare in assertion plugin

nicoddemus:
Previously it was an alias, which makes it unnecessary hard to find
all implementations (either by IDE or using a simple search).

### DIFF
--- a/src/_pytest/assertion/__init__.py
+++ b/src/_pytest/assertion/__init__.py
@@ -163,5 +163,5 @@ def pytest_sessionfinish(session):
             assertstate.hook.set_session(None)
 
 
-# Expose this plugin's implementation for the pytest_assertrepr_compare hook
-pytest_assertrepr_compare = util.assertrepr_compare
+def pytest_assertrepr_compare(config, op, left, right):
+    return util.assertrepr_compare(config=config, op=op, left=left, right=right)


src/_pytest/assertion/__init__.py
```diff
@@ -164,4 +164,5 @@ def pytest_sessionfinish(session):
 
 
 # Expose this plugin's implementation for the pytest_assertrepr_compare hook
```
blueyed: Doesn't need the comment then anymore.

nicoddemus:
> Why not move util.assertrepr_compare to it?

I suppose you mean move `util.assertrepr_compare`'s implementation to here right? Doing so would introduce unnecessary conflicts with other PRs which might be changing `util.assertrepr_compare`, so decided to keep it small.

blueyed:
> I suppose you mean move `util.assertrepr_compare`'s implementation to here right?

Would only be from me I guess, if any?
And conflicts in this regard would be trivial, since it is only moved.

nicoddemus:
Agree it would be trivial, but then I would also have to move a bunch of imports too (all simple to do of course), but I want to make this a really "small nobrainer patch". 😁 

Thanks for the suggestion though!

PR #6141, opened 2019-11-06
Add type annotations to _pytest.{warning_types,_code.source,pytester}

bluetech:
Some more annotations. `pytester` doesn't make sense in terms of progression, but I got to it somehow (only did about a half).

### DIFF
--- a/src/_pytest/_code/source.py
+++ b/src/_pytest/_code/source.py
@@ -7,9 +7,16 @@ import tokenize
 import warnings
 from ast import PyCF_ONLY_AST as _AST_FLAG
 from bisect import bisect_right
+from types import FrameType
 from typing import List
+from typing import Optional
+from typing import Sequence
+from typing import Tuple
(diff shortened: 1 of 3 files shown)


src/_pytest/_code/source.py
```diff
@@ -48,7 +55,15 @@ def __eq__(self, other):
...
     __hash__ = None  # type: ignore
 
-    def __getitem__(self, key):
+    @overload
+    def __getitem__(self, key: int) -> str:
+        ...
+
+    @overload  # noqa: F811
+    def __getitem__(self, key: slice) -> "Source":
+        ...
```
blueyed: With regard to coverage you could use `raise NotImplementedError()` here (ignored via config), or we could also add "..." to be ignored by default.

bluetech: Thanks, changed to `raise NotImplementedError()`.

PR #6145, opened 2019-11-07
tests: ids

blueyed:


bluetech:
The `id` argument of `pytest.param` is well hidden in the docs. It's nicer than using `pytest.mark.parametrize(ids=[...])`.

PR #6150, opened 2019-11-07
Two minor tweaks in MarkDecorator's implementation

bluetech:
Please see the commit messages.

### DIFF
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -2,7 +2,6 @@ import inspect
 import warnings
 from collections import namedtuple
 from collections.abc import MutableMapping
-from operator import attrgetter
 from typing import Set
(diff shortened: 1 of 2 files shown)


src/_pytest/mark/structures.py
```diff
@@ -205,17 +194,25 @@ def test_function():
...
-    kwargs = alias("mark.kwargs")
+    @property
+    def name(self):
+        """alias for mark.name"""
+        return self.mark.name
+
+    @property
+    def args(self):
+        """alias for mark.args"""
+        return self.mark.args
```
blueyed: This appears to be unused also, at least not covered.

bluetech: I added a test for the aliases.

PR #6151, opened 2019-11-08
_compare_eq_iterable: use AlwaysDispatchingPrettyPrinter

blueyed:
This fixes/removes the previous hack of re-trying with minimum width,
which fails short when it splits strings.

This inherits from `pprint.PrettyPrinter` to override `_format` in a
minimal way to always dispatch, regardless of the given width.

Code ref: https://github.com/python/cpython/blob/5c0c325453a175350e3c18ebb10cc10c37f9595c/Lib/pprint.py#L170-L178

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -28,6 +28,27 @@ _reprcompare = None  # type: Optional[Callable[[str, object, object], Optional[s
 _assertion_pass = None  # type: Optional[Callable[[int, str, str], None]]
 
 
+class AlwaysDispatchingPrettyPrinter(pprint.PrettyPrinter):
+    """PrettyPrinter that always dispatches (regardless of width)."""
+
+    def _format(self, object, stream, indent, allowance, context, level):
+        p = self._dispatch.get(type(object).__repr__, None)
+
+        objid = id(object)
+        if objid in context or p is None:
+            return super()._format(object, stream, indent, allowance, context, level)
+
+        context[objid] = 1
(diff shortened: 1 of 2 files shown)


blueyed:
No changelog, since not released.
I'm not too happy with it - feel free to reject, but then the previous hack should also be reverted.

nicoddemus:
> Code ref: https://github.com/python/cpython/blob/5c0c325453a175350e3c18ebb10cc10c37f9595c/Lib/pprint.py#L170-L178

Thanks for the reference, but (crazy idea sorry, very late now) what if we overwrite `_format`, like you did, but just call `super()` with `allowance=-1e999` (a huge negative number)? This would make `max_width` always a very large number so it would never enter the `if len(rep) > max_width` condition. 

> I'm not too happy with it - feel free to reject

I find it more elegant subclassing, but would like to avoid duplicating internal code as much as possible, hence the crazy suggestion above.

> but then the previous hack should also be reverted.

Why? Although I see it does not work as well for strings, it is still a huge improvement over what we had before, IMHO.

blueyed:
@nicoddemus 
Your idea might work, but only when used with the second call still.
But I can imagine / think that it would actually not wrap long string then correctly.
Feel free to try it based on this, but from my side I'd rather keep it like this.

> Why? Although I see it does not work as well for strings, it is still a huge improvement over what we had before, IMHO.

Yeah, but this is the 2nd fix to it already, and I do not really like causing regressions / annoyances (as you might know ;)).

nicoddemus:
I see, let's go ahead with it then

PR #6152, opened 2019-11-08
use __name__ attribute in the parametrize id for modules as well

grlee77:
`pytest.mark.parametrize` will currently use the name of a function or class as the parameter "id" by default. This PR extends this behavior to modules (or other objects defining `__name__`) as well.

An alternative implementation would have been to add an (`or inspect.ismodule(var)`) to the existing case, but I thought perhaps there may be other objects with `__name__` as well that could benefit from this behavior.

A concrete use case where I wanted these was for some tests that parametrize using different numerical array modules with matching API (NumPy and CuPy). It is preferable for the name of the module to show up in the id rather than "module0" or similar.

- [ x ] Target the `features` branch for new features, improvements, and removals/deprecations.
- [ x ] Include new tests or update existing tests when applicable.

- [ x ] Add yourself to `AUTHORS` in alphabetical order;

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -103,6 +103,7 @@ George Kussumoto
 Georgy Dyuldin
 Graham Horler
 Greg Price
+Gregory Lee
 Grig Gheorghiu
 Grigorii Eremeev (budulianin)
 Guido Wesdorp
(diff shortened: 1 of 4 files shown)


grlee77:
> The only thing missing was a CHANGELOG entry, but I took the liberty of including one myself.

Great, thanks for the quick review!

PR #6155, opened 2019-11-08
doctests: use real path instead of symlink path

### LINKED ISSUE #6146: Symlinks break conftest.py imports in the doctest plugin
The `conftest.py` import logic in the doctest plugin doesn't handle some layouts that contain multiple levels of symlinks (such as when used with bazel).

When the doctest plugin is disabled, `pytest` runs fine but when I enable doctests with `--doctest-modules` I get the following error:

```
____________________ ERROR collecting test/rtm/conftest.py _____________________
../pypi__pytest__4_6_4/_pytest/config/__init__.py:440: in _importconftest
    return self._conftestpath2mod[conftestpath]
E   KeyError: local('/private/var/tmp/_bazel_sitatkif/715d015e3d448dfcaced93a8d1b1a5f5/execroot/__main__/bazel-out/darwin-fastbuild/bin/test/test.runfiles/__main__/test/rtm/conftest.py')

../pypi__pytest__4_6_4/_pytest/runner.py:226: in from_call
    result = func()
../pypi__pytest__4_6_4/_pytest/runner.py:253: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
../pypi__pytest__4_6_4/_pytest/doctest.py:421: in collect
    module = self.config.pluginmanager._importconftest(self.fspath)
../pypi__pytest__4_6_4/_pytest/config/__init__.py:471: in _importconftest
    assert mod not in mods
E   AssertionError
```

Upon investigation, this is due to the fact that the mods dict contains the `realpath` of my `conftest.py` but the doctest plugin is trying to load it again using a symlinked path.

I noticed that core `pytest` uses the realpath to load the project's conftest.py module(s):

https://github.com/pytest-dev/pytest/blob/d3549df5b94137453b832345a9b8074f518731b0/src/_pytest/config/__init__.py#L424

...while the doctest plugin does not resolve the realpath:

https://github.com/pytest-dev/pytest/blob/d3549df5b94137453b832345a9b8074f518731b0/src/_pytest/doctest.py#L421

Changing the doctest line above with the code below solves the problem for me:

```python
module = self.config.pluginmanager._importconftest(self.fspath.realpath())
```

(note: the code on the master branch changed a bit but the behavior is the same. See https://github.com/pytest-dev/pytest/blob/85288b53218d4e0037dce9c34bd37f86fa9e212d/src/_pytest/config/__init__.py#L424-L432 for core and https://github.com/pytest-dev/pytest/blob/85288b53218d4e0037dce9c34bd37f86fa9e212d/src/_pytest/doctest.py#L414 for doctest).

JonathanGrant:
Avoids symlink problems. Closes #6146

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -470,8 +470,10 @@ class PytestPluginManager(PluginManager):
             if dirpath in self._dirpath2confmods:
                 for path, mods in self._dirpath2confmods.items():
                     if path and path.relto(dirpath) or path == dirpath:
-                        assert mod not in mods
-                        mods.append(mod)
+                        if mod in mods:
+                            assert Path(str(path)).resolve() == key, (path, key)
+                        else:
+                            mods.append(mod)
             self.trace("loaded conftestmodule %r" % (mod))
             self.consider_conftest(mod)
             return mod


blueyed:
Thanks!
Note that `_importconftest` resolves it already for the key (https://github.com/pytest-dev/pytest/blob/27b1773b40db5e1eca4889417d928bb56af494bf/src/_pytest/config/__init__.py#L458), so maybe that could be put to use?
Also the assertion should be fixed to use the realpath?  (https://github.com/pytest-dev/pytest/blob/27b1773b40db5e1eca4889417d928bb56af494bf/src/_pytest/config/__init__.py#L482)
Anyway, a test for this would be needed.

blueyed:
If you have a local test case you could try if this "fixes" it also (i.e. if the assertion holds then):
```
index 8b9af2bd3..44bc6ac05 100644
@@ -479,8 +479,10 @@ def _importconftest(self, conftestpath):
             if dirpath in self._dirpath2confmods:
                 for path, mods in self._dirpath2confmods.items():
                     if path and path.relto(dirpath) or path == dirpath:
-                        assert mod not in mods
-                        mods.append(mod)
+                        if mod in mods:
+                            assert Path(str(path)).resolve() == key, (path, key)
+                        else:
+                            mods.append(mod)
             self.trace("loaded conftestmodule %r" % (mod))
             self.consider_conftest(mod)
             return mod
```

I am not saying your fix is not correct, just that we should be careful since the collection mechanism is "fragile".

JonathanGrant:
Sorry I don't understand, what is the purpose of `_dirpath2confmods`?

blueyed:
A cache / and or to keep them unique I guess?

blueyed:
Do you have a test case, preferably in the form of a pytester/testdir based one?  (would make things easier to look into then from there)
FWIW I've tested the other assertion in https://github.com/blueyed/pytest/pull/97/commits/86a7de919896c1a191da73dc430d19ba8c78f6a1, which passed with current tests.

JonathanGrant:
I don't have a test, and I am new to this repo (as well as pytest for txt files). How can I add a test?

JonathanGrant:
Would this place be the right place to put such a test? https://github.com/pytest-dev/pytest/blob/master/testing/test_conftest.py

JonathanGrant:
Are the coverage percentages changing because I am not up to date with master branch?

nicoddemus:
Hi @JonathanGrant,

I think you can implement something similar to:

https://github.com/pytest-dev/pytest/blob/85288b53218d4e0037dce9c34bd37f86fa9e212d/testing/test_conftest.py#L109-L130

Where instead of creating a new conftest, you would make a symlink to the first one, and then when `_importconftest` gets called, you would ensure the same conftest module would be returned from the first call.

blueyed:
According to https://github.com/pytest-dev/pytest/issues/6146#issuecomment-559380843 it might be not necessary anymore / is fixed already in 5.1.2.

blueyed:
Closing as per previous comment.

PR #6157, opened 2019-11-08
[RFC] typing around terminal

blueyed:


src/_pytest/terminal.py
```diff
@@ -233,29 +238,30 @@ def get_location(self, config):
...
         import _pytest.config
 
         self.config = config
         self._numcollected = 0
-        self._session = None
+        self._session = None  # type: Optional[Session]
         self._showfspath = None
 
-        self.stats = {}
+        # XXX: needs TypedDict?  warnings => WarningReport, others => BaseReport
```
blueyed: Ref: https://www.python.org/dev/peps/pep-0589/

blueyed:
@bluetech any input here, please?

src/_pytest/reports.py
```diff
@@ -223,7 +223,7 @@ def __init__(
         #: a (filesystempath, lineno, domaininfo) tuple indicating the
         #: actual location of a test item - it might be different from the
         #: collected one e.g. if a method is inherited from a different module.
-        self.location = location
+        self.location = location  # type: Tuple[str, Optional[int], str]
```
bluetech: I *think* that adding an annotation to a function whose own signature is unannotated doesn't have an effect ([ref](https://github.com/python/mypy/issues/3948)).

So it is better to add this annotation to the `location` parameter instead, if possible. This will make mypy consider the function to be typed.

src/_pytest/terminal.py
```diff
@@ -514,7 +523,7 @@ def pytest_collection(self):
         elif self.config.option.verbose >= 1:
             self.write("collecting ... ", bold=True)
 
-    def pytest_collectreport(self, report):
+    def pytest_collectreport(self, report: CollectReport):
```
bluetech: `-> None`

src/_pytest/terminal.py
```diff
@@ -767,6 +778,7 @@ def getreports(self, name):
 
     def summary_warnings(self):
         if self.hasopt("w"):
+            # XXX: this needs to be fixed
```
bluetech: Not sure if you meant to keep this here or not.

src/_pytest/terminal.py
```diff
@@ -233,29 +238,30 @@ def get_location(self, config):
 
 
 class TerminalReporter:
-    def __init__(self, config, file=None):
+    def __init__(self, config: Config, file=None):
```
bluetech: `-> None`

src/_pytest/terminal.py
```diff
@@ -440,7 +446,7 @@ def pytest_runtest_logreport(self, report):
                     self._write_progress_information_filling_space()
             else:
                 self.ensure_newline()
-                self._tw.write("[%s]" % rep.node.gateway.id)
+                self._tw.write("[%s]" % rep.node.gateway.id)  # type: ignore
```
bluetech: Would be nice to have a short comment on why the `type: ignore` is needed here. This why we'll be able to tell later whether it's a "TODO" or a "workaround mypy" thing.

src/_pytest/terminal.py
bluetech: Can do it, only issue is that unless we want to bring in a *runtime* dependency on `typing_extension`, we'll need to add a compat for it, or always surround it with `if TYPE_CHECKING` and string annotations.

blueyed: I think it is rather out of scope, since it requires py38 (IIRC).
Better options?
Otherwise I will remove the XXX/TODO comments about it.

We could also have `stats_warnings` explicitly for this, but that's not B/C (but could be done via a property or thelike then).
But IIRC I've tried a property based approach (where I would define the returned type), which did not help really.

src/_pytest/terminal.py
blueyed: No.. it shows where the Union of the different report types causes a problem.

src/_pytest/terminal.py
blueyed: It's a workaround-xdist thing (I assume that xdist just puts it there).

src/_pytest/reports.py
blueyed: `reveal_type` gives me `'Tuple[builtins.str, Union[builtins.int, None], builtins.str]'` here (mypy 0.750+dev.84126ab9985844d8cda3809688034af895fdaf7c.dirty).

This appears to come / being enabled via BaseReport - adding it there instead.

src/_pytest/terminal.py
blueyed: Is there a way to have mypy complain about it?

src/_pytest/terminal.py
blueyed: I'll keep it as is (it is somehow obvious from "xdist" in there, and I do not like extra comments for this above).  Also it might get reported when not being used/needed later anymore (there's a mode in mypy to do so).

src/_pytest/reports.py
blueyed: But it is needed to fix "List or tuple expected as variable arguments" with https://github.com/blueyed/pytest/blob/0133097943e0140ab82e07d4d201405145d35e59/src/_pytest/terminal.py#L442.
But it seems strange that with "location: Tuple[str, Optional[int], str]" in `TestReport.__init__` and `rep` being `_pytest.reports.TestReport`, `rep.location` is `Union[Tuple[builtins.str, Union[builtins.int, None], builtins.str], None]` then?

The problem appears to be that BaseReport has `location = None  # type: Optional[Tuple[str, Optional[int], str]]`, and we cannot have `location: Tuple[str, Optional[int], str]` in TestReport (py35).
Adding an assert appears to work also, but then again it is better to keep the type comment there.

PR #6164, opened 2019-11-09
pytester: use no colors with inline runs by default

blueyed:
Sets `PY_COLORS=0` in the environment by default, which is used by pylib.

Via https://github.com/blueyed/pytest/pull/58


blueyed:
Needs more test adjustments (I guess).
Will do after approval, but also close after a few days if there's no interest.

blueyed:
Blocked by https://github.com/pytest-dev/pytest/issues/6213 for now.

PR #6166, opened 2019-11-09
pytester: set PYTEST_DISABLE_PLUGIN_AUTOLOAD=1

blueyed:
Via https://github.com/blueyed/pytest/pull/80

(cherry picked from commit e5fff3fad9a0c4ba0cd009c8c4b31808b074e7cb)

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -539,7 +539,11 @@ class Testdir:
 
         # Environment (updates) for inner runs.
         tmphome = str(self.tmpdir)
-        self._env_run_update = {"HOME": tmphome, "USERPROFILE": tmphome}
+        self._env_run_update = {
+            "HOME": tmphome,
+            "USERPROFILE": tmphome,
+            "PYTEST_DISABLE_PLUGIN_AUTOLOAD": "1",
+        }
 
     def __repr__(self):
         return "<Testdir {!r}>".format(self.tmpdir)
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -104,6 +104,8 @@ class TestGeneralUsage:
 
     @pytest.mark.parametrize("load_cov_early", [True, False])
     def test_early_load_setuptools_name(self, testdir, monkeypatch, load_cov_early):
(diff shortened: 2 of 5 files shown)


blueyed:
Needs https://github.com/blueyed/pytest/pull/78 (and some more likely).
Let me know if you are interested, otherwise I will close it in the next days.

The use case is to not have outer plugins interfere with tests.

nicoddemus:
Thanks.

This will break runninh with xdist and such, no? TBH not sure this is worthwhile, we should delegate having "clean" environments to tox runs. I don't think we need to guard against polluted environments when running outside `tox`. I fear there are many more situations where an unclean environment might interfere with tests, and would be very hard to enforce them all in `testdir`.

Closing then, but thanks anyway. 👍

blueyed:
Yes, IIRC we had that before, and that's why I also started working on an option (moved to https://github.com/blueyed/pytest/pull/71 for now).
IIRC I suggested using the env for what that option does (i.e. `1` means the same as empty, i.e. load none), and you can specify a list there then maybe.  Having both an option and env for the same purpose might be confusing, and it seems like handling it via an option on complicates things.

> tox

Yeah, but using it by itself is slower, and the idea here is that you can actually use other plugins with pytests owns tests failing (e.g. pytest-randomly etc).

nicoddemus:
> Yeah, but using it by itself is slower, and the idea here is that you can actually use other plugins with pytests owns tests failing (e.g. pytest-randomly etc).

That's fine, but I wouldn't like to go out of our way to support that; I know the PR here is simple enough, but I wouldn't like to guarantee support for that as this seems like a slippery slope to me.

blueyed:
Hmm, not sure I understand?
(I've edited s/with/without)
Do you mean not guaranteeing support to use other plugins with pytests own tests?
Yeah, that's pretty much the case already then.. :)

nicoddemus:
> Do you mean not guaranteeing support to use other plugins with pytests own tests?
> Yeah, that's pretty much the case already then.. :)

Exactly.

blueyed:
@nicoddemus 
JFI/for reference: I've moved it to `testing/conftest.py` myself for pytest's own tests only then (https://github.com/blueyed/pytest/pull/134).

PR #6173, opened 2019-11-12
Delineate syntactically that the 'match' argument to 'raises' is keyword-only

### LINKED ISSUE #6158: pytest.raises documention doesn't state upfront that 'match' is keyword only
Reading http://pytest.org/en/latest/reference.html#pytest-raises I thought the 'match' argument supporting positional specificity due to it being listed as `[, match]` in the parameter list and the argument specification not saying anything about it. But if you read a bit farther down you notice that it's mentioned it's keyword-only.

My suspicion is there's no way to change the syntax of that parameter list to use `*,` to make it clearer that `match` is positional-only? Otherwise I'm happy to write a PR to state that in its parameter documentation.

brettcannon:
Closes #6158

### DIFF
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -59,7 +59,7 @@ pytest.raises
 
 **Tutorial**: :ref:`assertraises`.
 
-.. autofunction:: pytest.raises(expected_exception: Exception, [match])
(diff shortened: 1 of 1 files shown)


brettcannon:
@nicoddemus @Zac-HD quite welcome! Just glad I could give back to pytest in some (very small) way. 😄

PR #6174, opened 2019-11-12
parametrized: ids: support generator/iterator

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/759

Via https://github.com/blueyed/pytest/pull/105.

### DIFF
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -2,6 +2,8 @@ import inspect
 import warnings
 from collections import namedtuple
 from collections.abc import MutableMapping
+from typing import List
+from typing import Optional
 from typing import Set
 
 import attr
@@ -144,7 +146,15 @@ class Mark:
     #: keyword arguments of the mark decorator
     kwargs = attr.ib()  # Dict[str, object]
 
-    def combined_with(self, other):
+    #: source Mark for ids with parametrize Marks
+    _param_ids_from = attr.ib(type=Optional["Mark"], default=None, repr=False)
+    #: resolved/generated ids with parametrize Marks
+    _param_ids_generated = attr.ib(type=Optional[List[str]], default=None, repr=False)
+
+    def _has_param_ids(self):
+        return "ids" in self.kwargs or len(self.args) >= 4
+
+    def combined_with(self, other: "Mark") -> "Mark":
         """
         :param other: the mark to combine with
         :type other: Mark
@@ -153,8 +163,20 @@ class Mark:
         combines by appending args and merging the mappings
         """
         assert self.name == other.name
+
+        # Remember source of ids with parametrize Marks.
+        param_ids_from = None  # type: Optional[Mark]
+        if self.name == "parametrize":
+            if other._has_param_ids():
+                param_ids_from = other
+            elif self._has_param_ids():
+                param_ids_from = self
+
         return Mark(
-            self.name, self.args + other.args, dict(self.kwargs, **other.kwargs)
+            self.name,
+            self.args + other.args,
+            dict(self.kwargs, **other.kwargs),
+            param_ids_from=param_ids_from,
         )
 
 
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -10,6 +10,7 @@ from collections.abc import Sequence
 from functools import partial
 from textwrap import dedent
 from typing import List
+from typing import Optional
 from typing import Tuple
 
 import py
@@ -36,6 +37,7 @@ from _pytest.deprecated import FUNCARGNAMES
 from _pytest.main import FSHookProxy
 from _pytest.mark import MARK_GEN
 from _pytest.mark.structures import get_unpacked_marks
+from _pytest.mark.structures import Mark
 from _pytest.mark.structures import normalize_mark_list
 from _pytest.outcomes import fail
 from _pytest.outcomes import skip
@@ -122,7 +124,7 @@ def pytest_cmdline_main(config):
 
 def pytest_generate_tests(metafunc):
     for marker in metafunc.definition.iter_markers(name="parametrize"):
-        metafunc.parametrize(*marker.args, **marker.kwargs)
+        metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker)
 
 
 def pytest_configure(config):
@@ -914,7 +916,16 @@ class Metafunc:
         warnings.warn(FUNCARGNAMES, stacklevel=2)
         return self.fixturenames
 
-    def parametrize(self, argnames, argvalues, indirect=False, ids=None, scope=None):
+    def parametrize(
+        self,
+        argnames,
+        argvalues,
+        indirect=False,
+        ids=None,
+        scope=None,
+        *,
+        _param_mark: Optional[Mark] = None
+    ):
         """ Add new invocations to the underlying test function using the list
         of argvalues for the given argnames.  Parametrization is performed
         during the collection phase.  If you need to setup expensive resources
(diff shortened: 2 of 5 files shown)


blueyed:
Fails on py35 due to random order:
```
E       Failed: nomatch: 'test_parametrize_iterator.py::test1[param0] PASSED'
E           and: '============================= test session starts =============================='
E           and: 'platform linux -- Python 3.5.6, pytest-5.2.3.dev233+gc8053b8d1, py-1.8.0, pluggy-0.13.0 -- /home/travis/build/pytest-dev/pytest/.tox/py35-xdist/bin/python'
E           and: 'cachedir: .pytest_cache'
E           and: "hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/travis/build/pytest-dev/pytest/.hypothesis/examples')"
E           and: 'rootdir: /tmp/pytest-of-travis/pytest-0/popen-gw0/test_parametrize_iterator0'
E           and: 'plugins: hypothesis-4.44.2, forked-1.1.3, xdist-1.30.0'
E           and: 'collecting ... collected 6 items'
E           and: ''
E           and: 'test_parametrize_iterator.py::test1[param2] PASSED'
E           and: 'test_parametrize_iterator.py::test1[param3] PASSED'
E           and: 'test_parametrize_iterator.py::test2[param0] PASSED'
E           and: 'test_parametrize_iterator.py::test2[param1] PASSED'
E           and: 'test_parametrize_iterator.py::test_converted_to_str[1] PASSED'
E           and: 'test_parametrize_iterator.py::test_converted_to_str[2] PASSED'
E           and: ''
E           and: '============================== 6 passed in 0.04s ==============================='
E       remains unmatched: 'test_parametrize_iterator.py::test1[param0] PASSED'
```
(https://travis-ci.org/pytest-dev/pytest/jobs/611099976#L6281)

blueyed:
For reference I'm planning to add `idsetfn` or something similar afterwards (https://github.com/blueyed/pytest/pull/104).

testing/python/metafunc.py
```diff
@@ -1784,3 +1832,39 @@ def test_foo(a):
...
+
+            @pytest.mark.parametrize("a, b", [(1, 2), (3, 4)], ids=itertools.count(1))
+            def test_converted_to_str(a, b):
+                pass
+        """
+        )
+        result = testdir.runpytest("-vv", "-s")
+        result.stdout.fnmatch_lines(
+            [
+                "test_parametrize_iterator.py::test1[param0] PASSED",
```
RonnyPfannschmidt: i would argue that this is wrong,
its misleading with infinite generators, and will break horrible with other generators
additionally this will potentially cause ids to be different based on how many test files are considered for collection

blueyed: @RonnyPfannschmidt 
Thanks for the feedback!

I see that you can shoot yourself in the foot with this somehow, however: collection happens on the whole file, so it's only relevant when using this across files, and then it would be the same when wrapping it in something to get the same outcome.

This PR/feature is about making it easier to use generators/iterators.

As for infinite generators that should not be a problem, or am I missing something?

But it certainly shows that collection is not stable for py35, likely due to some dict not being ordered somewhere there.  IIRC the test using the itertools fixture shows that it might sometimes collect/generate test2 before test1 there.

RonnyPfannschmidt: again, its already wrong on a single file and gets even more wrong on multiple files
as  far as i can tell, the only way to support iterators sanely would be to materealize them exactly once, that's not done here

blueyed: > again, its already wrong on a single file

Like I've said already this appears to be the case for py35 only, and also happens with:
```python
import pytest

id = 0


def idfn(val):
    global id
    id += 1
    return id


@pytest.mark.parametrize("a", [1], ids=idfn)
def test1(a):
    pass


@pytest.mark.parametrize("a", [2], ids=idfn)
def test2(a):
    pass
```
i.e. it's due to inconsistent collection there.
Or do you mean something else?

I can see that it could also be wrong when having some conditional code (i.e. test functions only defined on some conditions etc (where the decorator is used less often)) though.
But similar to using an iterator across multiple files this is something that happens by (abusing) the `id` as a function already.

> the only way to support iterators sanely would be to materealize them exactly once, that's not done here

I do not think that's possible really, is it?

Should we have something special to support the case of "just provide numeric ids, with a possible prefix"?

The next planned step here would have been to have a idsetfn function (or support it for `ids`), where this could be achieved through then maybe.
Should I go with `ids/idsetfn=lambda index: "param" + str(index)` then only?

RonnyPfannschmidt: You just use a broken example to justify a broken feature

It would be more apparent if the same mark was used,  and even more apparent if the iterator version was size limited to the length of the parameter set 

This is a no go

blueyed: I'm not sure I follow, but is this example less/not broken then?  (it also shows inconsistent results with py35):

```python
import pytest

id = 0


def idfn(val):
    global id
    id += 1
    return id


id_param = pytest.mark.parametrize("a", [1], ids=idfn)


@id_param
def test1(a):
    pass


@id_param
def test2(a):
    pass
```

Anyway, I've asked already which direction I should follow here, i.e. by not supporting iterators in the first place, but only a new / overloaded idfnset / ids function, which gets the index etc.

RonnyPfannschmidt: the example is broken in that it turns ids (which should related to parameters) into something thats related to a uncontrolled global variable 

what happens with the generator support you propose here is that this uncontrolled global variable with effects at a distance gets hidden in a mark 

as far as im concerned the only way to support iterators is bounded materialization that happens once, aka impossible with the current mark backend

blueyed: Ok.

I see three options then:

a) only allow generators in the top-most mark's args/kwargs (when generating the final mark), i.e. it would not allow them in the example above, raising an error.
b) store the generated results on the mark it was given for.  This would require some code changes (currently mark's args/kwargs are passed around, not the Mark itself).  I will push some WIP for this.
c) do not support this (as of now), closing https://github.com/pytest-dev/pytest/issues/759 as wont-fix.

src/_pytest/python.py
```diff
@@ -132,7 +132,25 @@ def pytest_generate_tests(metafunc):
...
+                save_ids = True
+            else:
+                save_ids = False
+                if len(args) > 3:
+                    args = args.copy()
+                    args[3] = ids
+                else:
+                    assert "ids" in kwargs
+                    kwargs = kwargs.copy()
+                    kwargs["ids"] = ids
```
blueyed: Having ids either as arg or kwarg is also not nice here.
Could be combined with `_has_param_ids` though at least.

src/_pytest/python.py
```diff
@@ -1016,6 +1042,7 @@ def parametrize(self, argnames, argvalues, indirect=False, ids=None, scope=None)
                 )
                 newcalls.append(newcallspec)
         self._calls = newcalls
+        return ids
```
blueyed: Not happy about having to return this here (for storing it in `pytest_generate_tests` then).

src/_pytest/mark/structures.py
```diff
@@ -153,8 +161,19 @@ def combined_with(self, other):
...
+            elif self._has_param_ids():
+                param_ids_from = self
+
         return Mark(
-            self.name, self.args + other.args, dict(self.kwargs, **other.kwargs)
+            self.name,
+            self.args + other.args,
+            dict(self.kwargs, **other.kwargs),
+            param_ids_from=param_ids_from,
         )
```
blueyed: I've thought about having a specialized `ParametrizedMark` to handle the extra logic here.
But it was not clear to me if `MarkDecorator` would be a better fit to begin with, and then the name check would be done when generating/instantiating marks instead.

testing/python/metafunc.py
RonnyPfannschmidt: D) defer this for after object based markers

src/_pytest/python.py
RonnyPfannschmidt: Solving the wrong problem at the wrong place feels like that

src/_pytest/mark/structures.py
RonnyPfannschmidt: Object based markers would resolve that at mark creation time

testing/python/metafunc.py
blueyed: And how would it be done then?  From https://github.com/pytest-dev/pytest/issues/5424 it seems like this could be achieved for ParametrizedMark already (like mentioned below, where for `name == "parametrize"` it would instantiate that).

src/_pytest/python.py
blueyed: The main problem here is passing `args` and `kwargs`, but not the mark.
This could be done instead here also - but the function can be used by itself also (without a mark), so I wanted to keep it apart.  But having an optional `store_resolved_ids_on_mark` kwarg might suffice then.

Returning the generated ids is not too bad also after all, but I've felt that it might be limiting in the future, and should therefore return the generated calls also etc.

testing/python/metafunc.py
RonnyPfannschmidt: its a potentially acceptable intermediate hack i would say

src/_pytest/mark/structures.py
```diff
@@ -144,7 +146,13 @@ class Mark:
     #: keyword arguments of the mark decorator
     kwargs = attr.ib()  # Dict[str, object]
 
-    def combined_with(self, other):
+    _param_ids_from = attr.ib(type=Optional["Mark"], default=None)
```
RonnyPfannschmidt: lets ensure those are not part of the repr

src/_pytest/mark/structures.py
```diff
@@ -153,8 +161,19 @@ def combined_with(self, other):
         combines by appending args and merging the mappings
         """
         assert self.name == other.name
+
```
RonnyPfannschmidt: we might want to add a comment that this is intended to be a temporary hack to be resolved with object based marks

blueyed: :+1: 
Will revisit before merging.

Thanks for the approval - I'm glad we could figure out something in the end.

PR #6176, opened 2019-11-13
pytester: Hookrecorder: improve assertoutcome

blueyed:
Before:

        def assertoutcome(self, passed: int = 0, skipped: int = 0, failed: int = 0) -> None:
            realpassed, realskipped, realfailed = self.listoutcomes()
            assert passed == len(realpassed)
    >       assert skipped == len(realskipped)
    E       assert 1 == 0
    E        +  where 0 = len([])

After:

    >       reprec = testdir.inline_run(testpath, "-s")
    E       AssertionError: ([], [], [<TestReport 'nodeid' when='call' outcome='failed'>])
    E       assert {'failed': 1, 'passed': 0, 'skipped': 0} == {'failed': 0, 'passed': 0, 'skipped': 1}

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -332,10 +332,17 @@ class HookRecorder:
         return [len(x) for x in self.listoutcomes()]
 
     def assertoutcome(self, passed: int = 0, skipped: int = 0, failed: int = 0) -> None:
-        realpassed, realskipped, realfailed = self.listoutcomes()
-        assert passed == len(realpassed)
-        assert skipped == len(realskipped)
-        assert failed == len(realfailed)
+        __tracebackhide__ = True
+
+        outcomes = self.listoutcomes()
+        realpassed, realskipped, realfailed = outcomes
+        obtained = {
+            "passed": len(realpassed),
+            "skipped": len(realskipped),
+            "failed": len(realfailed),
+        }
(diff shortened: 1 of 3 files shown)


src/_pytest/pytester.py
```diff
@@ -332,10 +332,15 @@ def countoutcomes(self) -> List[int]:
...
     def assertoutcome(self, passed: int = 0, skipped: int = 0, failed: int = 0) -> None:
-        realpassed, realskipped, realfailed = self.listoutcomes()
-        assert passed == len(realpassed)
-        assert skipped == len(realskipped)
-        assert failed == len(realfailed)
+        __tracebackhide__ = True
+
+        outcomes = self.listoutcomes()
+        realpassed, realskipped, realfailed = outcomes
+        assert {"passed": passed, "skipped": skipped, "failed": failed} == {
```
nicoddemus: Through the tests we usually write "assert obtained == expected" (for example `assert f() == 4`), but here the comparison is swapped... for consistency I think this should be swapped as well.

```python
expected = {"passed": passed, "skipped": skipped, "failed": failed}
obtained = {
    "passed": len(realpassed),
    "skipped": len(realskipped),
    "failed": len(realfailed),
}
assert obtained == expected
```

blueyed: Amended.

nicoddemus:
Thanks!

PR #6184, opened 2019-11-13
Show a better message when 'request' is used in parametrize

### LINKED ISSUE #6183: Raise an exception when `"request"` is used as a parametrization argument
I just found, accidentally, that this
```python
@pytest.mark.parametrize('request, expected', [(1, 1)])
def test_foo(request, expected):
    assert request == expected
```
would raise an error
```
...
    def get_direct_param_fixture_func(request):
>       return request.param
E       AttributeError: 'int' object has no attribute 'param'
```

Wouldn't it be helpful if there was a helpful exception raised here https://github.com/pytest-dev/pytest/blob/master/src/_pytest/fixtures.py#L881-L882?

For example, if 
```python
            if argname != "request":
                fixturedef.addfinalizer(functools.partial(self.finish, request=request))
            else:
                raise ValueError('`request` cannot be use as a parametrization name')
```
(or a better message) in the case of the initial example gives
```
self = <FixtureDef name='request' scope='function' baseid=''>
request = <SubRequest 'request' for <Function 'test_foo[1-1]'>>

    def execute(self, request):
        # get required arguments and register our own finish()
        # with their finalization
        for argname in self.argnames:
            fixturedef = request._get_active_fixturedef(argname)
            if argname != "request":
                fixturedef.addfinalizer(functools.partial(self.finish, request=request))
            else:
>               raise ValueError('`request` cannot be use as a parametrization name')
E               ValueError: `request` cannot be use as a parametrization name
```

The pytest version i'm using is 3.10.0

nicoddemus:
Fix #6183

### DIFF
--- a/testing/python/metafunc.py
+++ b/testing/python/metafunc.py
@@ -72,6 +72,19 @@ class TestMetafunc:
         ):
             metafunc.parametrize("x", [1], scope="doggy")
 
+    def test_parametrize_request_name(self, testdir):
+        """Show proper error  when 'request' is used as a parameter name in parametrize (#6183)"""
+
+        def func(request):
+            raise NotImplementedError()
+
+        metafunc = self.Metafunc(func)
+        with pytest.raises(
+            pytest.fail.Exception,
+            match=r"'request' is a reserved name and cannot be used in @pytest.mark.parametrize",
+        ):
+            metafunc.parametrize("request", [1])
+
     def test_find_parametrized_scope(self):
         """unittest for _find_parametrized_scope (#3941)"""
         from _pytest.python import _find_parametrized_scope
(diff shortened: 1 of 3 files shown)


testing/python/metafunc.py
```diff
@@ -72,6 +72,19 @@ def func(x):
         ):
             metafunc.parametrize("x", [1], scope="doggy")
 
+    def test_parametrize_request_name(self, testdir):
+        """Show proper error  when 'request' is used as a parameter name in parametrize (#6183)"""
+
+        def func(request):
+            pass
```
asottile: to have coverage be less mad this can be `raise NotImplementedError()`

PR #6188, opened 2019-11-14
Update text in PR template

hugovk:

- [n/a, small docs fix] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>.<TYPE>.rst`. See [changelog/README.rst](https://github.com/pytest-dev/pytest/blob/master/changelog/README.rst) for details.

---

* The "(please delete...)" text is redundant and can be removed; it's inside comments so won't be shown anyway.

* Replace the semicolon with a fullstop. It's not necessary here, you don't use them at the end of a sentence; they would only go in the middle of a sentence, a bit like a comma.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,6 @@
 Thanks for submitting a PR, your contribution is really appreciated!
 
 Here is a quick checklist that should be present in PRs.
--->
 
 - [ ] Target the `master` branch for bug fixes, documentation updates and trivial changes.
 - [ ] Target the `features` branch for new features, improvements, and removals/deprecations.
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,15 +2,14 @@
 Thanks for submitting a PR, your contribution is really appreciated!
 
 Here is a quick checklist that should be present in PRs.
-(please delete this text from the final description, this is just a guideline)
```
blueyed: This might be a misunderstanding then - I've assumed the whole text should be deleted - or at least we should not end up with open TODOs always (since you cannot target both master/features).
Can we have the closing `-->` at the end maybe, so everything becomes a comment / invisble?

hugovk: Sure, updated!

blueyed:
Thanks! :heart:

asottile:
hmmm if we're going to hide the checklist entirely what's the point? now I can't see if PRs are ready

hugovk:
An argument for hiding the checklist completely: you don't get these inconsistent progress bars. 

image

An argument for keeping it: maintainers can see which tasks have been done.

image

Would it be best to keep the checklist visible, and something like this in the comment block?

> Please delete the checkboxes which don't apply.

PR #6190, opened 2019-11-14
Preparing release version 5.2.3

nicoddemus:


nicoddemus:
Not sure, but travis seems to have missed my update, but the previous commit passed fine so I think we're good to go here.

blueyed:
> Not sure, but travis seems to have missed my update

Because of "[skip ci]".. ;)

> but the previous commit passed fine so I think we're good to go here.

:+1:

nicoddemus:
Rá duh

nicoddemus:
Oh no, this will probably cause it to also skip the deploy when I push the tag... will have to ammend I'm afraid. 😓

PR #6192, opened 2019-11-14
Remove report_log in favor of pytest-reportlog

### LINKED ISSUE #6180: Move report-log to an external plugin
Let's move report-log to an external plugin for now:

* It is always easy to merge a plugin rather than remove it later.
* The format is very much a work in progress (file size might is a concern when compared with result-log)
* It will allow the external plugin to gain further capabilities, like posting each report line to a server instead of writing to a file (which might incur extra dependencies which are not desired in pytest's core)
* We should update the warning to point users to the new plugin.

We might decide to merge the plugin later at any time, or not, as the functionality is quite niche and does not really need to be in the core.

nicoddemus:
Fix #6180


nicoddemus:
I plan to make the first release of https://github.com/pytest-dev/pytest-reportlog once this gets merged. 👍

blueyed:
@nicoddemus 
Cannot really tell currently, but FYI: https://github.com/pytest-dev/pytest/pull/6193 (in the new plugin then I guess/assume)

nicoddemus:
Thanks @blueyed: https://github.com/pytest-dev/pytest-reportlog/pull/3

PR #6193, opened 2019-11-15
Import Path from _pytest.pathlib for py35

blueyed:
This is important for `isinstance` checks etc.

TODO:

~~- [ ] after #6192 (conflicts there, https://github.com/pytest-dev/pytest/pull/6192/files#diff-af40f12336a069d7f9538cf71a4010ecL2)~~


blueyed:
Removed/ignored src/_pytest/report_log.py to avoid conflicts.

PR #6199, opened 2019-11-15
tests: revisit test_cacheprovider

blueyed:
- Setting PYTHONDONTWRITEBYTECODE in the environment does not change it
for the current process.
- Revisit tests in general, mainly formatting.

### DIFF
--- a/testing/test_cacheprovider.py
+++ b/testing/test_cacheprovider.py
@@ -59,7 +59,8 @@ class TestNewAPI:
     @pytest.mark.filterwarnings(
         "ignore:could not create cache path:pytest.PytestWarning"
     )
-    def test_cache_failure_warns(self, testdir):
+    def test_cache_failure_warns(self, testdir, monkeypatch):
+        monkeypatch.setenv("PYTEST_DISABLE_PLUGIN_AUTOLOAD", "1")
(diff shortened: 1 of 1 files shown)


testing/test_cacheprovider.py
```diff
@@ -300,22 +278,8 @@ def test_3():
...
-        )
-        testdir.tmpdir.join("test_b.py").write(
-            textwrap.dedent(
-                """\
-                def test_always_fails():
-                    assert 0
-                """
-            )
-        )
+        testdir.makepyfile(**{"test_a.py": "def test_always_passes(): pass"})
```
bluetech: Is the `.py` in `test_a.py` still needed with `makepyfile` (rather than `makefile`)?

testing/test_cacheprovider.py
```diff
@@ -345,18 +303,15 @@ def test_always_fails():
...
-            test_a="""\
-            def test_a1():
-                assert 0
-            def test_a2():
-                assert 1
-            """,
-            test_b="""\
-            def test_b1():
-                assert 0
+            **{
```
bluetech: Here can use regular kwargs I suppose?

testing/test_cacheprovider.py
blueyed: No, and can also use kwargs directly.  Amended.

PR #6205, opened 2019-11-16
Add type annotations to _pytest.compat and _pytest._code.code

bluetech:
This PR starts with some small prerequisite commits, and then adds type annotations to `_pytest.compat` and `_pytest._code.code`.

Let me know if this is too big for one PR, and I will split it. Note that some part of the diff is just adding `-> None` to some test functions, so that mypy will check them without `check_untyped_defs = True` (which needs more work).

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -7,13 +7,17 @@ from inspect import CO_VARKEYWORDS
 from io import StringIO
 from traceback import format_exception_only
 from types import CodeType
+from types import FrameType
 from types import TracebackType
 from typing import Any
+from typing import Callable
 from typing import Dict
 from typing import Generic
+from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Pattern
+from typing import Sequence
 from typing import Set
 from typing import Tuple
 from typing import TypeVar
@@ -27,9 +31,16 @@ import py
 import _pytest
 from _pytest._io.saferepr import safeformat
 from _pytest._io.saferepr import saferepr
+from _pytest.compat import overload
 
 if False:  # TYPE_CHECKING
     from typing import Type
+    from typing_extensions import Literal
+    from weakref import ReferenceType  # noqa: F401
+
+    from _pytest._code import Source
+
+    _TracebackStyle = Literal["long", "short", "no", "native"]
 
 
 class Code:
@@ -38,13 +49,12 @@ class Code:
     def __init__(self, rawcode) -> None:
         if not hasattr(rawcode, "co_filename"):
             rawcode = getrawcode(rawcode)
-        try:
-            self.filename = rawcode.co_filename
-            self.firstlineno = rawcode.co_firstlineno - 1
-            self.name = rawcode.co_name
-        except AttributeError:
+        if not isinstance(rawcode, CodeType):
             raise TypeError("not a code object: {!r}".format(rawcode))
-        self.raw = rawcode  # type: CodeType
+        self.filename = rawcode.co_filename
+        self.firstlineno = rawcode.co_firstlineno - 1
+        self.name = rawcode.co_name
+        self.raw = rawcode
 
     def __eq__(self, other):
         return self.raw == other.raw
@@ -72,7 +82,7 @@ class Code:
         return p
 
     @property
-    def fullsource(self):
+    def fullsource(self) -> Optional["Source"]:
         """ return a _pytest._code.Source object for the full source file of the code
         """
         from _pytest._code import source
@@ -80,7 +90,7 @@ class Code:
         full, _ = source.findsource(self.raw)
         return full
 
-    def source(self):
+    def source(self) -> "Source":
         """ return a _pytest._code.Source object for the code object's source only
         """
         # return source only for that part of code
@@ -88,7 +98,7 @@ class Code:
 
         return _pytest._code.Source(self.raw)
 
-    def getargs(self, var=False):
+    def getargs(self, var: bool = False) -> Tuple[str, ...]:
         """ return a tuple with the argument names for the code object
 
             if 'var' is set True also return the names of the variable and
@@ -107,7 +117,7 @@ class Frame:
     """Wrapper around a Python frame holding f_locals and f_globals
     in which expressions can be evaluated."""
 
-    def __init__(self, frame):
+    def __init__(self, frame: FrameType) -> None:
         self.lineno = frame.f_lineno - 1
         self.f_globals = frame.f_globals
         self.f_locals = frame.f_locals
@@ -115,7 +125,7 @@ class Frame:
         self.code = Code(frame.f_code)
 
     @property
-    def statement(self):
+    def statement(self) -> "Source":
         """ statement this frame is at """
         import _pytest._code
 
@@ -134,7 +144,7 @@ class Frame:
         f_locals.update(vars)
         return eval(code, self.f_globals, f_locals)
 
-    def exec_(self, code, **vars):
+    def exec_(self, code, **vars) -> None:
         """ exec 'code' in the frame
 
             'vars' are optional; additional local variables
@@ -143,7 +153,7 @@ class Frame:
         f_locals.update(vars)
         exec(code, self.f_globals, f_locals)
 
-    def repr(self, object):
+    def repr(self, object: object) -> str:
         """ return a 'safe' (non-recursive, one-line) string repr for 'object'
         """
         return saferepr(object)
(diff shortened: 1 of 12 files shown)


src/_pytest/_code/source.py
```diff
@@ -60,7 +60,7 @@ def __getitem__(self, key: int) -> str:
         raise NotImplementedError()
 
     @overload  # noqa: F811
-    def __getitem__(self, key: slice) -> "Source":
+    def __getitem__(self, key: slice) -> "Source":  # noqa: F811
```
bluetech: There seems to be a discrepancy between where CI flake8 and my flake8 report this - on the decorator or the `def`. Maybe because of Python 3.8? So for now I add it to both.

blueyed: > Maybe because of Python 3.8?

Very likely, yes.
Isn't there a flake8 plugin to handle this automatically maybe?

src/_pytest/_code/code.py
```diff
@@ -808,7 +845,10 @@ def _truncate_recursive_traceback(self, traceback):
                 max_frames=max_frames,
                 total=len(traceback),
             )  # type: Optional[str]
-            traceback = traceback[:max_frames] + traceback[-max_frames:]
+            # Type ignored because mypy thinks the slice is a List[TracebackEntry]
+            # instead of the Traceback subclass, even though we override __getitem__().
+            # Should figure out why it does that.
+            traceback = traceback[:max_frames] + traceback[-max_frames:]  # type: ignore
```
blueyed: The slice is a `Traceback`, but only the expression `traceback[:max_frames] + traceback[-max_frames:]` is a `List[…]` then.
No idea how to fix it, but it might help you.. :)

src/_pytest/_code/code.py
```diff
@@ -959,15 +1011,22 @@ def __init__(self, tblines):
...
 
 class ReprEntry(TerminalRepr):
-    def __init__(self, lines, reprfuncargs, reprlocals, filelocrepr, style):
+    def __init__(
+        self,
+        lines: Sequence[str],
+        reprfuncargs: Optional["ReprFuncArgs"],
+        reprlocals: Optional["ReprLocals"],
+        filelocrepr: Optional["ReprFileLocation"],
+        style: str,
```
blueyed: `Literal` for `style`?

src/_pytest/_code/code.py
```diff
@@ -976,6 +1035,7 @@ def __init__(self, lines, reprfuncargs, reprlocals, filelocrepr, style):
 
     def toterminal(self, tw) -> None:
         if self.style == "short":
+            assert self.reprfileloc is not None
```
blueyed: Could (additionally) be checked/validated in `__init__`?  But it's fine like that also.

src/_pytest/assertion/rewrite.py
```diff
@@ -1074,13 +1074,14 @@ def try_makedirs(cache_dir) -> bool:
 
 def get_cache_dir(file_path: Path) -> Path:
     """Returns the cache directory to write .pyc files for the given .py file path"""
-    if sys.version_info >= (3, 8) and sys.pycache_prefix:
+    # Type ignored until added in next mypy release.
```
blueyed: Would be good to link an issue in general, but ok for now.

src/_pytest/compat.py
```diff
@@ -342,39 +352,26 @@ def safe_isclass(obj):
...
-    @property
-    def funcargnames(self):
-        """ alias attribute for ``fixturenames`` for pre-2.3 compatibility"""
-        import warnings
-        from _pytest.deprecated import FUNCARGNAMES
-
-        warnings.warn(FUNCARGNAMES, stacklevel=2)
-        return self.fixturenames
-
-
```
blueyed: I'm ok with inlining, but wondered how it made typing harder (from the commit message)?

testing/code/test_source.py
```diff
@@ -96,28 +99,30 @@ def test_source_putaround():
...
 
 
-def test_syntaxerror_rerepresentation():
+def test_syntaxerror_rerepresentation() -> None:
     ex = pytest.raises(SyntaxError, _pytest._code.compile, "xyz xyz")
+    assert ex is not None
     assert ex.value.lineno == 1
     assert ex.value.offset in {5, 7}  # cpython: 7, pypy3.6 7.1.1: 5
+    assert ex.value.text is not None
     assert ex.value.text.strip(), "x x"
```
blueyed: Could check the full expected text instead?
`assert ex.value.text == "xyz xyz\n"`

blueyed:
> Note that some part of the diff is just adding `-> None` to some test functions, so that mypy will check them without `check_untyped_defs = True` (which needs more work).

IIRC this (`check_untyped_defs`) could be enabled based on modules also?

src/_pytest/_code/source.py
bluetech: There seems to be some discussion regarding this [here](https://github.com/PyCQA/pyflakes/issues/488), I'm too lazy to read through it though...

src/_pytest/_code/code.py
```diff
@@ -38,13 +46,12 @@ class Code:
     def __init__(self, rawcode) -> None:
```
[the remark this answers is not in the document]
bluetech: Not really. I spent some time trying to type `getrawcode` but left it aside for now.

src/_pytest/_code/code.py
```diff
@@ -134,7 +141,7 @@ def eval(self, code, **vars):
         f_locals.update(vars)
         return eval(code, self.f_globals, f_locals)
 
-    def exec_(self, code, **vars):
+    def exec_(self, code, **vars) -> None:
```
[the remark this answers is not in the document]
bluetech: They can be anything really AFAIK, but I'm not entirely sure, so I left it out.

src/_pytest/_code/code.py
```diff
@@ -273,33 +279,40 @@ def __str__(self):
...
+class Traceback(List[TracebackEntry]):
     """ Traceback objects encapsulate and offer higher level
         access to Traceback entries.
     """
 
-    Entry = TracebackEntry
-
-    def __init__(self, tb, excinfo=None):
+    def __init__(
+        self, tb: Union[TracebackType, Iterable[TracebackEntry]], excinfo=None
```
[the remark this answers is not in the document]
bluetech: Yes, will add.

src/_pytest/_code/code.py
```diff
@@ -273,33 +279,40 @@ def __str__(self):
...
-            list.__init__(self, tb)
+            super().__init__(tb)
 
-    def cut(self, path=None, lineno=None, firstlineno=None, excludepath=None):
+    def cut(
+        self,
+        path=None,
+        lineno: Optional[int] = None,
+        firstlineno: Optional[int] = None,
+        excludepath=None,
```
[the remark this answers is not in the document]
bluetech: It's can be a `py.path` AFAIU, so I'm leaving it unannotated.

src/_pytest/_code/code.py
bluetech: Right, I misunderstood the situation. Looks like a deficiency in mypy/typeshed for now. I'll update the comment.

src/_pytest/_code/code.py
bluetech: Yes, will add.

src/_pytest/_code/code.py
bluetech: Wouldn't happen unless the code is buggy; seems fine to only do it here.

src/_pytest/assertion/rewrite.py
bluetech: mypy can detect outdated `type: ignore`s so it's easy to get rid of when updating mypy.

src/_pytest/compat.py
bluetech: It's a mixin class, it uses `fixturenames` which is not defined in the class so mypy gets confused. There are ways to type this but it's not worth the bother in this case IMO.

testing/code/test_source.py
bluetech: Yes, that's better.

bluetech:
> IIRC this (check_untyped_defs) could be enabled based on modules also?

Yes, that's an option to. The `-> None` is not too bad, though.

bluetech:
Thanks for the review @blueyed, [updated](https://github.com/pytest-dev/pytest/compare/51e3be21e81bb1164d11dfd15755908bae5c9970..eaa34a9df0fa341c1b21bd3b232928812e6e8a06).

src/_pytest/compat.py
blueyed: Agreed.

PR #6215, opened 2019-11-18
_idval: remove trailing newline from exception

blueyed:
It might be worth including the error there, but it gets chained, so it's
ok I assume.


blueyed:
@togo can you please review/approve?

Issue #5199, opened 2019-05-02
Assertion error truncates byte strings


nickwilliams-eventbrite:
- [x] a detailed description of the bug or suggestion
  Even when using arguments `-vvvv --full-trace`, an explicit assertion error is truncated. This is different from comparisons/diffs. Those are not truncated. But when an explicit assertion error message is provided, its contents are truncated. Example:
  ```
  def test_get_events_for_user(self):
      response = self.get(f'/v3/organizations/{self.test_organization_id}/events/')
      assert response.status_code == 200, response.content
  ```

  Where `self.get` returns a Python Requests `Response` object. The point of this assertion line is that, if the status code is not correct, the entire response content is dumped so that the error code(s) and message(s) is the JSON response can be viewed. However, this is the result, even with full verbosity:

  ```
  >       assert response.status_code == 200, response.content
  E       AssertionError: b'{"error": "SERVER_ERROR", "error_description": "Error(code=u\'SERVER_ERROR\', message=u\\"Internal server error: \'E...54, in get_response\
  E             raise self.JobError(response.errors)\
  E         JobError: Error executing job:", "variables": null}]}'
  E       assert 500 == 200
  E         -500
  E         +200
  ```

  The response content, in this case, was ~5000 bytes, but only ~250 are displayed, with an ellipse in the middle. With anything `-vv` or greater, the entire assertion message should be displayed.
- [x] output of `pip list` from the virtual environment you are using
  ```
  pytest                 4.4.1             
  pytest-cov             2.6.1             
  pytest-django          3.4.8             
  pytest-runner          4.4
  ...
  requests               2.21.0 
  ```
- [x] pytest and operating system versions
  Ubuntu 16.04
  ```
  $ python --version
  Python 3.7.1
  $ pytest --version
  This is pytest version 4.4.1, imported from /usr/local/lib/python3.7/site-packages/pytest.py
  setuptools registered plugins:
    pytest-cov-2.6.1 at /usr/local/lib/python3.7/site-packages/pytest_cov/plugin.py
    pytest-django-3.4.8 at /usr/local/lib/python3.7/site-packages/pytest_django/plugin.py
- [x] minimal example if possible
  See above.

nicoddemus:
Thanks @nickwilliams-eventbrite for the report!

Closing as a duplicate of #5192

nickwilliams-eventbrite:
[...]

PR #6225, opened 2019-11-18
_format_assertmsg: use safeformat instead of saferepr

blueyed:
Custom assertion messages should not get shortened via `saferepr`.

Via https://github.com/blueyed/pytest/pull/125.
Fixes https://github.com/pytest-dev/pytest/issues/5199.

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -19,6 +19,7 @@ from typing import Optional
 from typing import Set
 from typing import Tuple
 
+from _pytest._io.saferepr import safeformat
 from _pytest._io.saferepr import saferepr
 from _pytest._version import version
 from _pytest.assertion import util
@@ -381,8 +382,7 @@ def _format_assertmsg(obj):
 
     For strings this simply replaces newlines with '\n~' so that
     util.format_explanation() will preserve them instead of escaping
-    newlines.  For other objects saferepr() is used first.
-
+    newlines.  For other objects ``safeformat()`` is used first.
     """
     # reprlib appears to have a bug which means that if a string
     # contains a newline it gets escaped, however if an object has a
@@ -390,8 +390,7 @@ def _format_assertmsg(obj):
     # However in either case we want to preserve the newline.
     replaces = [("\n", "\n~"), ("%", "%%")]
     if not isinstance(obj, str):
-        obj = saferepr(obj)
-        replaces.append(("\\n", "\n~"))
+        obj = safeformat(obj).replace("\\n", "\n~")
 
     for r1, r2 in replaces:
         obj = obj.replace(r1, r2)
--- a/testing/test_assertion.py
+++ b/testing/test_assertion.py
@@ -72,10 +72,11 @@ class TestImportHookInstallation:
         result = testdir.runpytest_subprocess()
         result.stdout.fnmatch_lines(
             [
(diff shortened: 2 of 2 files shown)


nicoddemus:
Hi @blueyed,

Thanks for the patch!

The underlying reasoning is sound, but the test doesn't seem to be harden enough.

Here's the output of `test_AssertionError_message` on `features`:

```
___________________________________________ test_bytes ____________________________________________

    def test_bytes():
>       assert 0, b'b' * 80
E       AssertionError: b'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'
E       assert 0

test_AssertionError_message.py:6: AssertionError
```

And here's the output from your branch:

```
___________________________________________ test_bytes ____________________________________________

    def test_bytes():
>       assert 0, b'b' * 80
E       AssertionError: (b'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'
E          b'bbbb')
E       assert 0

test_AssertionError_message.py:6: AssertionError
```

Both are not truncating the message, and in fact the message from `features` looks better to me. Thoughts?

blueyed:
@nicoddemus 
`saferepr` truncates at 240 characters.

I've thought it would be enough to show that safeformat gets used, which here wraps at 80 characters.

I've initially done this to fix:
```
>>> saferepr(list(range(0, 10)))
'[0, 1, 2, 3, 4, 5, ...]'
```

vs

```
>>> safeformat(list(range(0, 10)))
'[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]'
```

(Note that this patch would also look better with wider columns, not using 80, of course.
That's something to do as a follow-up though.)

I've also thought to use just `str` in the first place, but then thought that formatting is good in general.

Also note that `safeformat` does not wrap strings.

blueyed:
Should maybe use _pformat_dispatch from cc503c182?

nicoddemus:
> I've thought it would be enough to show that safeformat gets used, which here wraps at 80 characters.

I see, thanks for the explanation. Do you know why it is showing the `byte` around `()` though?

> Should maybe use _pformat_dispatch from cc503c1?

Hmm sounds reasonable.

blueyed:
> > I've thought it would be enough to show that safeformat gets used, which here wraps at 80 characters.
> 
> I see, thanks for the explanation. Do you know why it is showing the `byte` around `()` though?

This is because it gets split into multiple lines.

Also happens for strings, but requires spaces there it seems:
```
>>> pprint.pprint(('b'*10 + ' ') * 8)
('bbbbbbbbbb bbbbbbbbbb bbbbbbbbbb bbbbbbbbbb bbbbbbbbbb bbbbbbbbbb bbbbbbbbbb '
 'bbbbbbbbbb ')
```

> > Should maybe use _pformat_dispatch from [cc503c1](https://github.com/pytest-dev/pytest/commit/cc503c1821e709376d9a05ee4f6459e4e4f41a0c)?
> 
> Hmm sounds reasonable.

Not so sure myself anymore.  Anything fitting 80 chars (to be the terminal width in the long run) should be ok to not wrap probably.

So I'd rather leave it as is, maybe adding the extra test for not truncating the tuple/list?

PR #6234, opened 2019-11-19
Revert "A warning is now issued when assertions are made for `None`"

### LINKED ISSUE #4639: Warning about assert None from XFAIL tests

asottile:
Resolves #4639 

CC @Tadaboody

### DIFF
--- a/testing/test_warnings.py
+++ b/testing/test_warnings.py
@@ -543,7 +543,7 @@ class TestAssertionWarnings:
 
     def test_tuple_warning(self, testdir):
         testdir.makepyfile(
-            """
+            """\
             def test_foo():
                 assert (1,2)
             """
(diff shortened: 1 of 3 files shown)


testing/test_warnings.py
```diff
@@ -543,7 +543,7 @@ def assert_result_warns(result, msg):
 
     def test_tuple_warning(self, testdir):
         testdir.makepyfile(
-            """
+            """\
```
blueyed: Please do not change code unnecessary, especially when it is only to please your editor in some way (IIRC).
For all others this adds visual disbalance, and often gets highlighted in a special way (line continuation).

nicoddemus:
This can go to `master` btw (it can't possibly break anything).

asottile:
rebased on master!

testing/test_warnings.py
blueyed: ?

Tadaboody:
Oof another one bites the dust then (honestly it’s pretty satisfying seeing the revert being this smooth).

Is it worth it to try and eliminate the false positive? Maybe try and only warn if it fails the test? (Some kinda `try` wrapping it)

nicoddemus:
@Tadaboody I don't think so, most cases the test will at least fail with a message anyway, as @asottile pointed out.

@blueyed anything else here?

blueyed:
> @blueyed anything else here?

I do not like the (unnecessary) code style change (and being ignored for pointing it out (across a rebase where it could have been fixed)), and will likely investigate in what I've said in https://github.com/pytest-dev/pytest/issues/4639#issuecomment-555845784.
I am not blocking this though, but will likely carry this into my fork then also.. :+1:

PR #5934, opened 2019-10-09
Improve ExceptionInfo.__repr__

blueyed:

PR #6238, opened 2019-11-20
Fix rendering of Before/After in changelog

asottile:
Apparently the version of sphinx that rtd uses is a little more strict about whether an anonymous `code-block` can happen

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -113,7 +113,7 @@ Improvements
 
   Before:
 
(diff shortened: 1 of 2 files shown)


blueyed:
Can you include https://github.com/pytest-dev/pytest/pull/5934#pullrequestreview-319510256 here then also?

nicoddemus:
We can use just `::` for monospace blocks. Included #5934 as well.

PR #6242, opened 2019-11-20
Add link to testdir fixture in CHANGELOG entry

nicoddemus:
As per https://github.com/pytest-dev/pytest/pull/5914#issuecomment-556063975

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -67,7 +67,7 @@ Features
   rather than implicitly.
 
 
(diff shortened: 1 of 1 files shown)


blueyed:
There are quite a few more mentions of pytester, where at least some are also specific to testdir.

blueyed:
btw: https://github.com/pytest-dev/pytest/issues/4283

nicoddemus:
> There are quite a few more mentions of pytester, where at least some are also specific to testdir.

I see, thanks for pointing that out, but there are quite a few of them and I'm not sure it is worth updating the CHANGELOG for very old versions.

PR #6244, opened 2019-11-20
tests: skip tests without xmlschema

blueyed:


bluetech:
Given that this package is in the `testing` extra, when will it not be available?

blueyed:
It does not have to be installed necessarily, i.e. when just doing `pip install -e .`.
This is just for convenience, and leaves a better impression when tests are working out of the box.
(it will still fail due to hypothesis here, but that's not easy to fix)
Feel free to close, I just thought it's easy to upstream.

nicoddemus:
Thanks @blueyed, but this means most (if not all) tests from `junitxml` will get skipped if the user does not have `xmlschema` installed. This might be frustrating because a user will inavertidely get a successfull test run, but there might be broken `junitxml` tests which will only appear then in CI.

I prefer we advertise users working on the tests to always use `pip install -e .[testing]` to avoid our test suite having to work around incomplete virtual environments.

PR #6249, opened 2019-11-20
Add develop instructions to CONTRIBUTING

nicoddemus:
From: https://github.com/pytest-dev/pytest/pull/6244

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -262,6 +262,19 @@ Here is a simple overview, with pytest-specific bits:
 
    When committing, ``pre-commit`` will re-format the files if necessary.
 
+#. If instead of using ``tox`` you prefer to run the tests directly, then we suggest to create a virtual environment and use
+   an editable install with the ``testing`` extra::
+
+       $ python3 -m venv .venv
+       $ source .venv/bin/activate  # Linux
+       $ .venv/Scripts/activate.bat  # Windows
+       $ pip install -e ".[testing]"
+
+   Afterwards, you can edit the files and run pytest normally::
+
+       $ pytest testing/test_config.py
+
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -262,6 +262,19 @@ Here is a simple overview, with pytest-specific bits:
 
    When committing, ``pre-commit`` will re-format the files if necessary.
 
+#. If instead of using ``tox`` you prefer to run the tests directly, then we suggest to create a virtual environment and use
+   an editable install with the ``testing`` extra::
+
+       $ python3 -m venv .env
```
blueyed: s/.env/.venv/

`.env` is often used for envdirs etc, and `.venv` is more common from what I can tell.

CONTRIBUTING.rst
```diff
@@ -262,6 +262,19 @@ Here is a simple overview, with pytest-specific bits:
...
+       $ source .env/bin/activate  # linux
```
blueyed: s/linux/Linux

CONTRIBUTING.rst
```diff
@@ -262,6 +262,19 @@ Here is a simple overview, with pytest-specific bits:
...
+       $ .env/Scripts/activate.bat  # Windows
+       $ pip install -e .[testing]
```
blueyed:
```suggestion
       $ pip install -e ".[testing]"
```

`[` might trigger globbing in shells, resulting in an error then.

nicoddemus:
Cool, thanks for the review. 👍

nicoddemus:
> If we recommend .venv/, it will be good to add it to .gitignore.

Yeah, now that you mention, that's why my gut feeling was using `.env` in the first place (that's on `.gitignore` already).

But no worries, added `.venv` to `.gitignore` as well.

PR #6272, opened 2019-11-25
docs: allow for Sphinx directives in changelogs

blueyed:
Fixes #4283.

TODO:

- [x] update doc with regard to using towncrier to view drafts, should recommend uisng `tox -e docs-draft` instead.

2019-11-25-062631_1070x992_escrotum

PR #6287, opened 2019-11-28
docs: move changelog to docs/en and allow sphinx directives

nicoddemus:
Now `tox -e docs` will also include the draft changelog for the
next version (locally only).

`CHANGELOG.rst` now only points to the changelog on READTHEDOCS so
sphinx diretives can be used.

Followup to https://github.com/pytest-dev/pytest/pull/6272


changelog/README.rst
```diff
@@ -31,6 +31,7 @@ changelog using that instead.
 If you are not sure what issue type to use, don't hesitate to ask in your PR.
 
 ``towncrier`` preserves multiple paragraphs and formatting (code blocks, lists, and so on), but for entries
-other than ``features`` it is usually better to stick to a single paragraph to keep it concise. You can install
-``towncrier`` and then run ``towncrier --draft``
-if you want to get a preview of how your change will look in the final release notes.
+other than ``features`` it is usually better to stick to a single paragraph to keep it concise.
+
+You can also run ``tox -e docs`` to build the documentation
+with the draft changelog (``doc/en/_build/changelog.html``) if you want to get a preview of how your change will look in the final release notes..
```
blueyed:
```suggestion
with the draft changelog (``doc/en/_build/changelog.html``) if you want to get a preview of how your change will look in the final release notes.
```

nicoddemus:
> Isn't it included always now? (since the Sphinx tag is used always - but maybe the idea is that RTD does not use the tag?)

Exactly! I've added comments to make that more clear. 👍

changelog/README.rst
nicoddemus: Thanks, done

PR #6295, opened 2019-11-30
[4.6] Include <testsuites> root tag in generated XML (#5550)

nicoddemus:
From: https://github.com/pytest-dev/pytest/pull/5550#issuecomment-558797460


nicoddemus:
Merging because the windows failure (twisted-py27) only happened because a network hiccup on chocolatey installing vcpython27 on Windows.

PR #6304, opened 2019-12-02
Convert pytest.py into a package

nicoddemus:
As discussed in https://github.com/pytest-dev/pytest/issues/3342, this is the first step to make pytest fully support static typing.

Had a few minutes on my hands so decide to tackle this.

### DIFF
--- a/src/pytest/__main__.py
+++ b/src/pytest/__main__.py
@@ -1,0 +1,7 @@
+"""
+pytest entry point
+"""
+import pytest
+
+if __name__ == "__main__":
+    raise SystemExit(pytest.main())
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,8 +38,8 @@ packages =
     _pytest.assertion
     _pytest.config
     _pytest.mark
+    pytest
 
-py_modules = pytest
 python_requires = >=3.5
 
 [options.entry_points]
--- a/src/pytest/__init__.py
+++ b/src/pytest/__init__.py
@@ -4,6 +4,7 @@ pytest: unit and functional testing with Python.
 """
 from _pytest import __version__
 from _pytest.assertion import register_assert_rewrite
+from _pytest.compat import _setup_collect_fakemodule
 from _pytest.config import cmdline
 from _pytest.config import hookimpl
 from _pytest.config import hookspec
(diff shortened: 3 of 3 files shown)


bluetech:
My PR from yesterday https://github.com/pytest-dev/pytest/pull/6302 is pretty much the same. It failed CI twice on the Azure py37 job, timing out on the `testing/test_assertion.py::test_exception_handling_no_traceback` test. It *seems* like this one is going down the same path. I haven't looked but not sure how this affects this test, and it doesn't reproduce locally.

nicoddemus:
Oh really sorry @bluetech, I've completely missed the notification on that (in fact I still can't see it in my Inbox, weird), otherwise I wouldn't have opened this one. 😓

CaselIT:
Sorry if already discusses somewhere, the long term plan is to remove the `_pytest` package in favor of this one?

nicoddemus:
> Sorry if already discusses somewhere, the long term plan is to remove the _pytest package in favor of this one?

No, the `pytest` package contains only symbols of the public API, while `_pytest` contains the actual codebase.

nicoddemus:
It is incredible how it is basically impossible to restart an Azure job... Clicking on `Rebuild` on the menu below does absolutely nothing:

image

src/pytest/__main__.py
```diff
@@ -0,0 +1,6 @@
+"""
+pytest entry point
+"""
+import pytest
+
```
RonnyPfannschmidt: i vaguely recall its necessary to check for `__name__ == "__main__"` to guard against certain user mistakes

asottile:
the failure is quite weird, I can't reproduce it w/ 3.7.5x64 locally running `tox -e py37-twisted-numpy` :thinking:

src/pytest/__main__.py
nicoddemus: OK, will update it later, thanks

nicoddemus:
> the failure is quite weird, I can't reproduce it w/ 3.7.5x64 locally running

Me neither... just pushed a skip for that test to see if it is that test the culprit, or something else.

nicoddemus:
Weird, it is just that test who seems to be stucking CI... 🤔

nicoddemus:
Rá looking at the test things are clearer: the test uses `multiprocessing`, so having the `__main__` guard as @RonnyPfannschmidt pointed out is mandatory. 😁

PR #6307, opened 2019-12-02
Update mypy 0.740 -> 0.750

bluetech:
Release notes:
https://mypy-lang.blogspot.com/2019/11/mypy-0.html

Can remove existing workarounds, but unfortunately need to add some new ones.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -479,7 +479,7 @@ class ExceptionInfo(Generic[_E]):
         assert tup[1] is not None, "no current exception"
         assert tup[2] is not None, "no current exception"
         exc_info = (tup[0], tup[1], tup[2])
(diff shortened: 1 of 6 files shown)


src/_pytest/_code/code.py
```diff
@@ -479,7 +479,7 @@ def from_current(
         assert tup[1] is not None, "no current exception"
         assert tup[2] is not None, "no current exception"
         exc_info = (tup[0], tup[1], tup[2])
-        return cls.from_exc_info(exc_info, exprinfo)
+        return ExceptionInfo.from_exc_info(exc_info, exprinfo)
```
bluetech: This is needed because it seems now `cls` is bound to the `_E` generic which is not compatible with `BaseException`. Using the type directly "unbounds" it.

PR #6311, opened 2019-12-03
Some type annotation & check_untyped_defs fixes

bluetech:
The first two commits mostly finish annotating `_pytest/code/`.

The other commits fix `check_untyped_defs` errors in a few test files. This discovered a couple of problems in the annotations, which are fixed.

Most of the diff is adding `-> None` to test functions. Also, because of a deficiency in mypy, exceptions defined on functions e.g. `pytest.fail.Exception` don't work, so I replace them with the direct classes instead.

### DIFF
--- a/testing/test_pytester.py
+++ b/testing/test_pytester.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 import sys
 import time
+from typing import List
 
 import py.path
 
@@ -9,6 +10,7 @@ import _pytest.pytester as pytester
 import pytest
 from _pytest.config import PytestPluginManager
 from _pytest.main import ExitCode
+from _pytest.outcomes import Failed
 from _pytest.pytester import CwdSnapshot
 from _pytest.pytester import HookRecorder
 from _pytest.pytester import LineMatcher
@@ -16,7 +18,7 @@ from _pytest.pytester import SysModulesSnapshot
 from _pytest.pytester import SysPathsSnapshot
 
 
-def test_make_hook_recorder(testdir):
+def test_make_hook_recorder(testdir) -> None:
     item = testdir.getitem("def test_func(): pass")
     recorder = testdir.make_hook_recorder(item.config.pluginmanager)
     assert not recorder.getfailures()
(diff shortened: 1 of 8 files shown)


testing/test_pytester.py
```diff
@@ -24,51 +24,52 @@ def test_make_hook_recorder(testdir) -> None:
...
-    assert numpassed == 0
-    assert numskipped == 1
-    assert numfailed == 1
-    assert len(recorder.getfailedcollections()) == 1
-
-    recorder.unregister()
-    recorder.clear()
-    recorder.hook.pytest_runtest_logreport(report=rep3)
-    pytest.raises(ValueError, recorder.getfailures)
+    if False:  # pragma: no cover
```
blueyed: I'd rather leave the code unchanged and ignore the hint from codecov that this is uncovered.
(this avoids unnecessary diff churn etc)

bluetech: OK, removed. But from my understanding, because of how the coverage check works, whenever the file is touched, it will fail.

bluetech:
Rebased on latest features (no conflicts).

testing/test_pytester.py
blueyed: It "fails" when this part is contained in a diff.

testing/test_runner.py
```diff
@@ -11,11 +14,17 @@
...
 from _pytest.outcomes import OutcomeException
+from _pytest.outcomes import Skipped
+
+if False:  # TYPE_CHECKING
+    from typing import Type
 
 
 class TestSetupState:
-    def test_setup(self, testdir):
+    def test_setup(self, testdir) -> None:
```
blueyed: Wouldn't it be good to also add a type annotation for fixtures, i.e. "testdir" here?
In the long run it would be good to have this via a mypy plugin automatically (https://github.com/python/mypy/issues/6760).

bluetech: Definitely makes sense. In this PR in this file I just did a mechanical `-> None` + fixups. Fixtures like `testdir` can have a mechanical treatment too but maybe in another PR. I'm hoping to get this one merged so I can continue with more typing work.

blueyed:
> Most of the diff is adding `-> None` to test functions.

I'd rather see this enabled via mypy's config (see #6452), mostly to avoid diff churn.
What do you think?

testing/test_pytester.py
```diff
@@ -36,23 +38,23 @@ class rep:
...
         failed = False
         skipped = True
         when = "call"
 
-    rep.passed = False
-    rep.skipped = True
-    recorder.hook.pytest_runtest_logreport(report=rep)
+    rep2.passed = False
+    rep2.skipped = True
+    recorder.hook.pytest_runtest_logreport(report=rep2)
```
blueyed: Diff churn like this makes me think that `--allow-redefinition` would be better ([docs](https://mypy.readthedocs.io/en/latest/command_line.html#cmdoption-mypy-allow-redefinition)).
If not for "src" then maybe for "testing" at least, and maybe only per module there then?

testing/test_runner.py
```diff
@@ -537,33 +550,33 @@ def pytest_runtest_teardown(item):
...
+    with pytest.raises(Exit) as excinfo:
         pytest.exit("hello")
-    assert excinfo.errisinstance(pytest.exit.Exception)
+    assert excinfo.errisinstance(Exit)
 
 
-def test_pytest_fail():
-    with pytest.raises(pytest.fail.Exception) as excinfo:
+def test_pytest_fail() -> None:
+    with pytest.raises(Failed) as excinfo:
```
blueyed: Maybe some extra check like `assert isinstance(Failed, pytest.fail.Exception)` might make sense (ensuring that this is still the case really)?

blueyed:
@bluetech 
I also think it makes sense to rebase this against master.

testing/test_pytester.py
bluetech: I personally think this makes the code less confusing. I generally prefer to fix than to use `allow-redefinition`, for clarity and for the occasional errors it catches. Though it might be slightly annoying sometimes.

testing/test_runner.py
bluetech: Sure, added.

PR #6313, opened 2019-12-03
Fix assertion rewriting module detection for egg dists

### LINKED ISSUE #6301: Module re-write doesn't work with non dist-info based installations
More context behind this issue is available at: https://github.com/pytest-dev/pytest-mock/issues/167

TL;DR:

The function `_iter_rewritable_modules` doesn't detect modules that can be rewritten, if they are installed by a method which doesn't adopt the dist-info format.

An easy way to reproduce the problem is to: `pip install pytest-mock` in one environment and `pip install -e 'git+https://github.com/pytest-dev/pytest-mock#egg=pytest-mock'` in another and compare the output for:
```python
import importlib_metadata
from _pytest.config import _iter_rewritable_modules

for x in importlib_metadata.distributions():
  if x.metadata['Name']=='pytest-mock':
    for _file in x.files:
      print("file: {}; module_or_pkg_name: {}".format(str(_file), list(_iter_rewritable_modules([str(_file)]))))
```

Because of this problem, rpm maintainers are unable to run the tests for pytest-mock, since they rely on `python setup.py install` which creates egg-info directories.

nicoddemus:
Fix #6301

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -630,15 +630,66 @@ notset = Notset()
 
 
 def _iter_rewritable_modules(package_files):
+    """
+    Given an iterable of file names in a source distribution, return the "names" that should
+    be marked for assertion rewrite (for example the package "pytest_mock/__init__.py" should
+    be added as "pytest_mock" in the assertion rewrite mechanism.
+
+    This function has to deal with dist-info based distributions and egg based distributions
+    (which are still very much in use for "editable" installs).
+
+    Here are the file names as seen in a dist-info based distribution:
+
+        pytest_mock/__init__.py
+        pytest_mock/_version.py
+        pytest_mock/plugin.py
+        pytest_mock.egg-info/PKG-INFO
+
+    Here are the file names as seen in an egg based distribution:
+
+        src/pytest_mock/__init__.py
+        src/pytest_mock/_version.py
+        src/pytest_mock/plugin.py
+        src/pytest_mock.egg-info/PKG-INFO
+        LICENSE
+        setup.py
+
+    We have to take in account those two distribution flavors in order to determine which
+    names should be considered for assertion rewriting.
+
+    More information:
+        https://github.com/pytest-dev/pytest-mock/issues/167
+    """
+    package_files = list(package_files)
+    seen_some = False
     for fn in package_files:
         is_simple_module = "/" not in fn and fn.endswith(".py")
         is_package = fn.count("/") == 1 and fn.endswith("__init__.py")
         if is_simple_module:
             module_name, _ = os.path.splitext(fn)
-            yield module_name
+            # we ignore "setup.py" at the root of the distribution
+            if module_name != "setup":
+                seen_some = True
+                yield module_name
         elif is_package:
             package_name = os.path.dirname(fn)
+            seen_some = True
             yield package_name
+
+    if not seen_some:
+        # at this point we did not find any packages or modules suitable for assertion
+        # rewriting, so we try again by stripping the first path component (to account for
+        # "src" based source trees for example)
+        # this approach lets us have the common case continue to be fast, as egg-distributions
+        # are rarer
+        new_package_files = []
+        for fn in package_files:
+            parts = fn.split("/")
+            new_fn = "/".join(parts[1:])
+            if new_fn:
+                new_package_files.append(new_fn)
+        if new_package_files:
+            yield from _iter_rewritable_modules(new_package_files)
 
 
 class Config:
--- a/changelog/6301.bugfix.rst
+++ b/changelog/6301.bugfix.rst
@@ -1,0 +1,1 @@
+Fix assertion rewriting for egg-based distributions and ``editable`` installs (``pip install --editable``).
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -422,15 +422,21 @@ class TestConfigAPI:
     @pytest.mark.parametrize(
         "names, expected",
         [
+            # dist-info based distributions root are files as will be put in PYTHONPATH
             (["bar.py"], ["bar"]),
-            (["foo", "bar.py"], []),
-            (["foo", "bar.pyc"], []),
-            (["foo", "__init__.py"], ["foo"]),
-            (["foo", "bar", "__init__.py"], []),
+            (["foo/bar.py"], ["bar"]),
+            (["foo/bar.pyc"], []),
+            (["foo/__init__.py"], ["foo"]),
+            (["bar/__init__.py", "xz.py"], ["bar", "xz"]),
+            (["setup.py"], []),
+            # egg based distributions root contain the files from the dist root
+            (["src/bar/__init__.py"], ["bar"]),
+            (["src/bar/__init__.py", "setup.py"], ["bar"]),
+            (["source/python/bar/__init__.py", "setup.py"], ["bar"]),
         ],
     )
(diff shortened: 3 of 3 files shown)


src/_pytest/config/__init__.py
```diff
@@ -630,16 +630,67 @@ def __repr__(self):
 
 
 def _iter_rewritable_modules(package_files):
+    """
+    Given an iterable of file names in a source distributions, return the "names" that should
```
bluetech:
```suggestion
    Given an iterable of file names in a source distribution, return the "names" that should
```

nicoddemus:
> I wonder if it is not possible to unify the two cases by using importlib_metadata a bit more? Maybe by getting the file name using file.locate() instead of str(file) in _mark_plugins_for_rewrite(), or maybe some other way - I'm not really familiar with that library.

Good point, that's something that looks like worth pursuing. @asottile (which knows a lot more about `importlib_metadata` than I) any input here?

> Is there a reason not to just strip to at most the two last components (so it catches foo.py and foo/__init__.py) instead of stripping one component at a time and trying that? So source/python/bar/__init__.py -> bar/__init__.py instead of source/python/bar/__init__.py -> python/bar/__init__.py -> bar/__init__.py.

> Does the function need to handle subpackages, e.g. pytest_mock.foo._version? If such a file existed, wouldn't it conflict with pytest_mock._version? Or maybe subpackages are considered different "distributions"?

All good points, the original intent was to only mark for rewrite the top level package, because `register_assert_rewrite` only needs the top-level package, but adding the subpackages is harmless (it will only increase the size of the set() used to hold the names). 

But on second thought, perhaps we don't need to be so nit-picky here and just add all modules/packages we find?

src/_pytest/config/__init__.py
```diff
@@ -630,16 +630,67 @@ def __repr__(self):
...
+                yield module_name
         elif is_package:
             package_name = os.path.dirname(fn)
+            seen_some = True
             yield package_name
 
+    if not seen_some:
+        # at this point we did not find any packages or modules suitable for assertion
+        # rewriting, so we try again by stripping the first path component (to account for
+        # "src" based source trees for example)
```
asottile: we should not do this arbitrarily -- `src` is just a use of `package_dir` but it could be anywhere or any number of directories deep

perhaps a better approach would be to use `sys.path` to determine whether they are modules or not?

src/_pytest/config/__init__.py
```diff
@@ -630,16 +630,67 @@ def __repr__(self):
...
+    names should be considered for assertion rewriting.
+
+    More information:
+        https://github.com/pytest-dev/pytest-mock/issues/167
+    """
+    package_files = list(package_files)
+    seen_some = False
     for fn in package_files:
         is_simple_module = "/" not in fn and fn.endswith(".py")
         is_package = fn.count("/") == 1 and fn.endswith("__init__.py")
```
asottile: hmmm the existing code seems to be a pre-PEP420 world, I wonder if we could update this to not be so dependent on `__init__.py` files

asottile:
though it is interesting that `importlib_metadata` is handing out the `SOURCES` instead of the `installed-files` metadata, that feels like a bug -- maybe @jaraco has some ideas there

src/_pytest/config/__init__.py
nicoddemus: > we should not do this arbitrarily -- src is just a use of package_dir but it could be anywhere or any number of directories deep

That's just a comment, we are not hard-coding `src` anywhere... we do strip the first directory (if we don't find a suitable mode/package) and try again in the next iteration.

src/_pytest/config/__init__.py
nicoddemus: Not sure, pytest itself doesn't seem to support this and AFAIR many tools were struggling to add support to this as well...

nicoddemus:
> I'm of the opinion that we should not support eggs

That was my initial reaction as well, but `pip install --editable` still works with eggs, and that's something we definitely should support IMO (that was the origin of the issue, the assertion rewriter was not rewriting pytest-mock's own files in develop mode).

asottile:
makes sense, we can probably iterate `installed-files.txt` instead of what `importlib_metadata` hands us to fix this?

nicoddemus:
> makes sense, we can probably iterate installed-files.txt instead of what importlib_metadata hands us to fix this?

I'm not sure what you mean exactly, but I think whatever you think is appropriate here can be done; this code dates from our 2016 sprint, if there are improvements to be done to it we should.

Can you post an example of using `installed-files.txt`and what it outputs for a package? Also if you are inclined to push to this PR please do so. 😁

asottile:
here's an example:
```
$ tail -n999 venv/lib/python3.6/site-packages/pytest_mock-1.12.1-py3.6.egg-info/{installed-files,SOURCES}.txt
==> venv/lib/python3.6/site-packages/pytest_mock-1.12.1-py3.6.egg-info/installed-files.txt <==
../pytest_mock/__init__.py
../pytest_mock/__pycache__/__init__.cpython-36.pyc
../pytest_mock/__pycache__/_version.cpython-36.pyc
../pytest_mock/__pycache__/plugin.cpython-36.pyc
../pytest_mock/_version.py
../pytest_mock/plugin.py
PKG-INFO
SOURCES.txt
dependency_links.txt
entry_points.txt
requires.txt
top_level.txt

==> venv/lib/python3.6/site-packages/pytest_mock-1.12.1-py3.6.egg-info/SOURCES.txt <==
.gitignore
.pre-commit-config.yaml
CHANGELOG.rst
HOWTORELEASE.rst
LICENSE
README.rst
setup.cfg
setup.py
tox.ini
.github/FUNDING.yml
.github/workflows/main.yml
src/pytest_mock/__init__.py
src/pytest_mock/_version.py
src/pytest_mock/plugin.py
src/pytest_mock.egg-info/PKG-INFO
src/pytest_mock.egg-info/SOURCES.txt
src/pytest_mock.egg-info/dependency_links.txt
src/pytest_mock.egg-info/entry_points.txt
src/pytest_mock.egg-info/requires.txt
src/pytest_mock.egg-info/top_level.txt
tests/test_pytest_mock.py
```
though hmmm, that file doesn't show up for editable installs so now I'm just confused :thinking:

nicoddemus:
> here's an example

Hmmm can I access that contents through importlib-metadata, or only by accessing the filesystem directly?

nicoddemus:
I think we can go with the PR as is: it solves the existing problem without incurring additional overhead; we can improve this later if we find a more elegant way to accomplish this. 👍

nicoddemus:
Backported in https://github.com/pytest-dev/pytest/pull/6368

amatissart:
This change causes a new warning in one of our projects, that uses `lovely-pytest-docker==0.1.0` as a plugin. It looks like this plugin is packaged with a submodule named `pytest`: 
```
lovely/pytest/__init__.py
lovely/pytest/docker/__init__.py
lovely/pytest/docker/compose.py
...
```

As a result this warning is now triggered
```
/lib/python3.6/site-packages/_pytest/config/__init__.py:891
  /lib/python3.6/site-packages/_pytest/config/__init__.py:891: PytestAssertRewriteWarning: 
Module already imported so cannot be rewritten: pytest
    self._mark_plugins_for_rewrite(hook)
```
which is probably harmless, but still confusing.

I am not familiar with the details of python packages ; is there a problem with how this specific dependency is packaged ?

asottile:
yeah they probably should not use `pytest` (our namespace) as their namespace

Issue #5471, opened 2019-06-23
Hostname and timestamp properties in generated JUnit XML reports


anaskhan96:
Pytest enables generating JUnit XML reports of the tests.

However, there are some properties missing, specifically `hostname` and `timestamp` from the `testsuite` XML element. Is there an option to include them?

Example of a pytest XML report:
```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="2" name="check" skipped="0" tests="4" time="0.049">
	<testcase classname="test_sample.TestClass" file="test_sample.py" line="3" name="test_addOne_normal" time="0.001"></testcase>
	<testcase classname="test_sample.TestClass" file="test_sample.py" line="6" name="test_addOne_edge" time="0.001"></testcase>
</testsuite>
```

Example of a junit XML report:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="location.GeoLocationTest" tests="2" skipped="0" failures="0" errors="0" timestamp="2019-04-22T10:32:27" hostname="Anass-MacBook-Pro.local" time="0.048">
  <properties/>
  <testcase name="testIoException()" classname="location.GeoLocationTest" time="0.044"/>
  <testcase name="testJsonDeserialization()" classname="location.GeoLocationTest" time="0.003"/>
  <system-out><![CDATA[]]></system-out>
  <system-err><![CDATA[]]></system-err>
</testsuite>
```

samueljsb:
Fixed by #5692

anaskhan96:
@samueljsb thanks!

KellyWalker:
Can the fix for this be back-ported to a 4.6 patch release?

@nicoddemus did something similar for #5594 

See my request in that ticket for details on my request:
https://github.com/pytest-dev/pytest/pull/5550#issuecomment-558797460

nicoddemus:
Thanks @KellyWalker for the ping. 

Opened: https://github.com/pytest-dev/pytest/pull/6332

nicoddemus:
Done! It will be available on the next 4.6.X release. 👍

KellyWalker:
Most excellent!

symbiont-eli-wang:
Can we add timestamp to each test case?

nicoddemus:
> Can we add timestamp to each test case?

You mean each `TestCase` subclass? Hmmm it is not trivial to do it, I think.

KellyWalker:
[...]

PR #6332, opened 2019-12-10
[4.6] Add hostname and timestamp to JUnit XML testsuite tag (#5692)

nicoddemus:
Add hostname and timestamp to JUnit XML testsuite tag

Conflicts:
  	testing/test_junitxml.py

Requested from https://github.com/pytest-dev/pytest/issues/5471#issuecomment-563500493

### DIFF
--- a/changelog/5471.feature.rst
+++ b/changelog/5471.feature.rst
@@ -1,0 +1,1 @@
+JUnit XML now includes a timestamp and hostname in the testsuite tag.
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -15,9 +15,11 @@ from __future__ import print_function
 
 import functools
 import os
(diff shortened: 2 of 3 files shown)


nicoddemus:
OK to merge: the py34 failure on Azure can be ignore, it seems Azure no longer supports Python 3.4:

```
##[error]Version spec 3.4 for architecture x64 did not match any version in Agent.ToolsDirectory.
Versions in C:/hostedtoolcache/windows:
2.7.16 (x86)
3.5.4 (x86)
3.6.8 (x86)
3.7.5 (x86)
3.8.0 (x86)
2.7.16 (x64)
3.5.4 (x64)
3.6.8 (x64)
3.7.5 (x64)
3.8.0 (x64)
```

But it passes on Travis so we're good.

PR #6333, opened 2019-12-11
Deprecate no print logs

vinaycalastry:
Deprecated the --no-print-logs option.
Added 2 unit tests, for ini and command line.
Added 3238.deprecation.rst file.
Updated the deprecations.rst doc.

--Recreated PR as i messed up the old one

### DIFF
--- a/doc/en/deprecations.rst
+++ b/doc/en/deprecations.rst
@@ -20,6 +20,20 @@ Below is a complete list of all pytest features which are considered deprecated.
 :ref:`standard warning filters <warnings>`.
 
 
+``--no-print-logs`` command-line option
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 5.4
+
+
+Option ``--no-print-logs`` is deprecated and meant to be removed in a future release. If you use ``--no-print-logs``, please try out ``--show-capture`` and
+provide feedback.
+
+``--show-capture`` command-line option was added in ``pytest 3.5.0` and allows to specify how to
+display captured output when tests fail: ``no``, ``stdout``, ``stderr``, ``log`` or ``all`` (the default).
+
+
+
 Node Construction changed to ``Node.from_parent``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/testing/deprecated_test.py
+++ b/testing/deprecated_test.py
@@ -90,3 +90,44 @@ def test_node_direct_ctor_warning():
         nodes.Node(name="test", config=ms, session=ms, nodeid="None")
     assert w[0].lineno == inspect.currentframe().f_lineno - 1
     assert w[0].filename == __file__
+
+
+def assert_no_print_logs(testdir, args):
+    result = testdir.runpytest(*args)
+    result.stdout.fnmatch_lines(
+        [
+            "*--no-print-logs is deprecated and scheduled for removal in pytest 6.0*",
+            "*Please use --show-capture instead.*",
+        ]
+    )
+
+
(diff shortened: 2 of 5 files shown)


asottile:
looks like your branch is accidentally based on `master` but targetting `features` -- could you rebase on top of features?

should be able to

```
git rebase -i upstream/features
```
and then remove any commits which aren't yours

(let me know and I can do this for you and push to your branch)

vinaycalastry:
I think I've fixed this now. Could you please recheck.

testing/deprecated_test.py
```diff
@@ -90,3 +90,34 @@ class MockConfig:
...
+        args = ()
+        testdir.makeini(
+            """
+            [pytest]
+            log_print=False
+            """
+        )
+    else:
+        assert config_mode == "cmdline"
+        args = ("--no-print-logs",)
```
asottile: I tend to try and avoid logic in tests, I'd probably write this as two separate tests instead of this

tests should be as simple as possible for the programmer to read

here's a blog post I often link on this subject (a little more severe of a case there, but still the same sentiment): https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

nicoddemus: Originally @vinaycalastry wrote two separate tests, and it was I who asked him to unify in a single one.

I agree we should avoid too much logic in tests in general, but I tend to parametrize tests when only the setup logic is involved (like here).

nicoddemus: Btw if you feel strongly here I'm perfectly fine to change it back to two tests, as I also appreciate the simplicity, although I'm not a big fan of the copy/paste involved then.

asottile: I'd rather see the assertion as a helper function than a stringly typed if statement -- no need for duplication and it's very clear the execution path

vinaycalastry: I can add a helper function and amend my commit.

vinaycalastry: Changed to use a helper function. Please review

asottile: I meant for a helper for the bottom half and don't use parametrize or an if statement

nicoddemus: (@asottile I suggest using the "request changes" feature so it is explicit that this PR is awaiting further changes 👍)

vinaycalastry: Changed it.

vinaycalastry: Please review

nicoddemus:
Thanks a lot @vinaycalastry!

nicoddemus:
Merging as the failures are unrelated. 👍

doc/en/deprecations.rst
```diff
@@ -20,6 +20,20 @@ Below is a complete list of all pytest features which are considered deprecated.
...
+``--no-print-logs`` command-line option
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 5.4
+
+
+Option ``--no-print-logs`` is deprecated and meant to be removed in a future release. If you use ``--no-print-logs``, please try out ``--show-capture`` and
+provide feedback.
+
+``--show-capture`` command-line option was added in ``pytest 3.5.0` and allows to specify how to
```
TheMeaningfulEngineer, line 32: Hi there,
all the options aren't documented in the [latest docs](https://docs.pytest.org/en/latest/capture.html#setting-capturing-methods-or-disabling-capturing) at the time of writing.
However the options mentioned here do seem to work regardless.

i.e. 
`pytest --show-capture=stdout` does work but isn't mentioned in the docs.

TheMeaningfulEngineer: Mixed up `show-capture` and `capture`. Deleting comments :)

Issue #6352, opened 2019-12-18
#5990 breaks pytester


aragilar:
It looks like #5990 broke pytester, by changing "error" to "errors". 

MWE follows (use `pytest -p pytester test_fail.py` to run the code below)

```python
CODE = """
    import pytest

    @pytest.fixture
    def bad_fixture():
        raise Exception("bad")

    def test_error1(bad_fixture):
        pass

    def test_error2(bad_fixture):
        pass
"""

def test_failfail(testdir):
    testdir.makepyfile(CODE)
    result = testdir.runpytest()
    result.assert_outcomes(error=2)
```

Zac-HD:
Hmm, not quite sure *when* it broke, but changing 

https://github.com/pytest-dev/pytest/blob/8be9684ab23220cc85abb06e06414961667340b3/src/_pytest/pytester.py#L459

to 

```python
        "error": d.get("error", 0) + d.get("errors", 0), 
```

should fix it :smile:

nicoddemus:
Yep, anybody up for a quick PR? 😁

mulatinho:
I would like to be more involved with pytest, here's a PR :)

PR #6353, opened 2019-12-18
pytester: quick fix error introduced in #5990

mulatinho:
- Fix error introduced on #5990 

Fixes: #6352

### DIFF
--- a/changelog/6532.bugfix.rst
+++ b/changelog/6532.bugfix.rst
@@ -1,0 +1,1 @@
+Fix problem with ``testdir`` not recognizing errors correctly in runs with a single test.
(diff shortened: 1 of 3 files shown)


changelog/6532.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixing error introducted by #5990
```
nicoddemus:
```suggestion
Fix problem with ``testdir`` not recognizing errors correctly in runs with a single test.
```

mulatinho:
@nicoddemus thanks for the feedback, I added one test to avoid that as you requested.

PR #6355, opened 2019-12-18
Add GitHub actions for CI

nicoddemus:
This includes our current full matrix (windows, linux and macos), for evaluting
purposes.

We should disconsider failures when evaluating PRs.

TODO:

- deploy
- coverage
- github release notes

Even with the above missing, I still believe it would be nice to merge
this and have GitHub actions working in parallel so we can evaluate performance
and usability from now on.

Unfortunately due to the way GitHub actions work, the jobs are not triggered until GitHub actions are enabled on `master`, but the results of this build can be seen here:

https://github.com/nicoddemus/pytest/pull/11

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,123 @@
+# evaluating GitHub actions for CI, disconsider failures when evaluating PRs
+#
+# this is still missing:
+# - deploy
+# - coverage
+# - upload github notes
+#
+name: main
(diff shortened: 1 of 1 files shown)


nicoddemus:
> Is it worth adding MacOS for python 3.8 while you're here?

Done! It is now building, hopefully MacOS Python 3.8 will also pass. 🤞

PR #6357, opened 2019-12-19
invocation in last section 'pythonpath.rst' title swapped

captainCapitalism:
The order of invocations 'python -m pytest' and 'pytest' are different in the header and the explanation. Me being lazy reading about the behaviour of 'former' looked up quickly the title and rushed to implementation to discover it actually works the other way - as stated in the documentation. So I propose to switch the order in the title to achieve consistent ordering and not confusing somebody like me again! :)

### DIFF
--- a/doc/en/pythonpath.rst
+++ b/doc/en/pythonpath.rst
@@ -74,7 +74,7 @@ This is also discussed in details in :ref:`test discovery`.
 
 .. _`pytest vs python -m pytest`:
 
-Invoking ``pytest`` versus ``python -m pytest``
+Invoking ``python -m pytest`` versus ``pytest`` 
 -----------------------------------------------
 
 Running pytest with ``python -m pytest [...]`` instead of ``pytest [...]`` yields nearly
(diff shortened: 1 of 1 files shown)


doc/en/pythonpath.rst
```diff
@@ -74,7 +74,7 @@ This is also discussed in details in :ref:`test discovery`.
 
 .. _`pytest vs python -m pytest`:
 
-Invoking ``pytest`` versus ``python -m pytest``
+Invoking ``python -m pytest`` versus ``pytest``
```
asottile:
```suggestion
Invoking ``python -m pytest`` versus ``pytest``
```

the `pre-commit` hooks can fix this too :)

blueyed:
Note that the anchor is "pytest vs python -m pytest" (and should stay that way (permalink)), and therefore it might be better to change the order in the text, not the heading.

nicoddemus:
> Note that the anchor is "pytest vs python -m pytest" (and should stay that way (permalink)), and therefore it might be better to change the order in the text, not the heading.

Done

nicoddemus:
Thanks @captainCapitalism!

PR #6360, opened 2019-12-19
Preparing release version 4.6.8

asottile:


asottile:
I'm going to finish the release when travis completes -- the coverage issue is trickier to work around than I thought due to the way we conditionally install `coverage-enable-subprocess` (I'd want to add `coverage!=5.0` on windows, but there isn't a way to do that without either (1) always installing coverage or (2) rewriting the conditional stuff)

nicoddemus:
Sounds good to me!

PR #6384, opened 2019-12-30
Make --showlocals work together with --tb=short

### LINKED ISSUE #494: "-l" does not dump variables where "--tb" is not "long"
Originally reported by: **Ionel Cristian Mărieș (BitBucket: [ionelmc](http://bitbucket.org/ionelmc), GitHub: [ionelmc](http://github.com/ionelmc))**

---

Example output

```
    > assert hist == ['before', 'the-return-valuex', 'after', 'finally', 'closed']
    E assert ['before', 't...ly', 'closed'] == ['before', 'th...ly', 'closed']
    E At index 1 diff: 'the-return-value' != 'the-return-valuex'
```

I use these options:

```
 -rxEfsl
--strict
--ignore docs/conf.py
--ignore setup.py
--doctest-modules
--doctest-glob \*.rst
--tb short
```

---
- Bitbucket: https://bitbucket.org/pytest-dev/pytest/issue/494

pv:
Enable showing local variables when asked to do so in the short
traceback mode.

Fixes #494

Looks like this in practice:
```
$ python3 -mpytest scipy/_lib/tests/test__util.py --tb=short -l
...
================================ FAILURES =================================
___________________________ test__aligned_zeros ___________________________
scipy/_lib/tests/test__util.py:49: in test__aligned_zeros
    check(shape, dtype, order, align)
        align      = 1
        check      = <function test__aligned_zeros.<locals>.check at 0x7f866a141560>
        dtype      = <class 'numpy.uint8'>
        j          = 0
        n          = 0
        niter      = 10
        order      = 'C'
        shape      = 0
scipy/_lib/tests/test__util.py:22: in check
    assert dtype is None
E   AssertionError: assert <class 'numpy.uint8'> is None
        align      = 1
        dtype      = <class 'numpy.uint8'>
        err_msg    = "(0, <class 'numpy.uint8'>, 'C', 1)"
        order      = 'C'
        shape      = 0
        x          = array([], dtype=uint8)
====================== 1 failed, 11 passed in 0.23s =======================
```

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -785,9 +785,7 @@ class FormattedExcinfo:
                 message = excinfo and excinfo.typename or ""
             path = self._makepath(entry.path)
             filelocrepr = ReprFileLocation(path, entry.lineno + 1, message)
-            localsrepr = None
-            if not short:
-                localsrepr = self.repr_locals(entry.locals)
+            localsrepr = self.repr_locals(entry.locals)
             return ReprEntry(lines, reprargs, localsrepr, filelocrepr, style)
         if excinfo:
             lines.extend(self.get_exconly(excinfo, indent=4))
@@ -1044,6 +1042,8 @@ class ReprEntry(TerminalRepr):
             for line in self.lines:
                 red = line.startswith("E   ")
                 tw.line(line, bold=True, red=red)
+            if self.reprlocals:
+                self.reprlocals.toterminal(tw, indent=" " * 8)
             return
         if self.reprfuncargs:
             self.reprfuncargs.toterminal(tw)
@@ -1085,9 +1085,9 @@ class ReprLocals(TerminalRepr):
     def __init__(self, lines: Sequence[str]) -> None:
         self.lines = lines
 
-    def toterminal(self, tw) -> None:
+    def toterminal(self, tw, indent="") -> None:
         for line in self.lines:
-            tw.line(line)
+            tw.line(indent + line)
 
 
 class ReprFuncArgs(TerminalRepr):
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -670,6 +670,26 @@ class TestTerminalFunctional:
             ]
         )
 
+    def test_showlocals_short(self, testdir):
+        p1 = testdir.makepyfile(
+            """
+            def test_showlocals_short():
+                x = 3
+                y = "xxxx"
+                assert 0
+        """
+        )
+        result = testdir.runpytest(p1, "-l", "--tb=short")
+        result.stdout.fnmatch_lines(
+            [
+                "test_showlocals_short.py:*",
+                "    assert 0",
+                "E   assert 0",
+                "        x          = 3",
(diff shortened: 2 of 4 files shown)


changelog/6384.bugfix.rst
```diff
@@ -0,0 +1 @@
+Make --showlocals work also with --tb=short
```
blueyed:
```suggestion
Make `--showlocals` work also with `--tb=short`.
```

testing/test_terminal.py
```diff
@@ -668,6 +668,26 @@ def test_showlocals():
...
+        """
+        )
+        result = testdir.runpytest(p1, "-l", "--tb=short")
+        result.stdout.fnmatch_lines(
+            [
+                "test_showlocals_short.py:*",
+                "    assert 0",
+                "E   assert 0",
+                "        x* = 3",
+                "        y* = 'xxxxxx*",
```
blueyed: The `"*"` is not needed with `"x*"` and `"y*"`, is it?  (I think it is better to have explicit matching here)
Also with regard to the `"x" * 5000` - I would expect this to be not truncated (at least in verbose mode), so you could add it there via string formatting.

src/_pytest/_code/code.py
```diff
@@ -1044,6 +1042,8 @@ def toterminal(self, tw) -> None:
             for line in self.lines:
                 red = line.startswith("E   ")
                 tw.line(line, bold=True, red=red)
+            if self.reprlocals:
+                self.reprlocals.toterminal(tw, indent=" " * 8)
```
blueyed: I think this should have consistent indentation as with the case below, i.e. none currently.
(but when adding indentation it might be better to have 4 here (and below then probably also) maybe)

pv: I believe 8 is better here, as the source code is intended to 4, which makes it harder to visually scan it.

changelog/6384.bugfix.rst
pv: ok

testing/test_terminal.py
pv: Replaced with explicit test. The output here is truncated since there's no `-vv` there, but probably truncation is orthogonal to the change here.

changelog/6384.bugfix.rst
blueyed: Missed the dot.. ;)
(please squash into a single commit then also with the next update - I'd await approval/feedback from others first)

pv: Right

changelog/6384.bugfix.rst
```diff
@@ -0,0 +1 @@
+Make `--showlocals` work also with `--tb=short`.
```
nicoddemus: * I think this should be an `improvement`, the code clearly intended to hide locals when using `--tb=short`. 

* Because of the above, this should target `features`.

Other than that LGTM 👍

pv: Moved to features.

blueyed:
Thank you!

PR #6388, opened 2020-01-01
Update mypy 0.750 -> 0.761

bluetech:
This fixes some type: ignores due to typeshed update.

Newer mypy seem to ignore unannotated functions better, so add a few
minor annotations so that existing correct type:ignores make sense.


blueyed:
`testing/code/test_source.py::TestSourceParsingAndCompiling::test_compilefuncs_and_path_sanity` fails.

bluetech:
Thanks for review @blueyed!

Issue #6363, opened 2019-12-20
strange handling of test ids for parametrize


graingert:
```
$ cat tests/test_toast.py
import pytest

@pytest.mark.parametrize(
    "ham,spam",
    [
        ("/bacon/../", u"//bacon/../"),
        ("/bacon/../../", u"//bacon/../../"),
        ("/bacon//", u"//bacon//"),
    ],
)
def test_eggs(ham,spam):
    pass
==================================== test session starts ====================================
platform linux -- Python 3.7.5, pytest-5.3.2, py-1.8.0, pluggy-0.12.0 -- /home/graingert/.virtualenvs/pytest/bin/python3.7
cachedir: .pytest_cache
rootdir: /home/graingert/projects/sausage/tests, inifile: pytest.ini
collected 3 items                                                                           

tests/test_toast.py::test_eggs[/-/] PASSED                                            [ 33%]
tests/] PASSED                                                                        [ 66%]
tests/test_toast.py::test_eggs[/bacon/-/bacon/] PASSED                                [100%]

===================================== 3 passed in 0.01s =====================================
```

- [x] a detailed description of the bug or suggestion
- [ ] output of `pip list` from the virtual environment you are using
- [x] pytest and operating system versions
- [x] minimal example if possible

graingert:
of particular concern is `("/bacon/../../", u"//bacon/../../")` -> `tests/] PASSED                                                                        [ 66%]`

graingert:
this also happens on pytest==4.6.5

Zac-HD:
FYI you can [syntax-highlight code blocks](https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks).

Not sure what's happening here but I agree it's not good!

graingert:
> FYI you can [syntax-highlight code blocks](https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks).

I know but GitHub bash syntax highlighting is not smart enough to detect `cat` and file extensions

asottile:
if you use


```markdown

    ```console
    $ foo
    bar
    ```
```

it will highlight as

```console
$ foo
bar
```

blueyed:
@graingert 
Not sure what the problem here is, but it looks like this for me, i.e. the params are joined using "-" (which appears to not be the case for you I guess?):

```
============================= test session starts ==============================
[...]

PR #6394, opened 2020-01-03
add test for #6363 (strange handling of test ids for parametrize)

graingert:
https://github.com/pytest-dev/pytest/issues/6363

### DIFF
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -97,6 +97,31 @@ class TestTerminal:
             ["    def test_func():", ">       assert 0", "E       assert 0"]
         )
 
+    def test_parametrize_relative_path(self, testdir):
+        """
+        Test demonstrating #6363
+        """
+        tests_subdir = testdir.mkdir("tests")
+        tests_subdir.join("pytest.ini").write("")
+        tests_subdir.join("test_f.py").write(
+            textwrap.dedent(
+                """\
+                import pytest
+
+                @pytest.mark.parametrize("p", ["/p/../", "/p/../../", "/p//"])
+                def test_fn(p):
+                    pass
+                """
+            )
+        )
+        testdir.runpytest("tests", "--verbosity=1").stdout.re_match_lines(
+            [
+                r"tests/test_f.py::test_fn\[/p/\.\./] PASSED \s+ \[ 33%\]",
+                r"tests/test_f.py::test_fn\[/p/\.\./\.\./] PASSED \s+ \[ 66%\]",
+                r"tests/test_f.py::test_fn\[/p//] PASSED \s+ \[100%\]",
+            ]
+        )
+
     def test_internalerror(self, testdir, linecomp):
         modcol = testdir.getmodulecol("def test_one(): pass")
         rep = TerminalReporter(modcol.config, file=linecomp.stringio)
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -827,11 +827,15 @@ class Config:
                 sys.stderr.flush()
 
     def cwd_relative_nodeid(self, nodeid):
+        if self.invocation_dir == self.rootdir:
+            return nodeid
         # nodeid's are relative to the rootpath, compute relative to cwd
-        if self.invocation_dir != self.rootdir:
-            fullpath = self.rootdir.join(nodeid)
-            nodeid = self.invocation_dir.bestrelpath(fullpath)
-        return nodeid
+        filepath, separator, rest = nodeid.partition("::")
+        return (
+            self.invocation_dir.bestrelpath(self.rootdir.join(filepath))
+            + separator
+            + rest
+        )
 
     @classmethod
     def fromdictargs(cls, option_dict, args):


testing/test_terminal.py
```diff
@@ -97,6 +97,20 @@ def test_func():
...
+            """
+            import pytest
+
+            @pytest.mark.parametrize("param", [("/p/../../", )])
+            def test_func(param):
+                pass
+            """
+        )
+        testdir.runpytest(*Option(verbosity=1).args).stdout.fnmatch_lines(
+            ["*test_parametrize_relative_path.py::test_func[/p/../../] PASSED*"]
```
graingert: actually not sure what this should be - perhaps 
```suggestion
            ["*test_parametrize_relative_path.py::test_func[p] PASSED*"]
```

graingert: not sure what cleanup is *intended*

testing/test_terminal.py
```diff
@@ -97,6 +97,40 @@ def test_func():
...
+    @pytest.mark.parametrize(
+        "write_pytest_ini,specify_path",
+        [
+            (False, False),
+            (False, True),
+            (True, False),
+            pytest.param(True, True, marks=pytest.mark.xfail),
+        ],
+    )
+    def test_parametrize_relative_path(self, testdir, write_pytest_ini, specify_path):
```
asottile: I find this test incredibly difficult to read, please [avoid logic in tests](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html) -- they should be clear and readable

In this case there's a lot of things that make this particularly hard to comprehend:
- [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) `parametrize`
- branching logic in tests
- conditional xfail
- splat of ternary list

I'd suggest writing this as separate tests (I'm not even sure all four combinations are helpful here, you can probably just use the failing combination since all the other variants are already passing)

nicoddemus: FTR I've wrote many of tests using logic like this one, and nowadays I tend to agree with @asottile that it should be avoided in general, using free functions to reuse logic instead.

graingert: I used this layout to document the specific case that causes the issue, and what cases that do not. I intend to remove the other parameterize entries once it's fixed

nicoddemus: Oh OK sounds good. 👍 

Do we intend to merge this as is then? If so I think we should add a comment stating the above (removing the parametrization cases that pass).

graingert: sure, can you `` ```suggestion `` what you think the comment should be?

I'd imagine we'd change this test a lot when the root cause is determined

nicoddemus:
```suggestion
    def test_parametrize_relative_path(self, testdir, write_pytest_ini, specify_path):
        """Test demonstrating #6363; the last parametrized case demonstrates the failure,
        We can probably get rid of the others or refactor the test 
        as commented in https://github.com/pytest-dev/pytest/pull/6394/files#r362861185.
        """
```

nicoddemus: Sure

testing/test_terminal.py
```diff
@@ -97,6 +97,44 @@ def test_func():
             ["    def test_func():", ">       assert 0", "E       assert 0"]
         )
 
+    @pytest.mark.parametrize(
+        "write_pytest_ini,specify_path",
+        [
+            (False, False),
+            (False, True),
+            (True, False),
+            pytest.param(True, True, marks=pytest.mark.xfail),
```
nicoddemus:
```suggestion
            pytest.param(True, True, marks=pytest.mark.xfail(strict=True, reason="#6363")),
```

nicoddemus:
I've made another suggestion (actually I could swear I've done it together with the other suggestion), can you please accept it and fix linting locally?

graingert:
> I've made another suggestion (actually I could swear I've done it together with the other suggestion), can you please accept it and fix linting locally?

done

nicoddemus:
Thanks! Waiting for @asottile to do the final review. 👍

blueyed:
I think we should rather fix the issue (using the single test that triggers it from here) - there's not much gain in merging and/or bike-shedding over some xfail test(s).

blueyed:
@graingert 
This appears to fix it:
```diff
diff --git i/src/_pytest/config/__init__.py w/src/_pytest/config/__init__.py
index 038695d98..81743e518 100644
--- i/src/_pytest/config/__init__.py
+++ w/src/_pytest/config/__init__.py
@@ -829,8 +829,10 @@ def notify_exception(self, excinfo, option=None):
     def cwd_relative_nodeid(self, nodeid):
         # nodeid's are relative to the rootpath, compute relative to cwd
         if self.invocation_dir != self.rootdir:
-            fullpath = self.rootdir.join(nodeid)
-            nodeid = self.invocation_dir.bestrelpath(fullpath)
+            assert "::" in nodeid, nodeid
+            fpath, rest = nodeid.split("::", 1)
+            fullpath = self.rootdir.join(fpath)
+            nodeid = self.invocation_dir.bestrelpath(fullpath) + "::" + rest
         return nodeid

     @classmethod
```
Please include it here, and only keep the test that triggered it (the xfail one).

blueyed:
@graingert 
I also suggest to remove/fix the comment:

> # nodeid's are relative to the rootpath, compute relative to cwd

(s/cwd/rootdir)

graingert:
@blueyed now behaviour is different on windows

blueyed:
Windows failure:
```
E            and: 'collecting ... collected 3 items'
E            and: ''
E            and: 'tests\\test_f.py::test_fn[/p/../] PASSED                                  [ 33%]'
E            and: 'tests\\test_f.py::test_fn[/p/../../] PASSED                               [ 66%]'
E            and: 'tests\\test_f.py::test_fn[/p//] PASSED                                    [100%]'
E            and: ''
E            and: '============================== 3 passed in 0.03s =============================='
E       remains unmatched: 'tests/test_f.py::test_fn\\[/p/\\.\\./] PASSED \\s+ \\[ 33%\\]'
```
Isn't that expected (to have backslashes there)?
I think `fnmatch_lines` handles it.

```diff
diff --git i/testing/test_terminal.py w/testing/test_terminal.py
index a2777e5b1..7e0e9de6c 100644
--- i/testing/test_terminal.py
+++ w/testing/test_terminal.py
@@ -98,27 +98,26 @@ def test_func():
         )
 
     def test_parametrize_relative_path(self, testdir):
-        """
-        Test demonstrating #6363
-        """
-        tests_subdir = testdir.mkdir("tests")
-        tests_subdir.join("pytest.ini").write("")
-        tests_subdir.join("test_f.py").write(
-            textwrap.dedent(
-                """\
-                import pytest
+        testdir.makepyfile(**{
+            "tests/test_f": """
+            import pytest
 
-                @pytest.mark.parametrize("p", ["/p/../", "/p/../../", "/p//"])
-                def test_fn(p):
-                    pass
-                """
-            )
-        )
-        testdir.runpytest("tests", "--verbosity=1").stdout.re_match_lines(
-            [
-                r"tests/test_f.py::test_fn\[/p/\.\./] PASSED \s+ \[ 33%\]",
-                r"tests/test_f.py::test_fn\[/p/\.\./\.\./] PASSED \s+ \[ 66%\]",
-                r"tests/test_f.py::test_fn\[/p//] PASSED \s+ \[100%\]",
+            @pytest.mark.parametrize("p", [
+                "/p/../",
+                "/p/../../",
+                "/p//",
+            ])
+            def test_fn(p):
+                pass
+            """
+        })
+        testdir.makefile(".ini", **{"tests/pytest": ""})
+        result = testdir.runpytest("tests", "-v")
+        result.stdout.fnmatch_lines([
+            "rootdir: */tests, inifile: pytest.ini",
+            "tests/test_f.py::test_fn[[]/p/../[]] PASSED *",
+            "tests/test_f.py::test_fn[[]/p/../../[]] PASSED *",
+            "tests/test_f.py::test_fn[[]/p//[]] PASSED *",
             ]
         )
```

blueyed:
> also creating a test with the path "test/foo::bar/test_foo.py" results in "test\foo::bar/test_foo.py" on windows which is very confusing

Yeah, it does not handle "::" in fpath properly I guess, but there are likely other places falling short there.  Maybe `rsplit()` would work better for this, but the main problem is already that `cwd_relative_nodeid` only gets the nodeid, but that needs more cleanup/refactor.

blueyed:
> also creating a test with the path "test/foo::bar/test_foo.py" results in "test\foo::bar/test_foo.py" on windows which is very confusing

This is without verbose display then?  I've thought ":" is not valid in a filename on Windows?
Worth adding a test for it for sure.  Will update the diff, but likely missed the point there.

graingert:
on windows: https://stackoverflow.com/a/31976060/833093

> : (colon - sometimes works, but is actually NTFS Alternate Data Streams)

graingert:
it's definitely possible to have "::" in a filename in pytest on linux though

graingert:
nodeid.rpartition won't help either - because you can put :: in a parameter id

blueyed:
ok, let's keep using `partition` then (like all the other places using `split(…)[0]`.
In the long run it should use the fspath/loc there, which is available even IIRC - but I think for now it's ok to have it looking "weird" with "::" in the fname (like other places likely also).  Might make sense to have an xfail for that then maybe.
Of course, if you feel like fixing it properly I'd not be against it.. :)

blueyed:
btw: I've found `_pytest.nodes._splitnode`, which might be a good start to improve splitting of nodes.  The idea there is to split by "/" first, and then the last part by "::".
This still fails short for when there are "/" or "::" in params, but maybe those (`[.*]$`) could be ignored there then (for the splitting).
Anyway, it seems like a good idea to start using this instead of `.split("::")[0]` I think.

blueyed:
I think `test_itemreport_subclasses_show_subclassed_file` is a good test in this regard also (hardened/extended in https://github.com/pytest-dev/pytest/pull/6467), because it maps "nodeid <- fspath" in the output there.

For the issue on Windows: the problem there is that nodeids are normalized with regard to slashes, which then does not get done again when making the "guessed path" fspath from it relative.

nicoddemus:
Hi @graingert!

Recently we decided to close some PRs which haven't seen an update in a while with the intent to clear up our queue.

In light of that let's close this for now but please feel free to reopen it when you are ready to continue work the here. Thanks again for this (and many other) contributions! 👍 

Cheers,

PR #6396, opened 2020-01-03
Fix output capture in junit xml report, extend options available

dxtr85:

### DIFF
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -167,51 +167,22 @@ class _NodeReporter:
         content_out = report.capstdout
         content_log = report.caplog
         content_err = report.capstderr
+        if self.xml.logging == "no":
+            return
+        if content_log and self.xml.logging in ["log", "all"]:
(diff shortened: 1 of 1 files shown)


nicoddemus:
Hi @dxtr85, sorry about the silence here, but what was the purpose of the PR?

dxtr85:
Hi @nicoddemus I closed it, since multiple tests have failed.
The purpose for it was to enable user to set what should be logged into junit file since for some (me included) this file is used only for report generation, hence log/out/err are not always welcome, and can take lots of space.

I wanted user to pick between no|log|system-out|system-err|out-err|all and put selected output in xml file.

I am trying to fix/extend the tests so that all pass and this new parameters are also verified.
Right now I am trying to figure out why junitxml does not write logs at INFO level and if/where it is configured.

nicoddemus:
I see, thanks

> I closed it, since multiple tests have failed.

Well that's actually normal, and it is OK to ask maintainers for help in that case.

Feel free to open a new PR once you feel like it/has a better understanding. 👍

PR #6406, opened 2020-01-05
[4.6] Remove unused _pytest.code.Source.isparseable function (#6405)

nicoddemus:
Remove unused _pytest.code.Source.isparseable function

### DIFF
--- a/changelog/6404.trivial.rst
+++ b/changelog/6404.trivial.rst
@@ -1,0 +1,1 @@
+Removed unused ``_pytest.code.Source.isparseable`` function.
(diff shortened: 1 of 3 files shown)


nicoddemus:
You are right... we can revert it easily, but a quick search shows that it doesn't seem to be used by code outside `pytest` (and which uses pytest I mean):

https://github.com/search?l=Python&o=desc&q=%22.isparseable%28%29%22+language%3APython+language%3APython+language%3APython+language%3APython&s=indexed&type=Code

nicoddemus:
Well let's just revert it then, as that's easy enough. 👍

PR #6409, opened 2020-01-06
terminal: refactor, no yellow ("boring") for non-last item

blueyed:
TODO:

- [x] do not use "yellow" ("boring status") for non-final items (e.g. progress indicator): if the first test (file) is "xfail" the progress should not be yellow there.
- [x] test(s):
  - [x] test progress color output with single xfail test ("100%" should be yellow)
- [x] check if this breaks pytest-parallel (due to changes to `self.stats`)

~~Blocked by: https://github.com/pytest-dev/pytest/pull/6536 (needs to be merged into features..)~~


nicoddemus:
Hey @blueyed,

> do not use "yellow" ("boring status") for non-final items (e.g. progress indicator): if the first test (file) is "xfail" the progress should not be yellow there.

Can you explain the rationale here in more details? Why only non-final items?

blueyed:
> Can you explain the rationale here in more details? Why only non-final items?

Currently it uses "main color" also for the progress indicator.
That is "yellow" after the first test passed (but is "xfail").  It should be "green" instead.  See the test.

Only if there's only a single "xfail" test it should be "yellow" (could need a test for this).

src/_pytest/terminal.py
```diff
@@ -457,33 +465,38 @@ def pytest_runtest_logreport(self, report: TestReport) -> None:
...
 
+    @property
+    def _is_last_item(self):
+        return len(self._progress_nodeids_reported) == self._testscollected
+
+    # XXX
+    if "xdist" in sys.modules:
+
+        def pytest_xdist_node_collection_finished(self, node, ids):
+            self._testscollected = len(ids)
```
blueyed: This is needed since `_session` is gone (and used by xdist).
I think it's good to remove that "hack", but it needs to only add the `pytest_xdist_node_collection_finished` for when it is recognized.
Can this be improved?

nicoddemus:
> Currently it uses "main color" also for the progress indicator.
That is "yellow" after the first test passed (but is "xfail"). It should be "green" instead. See the test.

> Only if there's only a single "xfail" test it should be "yellow" (could need a test for this).

Not sure I like the discrepancy... it just seems to make things harder to explain to people. Why having a special rule for this? Can you explain better your reasoning?

blueyed:
It's an existing/old rule: https://github.com/pytest-dev/pytest/blob/14005384767985f6abd2e8ec62d1f66f2fc8f205/testing/test_terminal.py#L1357-L1360

It applies to the final color though only, not the progress indicator.

It is confusing to see "[  0%]" in yellow (when there have been xfailed ones only), when the next line becomes green (after there is a passed one) - and the final stats line also being green, isn't it?
(that's the bug being fixed/addressed here)

nicoddemus:
Ohh now I get it, thanks! I didn't even remember this behavior at all. I agree it is a bit confusing. 👍

blueyed:
@nicoddemus thanks for the review, I am finishing this currently via https://github.com/blueyed/pytest/pull/151, i.e. it will get a bit more refactoring here likely then.

blueyed:
@nicoddemus please merge it yourself if you still approve with the 2 new commits.

src/_pytest/terminal.py
```diff
@@ -1012,6 +1030,54 @@ def show_skipped(lines: List[str]) -> None:
...
+                    unknown_types.append(found_type)
+        self._known_types = list(KNOWN_TYPES) + unknown_types
+        self._main_color = self._determine_main_color(bool(unknown_types))
+        return self._main_color, self._known_types
+
+    def build_summary_stats_line(self) -> Tuple[List[Tuple[str, Dict[str, bool]]], str]:
+        main_color, known_types = self._get_main_color()
+
+        parts = []
+        for key in known_types:
```
nicoddemus: Following my comment about `known_types` might not being needed, can we use this instead?

```python
known_types = list(KNOWN_TYPES) + [x for x in self.stats if x not in KNOWN_TYPES]
```

src/_pytest/terminal.py
```diff
@@ -365,6 +378,12 @@ def section(self, title, sep="=", **kw):
     def line(self, msg, **kw):
         self._tw.line(msg, **kw)
 
+    def _add_stats(self, category: str, items: List) -> None:
+        set_main_color = category not in self.stats
```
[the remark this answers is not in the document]
nicoddemus: I just realized after reading the rest of the code that we support adding stats unknown to the terminal reporter, is that correct? If so this suggestion does not apply.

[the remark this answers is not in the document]
blueyed: Yes, any can be added.

src/_pytest/terminal.py
```diff
@@ -247,6 +258,8 @@ def __init__(self, config: Config, file=None) -> None:
         self._showfspath = None
 
         self.stats = {}  # type: Dict[str, List[Any]]
+        self._main_color = None  # type: Optional[str]
+        self._known_types = None  # type: Optional[List]
```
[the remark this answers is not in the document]
blueyed: This is what gets done, and cached/set there then: https://github.com/blueyed/pytest/blob/114de91ab780ce59b53176d0ba8d87dad66a0f3a/src/_pytest/terminal.py#L1058

src/_pytest/terminal.py
blueyed: Could be done maybe.. but a) order is important here (could be solved), but b) we have it already, so can use it.
This can be done in a next round of cleanup/refactoring then maybe.
Also keep in mind that e.g. pytest-parallel needs this to be usable via a multiprocessing proxy.
I'd rather leave it as is for now, but thanks for the suggestion and review, of course.

nicoddemus: Fair enough. 👍

nicoddemus:
LGTM. 👍

PR #6424, opened 2020-01-09
Fix wrong 'changelog' and 'reference' links in docs

### LINKED ISSUE #6397: Incorrect changelog link from docs homepage
On https://docs.pytest.org/en/latest/index.html, the Changelog is erroneously linked:

> Consult the [Changelog](https://docs.python.org/3/whatsnew/changelog.html#changelog) page for fixes and enhancements of each version.

It links to the Python language changelog (https://docs.python.org/3/whatsnew/changelog.html#changelog) rather than the Pytest changelog (https://docs.pytest.org/en/latest/changelog.html).

I'm not sure why; the `rst` source says the following and there is a changelog page.

```rst
Changelog
---------

Consult the :ref:`Changelog <changelog>` page for fixes and enhancements of each version.
```

Is there an issue with the doc generation process?

nicoddemus:
Both references were referencing links from Python because of our intersphinx
mapping in `conf.py`:

    intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}

Because Python's docs explicitly define both references, Sphinx fallbacks to
them instead of generating implicit references as was expected.

Found the two references in this PR by commenting the line above in `conf.py`.

Fix #6397

### DIFF
--- a/doc/en/changelog.rst
+++ b/doc/en/changelog.rst
@@ -1,3 +1,5 @@
+.. _`changelog`:
+
 =========
 Changelog
 =========
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1,3 +1,5 @@
+.. _`reference`:
+
 API Reference
 =============
 


nicoddemus:
Weird, Azure seems to be acting up

Issue #449, opened 2014-02-08
better indicate that there were xpassed test results in a test run


pytestbot:
Originally reported by: **Jurko Gospodnetić (BitBucket: [jurko](http://bitbucket.org/jurko), GitHub: [jurko](http://github.com/jurko))**

---

This is a usability related enhancement suggestion.

When running tests what I really want to know first is:
- Did anything unexpected happen?

which includes both:
- Whether there were any test failures/errors?
- Whether any tests marked `xfail` suddenly started passing?

The first I can get easily at first glance - either by running the tests with the `-x` option or by checking the final summary line color (green/red). The latter is a problem though because in order to find whether there were any `xpassed` test I have to concentrate a lot harder and either:
- Scan the test summary for `xpassed` test result indicators (capital letter `X`) which can be difficult to discern from expected `xfailed` (lower letter `x`) results.
- Read the final colored summary line to see if any `xpassed` test results occurred.
- Run the tests with the '-r X' option and read whether the summary displays any `xpassed` test results.

One of my projects has a lot of tests marked `xfail` and it started to bug me that I often waste a lot of time and interrupt my flow & concentration by having to check the test results in detail just to see if there were any `xpassed` test results.

My suggestion would be to:
- Use a different color (e.g. blue?) if you would otherwise color it green by at least one `xpassed` test result was encountered.
- Use an exit code other that a simple 0=success in such cases.
- Possibly color individual `error` (`E`), `failed` (`F`) & `xpassed` (`X`) test result indicators red or red/blue.
  
  You might not want to use the new `xpassed` result related coloring when running with disabled assertions (`-O`) and you displayed a warning about this possibly causing failing tests to be marked as passed, e.g. when running using an older Python interpreter version.
  
  The whole enhancement could possibly be made configurable as well.
  
  Hope this helps.
  
  Best regards,
    Jurko Gospodnetić

---
- Bitbucket: https://bitbucket.org/pytest-dev/pytest/issue/449

blueyed:
I'm re-opening this: by now there is not only red/green, but also yellow, which appears to be a good fit for when there are any "xpassed" tests.

blueyed:
Fixed in features.

PR #6425, opened 2020-01-09
terminal: use "yellow" with any "xpassed" tests

blueyed:
Closes https://github.com/pytest-dev/pytest/issues/449.

### DIFF
--- a/changelog/449.improvement.rst
+++ b/changelog/449.improvement.rst
@@ -1,0 +1,1 @@
+Use "yellow" main color with any XPASSED tests.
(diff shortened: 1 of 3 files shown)


changelog/449.improvement.rst
```diff
@@ -0,0 +1 @@
+Use "yellow" main color with any XPASSED tests.
```
blueyed:
```suggestion
Use yellow main color with any XPASSED tests.
```

bluetech: IIUC there is no actual change here, because the previous code (probably unintentionally) used the fallback color, which is also yellow. So the changelog can be dropped as it might be confusing.

blueyed: ..only if there is no passed one.
So now it is yellow always with an xpass.

bluetech: Oh right, I missed that this talks about the *main* color.

PR #6430, opened 2020-01-10
[WIP] cacheprovider: pytest_collectreport: do not add to lastfailed with passes

blueyed:
Not clear what this is meant to do - it will usually get cleared with
test reports afterwards already.

Replacing it with `assert 0` causes a test failure, but not when doing
so with the commit that merged the cache plugin (e20216a1a).

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -210,8 +210,7 @@ class LFPlugin:
     def pytest_collectreport(self, report):
         passed = report.outcome in ("passed", "skipped")
         if passed:
-            if report.nodeid in self.lastfailed:
-                self.lastfailed.pop(report.nodeid)
+            self.lastfailed.pop(report.nodeid, None)
         else:
             self.lastfailed[report.nodeid] = True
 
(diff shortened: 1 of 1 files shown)


src/_pytest/cacheprovider.py
```diff
@@ -210,9 +210,7 @@ def pytest_runtest_logreport(self, report):
     def pytest_collectreport(self, report):
         passed = report.outcome in ("passed", "skipped")
         if passed:
-            if report.nodeid in self.lastfailed:
-                self.lastfailed.pop(report.nodeid)
-                self.lastfailed.update((item.nodeid, True) for item in report.result)
```
nicoddemus: This line is pretty weird... wasn't it just adding the `nodeids` back to `lastfailed`? 🤔 

`git blame` says @RonnyPfannschmidt was the previous author, do you remember what's this line suppose to do @RonnyPfannschmidt?

blueyed: > git blame

Do you have a commit?

The other day when I've looked at this I've noticed that this is for collection: if collection failed before it gets popped and then all tests from the collection report are considered to be failures (so they are re-run I assume).
So it might make sense to have in general, but should have a test.

nicoddemus: Thanks for the clarification 👍

PR #6437, opened 2020-01-10
Use cached_property with some code properties

blueyed:


bluetech:
Presumably this increases memory usage, does it improve performance for anything in particular?

blueyed:
@bluetech 
Good call.  I assume that it only helps with failures in general, e.g. when there are many (long) tracebacks etc.
Needs more investigation, closing for now.

Issue #6426, opened 2020-01-09
Azure is failing to clone


nicoddemus:
Recently Azure is failing with:

```
git -c http.extraheader="AUTHORIZATION: bearer ***" fetch --force --tags --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/pull/6387/merge:refs/remotes/pull/6387/merge
fatal: unable to access 'https://github.com/pytest-dev/pytest/': The requested URL returned error: 400
##[warning]Git fetch failed with exit code 128, back off 3.171 seconds before retry.
git -c http.extraheader="AUTHORIZATION: bearer ***" fetch --force --tags --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/pull/6387/merge:refs/remotes/pull/6387/merge
fatal: unable to access 'https://github.com/pytest-dev/pytest/': The requested URL returned error: 400
##[warning]Git fetch failed with exit code 128, back off 9.696 seconds before retry.
git -c http.extraheader="AUTHORIZATION: bearer ***" fetch --force --tags --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/pull/6387/merge:refs/remotes/pull/6387/merge
fatal: unable to access 'https://github.com/pytest-dev/pytest/': The requested URL returned error: 400

```

A quick Google search did not turn anything... will try to investigate later.

PR #6438, opened 2020-01-10
Fixtures now register finalizers with all fixtures before them in the stack

### LINKED ISSUE #6436: Autouse and parameterization results in fixtures not being torn down, or even being torn down out of order
```
$ pip list --format=columns
Package            Version
------------------ -------
attrs              19.3.0 
importlib-metadata 1.3.0  
more-itertools     8.0.2  
packaging          20.0   
pip                9.0.1  
pkg-resources      0.0.0  
pluggy             0.13.1 
py                 1.8.1  
pyparsing          2.4.6  
pytest             5.3.2  
selenium           3.141.0
setuptools         39.0.1 
six                1.13.0 
urllib3            1.25.7 
wcwidth            0.1.8  
zipp               0.6.0
```

I noticed that when autouse and parameterized fixtures of multiple scope levels, it can result in fixtures not being torn down every time a parameterized fixture is torn down if they were executed after the parameterized fixture was setup (and before it was torn down). I also saw in more extreme cases that fixtures would even be torn down in the wrong order.

Here's a simple example:

```py
import pytest

@pytest.fixture(scope="module", params=["m1", "m2"], autouse=True)
def m_fix(request):
    yield

def c_fix():
    yield

class TestFixtures:
    def test_a(self, c_fix):
        pass
```

I would expect the setup plan for this to be:

```
tests/test_auto.py 
    SETUP    M m_fix[m1]
      SETUP    C c_fix
        tests/test_auto.py::TestFixtures::test_a[m1] (fixtures used: c_fix, m_fix, request)
      TEARDOWN C c_fix
    TEARDOWN M m_fix[m1]
    SETUP    M m_fix[m2]
      SETUP    C c_fix
        tests/test_auto.py::TestFixtures::test_a[m2] (fixtures used: c_fix, m_fix, request)
      TEARDOWN C c_fix
    TEARDOWN M m_fix[m2]
```

but it's actually this:

```
tests/test_auto.py 
    SETUP    M m_fix[m1]
      SETUP    C c_fix
        tests/test_auto.py::TestFixtures::test_a[m1] (fixtures used: c_fix, m_fix, request)
    TEARDOWN M m_fix[m1]
    SETUP    M m_fix[m2]
        tests/test_auto.py::TestFixtures::test_a[m2] (fixtures used: c_fix, m_fix, request)
      TEARDOWN C c_fix
    TEARDOWN M m_fix[m2]
```

Notice that the class-level fixture is not torn down along with the parameterized, autouse module-level fixture, despite being affected by it.

SalmonMode:
Closes #6436 

Turns out the root cause was because fixtures were only considering the fixtures requested in their argnames when trying to find out where to register their finalizers at execute time. They weren't considering autouse fixtures, so when this was combined with parameterization, this resulted in fixtures not being torn down when a parameterized, autouse fixture was being torn down to go from one param set to the next.

This also resulted in fixtures being torn down out of order in extreme cases. You can see an example of this in the test that I added (if you run it without this fix).

The solution ultimately required me to have fixtures register their finalizers with all fixtures that came before them in the stack. This list of fixtures came from using `request.fixturenames` in combination with the parent requests (when relevant) and the fixture's requested argnames to determine a good slicing point. Since this logic could be summed up on its own, I moved it into its own private method with a beefy docstring to explain it.

### DIFF
--- a/changelog/6436.bugfix.rst
+++ b/changelog/6436.bugfix.rst
@@ -1,0 +1,3 @@
+:class:`FixtureDef <_pytest.fixtures.FixtureDef>` objects now properly register their finalizers with autouse and
+parameterized fixtures that execute before them in the fixture stack so they are torn
+down at the right times, and in the right order.
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -882,9 +882,7 @@ class FixtureDef:
             self._finalizers = []
 
     def execute(self, request):
-        # get required arguments and register our own finish()
-        # with their finalization
-        for argname in self.argnames:
+        for argname in self._dependee_fixture_argnames(request):
             fixturedef = request._get_active_fixturedef(argname)
             if argname != "request":
                 fixturedef.addfinalizer(functools.partial(self.finish, request=request))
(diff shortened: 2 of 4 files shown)


SalmonMode:
I'm not sure why it's failing. Could it be because I pushed a fix for the string formatting issue before the build was completely done?

blueyed:
@SalmonMode 
pytest-CI is broken currently (#6426).

src/_pytest/fixtures.py
```diff
@@ -907,6 +905,61 @@ def execute(self, request):
...
+        ruled out. The argnames of all fixtures associated with a request in the
+        ``self._parent_request`` stack are found, and the lowest index argname is
+        considered the earliest point in the list of fixture argnames where everything
+        from that point onward can be considered to execute after this fixture.
+        Everything before this point can be considered fixtures that this fixture
+        depends on, and so this fixture should register its finalizer with all of them
+        to ensure that if any of them are to be torn down, they will tear this fixture
+        down first.
+
+        This is the fist part of the list of fixture argnames that is returned. The last
```
blueyed:
```suggestion
        This is the first part of the list of fixture argnames that is returned. The last
```

changelog/6436.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+``FixtureDef`` objects now properly register their finalizers with autouse and
```
blueyed:
```suggestion
``:class:FixtureDef`` objects now properly register their finalizers with autouse and
```

SalmonMode:
Thanks for catching those! All fixed. Also thanks for the heads up about pytest-CI :grin:

SalmonMode:
You got it! All set! :grin:

nicoddemus:
Thanks @SalmonMode, but there's a bunch of commits which shouldn't be there... can you recheck how you rebased/squashed please?

SalmonMode:
Ah, right you are. I pulled and merged my branch after rebasing without thinking haha. Looks like it's good now. I also made sure to get the very latest master and rebased on that again.

nicoddemus:
The rebase now looks correct, thanks!

Could you please also squash the commits into one? The history in that case is not really relevant. 😁 

(Let me know if you want me to do it, I will be glad to; our policy is to always ask before push-forcing into forks)

SalmonMode:
Sure thing! All set :grin:

nicoddemus:
Thanks @SalmonMode appreciate the patience!

SalmonMode:
No worries, @nicoddemus! Sorry for the poke. I just downloaded the new GitHub app and was messing around :P

nicoddemus:
Not at all 😁

src/_pytest/fixtures.py
```diff
@@ -907,6 +905,61 @@ def execute(self, request):
...
+            if hasattr(parent_request, "_fixturedef"):
+                parent_fix_name = parent_request._fixturedef.argname
+                if parent_fix_name in all_fix_names:
+                    parent_fixture_indexes.add(all_fix_names.index(parent_fix_name))
+            parent_request = parent_request._parent_request
+
+        stack_slice_index = min([current_fix_index, *parent_fixture_indexes])
+        active_fixture_argnames = all_fix_names[:stack_slice_index]
+        return {*active_fixture_argnames, *self.argnames}
+
```
blueyed: JFI/as a suggestion: we should strive for adding type annotations with new code, especially with hairy parts like this.

Thanks for the fix though anyway, of course.

SalmonMode: I wanted to, because it was getting pretty hard to track what was where, but I didn't see them being used elsewhere, and didn't want to break a build

blueyed: We require Python 3.5+, so that's fine.
There are large parts still not using them, but we're making progress.
I share your sentiment of them being useful also while working on code due to this, but also can see that it might be not easy and cause problems when trying to add them ("rabbit hole").
Just wanted to mention this since I think it's useful to ask for in code reviews and with own PRs.  /cc @nicoddemus @RonnyPfannschmidt

nicoddemus: Fair enough, good point

blueyed: See #6486 for running mypy with diff-cover on CI.

PR #6421, opened 2020-01-07
Include codecov on GitHub Actions

nicoddemus:
Follow up from https://github.com/pytest-dev/pytest/pull/6420

Issue #6369, opened 2019-12-26
Include deploy and coverage in GitHub actions


nicoddemus:
Follow up to https://github.com/pytest-dev/pytest/pull/6355:

- [x] Coverage
- [x] GitHub release notes
- [x] Deploy
- [x] Cleanup Azure and Travis files

nicoddemus:
https://github.com/pytest-dev/pytest/pull/6441#discussion_r366372520

webknjaz:
FTR it should be possible to request the missing interpreters from GH for GHA by opening an issue in their setup-python repo

The-Compiler:
I guess the other question is how far away we are from dropping 3.5.0/.1 (see e.g. #5751) or just 3.5 entirely (upstream support [ends this September](https://devguide.python.org/#status-of-python-branches)). Perhaps we should start showing deprecation warnings for Python 3.5 with pytest 5.1 or so? I think I'd rather just keep Travis around until we dropped it instead of investing time (both ours and Github's) in something which we'll remove a few months down the line.

nicoddemus:
> Perhaps we should start showing deprecation warnings for Python 3.5 with pytest 5.1 or so?

(I guess you mean pytest 6.1)

Not sure, we did this in the past with Python 2.7 support and a lot of people complained that they were sick of tools telling them all the time about Python 2.7 deprecation.

I don't think we need to explicitly tell users about the upcoming Python 3.5 support drop, due to `python_requires` Python 3.5 users will still get a supported version with `pip install`.

I'm closing this now, I think we will remove Travis files naturally as part of dropping Python 3.5 support. 👍

webknjaz:
> a lot of people complained that they were sick of tools telling them all the time about Python 2.7 deprecation.

LOL this reminds me of https://github.com/pyca/cryptography/issues/5335#issuecomment-663747001. OTOH I think emitting a deprecation is the right thing to do.

PR #6441, opened 2020-01-11
Implement code coverage in GitHub actions

nicoddemus:
This overwrites the `codecov.yml` file in the root of the repository with
`codecov-upstream.yml` file (which contains the code-cov token), so PRs
and branches on the repository can upload coverage.

Suggestion from here:

https://github.com/pytest-dev/pytest/pull/6421#issuecomment-571934112

Security concerns: the token might be misused, but only to upload bogus coverage
to `codecov.io`, so the team believe this is harmless. If we decide to fallback
from this decision , we just need to revoke the token.

Related to #6369

(Following up from https://github.com/pytest-dev/pytest/pull/6421, but now on my fork)


.github/workflows/main.yml
```diff
@@ -118,6 +120,37 @@ jobs:
...
+    - name: Test with coverage
+      if: "! matrix.skip_coverage"
+      env:
+        _PYTEST_TOX_COVERAGE_RUN: "coverage run -m"
+        COVERAGE_PROCESS_START: ".coveragerc"
+        _PYTEST_TOX_EXTRA_DEP: "coverage-enable-subprocess"
+      run: "tox -e ${{ matrix.tox_env }}"
+
+    - name: Prepare coverage token
+      if: success() && !matrix.skip_coverage && ( github.repository == 'pytest-dev/pytest' || github.event_name == 'pull_request' )
```
nicoddemus: @hugovk small difference here: we still upload coverage on PRs. Of course this may happen on PRs on fork -> fork, but this should be rare and not a problem.

hugovk: OK, should be fine.

So, when `github.event_name == 'pull_request'`, is it something like `github.repository == 'nicoddemus/pytest'`?

nicoddemus: Didn't check, but I think so

nicoddemus:
I've tried to remove the Azure pipelines/add it back (to see if I could fix the failures), but the Pipelines page no longer shows `pytest-dev/pytest` for me to select... weird.

Anyway, I think we can disable azure entirely now because all Windows builds will be done on GitHub actions anyway.

Still keeping Travis for deploy, at least until #6369 is fully resolved.

nicoddemus:
It would be nice to merge this soon, as Azure is broken.

blueyed:
I do not get why we need `.github/codecov-upstream.yml`.
We could also skip reporting coverage for when it would not be copied, no?

blueyed:
> Anyway, I think we can disable azure entirely now because all Windows builds will be done on GitHub actions.

Makes sense already now if it cannot be fixed.

codecov.yml
```diff
@@ -1,3 +1,5 @@
+# note: `.github/codecov-upstream.yml` is basically a copy of this file, please propagate
+# changes as needed
```
blueyed: If (the concept of) `.github/codecov-upstream.yml` needs to be kept it could also just append necessary lines to this file, making this comment unnecessary.

nicoddemus:
> I do not get why we need .github/codecov-upstream.yml.
We could also skip reporting coverage for when it would not be copied, no?

From here: https://github.com/pytest-dev/pytest/pull/6421#issuecomment-571934112

> If you put it inline, builds on forks (even if no PR is made) will be sent to the pytest's own Codecov, which clutters up the logs somewhat, and fork users cannot see their own coverage before creating PRs.

Not sure how common it is to use codecov on forks though.

codecov.yml
nicoddemus: > append necessary lines to this file

What do you mean? Is there a way to merge the two files?

blueyed:
Ok, let's only add the token manually then for now.
But I suggest just appending it instead of having a separate file.

codecov.yml
blueyed:
```sh
cat >> codecov.yml <<EOF
codecov:
￼  # token from: https://codecov.io/gh/pytest-dev/pytest/settings
￼  # use same URL to regenerate it if needed
￼  token: "1eca3b1f-31a2-4fb8-a8c3-138b441b50a7"
EOF
```

blueyed: /cc @hugovk in case you want to adopt it also.

nicoddemus:
> But I suggest just appending it instead of having a separate file.

But won't the below be a problem?

> If you put it inline, builds on forks (even if no PR is made) will be sent to the pytest's own Codecov, which clutters up the logs somewhat

blueyed:
@nicoddemus 
It's not put inline / used always, but instead of having a separate file and copying it you'll do the `cat` (with the same pre-conditions).

hugovk:
Appending should be fine too (with the same pre-conditions), it avoids needing to duplicate and keep the rest of the file in sync. 

The `cat`ted snippet could also be in a file of you'd prefer to keep it out of the yml config.

codecov.yml
nicoddemus: Good idea, done. 👍 

Used a python script so I can put docs, do assertions, etc.

blueyed:
> The `cat`ted snippet could also be in a file of you'd prefer to keep it out of the yml config.

Can be even rewritten in Python.. :)

blueyed:
@nicoddemus 
GH actions are not done for PRs to `features`?

nicoddemus:
> GH actions are not done for PRs to features?

Not yet it seems. I will go ahead and merge this one.

nicoddemus:
Sorry, just realized we still needed a small change in `main.yaml`, done it in https://github.com/pytest-dev/pytest/pull/6458 (which only after pushing realized you had already opened https://github.com/pytest-dev/pytest/pull/6457, sorry about the wasted time).

blueyed:
> Please squash.

Guess I should request changes for this in the future.

hugovk:
🎉 GitHub Actions is now showing up under the builds for new commits on Codecov!

* https://codecov.io/gh/pytest-dev/pytest/commit/d291905825d2ff3e5bbbaa245d8fecd24c980db9/build

Compare to an older commit from last week:

* https://codecov.io/gh/pytest-dev/pytest/commit/36531599a4192c9c4f4f39ecdf2cdb32cde91628/build

nicoddemus:
> Guess I should request changes for this in the future.

Sorry didn't see that; my initial take was to keep the commit, but in retrospect I agree it would have been better to squash it.

.github/workflows/main.yml
```diff
@@ -1,8 +1,7 @@
-# evaluating GitHub actions for CI, disconsider failures when evaluating PRs
+# evaluating GitHub actions for CI, disregard failures when evaluating PRs
 #
 # this is still missing:
 # - deploy
```
webknjaz: FYI: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ / https://github.com/marketplace/actions/pypi-publish

nicoddemus: Thanks for the links!

webknjaz: Feel free to ping me for help/review. I'm kinda developing my own framework for GH Apps&Actions + I authored that guide & action...

nicoddemus: Sure, will do! I don't expect many problems, I've already setup deployment for a number of projects (including `pytest-mock`), but you never know!

PR #6446, opened 2020-01-12
tox: add mypy toxenv

blueyed:
This is different from what pre-commit (in "linting") runs in that it
uses our config (alone), and stubs from dependencies.

It would make sense to run this on CI additionally (since there is no
"pre-commit --skip mypy", and a separate config is not worth it).
Currently it triggers a false positive though anyway
(https://github.com/erikrose/more-itertools/pull/374).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
     -   id: pyupgrade
         args: [--py3-plus]
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.761
+    rev: v0.761  # NOTE: keep this in sync with setup.py.
     hooks:
     -   id: mypy
         files: ^(src/|testing/)
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,10 @@ def main():
                 "nose",
                 "requests",
                 "xmlschema",
-            ]
+            ],
+            "checkqa-mypy": [
+                "mypy==v0.761",  # keep this in sync with .pre-commit-config.yaml.
+            ],
         },
         install_requires=INSTALL_REQUIRES,
     )
--- a/tox.ini
+++ b/tox.ini
@@ -55,6 +55,10 @@ basepython = python3
 deps = pre-commit>=1.11.0
 commands = pre-commit run --all-files --show-diff-on-failure {posargs:}
 
+[testenv:mypy]
+extras = checkqa-mypy, testing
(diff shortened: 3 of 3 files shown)


bluetech:
> This is different from what pre-commit (in "linting") runs in that it uses our config (alone)

Does pre-commit add its own configurations?

> and stubs from dependencies.

This will be nice to have, `tox linting` indeed doesn't include the deps.

IIUC, this currently only includes the non-testing dependencies. But the mypy run also includes `testing/`, which can import code from testing dependencies, which can have types. I think hypothesis is one. So does it make sense to install these deps too?

.pre-commit-config.yaml
```diff
@@ -37,7 +37,7 @@ repos:
     -   id: pyupgrade
         args: [--py3-plus]
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.740
+    rev: v0.740  # NOTE: keep this in sync with setup.py.
```
bluetech: Just for anyone doing the merge from features, note that features is already at `v0.761`.

setup.py
```diff
@@ -29,7 +29,10 @@ def main():
                 "nose",
                 "requests",
                 "xmlschema",
-            ]
+            ],
+            "mypy": [
+                "mypy==v0.740",  # keep this in sync with .pre-commit-config.yaml.
```
bluetech: What's the advantage of adding an extra over specifying a `deps` in ` [testenv:mypy]` tox?

blueyed: If e.g. pre-commit could also use the extra, the version could be changed just there.
Given requirements.txt, tox.ini etc I often find it more sane to have this in a more central place like setup.py.

blueyed:
> Does pre-commit add its own configurations?

I thought it would have additional args, but does not have them.  So the only diff is with the deps.

Added the testing extra also.

setup.py
bluetech: OK, makes sense.

I fear though that people will see it and start doing `pytest[mypy]` due to a misconception. There are probably tools which parse the extras and show them to the user. So WDYT about changing the name to `mypy-internal` or `_mypy` or something like that?

blueyed: Good point.  What about `checkqa-mypy`?

bluetech: That's clear enough I think.

blueyed: Great, amended.

PR #6448, opened 2020-01-13
LFPlugin: use sub-plugins to deselect during collection

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/5301

Two subplugins are necessary due to limitations in pluggy (https://github.com/pytest-dev/pluggy/pull/246).

TODO:

- [x] changelog; NOTE: might only fix later issue reported in #5301 (https://github.com/pytest-dev/pytest/issues/5301#issuecomment-572524754)

### DIFF
--- a/testing/test_cacheprovider.py
+++ b/testing/test_cacheprovider.py
@@ -265,7 +265,13 @@ class TestLastFailed:
             """
         )
         result = testdir.runpytest(str(p), "--lf")
-        result.stdout.fnmatch_lines(["*2 passed*1 desel*"])
+        result.stdout.fnmatch_lines(
+            [
+                "collected 2 items",
+                "run-last-failure: rerun previous 2 failures",
+                "*= 2 passed in *",
+            ]
+        )
         result = testdir.runpytest(str(p), "--lf")
         result.stdout.fnmatch_lines(
             [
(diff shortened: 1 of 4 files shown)


testing/test_cacheprovider.py
```diff
@@ -793,9 +818,9 @@ def test_lastfailed_with_known_failures_not_being_selected(self, testdir):
         result = testdir.runpytest("--lf")
         result.stdout.fnmatch_lines(
             [
-                "collected 1 item",
-                "run-last-failure: 1 known failures not in selected tests (skipped 1 file)",
-                "* 1 failed in *",
+                "collected 2 items",
+                "run-last-failure: 1 known failures not in selected tests",
+                "* 1 failed, 1 passed in *",
```
blueyed: Not really sure, but this might show the fix for https://github.com/pytest-dev/pytest/issues/5301.
Could use a more explicit test otherwise, of course.

nicoddemus:
Thanks, sorry for the delay.

At first glance the approach seems to work, but also seems a bit more complicated than I would like, so before we discuss this PR further I would like for us to consider the alternative using mtime you proposed in https://github.com/pytest-dev/pytest/issues/5301#issuecomment-577418161. 👍

blueyed:
@nicoddemus 
I'm not really interested in discussing alternatives - I think this approach is fine.

blueyed:
@nicoddemus 
Have you tested this with regard to https://github.com/pytest-dev/pytest/issues/5301#issuecomment-572574330?

nicoddemus:
Yes, have you?

blueyed:
No.  But I've guessed that an existing (changed) test would cover it already (https://github.com/pytest-dev/pytest/pull/6448#pullrequestreview-341614837).  Not sure though, and it might be good to review/extend it with regard to that.

nicoddemus:
I suggest to make use of reproducible use cases when they are provided when opening PRs claiming to fix a bug ;)

Either way, the change seems fine so feel free to merge it.

PR #6449, opened 2020-01-13
[WIP] typing improvements

blueyed:


src/_pytest/config/__init__.py
```diff
@@ -203,7 +206,10 @@ def get_plugin_manager():
     return get_config().pluginmanager
 
 
-def _prepareconfig(args=None, plugins=None):
+def _prepareconfig(
+    args: Optional[Union[List, Tuple]] = None,
```
bluetech: Looks like this can also be a `py.path.local`. Even though it's not a real type yet, if you recall we decided to add it anyway so once typing is added to `py.path` it will start working.

src/_pytest/config/__init__.py
```diff
@@ -172,7 +172,10 @@ def directory_arg(path, optname):
 builtin_plugins.add("pytester")
 
 
-def get_config(args=None, plugins=None):
+def get_config(
+    args: Optional[Union[List, Tuple]] = None,
```
bluetech: `List` and `Tuple` without a `[]` is equivalent to `List[Any]` and `Tuple[Any]`. Are there narrower types that can be used?

If yes, it's OK to leave them out for now, leaving them out indicates a TODO.

If no, I prefer to put explicit `List[Any]`, this it's clear it can really be anything and is not just omitted/TODO.

src/_pytest/config/__init__.py
```diff
@@ -732,7 +738,9 @@ class InvocationParams:
         plugins = attr.ib()
         dir = attr.ib(type=Path)
 
-    def __init__(self, pluginmanager, *, invocation_params=None) -> None:
+    def __init__(
+        self, pluginmanager: "PytestPluginManager", *, invocation_params=None
```
bluetech: Maybe `invocation_params: Optional[Config.InvocationParams]`?

(OK if you left it out intentionally).

src/_pytest/config/__init__.py
```diff
@@ -203,7 +206,10 @@ def get_plugin_manager():
     return get_config().pluginmanager
 
 
-def _prepareconfig(args=None, plugins=None):
+def _prepareconfig(
+    args: Optional[Union[List, Tuple]] = None,
+    plugins: Optional[Sequence[Union[str, object]]] = None,
```
bluetech: When I (tried to) [add types to pluggy](https://github.com/pytest-dev/pluggy/pull/225), I also typed plugins as `object`. This is the best type for completely opaque things.

I wonder if for clarity we might want to add an alias `Plugin = object` to pluggy and then use it, this why it'd be a bit clearer when we refer to plugins. But just a thought, not for this PR.

src/_pytest/config/__init__.py
blueyed: Will change it to `args: Optional[Sequence[Union[str, py.path.local]]] = None,`.

src/_pytest/config/__init__.py
blueyed: Also changing to `args: Optional[Sequence[Union[str, py.path.local]]] = None,`.

src/_pytest/config/__init__.py
blueyed: > I wonder if for clarity we might want to add an alias `Plugin = object` to pluggy and then use it, this why it'd be a bit clearer when we refer to plugins. But just a thought, not for this PR.

Makes total sense, also for this PR already then, of course.

But we should add it here, given that pluggy is not typed?

src/_pytest/config/__init__.py
blueyed: JFI: postponing this for now, not trivial it seems (also the one above, will delete the comments).

src/_pytest/fixtures.py
```diff
@@ -178,30 +183,43 @@ def getfixturemarker(obj):
         return None
 
 
-def get_parametrized_fixture_keys(item, scopenum):
+def get_parametrized_fixture_keys(
```
bluetech: Depending on how much effort you're willing to expand on this function, it looks like this can be something like this:

```py
@overload
def get_parametrized_fixture_keys(item: "nodes.Item", scopenum: Literal[0]) -> Iterator[Tuple[str, int]]: ...

@overload
def get_parametrized_fixture_keys(item: "nodes.Item", scopenum: Literal[1, 2]) -> Iterator[Tuple[str, int, py.path.local]]: ...

@overload
def get_parametrized_fixture_keys(item: "nodes.Item", scopenum: Literal[4]) -> Iterator[Tuple[str, int, py.path.local, "Class"]]: ...
```

Whether it needs `scopenum: int` fallback depends on the callers.

.pre-commit-config.yaml
```diff
@@ -25,7 +25,12 @@ repos:
     hooks:
     -   id: flake8
         language_version: python3
-        additional_dependencies: [flake8-typing-imports==1.3.0]
+        additional_dependencies: [
+          flake8-typing-imports==1.3.0,
+          # To handle PEP 484 type comments with following text.
+          # (https://github.com/PyCQA/pyflakes/pull/455)
+          'git+https://github.com/PyCQA/pyflakes.git@1911c20#egg=pyflakes',
```
bluetech: Maybe we should wait until it's officially released? For example in my development venv I just do `pip install flake8` which means my editor will still complain. Also maybe things will change before the release.

src/_pytest/config/__init__.py
bluetech: Yeah, adding here for now makes sense -- better than nothing.

.pre-commit-config.yaml
blueyed: Yeah.. but from what I can tell it might still take months, and we should not let that hold us back.
Let's ignore the flake8 error then instead?  (this is what mypy itself does)

bluetech: > Let's ignore the flake8 error then instead?

:+1: 

Next flake8 release will be nice for all the typing fixes :)

src/_pytest/config/__init__.py
```diff
@@ -47,6 +47,8 @@
 if TYPE_CHECKING:
     from typing import Type
 
+    _PluggyPlugin = object
+
```
blueyed: @bluetech since this might be used elsewhere - where should be put this?

bluetech: As long is we keep it internal for now (I mean, not intended to be exposed in the "public" typing in the future), I guess putting it where `PytestPluginManager` is defined makes most sense?

A short comment here would be helpful, something like

> A type to represent plugin objects. Plugins can be any namespace, so we can't narrow it down much, but we use an alias to make the intent clear. Ideally this type would be provided by pluggy itself.

BTW, I wouldn't put it under `TYPE_CHECKING`, this way we can use it without the annoying quotes.

blueyed: Thanks!  Amended in a39b4e2.

bluetech:
@blueyed Are you working on this PR?

blueyed:
> @blueyed Are you working on this PR?

No.  Some things have been taken out of it already, and I plan to revisit it to see if anything is left.

blueyed:
Closing for now.

blueyed:
@bluetech 
Feel free to pick anything from here, of course.

PR #6450, opened 2020-01-13
[WIP] typing for getcfg and determine_setup

blueyed:


bluetech:
Somehow some comments seem to have been lost...

Did we get a conclusion on using an empty `SectionWrapper` instead of the `Dict`? If a dict is necessary, can it be typed as `Dict[str, str]` instead?

blueyed:
> Somehow some comments seem to have been lost...

I've pulled this out of https://github.com/pytest-dev/pytest/pull/6447.

> Did we get a conclusion on using an empty `SectionWrapper` instead of the `Dict`? If a dict is necessary, can it be typed as `Dict[str, str]` instead?

Given that SectionWrapper is not importable I've found this to be less useful / possible.

bluetech:
Do the `py` stubs change things for this PR?

In the `py` stubs we made `SectionWrapper` hidden - `_SectionWrapper` - because it isn't exposed in the public API (not importable).

blueyed:
@bluetech py stubs are helpful / change things.
I've worked a bit more on this in https://github.com/blueyed/pytest/pull/251, but IIRC used the `_vendored` import also for instance checks there.

bluetech:
OK, feel free to ping me once this is ready.

nicoddemus:
Closing for now.

PR #6459, opened 2020-01-14
Use stanford theme

nicoddemus:
Use stanford theme for evaluation (#6402)


The-Compiler:
Looking at https://github.com/LinxiFan/Sphinx-theme stanford mostly seems to be an outdated version of rtd:

> This branch is 25 commits ahead, 649 commits behind readthedocs:master.

Last update was over three years ago, so the chances of maintenance happening when needed for some reason (e.g. Sphinx upgrades?) are pretty slim. Also I don't think the colors really fit well for pytest.

nicoddemus:
😞 shame, I liked it for some reason.

blueyed:
Let's drop it from the options to chose from?

nicoddemus:
Sure, done. 👍

Issue #1112, opened 2015-10-05
Reorganize content page


The-Compiler:
I always found the pytest docs quite hard to navigate - and because of someone with the same issue in IRC just now I realized why: The stuff which is probably most relevant (`assert.rst` and `fixture.rst`) is hidden on `examples.rst` but not mentioned at all in `content.rst`.

Some other stuff in `content.rst` is probably not that relevant... I think the whole thing should have some reorganization.

I'll take a look in a few weeks (still have a growing todo list), but if someone else wants to pick this up, please go ahead!

Zearin:
I came to this repo hoping to see if I was missing an easier-to-navigate piece of documentation (or any helpful GitHub Wiki content).  

Definitely :+1: on this.  

I do not have the free time to contribute at the moment, but if I can make a suggestion...

I find that the best documentation tends to have content organized something like this:

- **Guides**: step-by-step instructions on accomplishing the most common tasks.
    - These are useful to newcomers and early-to-intermediate users. 
    - They should contain stuff like “How to Install”, “Your First Project”, and “Going Further”.  
    - Sometimes they contain stuff like “Advanced Uses” or “Kung-Fu Tricks”, which are useful for advanced users.
- **Quick Reference**: a list of things for fast lookup, usually human-written.
    - These are useful to everyone.  (Even advanced programmers regularly look stuff up! ☻)
    - They should contain stuff like (in PyTest’s case), “Command Line Reference”, “INI Config Reference”, “conftest.py”, “List of PyTest Decorators”, “Markers Reference”, and “PyTest Special Variables” (listing all cases that require PyTest to find a specially-named variable in a specific place to make it do something).  
    - There may be plenty of overlap between “Quick Reference” and the other two sections.  The distinguishing feature is that these tend to be organized as “get in, get out” info.  Depending on the project, they may tend to be written as lists, glossaries, simple diagrams, or cheat sheets.
    - A “Conceptual Overview” is also useful here.  This should give a “10,000 foot overview” of the project’s major components and their roles.
- **Full Reference**/**API Docs**: the complete documentation or API, usually generated.
    - These are useful to advanced users and developers.  They contain the nitty-gritty to anything the code is capable of.

That’s as much time as I can spare to help.  Wish I had time for more!

nicoddemus:
Thanks for taking the time to write this down! This was somewhat what I had in mind (and I suspect @The-Compiler as well), so it is good to have it in writing so we don't forget anything. :grin:

PR #6460, opened 2020-01-14
Use rtd theme

nicoddemus:
Use rtd theme for evaluation (#6402)


blueyed:
Answering to https://github.com/pytest-dev/pytest/issues/6402#issuecomment-579521818 here.

@nicoddemus @The-Compiler 
> > I do like the RTD theme, but its sidebar seems quite messy compared to our current one - is this something we could port over to the RTD theme as well?
> 
> https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#how-the-table-of-contents-displays
> 
> Not sure how to configure it properly though...

It uses our TOC tree, which is not well organized it seems: https://github.com/pytest-dev/pytest/blame/ca8a1e47b0cc7fe44851b83ae0d225bf5b13e55d/doc/en/contents.rst#L10

Previously this was used: https://github.com/pytest-dev/pytest/blob/60f09d4600ebdc884fa5b8d2b20258a72e69b918/doc/en/_templates/globaltoc.html

It could still be used with the new layout, e.g. by extending/overriding the default blocks probably: https://github.com/readthedocs/sphinx_rtd_theme/blob/68a19ca510ca67fc0afefce155544ee008a23aa7/sphinx_rtd_theme/layout.html#L148-L169

It might be better to re-structure the TOC in general though, so that it ends up in maybe only 10 top level documents.

nicoddemus:
Thanks @blueyed for investigating this further!

> It uses our TOC tree, which is not well organized it seems

Why do you say that?

The objective of that reorganization was to have "flatter" topics so they are easier to find, from #1112:

> I always found the pytest docs quite hard to navigate - and because of someone with the same issue in IRC just now I realized why: The stuff which is probably most relevant (assert.rst and fixture.rst) is hidden on examples.rst but not mentioned at all in content.rst.

Of course a flat organization like that does mess the sidebar with this theme, but that IMHO is a fault of the theme that doesn't allow it to be customized like we did with the previous `globaltoc.html` file.

> It could still be used with the new layout, e.g. by extending/overriding the default blocks probably: readthedocs/sphinx_rtd_theme:sphinx_rtd_theme/layout.html@68a19ca#L148-L169

Oh that would be great! I don't really know how to do that though. 😬

If you are willing to give this a try, please feel free to push into this branch directly. 👍 

> It might be better to re-structure the TOC in general though, so that it ends up in maybe only 10 top level documents.

But then we are back to the original format, which was hard to find things...

blueyed:
> > It might be better to re-structure the TOC in general though, so that it ends up in maybe only 10 top level documents.
> 
> But then we are back to the original format, which was hard to find things...

I've not really looked into how it looked/appeared before, but thought that https://github.com/pytest-dev/pytest/commit/519f02b0149c3150c15bf240c09f362f5a391dd5#diff-6f267ce6423005e07dd261cb63ec96fa was just dumping it all into the top level.

nicoddemus:
IIRC you are right, while also changing the order so things make more sense.

blueyed:
So the blocker here is that we have no structure for the docs apprently.
Should we use the current sidebar as top-level elements then?

nicoddemus:
Closing in favor of #6453

PR #6465, opened 2020-01-15
doc: revisit section about rootdir

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/6376

(had this in a stash from looking at it the other day)

### DIFF
--- a/doc/en/customize.rst
+++ b/doc/en/customize.rst
@@ -20,8 +20,6 @@ which were registered by installed plugins.
 Initialization: determining rootdir and inifile
 -----------------------------------------------
 
-
-
 pytest determines a ``rootdir`` for each test run which depends on
 the command line arguments (specified test files, paths) and on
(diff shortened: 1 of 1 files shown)


nicoddemus:
@erheron you might want to take a look at this. 👍

erheron:
@nicoddemus , sorry for answering such late. After reading changes I've got a feeling that there should be done something slightly more than that. Frankly, this opinion comes from looking at diff and... to me it looks like not so big change. I don't have a clear idea about "more changes" i'm talking about, but... maybe some example (example of how rootdir shouldn't be used?) would help? It could be taken from one of the issues, where such confusing interaction was noticed :thinking:

blueyed:
@erheron 
> to me it looks like not so big change

It was not meant to fix/address https://github.com/pytest-dev/pytest/issues/6376 (I've done it before that).  Please consider creating a followup PR - I might have integrated changes here, but then rather like to have not too many PRs open myself (so I've merged it already).

PR #6469, opened 2020-01-15
Allow user to select if he wants log, std-out/err in junit capture xml file, and respect his decision

dxtr85:
The purpose for this PR is enabling user to set what should be contained in generated xml file since for some (me included) this file is used only for report generation, hence log/out/err are not always welcome, and can take lots of space.

User can pick between no|log|system-out|system-err|out-err|all in order to put selected output in xml file.

### DIFF
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -1297,7 +1297,11 @@ def test_tee_stdio_captures_and_live_prints(testdir):
     """
     )
     result = testdir.runpytest_subprocess(
-        testpath, "--capture=tee-sys", "--junitxml=output.xml"
+        testpath,
+        "--capture=tee-sys",
+        "--junitxml=output.xml",
+        "-o",
+        "junit_logging=all",
     )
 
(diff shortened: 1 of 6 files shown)


testing/acceptance_test.py
```diff
@@ -1297,7 +1297,11 @@ def test_simple():
     """
     )
     result = testdir.runpytest_subprocess(
-        testpath, "--capture=tee-sys", "--junitxml=output.xml"
+        testpath,
+        "--capture=tee-sys",
+        "--junitxml=output.xml",
+        "-o",
+        "junit_logging=all",
```
nicoddemus: Nice, the test was actually buggy: `junit_logging` by default is `no`, so it should not be writing stdout contents on the xml file. 👍

dxtr85:
Changes are squashed now down to 1 commit.

nicoddemus:
Thanks again @dxtr85!

dxtr85:
Thank you, @nicoddemus !

PR #6474, opened 2020-01-16
ci: github actions: only deploy pytest-dev/pytest

blueyed:


.github/workflows/main.yml
```diff
@@ -159,7 +159,7 @@ jobs:
         name: ${{ matrix.name }}
 
   deploy:
-    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && github.repository == 'pytest-dev/pytest'
```
nicoddemus: Why do you think this is necessary?

* The secrets won't be available outside pytest-dev anyway
* I've tested this on my fork, so this change would prevent that in the future

I'm not entirely against it, just would like to know the actual reason for it. 😁

blueyed: > The secrets won't be available outside pytest-dev anyway

Sure, I expect this to cause an error then.  I.e. no harm, but "confusing".

> I've tested this on my fork, so this change would prevent that in the future

That's similar how you would have to enable codecov explicitly when testing on a fork, no?

I think forks should have to enable this explicitly (e.g. in my fork), especially given that "deploy" can be more harmful than "uploading coverage" - although I can see that the coverage token is in the repo itself, while it is not there for "deploy".
So this is merely to avoid an error on forks I guess.

I'm +0 on this myself, feel free to close.

nicoddemus: > Sure, I expect this to cause an error then. I.e. no harm, but "confusing".

No, it just skips the task.

> So this is merely to avoid an error on forks I guess.

This would only happen if a pytest maintainer explicitly added a secret with their upload token to their own fork.

> I'm +0 on this myself, feel free to close.

Nah no worries, let's have the extra check. If one needs to test it again in the future, it is just a matter of commenting out that condition. :+1:

PR #6476, opened 2020-01-16
Fix test_config: wrong assertions, lint, unused fixtures

blueyed:


blueyed:
Well, most of the credit goes to pylint.

PR #6412, opened 2020-01-07
Fix serialization of 'None' reprcrashes

nicoddemus:
Tracebacks coming from remote processes crated by the multiprocess module
will contain "RemoteTracebacks" which don't have a 'reprcrash' attribute.

I've made sure to test this fixes in xdist.

Fix #5971

PR #6479, opened 2020-01-16
[master] Use a dummy RemoteTraceback for test in Python 3.5 Windows

blueyed:
Somehow in Python 3.5 on Windows this test fails with:
   File "c:\hostedtoolcache\windows\python\3.5.4\x64\Lib\multiprocessing\connection.py", line 302, in _recv_bytes
     overlapped=True)
OSError: [WinError 6] The handle is invalid

This only happens in this platform and Python version, decided to use
a dummy traceback as originally done in #6412.

(cherry picked from commit b9c136b809a33009ad514672666c515953957b33)

### DIFF
--- a/testing/test_reports.py
+++ b/testing/test_reports.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 from _pytest._code.code import ExceptionChainRepr
 from _pytest.pathlib import Path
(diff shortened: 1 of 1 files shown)


blueyed:
IIRC there was the same failure this was done for with a PR for master.
@nicoddemus Was this flaky in the first place only?

nicoddemus:
> Was this flaky in the first place only?

I've never seen it fail before trying to merge into features... but good thing it also breaks on master I guess, at least things are consistent.

Thanks

PR #6482, opened 2020-01-17
typing: fix _TracebackStyle

blueyed:
Ref: https://github.com/blueyed/pytest/pull/163

TODO:

- [x] rebase to fix commit message
- [ ] check my stashes

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -41,7 +41,7 @@ if TYPE_CHECKING:
 
     from _pytest._code import Source
 
-    _TracebackStyle = Literal["long", "short", "no", "native"]
+    _TracebackStyle = Literal["long", "short", "line", "no", "native"]
 
 
 class Code:
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -15,7 +15,9 @@ from .reports import CollectErrorRepr
 from .reports import CollectReport
 from .reports import TestReport
 from _pytest._code.code import ExceptionInfo
+from _pytest._code.code import ExceptionRepr
 from _pytest.compat import TYPE_CHECKING
+from _pytest.nodes import Collector
 from _pytest.nodes import Node
 from _pytest.outcomes import Exit
 from _pytest.outcomes import Skipped
@@ -251,7 +253,7 @@ def pytest_runtest_makereport(item, call):
     return TestReport.from_item_and_call(item, call)
 
 
-def pytest_make_collect_report(collector) -> CollectReport:
+def pytest_make_collect_report(collector: Collector) -> CollectReport:
     call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
     longrepr = None
     if not call.excinfo:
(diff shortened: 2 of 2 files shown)


src/_pytest/_code/code.py
```diff
@@ -41,7 +41,7 @@
 
     from _pytest._code import Source
 
-    _TracebackStyle = Literal["long", "short", "no", "native"]
+    _TracebackStyle = Literal["long", "short", "line", "no", "native"]
```
blueyed: "native" in not used internally with `FormattedExcinfo` at least (https://github.com/blueyed/pytest/blob/f1609d6e956bdf4d712610d8e6ca9fbb0b2ed108/src/_pytest/_code/code.py#L612-L631), so it might make sense to have two different types here (but it makes sense to match the `--tb` option here I guess (currently (with this patch) missing "auto")) - but that would be a followup.

src/_pytest/runner.py
```diff
@@ -264,7 +266,10 @@ def pytest_make_collect_report(collector) -> CollectReport:
             skip_exceptions.append(unittest.SkipTest)  # type: ignore
         if call.excinfo.errisinstance(tuple(skip_exceptions)):
             outcome = "skipped"
-            r = collector._repr_failure_py(call.excinfo, "line").reprcrash
+            r_ = collector._repr_failure_py(call.excinfo, "line")
```
bluetech: This looks good to me, but looking at the type of `_repr_failure_py`, I wonder:

1. Why do `_repr_failure_py` and `repr_failure` both exist? `repr_failure` just forwards directly `_repr_failure_py` so why not just inline it?
2. Is it correct that it can only take `Failed` and `FixtureLookupError` and not other exception types?
3. Is it possible to add overloads to it so this assert is not needed?
4. Should its `style` argument be typed with `_TracebackStyle`?

src/_pytest/_code/code.py
bluetech: Should `line` be added to `ExceptionInfo.getrepr()` docstring?

blueyed: I do not think it is used there - i.e. I am not sure.
I think it is only used here: https://github.com/pytest-dev/pytest/blob/719e0e354bd9e13c5a9ecc3f1ac9839d3e7a8704/src/_pytest/terminal.py#L868

src/_pytest/runner.py
blueyed: 1. it is used by other subclasses etc - it is rather tangled (I have some WIP to fix this, also for 4.)
2. IIRC we/I've limited this intentionally for what has been observed, but from the code it is clear that this is not correct: https://github.com/pytest-dev/pytest/blob/719e0e354bd9e13c5a9ecc3f1ac9839d3e7a8704/src/_pytest/nodes.py#L288-L292
3. overload might be a good idea, but in general it would be better to not return a string there.  Also it is based on `excinfo.value.pytrace` - so would need separate types for this I assume.
4. yes, I have that in the WIP to clean up handling of `--fulltrace`

PR #6486, opened 2020-01-17
Add mypy-diff to tox/CI

blueyed:
Tested in https://github.com/blueyed/pytest/pull/164.
Via https://github.com/pytest-dev/pytest/pull/6438#discussion_r367594687.

PR #6488, opened 2020-01-17
typing: Node.reportinfo: might return py.path.local via fspath

blueyed:


src/_pytest/fixtures.py
```diff
@@ -424,7 +424,7 @@ def module(self):
         return self._pyfuncitem.getparent(_pytest.python.Module).obj
 
     @scopeproperty()
-    def fspath(self):
+    def fspath(self) -> py.path.local:
```
blueyed: Note that `mypy-diff` (https://github.com/pytest-dev/pytest/pull/6486) would complain here for this to be missing still (since mypy does not know about py.path.local).

PR #6490, opened 2020-01-17
Skip most jobs when making a release

nicoddemus:
This adds a new optional matrix variable 'release_checker': when true,
the job also executes when we are pushing a tag, otherwise it is skipped.

This is similar to what we did when deploys where done by Travis.


nicoddemus:
Hmm bummer apparently you can't use matrix variables in `jobs.<job_id>.if`... will have to think of some other approach (suggestions welcome).

PR #6491, opened 2020-01-17
Clear node collection cache after collection is done

nicoddemus:
Also rename the variable to convey its intent better

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -6,19 +6,27 @@ import importlib
 import os
 import sys
 from typing import Dict
+from typing import FrozenSet
+from typing import List
 
 import attr
 import py
 
 import _pytest._code
 from _pytest import nodes
+from _pytest.compat import TYPE_CHECKING
 from _pytest.config import directory_arg
 from _pytest.config import hookimpl
 from _pytest.config import UsageError
 from _pytest.fixtures import FixtureManager
+from _pytest.nodes import Node
 from _pytest.outcomes import exit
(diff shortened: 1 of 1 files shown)


nicoddemus:
> I think the renaming is not necessary

I find it makes the intent of the cache clearer (only meant to use during collection); unless you are against it I would like to keep it.

I will add the type hints, thanks. 👍

src/_pytest/main.py
```diff
@@ -392,7 +393,7 @@ def __init__(self, config):
         self.startdir = config.invocation_dir
         self._initialpaths = frozenset()
         # Keep track of any collected nodes in here, so we don't duplicate fixtures
-        self._node_cache = {}
+        self._node_collection_cache = {}  # type: Dict[str, List["Node"]]
```
blueyed:
```suggestion
        self._node_collection_cache = {}  # type: Dict[str, List[Node]]
```

Also `-> None` then for `__init__` if it works.

nicoddemus:
Done, thanks for the pointers!

Had to add a few more type hints and also applied the same logic to the old `_pkg_roots` attribute. 

I want to eventually refactor the entire logic out of the `Session` class so the `session` instance is not so polluted with temporary variables/caches.

nicoddemus:
> Is this meant to fix a correctness issue or just optimization/reduce memory usage?

The latter. 😁

nicoddemus:
Rebased this on features too, as it should be.

PR #6501, opened 2020-01-18
tests: terminal: harden test_color_yes

blueyed:


blueyed:
Decided to skip the test for py35, since markup is not ordered there (instead of fiddling with it).

PR #6504, opened 2020-01-18
[WIP] Fix exponential recursion

SalmonMode:
Should resolve the issues with extreme recursion but only allowing a fixture to register its finalizer with another fixture once.


blueyed:
@SalmonMode 
Should include this in https://github.com/pytest-dev/pytest/pull/6551 probably?
Can you push there?

SalmonMode:
Sure. I'll push it there now.

PR #6509, opened 2020-01-19
typing: minor improvements

blueyed:
Taken out of https://github.com/pytest-dev/pytest/pull/6449.


src/_pytest/main.py
```diff
@@ -378,6 +379,7 @@ class Session(nodes.FSCollector):
     Interrupted = Interrupted
     Failed = Failed
     _setupstate = None  # type: SetupState
+    _fixturemanager = None  # type: FixtureManager
```
bluetech: Since `_fixturemanager` is nowhere to be seen in this file, a comment `Set on the session by pytest_sessionstart().`  would be very helpful here.

PR #6510, opened 2020-01-19
typing: fix some "incompatible types in assignment" with py

blueyed:


src/_pytest/config/findpaths.py
```diff
@@ -121,7 +121,9 @@ def determine_setup(
         sections = ["tool:pytest", "pytest"] if is_cfg_file else ["pytest"]
         for section in sections:
             try:
-                inicfg = iniconfig[section]
+                inicfg = iniconfig[
+                    section
+                ]  # type: Optional[py.iniconfig._SectionWrapper]
```
bluetech: `py.iniconfig._SectionWrapper` will only be available in the new `py` right? Also, it will be inferred then. So I think it will be better to leave this out?

blueyed: Yes, this is with the new stubs.
Do you mean to not type it altogether then?  (it is meant to fix the assignment error, since it gets set to `None` below - the type there is what mypy told me it is)

bluetech: Huh I forgot about that. I actually hit the same [here](https://github.com/bluetech/pytest/commit/82b785887fee82da34b4a3b7d4460582ab7884f6#diff-08c0a98f1819c3df52b7b81ff7f4c173R127).

Given it'd be any without py types and correct with py types, this should be good.

blueyed: I've totally missed https://github.com/pytest-dev/pytest/compare/master...bluetech:py-typed-fixes though.. :/

bluetech: I'll submit it now, no reason to wait I guess.

blueyed: :+1:

PR #6511, opened 2020-01-19
Fix type errors after adding types to the `py` dependency

bluetech:
This is in preparation for adding type stubs to `py`. See https://github.com/pytest-dev/py/pull/232.

These changes make mypy clean both before and after.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -351,7 +351,7 @@ class FixtureRequest:
         self.fixturename = None
         #: Scope string, one of "function", "class", "module", "session"
         self.scope = "function"
-        self._fixture_defs = {}  # argname -> FixtureDef
+        self._fixture_defs = {}  # type: Dict[str, FixtureDef]
         fixtureinfo = pyfuncitem._fixtureinfo
         self._arg2fixturedefs = fixtureinfo.name2fixturedefs.copy()
         self._arg2index = {}
@@ -426,7 +426,8 @@ class FixtureRequest:
     @scopeproperty()
     def fspath(self) -> py.path.local:
         """ the file system path of the test module which collected this test. """
-        return self._pyfuncitem.fspath
+        # TODO: Remove ignore once _pyfuncitem is properly typed.
+        return self._pyfuncitem.fspath  # type: ignore
 
     @property
     def keywords(self):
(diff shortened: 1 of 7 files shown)


src/_pytest/fixtures.py
```diff
@@ -426,7 +426,8 @@ def module(self):
     @scopeproperty()
     def fspath(self) -> py.path.local:
         """ the file system path of the test module which collected this test. """
-        return self._pyfuncitem.fspath
+        # TODO: Remove ignore once _pyfuncitem is properly typed.
+        return self._pyfuncitem.fspath  # type: ignore
```
blueyed: Use the error code here?  (see `--show-error-codes`, could be added to the config probably)  (with flake-ignore then)

src/_pytest/nodes.py
```diff
@@ -462,6 +462,7 @@ def reportinfo(self) -> Tuple[Union[py.path.local, str], Optional[int], str]:
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
+        assert isinstance(location[0], py.path.local), location[0]
```
blueyed: You've removed the change from reportinfo, right?  (from the initial Git tree)

testing/python/collect.py
```diff
@@ -1018,18 +1018,18 @@ class TestReportInfo:
     def test_itemreport_reportinfo(self, testdir):
         testdir.makeconftest(
             """
-            import pytest
+            import pytest, py
             class MyFunction(pytest.Function):
                 def reportinfo(self):
-                    return "ABCDE", 42, "custom"
+                    return py.path.local("foo"), 42, "custom"
```
blueyed: That's kind of an API change - I've noticed this myself that the tests seem to be the only things using `str`.
Ok with me in general, just wanted to mention this, i.e. some plugins might return strings there (but likely fail then afterwards, depending on where it is used I guess?)
Maybe this could/should have a test?

src/_pytest/nodes.py
bluetech: Yes.

testing/python/collect.py
bluetech: The `_bestrelpathcache` backing this only works with `py.path.local`, that it returned something for `str` was accidental and incorrect. So I think the assert just codifies the current behavior.

Only this test trigerred the assert, and it's just a mock, so I think it should be fine.

src/_pytest/fixtures.py
bluetech: We should probably do them all at once, maybe on the next mypy bump.

PR #6514, opened 2020-01-20
[WIP] Add reference for issue with rootdir

erheron:
Ref. #6376 
I was thinking about adding some example, but maybe just referencing some "common" issue would help?

### DIFF
--- a/doc/en/customize.rst
+++ b/doc/en/customize.rst
@@ -20,7 +20,9 @@ which were registered by installed plugins.
 Initialization: determining rootdir and inifile
 -----------------------------------------------
 
-pytest determines a ``rootdir`` for each test run which depends on
+Pytest defines ``rootdir`` parameter which is determined for each test run.
+It should't be consufed with ``--rootdir`` command-line argument (please see `https://github.com/pytest-dev/pytest/issues/6359` for example).
+Pytest determines a ``rootdir`` for each test run which depends on
 the command line arguments (specified test files, paths) and on
 the existence of *ini-files*.  The determined ``rootdir`` and *ini-file* are
 printed as part of the pytest header during startup.
@@ -118,8 +120,6 @@ check for ini-files as follows:
 
 .. _`how to change command line options defaults`:
 .. _`adding default options`:
-
-
 
 How to change command line options defaults
 ------------------------------------------------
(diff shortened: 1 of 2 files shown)


doc/en/customize.rst
```diff
@@ -20,7 +20,9 @@ which were registered by installed plugins.
 Initialization: determining rootdir and inifile
 -----------------------------------------------
 
-pytest determines a ``rootdir`` for each test run which depends on
+Pytest defines ``rootdir`` parameter which is determined for each test run.
+It should't be consufed with ``--rootdir`` command-line argument (please see `https://github.com/pytest-dev/pytest/issues/6359` for example).
```
hugovk:
```suggestion
It should't be confused with ``--rootdir`` command-line argument (please see `https://github.com/pytest-dev/pytest/issues/6359` for example).
```

doc/en/customize.rst
```diff
@@ -20,7 +20,9 @@ which were registered by installed plugins.
 Initialization: determining rootdir and inifile
 -----------------------------------------------
 
-pytest determines a ``rootdir`` for each test run which depends on
+Pytest defines ``rootdir`` parameter which is determined for each test run.
```
hugovk:
```suggestion
Pytest defines a ``rootdir`` parameter which is determined for each test run.
```

doc/en/customize.rst
blueyed: > It should't be confused with ``--rootdir`` command-line argument

That sounds wrong: `--rootdir` sets `rootdir` (manually).

You are more likely wanting to explain:

> Unfortunately I agree that the current interaction between rootdir and ini files is far from clear, and someone just looking at --rootdir would assume that it should point to the root of your repository/project, and from there on pytest would find pytest.ini files and start collection, but this is currently not the case.

erheron: Oops! Wanted to write exactly that, but in shortenned manner :/ Maybe the better option would be re-writing sentence of @nicoddemus like "It shouldn't be confused with the place, from where pytest would find pytest.ini files and start collection (for example,...)"?

bluetech:
I think that `rootdir` is confusing enough, that adding a "should not be confused with ..." would only be more confusing :smile:

I think the current documentation, which just lays out the two thing that `rootdir` is used for, is the best thing to do, short of changing the actual meaning of `rootdir` as discussed in #6943 

Given also that the PR has been open for a while, I'll close it. Thanks @erheron.

PR #6521, opened 2020-01-21
tests: improve test for `nose.raises`

blueyed:
This should probably get transferred into a `pytest.fail` really, but
tests/documents the current behavior.

### DIFF
--- a/src/_pytest/config/findpaths.py
+++ b/src/_pytest/config/findpaths.py
@@ -1,6 +1,9 @@
 import os
+from typing import Any
+from typing import Iterable
 from typing import List
 from typing import Optional
+from typing import Tuple
 
 import py
 
@@ -60,7 +63,7 @@ def getcfg(args, config=None):
     return None, None, None
 
 
-def get_common_ancestor(paths):
+def get_common_ancestor(paths: Iterable[py.path.local]) -> py.path.local:
     common_ancestor = None
     for path in paths:
         if not path.exists():
@@ -113,7 +116,7 @@ def determine_setup(
     args: List[str],
     rootdir_cmd_arg: Optional[str] = None,
     config: Optional["Config"] = None,
-):
+) -> Tuple[py.path.local, Optional[str], Any]:
     dirs = get_dirs_from_args(args)
     if inifile:
         iniconfig = py.iniconfig.IniConfig(inifile)
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -462,6 +462,7 @@ class Item(Node):
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
+        assert isinstance(location[0], py.path.local), location[0]
         fspath = self.session._node_location_to_relpath(location[0])
         assert type(location[2]) is str
         return (fspath, location[1], location[2])
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -462,7 +462,9 @@ class Item(Node):
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
-        assert isinstance(location[0], py.path.local), location[0]
-        fspath = self.session._node_location_to_relpath(location[0])
+        fspath = location[0]
+        if not isinstance(fspath, py.path.local):
+            fspath = py.path.local(fspath)
+        fspath = self.session._node_location_to_relpath(fspath)
         assert type(location[2]) is str
         return (fspath, location[1], location[2])
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -462,9 +462,10 @@ class Item(Node):
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
-        fspath = location[0]
-        if not isinstance(fspath, py.path.local):
-            fspath = py.path.local(fspath)
-        fspath = self.session._node_location_to_relpath(fspath)
+        if isinstance(location[0], py.path.local):
+            fspath = location[0]
+        else:
+            fspath = py.path.local(location[0])
+        relfspath = self.session._node_location_to_relpath(fspath)
         assert type(location[2]) is str
-        return (fspath, location[1], location[2])
+        return (relfspath, location[1], location[2])
--- a/testing/test_nose.py
+++ b/testing/test_nose.py
@@ -395,9 +395,12 @@ def test_raises(testdir):
             raise BaseException
         """
     )
-    result = testdir.runpytest()
+    result = testdir.runpytest("-vv")
     result.stdout.fnmatch_lines(
         [
+            "test_raises.py::test_raises_runtimeerror PASSED*",
+            "test_raises.py::test_raises_baseexception_not_caught FAILED*",
+            "test_raises.py::test_raises_baseexception_caught PASSED*",
             "*= FAILURES =*",
             "*_ test_raises_baseexception_not_caught _*",
             "",
--- a/testing/test_nose.py
+++ b/testing/test_nose.py
@@ -377,15 +377,45 @@ def test_skip_test_with_unicode(testdir):
     result.stdout.fnmatch_lines(["* 1 skipped *"])
 
 
-def test_issue_6517(testdir):
+def test_raises(testdir):
     testdir.makepyfile(
         """
         from nose.tools import raises
 
         @raises(RuntimeError)
-        def test_fail_without_tcp():
+        def test_raises_runtimeerror():
             raise RuntimeError
+
+        @raises(Exception)
+        def test_raises_baseexception_not_caught():
+            raise BaseException
+
+        @raises(BaseException)
+        def test_raises_baseexception_caught():
+            raise BaseException
         """
     )
     result = testdir.runpytest()
-    result.stdout.fnmatch_lines(["* 1 passed *"])
+    result.stdout.fnmatch_lines(
+        [
+            "*= FAILURES =*",
+            "*_ test_raises_baseexception_not_caught _*",
+            "",
+            "arg = (), kw = {}",
+            "",
+            "    def newfunc(*arg, **kw):",
+            "        try:",
+            ">           func(*arg, **kw)",
+            "",
+            "*/nose/*: ",
+            "_ _ *",
+            "",
+            "    @raises(Exception)",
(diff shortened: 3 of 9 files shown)


testing/test_nose.py
```diff
@@ -377,15 +377,48 @@ def test_io(self):
...
         """
     )
-    result = testdir.runpytest()
-    result.stdout.fnmatch_lines(["* 1 passed *"])
+    result = testdir.runpytest("-vv")
+    result.stdout.fnmatch_lines(
+        [
+            "test_raises.py::test_raises_runtimeerror PASSED*",
+            "test_raises.py::test_raises_baseexception_not_caught FAILED*",
+            "test_raises.py::test_raises_baseexception_caught PASSED*",
```
blueyed: Without the cast here it would display "test_raises.py::test_raises_runtimeerror <- /…/test_raises0/test_raises.py PASSED` here.

nicoddemus: Hmm sorry, what "cast" you mean?

nicoddemus:
LGTM but @bluetech specifically said he would like to revisit his original commit instead of reverting it, so let's wait for his review.

testing/test_nose.py
blueyed: @nicoddemus https://github.com/blueyed/pytest/blob/156a41a855d28ae9f6d83545af69b51fb26f199e/src/_pytest/terminal.py#L747-L752

Since `location` fails to make the path relative it is absolute, and then triggers that code.

blueyed:
> LGTM but @bluetech specifically said he would like to revisit his original commit instead of reverting it, so let's wait for his review.

That's why I've requested it.. ;)

From my point of view I've come up with similar / the same like he did initially, except for that I've noticed (and IIRC mentioned in the review) that it might not be 100% correct, what we've seen now.

Given that this has an additional explicit/hardened test for it I think it is good.
(I would still do the 5.3.5 hotfix release without it.. ;))

nicoddemus:
> (I would still do the 5.3.5 hotfix release without it.. ;))

You would get the exact same error in the next release anyway ;)

src/_pytest/nodes.py
```diff
@@ -462,6 +462,9 @@ def reportinfo(self) -> Tuple[Union[py.path.local, str], Optional[int], str]:
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
-        fspath = self.session._node_location_to_relpath(location[0])
+        fspath = location[0]
+        if not isinstance(fspath, py.path.local):
+            fspath = py.path.local(fspath)
```
blueyed: This might be triggered via https://github.com/pytest-dev/pytest/blob/156a41a855d28ae9f6d83545af69b51fb26f199e/src/_pytest/python.py#L288-L292.

nicoddemus: Thanks I was wondering this myself because a cursory look at `nose.py` did not bring anything to light.

bluetech:
Thanks @blueyed for hardening the test! The test improvements looks good to me, the rest I can't quite asses ATM.

> From my point of view I've come up with similar / the same like he did initially, except for that I've noticed (and IIRC mentioned in the review) that it might not be 100% correct, what we've seen now.

You did -- the revert PR message explains why I was sure the assert was safe, but I'll be more careful with these. Amusingly, this is just the sort of bugs that types help prevent, and we'll get there eventually, but the road can be bumpy :)

---

One thing I'm not sure about in general regarding this part of the code is: are custom `Item`s a thing? i.e. are there plugins which create subclasses of `Item`? I see that at least the internal doctest plugin does this, so I assume yes. And if so, is it considered to be public API? Is `reportinfo` public API (not allowed to break unscrupulously)?

blueyed:
> One thing I'm not sure about in general regarding this part of the code is: are custom `Item`s a thing? i.e. are there plugins which create subclasses of `Item`? I see that at least the internal doctest plugin does this, so I assume yes. And if so, is it considered to be public API? Is `reportinfo` public API (not allowed to break unscrupulously)?

Yes, I would say it is public, and therefore we need to handle `str`s returned from it, as does pytest itself internally still.
With your revert I've left out the adjustment to the test (`test_itemreport_reportinfo`) that you had to adjust, which kind of showed the issue already.
With the hardened test `-vv` triggers similar code.

blueyed:
@nicoddemus 
> > (I would still do the 5.3.5 hotfix release without it.. ;))
> 
> You would get the exact same error in the next release anyway ;)

But there would be more time to discover/fix it.
You see that we have to do a new hotfix release now, don't you?
And that this PR here would have been easier to review without the revert also?  (i.e. if we could leave master "broken" while sorting things out, instead of (re-)reverting there for hotfix releases).

bluetech:
> Yes, I would say it is public, and therefore we need to handle strs returned from it, as does pytest itself internally still.

OK. I think it will be better to use the same signature for all `reportinfo` functions, then. Here are some suggestions (untested...):

```diff
diff --git a/src/_pytest/doctest.py b/src/_pytest/doctest.py
index d7ca888cc..ffc1890d4 100644
--- a/src/_pytest/doctest.py
+++ b/src/_pytest/doctest.py
@@ -308,7 +308,7 @@ class DoctestItem(pytest.Item):
         else:
             return super().repr_failure(excinfo)
 
-    def reportinfo(self) -> Tuple[py.path.local, int, str]:
+    def reportinfo(self) -> Tuple[Union[str, py.path.local], int, str]:
         return self.fspath, self.dtest.lineno, "[doctest] %s" % self.name
 
 
diff --git a/src/_pytest/nodes.py b/src/_pytest/nodes.py
index 080f079cd..ab976efae 100644
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -462,9 +462,10 @@ class Item(Node):
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
-        fspath = location[0]
-        if not isinstance(fspath, py.path.local):
-            fspath = py.path.local(fspath)
-        fspath = self.session._node_location_to_relpath(fspath)
+        if isinstance(location[0], py.path.local):
+            fspath = location[0]
+        else:
+            fspath = py.path.local(location[0])
+        relfspath = self.session._node_location_to_relpath(fspath)
         assert type(location[2]) is str
-        return (fspath, location[1], location[2])
+        return (relfspath, location[1], location[2])
diff --git a/src/_pytest/python.py b/src/_pytest/python.py
index 3d2916c83..82dca3bcc 100644
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -369,7 +369,12 @@ class PyCollector(PyobjMixin, nodes.Collector):
                 if not isinstance(res, list):
                     res = [res]
                 values.extend(res)
-        values.sort(key=lambda item: item.reportinfo()[:2])
+
+        def sort_key(item):
+            fspath, lineno, _ = item.reportinfo()
+            return (str(fspath), lineno)
+
+        values.sort(key=sort_key)
         return values
 
     def _makeitem(self, name, obj):
```

nicoddemus:
> But there would be more time to discover/fix it.

Hardly, people rarely use `pytest` master, even more rarely to do that *and* mix nose constructs.

> You see that we have to do a new hotfix release now, don't you?

I don't. Seems to me like a bug that might happen in any minor release, breaking a small subset of suites (contrary to the one earlier which broke a majority). It can wait for a new release in the normal schedule, I think.

> And that this PR here would have been easier to review without the revert also?

Probably, that's why I asked if we should fix it instead of reverting it, but @bluetech decide to revert it right away and have `master` with working code, and I respect that decision.

blueyed:
> I think it will be better to use the same signature for all `reportinfo` functions, then.

Why?  If it is known to return a `str` only that should be reflected in the signature, no?

bluetech:
I think it adds to the confusion. When all the subclasses publicize the same interface that the base class dictates, there is less chance of errors creeping in when the types are interchanged. But you can leave it as just `py.path.local` if you disagree.

blueyed:
@bluetech 
I think it makes it clearer in general - especially with regard to removing the `py.path.local` instances in the long run it is good to know where only `str`s are returned by now already.
So I'd rather leave it as-is (i.e. more strict / describing the actual return values).

blueyed:
@bluetech are you OK with it then?

src/_pytest/config/findpaths.py
```diff
@@ -113,7 +116,7 @@ def determine_setup(
     args: List[str],
     rootdir_cmd_arg: Optional[str] = None,
     config: Optional["Config"] = None,
-):
+) -> Tuple[py.path.local, Optional[str], Any]:
```
bluetech: I see there is already a `type: py.iniconfig._SectionWrapper` below, so I think we can change the `Any` here to `Union[py.iniconfig._SectionWrapper, Dict[str, Optional[str]]]`.

src/_pytest/nodes.py
```diff
@@ -462,6 +462,9 @@ def reportinfo(self) -> Tuple[Union[py.path.local, str], Optional[int], str]:
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
-        fspath = self.session._node_location_to_relpath(location[0])
+        fspath = location[0]
```
bluetech: I prefer writing it as

```py
        if isinstance(location[0], py.path.local):
            fspath = location[0]
        else:
            fspath = py.path.local(location[0])
        relfspath = self.session._node_location_to_relpath(fspath)
```

This way there is no type confusion.

src/_pytest/config/findpaths.py
blueyed: This fails as follows then though with the return of `{}`:

> error: Incompatible return value type (got "Tuple[local, Optional[str], Union[_SectionWrapper, Dict[<nothing>, <nothing>]]]", expected "Tuple[local, Optional[str], Union[_SectionWrapper, Dict[str, Optional[str]]]]")

Will use just `Dict` there - we can follow up later with "empty dict only" later there (IIRC this requires py36).

blueyed: ..but it fails later with:

> src/_pytest/config/__init__.py:1052: error: Item "Dict[Any, Any]" of "Union[Any, Dict[Any, Any]]" has no attribute "config"

Let's leave it for https://github.com/pytest-dev/pytest/pull/6450.

bluetech: Yeah, let's leave it then.

src/_pytest/nodes.py
blueyed: :+1: Amended.

Issue #6454, opened 2020-01-14
[RFC] Use `-rfE` by default?  (reportchars)


blueyed:
I keep adding this to `addopts` for projects since it is very useful and nice to have.

Should this be enabled by default?

Unfortunately there is no way to disable this then (would require something like `-r=` or `-r-` then, see https://github.com/pytest-dev/pytest/issues/5066).

nicoddemus:
Yeah I also find it extremely useful to have a summary of failures at the end of the test session...

I would be 👍 if we find a way to disable this somehow, like you mention.

bluetech:
For reference for people who are not familiar with the option (like me):

```
  -r chars              show extra test summary info as specified by chars:
                        (f)ailed, (E)rror, (s)kipped, (x)failed, (X)passed,
                        (p)assed, (P)assed with output, (a)ll except passed
                        (p/P), or (A)ll. (w)arnings are enabled by default (see
                        --disable-warnings).
```

And the output looks like this added at the end:

```
================================================================================== short test summary info ===================================================================================
FAILED project/tests/test_quux.py::test_permissions - project.errors.NotAllowed
FAILED project/tests/test_foo.py::test_scenario - project.models.Bar.DoesNotExist: Bar matching query does not exist.
```
---

My opinion is that it will be nice to add by default. However, according to the help, `-ra` also includes expected things like X passed and skipped. IMO would be better not to show those by default.

nicoddemus:
How about this:

* Use `fEsxX` by default
* Include a new `N` option, meaning `None`

So to disable the summary completely user can enter `-rN`.

bluetech:
@nicoddemus I just think that showing skipped (`s`) and xfail (`x`) would be mildly annoying, because it does not indicate a problem, just something intentional the user did, and they will be shown even for successful sessions, which I think are better to keep without a summary.

nicoddemus:
Fair enough, updated by comment adding that option for consideration. 👍

bluetech:
(I think `X` should be included because it does indicate a failure - but maybe I'm confused, because pytest doesn't consider it a failure, just "yellow". I guess the intended use for xfail is more "TODO make it pass" than "ensure this test fails"?).

blueyed:
Yes, `X` (XPASS) is unexpected, but people appear to even use it for flaky things.
It was even green until recently (https://github.com/pytest-dev/pytest/pull/6425).
[...]

Issue #6471, opened 2020-01-16
Introduce new option to skip header and summary


nicoddemus:
#6454 reminds me that sometimes we want to skip header and summary entirely from the output (posting examples online for example). Currently the only way to skip headers is to use `-q`, but this is a global option which affects many other parts (like diff and progress reporting).

How about a `--lean-terminal` or something which only skips both header and summary?

nicoddemus:
A more concrete example:

```
λ python -m pytest src -ra
======================== test session starts ========================
platform win32 -- Python 3.6.8, pytest-5.3.2, py-1.8.1, pluggy-0.13.1
rootdir: C:\Users\bruno\hooksample
collected 1 item

src\pyi_hooksample\__pyinstaller\t\test_hooksample_packaging.py . [100%]

========================= warnings summary ==========================
src/pyi_hooksample/__pyinstaller/t/test_hooksample_packaging.py::test_pyi_hooksample
  C:\Users\bruno\hooksample\.env36\lib\site-packages\PyInstaller\lib\modulegraph\util.py:4: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=================== 1 passed, 1 warning in 4.81s ====================
```

Becomes:

```
λ python -m pytest src -ra --lean-terminal
======================== test session starts ========================
collected 1 item

src\pyi_hooksample\__pyinstaller\t\test_hooksample_packaging.py . [100%]

=================== 1 passed, 1 warning in 4.81s ====================
```

blueyed:
> How about a `--lean-terminal` or something which only skips both header and summary?

What about `--no-header`, for something explicit?

The summary gets skipped by default (except for warnings being on there by default), but `-rN` (or `-r-w` etc) would skip it there then (#6454).

nicoddemus:
> What about --no-header, for something explicit?

I'm using this suggestion in case we go for #6454 and the summary gets shown by default, but you are right that if `-rN` is a thing probably `--no-header` is better.

blueyed:
#6454 needs a way to disable it.
And I agree with you that it would be good to have a single option for it ("`--lean-terminal`"), but also think that it makes sense to have an option to skip the header by itself explicitly.
So instead of `-rN` we could also have `--no-summary` as a long option for clarity (but `-rN` would make sense still).

[...]

Issue #5066, opened 2019-04-07
Remove --disable-warnings option?


blueyed:
It was added in https://github.com/pytest-dev/pytest/commit/e04d9ff80be855164799bb315d23ca611c4b10f4 by @aostr, and its only purpose is to remove the "w" in reportopts.

In https://github.com/pytest-dev/pytest/issues/1668#issuecomment-228550742 it is mentioned that this was discussed with @hpk42.

The main intention here appears to be to enable them by default.

I suggest adding support for `-w` in reportchars instead.

This would also allow for something like `a-s` to have everything, without "skipped" easily.

(came here because I want to add `A` also to also include `pP` passed ones)

nicoddemus:
I like to be able to pass `-w` in front of a report char in order to hide it from the output.

To give some more context, I believe this option was added as a quick way to disable warnings because at that time we have decided to show internal warnings by default, and thought people would get annoyed (and people did). Since then a lot has changed, with us dropping the internal warnings system and showing all warnings by default (and more people got annoyed too hehehe).

But it doesn't seem necessary to actually remove the option, does it? Or is just a matter to cleanup what we believe is mostly an unused option?

blueyed:
> But it doesn't seem necessary to actually remove the option, does it? Or is just a matter to cleanup what we believe is mostly an unused option?

Yes, for cleanup.

> I like to be able to pass `-w` in front of a report char in order to hide it from the output.

Cool.

nicoddemus:
> Yes, for cleanup.

I see, thanks. 👍 

I agree, definitely sounds good. We just have to follow the normal deprecation procedure: emit warnings and only remove it effectively in 5.0. 😁

PR #6524, opened 2020-01-21
terminal: default to `fE` with `-r` (reportchars)

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/6454

### DIFF
--- a/doc/en/usage.rst
+++ b/doc/en/usage.rst
@@ -169,10 +169,10 @@ option you make sure a trace is shown.
 Detailed summary report
 -----------------------
 
-
-
 The ``-r`` flag can be used to display a "short test summary info" at the end of the test session,
 making it easy in large test suites to get a clear picture of all failures, skips, xfails, etc.
+
+It defaults to ``fE`` to list failures and errors.
 
 Example:
 
@@ -261,8 +261,12 @@ Here is the full list of available characters that can be used:
  - ``X`` - xpassed
  - ``p`` - passed
  - ``P`` - passed with output
+
+Special characters for (de)selection of groups:
+
  - ``a`` - all except ``pP``
  - ``A`` - all
+ - ``N`` - none, this can be used to display nothing (since ``fE`` is the default)
 
 More than one character can be used, so for example to only see failed and skipped tests, you can execute:
 
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -33,6 +33,8 @@ from _pytest.reports import TestReport
 
 REPORT_COLLECTING_RESOLUTION = 0.5
 
+_REPORTCHARS_DEFAULT = "fE"
+
 
 class MoreQuietAction(argparse.Action):
     """
@@ -88,12 +90,13 @@ def pytest_addoption(parser):
         "-r",
         action="store",
         dest="reportchars",
-        default="",
+        default=_REPORTCHARS_DEFAULT,
         metavar="chars",
         help="show extra test summary info as specified by chars: (f)ailed, "
         "(E)rror, (s)kipped, (x)failed, (X)passed, "
         "(p)assed, (P)assed with output, (a)ll except passed (p/P), or (A)ll. "
-        "(w)arnings are enabled by default (see --disable-warnings).",
+        "(w)arnings are enabled by default (see --disable-warnings), "
+        "'N' can be used to reset the list. (default: 'fE').",
     )
     group._addoption(
         "--disable-warnings",
@@ -166,24 +169,27 @@ def pytest_configure(config: Config) -> None:
 
 
 def getreportopt(config: Config) -> str:
-    reportopts = ""
     reportchars = config.option.reportchars
-    if not config.option.disable_warnings and "w" not in reportchars:
-        reportchars += "w"
-    elif config.option.disable_warnings and "w" in reportchars:
-        reportchars = reportchars.replace("w", "")
-    aliases = {"F", "S"}
+
+    old_aliases = {"F", "S"}
+    reportopts = ""
     for char in reportchars:
-        # handle old aliases
-        if char in aliases:
+        if char in old_aliases:
             char = char.lower()
         if char == "a":
-            reportopts = "sxXwEf"
+            reportopts = "sxXEf"
         elif char == "A":
-            reportopts = "PpsxXwEf"
-            break
+            reportopts = "PpsxXEf"
+        elif char == "N":
+            reportopts = ""
         elif char not in reportopts:
             reportopts += char
+
+    if not config.option.disable_warnings and "w" not in reportopts:
+        reportopts = "w" + reportopts
+    elif config.option.disable_warnings and "w" in reportopts:
+        reportopts = reportopts.replace("w", "")
+
     return reportopts
 
 
--- a/changelog/6454.bugfix.rst
+++ b/changelog/6454.bugfix.rst
@@ -1,0 +1,1 @@
+`--disable-warnings` is honored with `-ra` and `-rA`.
--- a/changelog/6454.feature.rst
+++ b/changelog/6454.feature.rst
@@ -1,0 +1,1 @@
+Changed default for `-r` to `fE`, which displays failures and errors in the :ref:`short test summary <pytest.detailed_failed_tests_usage>`.  `-rN` can be used to disable it (the old behavior).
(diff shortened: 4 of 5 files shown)


asottile:
I'm still opposed to making the (already bad) situation with needing to scroll for error messages worse

I think I'd be less opposed to this change if we reordered the output so that warnings / logging / etc. are above the stacktraces.

I realize that relative to output / warnings this tends to be only a few lines instead of hundreds, though in larger testsuites this still adds an unwieldy amount of output.

nicoddemus:
Thanks @asottile,

> I'm still opposed to making the (already bad) situation worse with needing to scroll for error messages worse

We are still talking about a few lines though. If you don't have any failures, the new default won't show a summary at all. If there are failures, in addition to several lines of tracebacks (hundreds unless using `--tb=short`) you will also get a 1 line summary per failed test.

I understand your sentiment of "this will make things a bit worse", but I think this is a matter of preference: when you see the current output you think that it is bad, so adding a few lines to the end probably won't make things *that much* worse for you anyway.

On the other hand it will make things more pleasant and useful for people who prefer to see a short summary of failures at the end (like myself and @blueyed). We already show a lot of information anyway by default, I believe a few more lines won't make it that worse for you.

I also agree with the sentiment in general of having a separate option to obtain a leaner output (#6471).

asottile:
I would have no opposition to this if it were an option, but I'd like to see pytest's _default_ be less noisy wherever possible (a slim option I don't think actually addresses the problem that pytest's output is a lot of noise and very little signal by default)

nicoddemus:
> I would have no opposition to this if it were an option

It is already an option, you can use `-rfE` explicitly.

> but I'd like to see pytest's default be less noisy wherever possible

I see, but having pytest being *less verbose by default* is a separate (and more controversial) proposal.

For example, suppose we can give a grade for verbosity, and currently pytest is 8. This proposal is a small increase to that, to say 9. But what you actually want is to reduce verbosity a lot by default, say 2 or 3. 

I understand the sentiment, but that's a different take and separate proposal from this one.

asottile:
if we take that since, the next small change will make it a 10 and then 11 and then 12 -- in my opinion directionally this pr is worse and sets a bad precedent

nicoddemus:
> if we take that since, the next small change will make it a 10 and then 11 and then 12 -- in my opinion directionally this pr is worse and sets a bad precedent

Well I disagree, but I think I've laid all the arguments I had to convince you otherwise, so that's that.

Personally the summary is the first thing I miss once I get more than 1 or 2 failures, so this PR is beneficial in my eyes.

Let's see what others have to say then, if more people feel the same we should then discard the proposal and the PR.

src/_pytest/terminal.py
```diff
@@ -181,7 +184,8 @@ def getreportopt(config: Config) -> str:
             reportopts = "sxXwEf"
         elif char == "A":
             reportopts = "PpsxXwEf"
-            break
+        elif char == "N":
+            reportopts = ""
```
The-Compiler: If I'm reading this right, it means that `-rN` also implies `--disable-warnings`. Is that intended?

blueyed: @The-Compiler yes.  See https://github.com/pytest-dev/pytest/issues/5066.
I've amended/extended the tests a bit.

doc/en/usage.rst
```diff
@@ -261,8 +261,12 @@ Here is the full list of available characters that can be used:
...
  - ``p`` - passed
  - ``P`` - passed with output
+
+Special characters for (de)selection of groups:
+
  - ``a`` - all except ``pP``
  - ``A`` - all
+ - ``N`` - none, this can be used to display nothing (since ``fE`` is the default)
 
 More than one character can be used, so for example to only see failed and skipped tests, you can execute:
```
blueyed: This could also somehow mention that the effective default is `wfE` without `--disable-warnings` actually?  (https://github.com/pytest-dev/pytest/issues/5066)

nicoddemus: Good point, that got me thinking and I think `-rw` should be removed from the docs.

Since we have merged `pytest-warnings` (which seems like decades ago), `-rw` is not really a thing, because independently of what you put in `-r` (except `N` now in this PR) warnings are always displayed:

No `-r` option:

```
λ pytest .tmp\test-reporting.py
...
========================= warnings summary ==========================
test-reporting.py::test_warn
  D:\projects\pytest\.tmp\test-reporting.py:28: UserWarning: some warning
    warnings.warn(UserWarning('some warning'))

====================== short test summary info ======================
FAILED .tmp\test-reporting.py::test_fail[0] - assert 0
FAILED .tmp\test-reporting.py::test_fail[1] - assert 0
FAILED .tmp\test-reporting.py::test_fail[2] - assert 0
=== 3 failed, 4 passed, 3 skipped, 3 xfailed, 1 warning in 0.11s ====
```

Some `-r` combination (note the missing `w`):

```
λ pytest .tmp\test-reporting.py -rXfs
...
========================= warnings summary ==========================
test-reporting.py::test_warn
  D:\projects\pytest\.tmp\test-reporting.py:28: UserWarning: some warning
    warnings.warn(UserWarning('some warning'))

====================== short test summary info ======================
FAILED .tmp\test-reporting.py::test_fail[0] - assert 0
FAILED .tmp\test-reporting.py::test_fail[1] - assert 0
FAILED .tmp\test-reporting.py::test_fail[2] - assert 0
SKIPPED [3] test-reporting.py:15: nas
=== 3 failed, 4 passed, 3 skipped, 3 xfailed, 1 warning in 0.12s ====
```

So the only way to skip the *warnings summary* section is by using either `-rN` or `--disable-warnings`, as of now.

Warnings being controlled in `-r` is historical left-over, so much that *warnings summary* and *short test summary* are separate today.

I propose then:

* Remove `-rw` from the docs, given it doesn't do anything (or mention that it is there for historical reasons but has no effect).
* Make `-rN` not affect warnings at all, warnings can already be disabled with `--disable-warnings`. No backward compatibility here to consider because `N` is new, and the only way to hide warnings was with `--disable-warnings`. This also makes `-rN` the exact behavior as before (hide *short test summary*, but *warnings summary* is still shown).

blueyed: @nicoddemus have you looked at https://github.com/pytest-dev/pytest/issues/5066 again?
There we wanted something like `-r-w` and remove/deprecate the option completely even.

nicoddemus: Ahh I didn't, even thought you mentioned it, my bad.

I've re-read that thread, and I still think what I wrote above holds better, specially because warnings and short summaries are reported and dealt separately. It also means we wouldn't need to deprecate anything really.

nicoddemus: Ahh I didn't, even thought you mentioned it, my bad.

I've re-read that thread, and I still think what I wrote above holds better, specially because warnings and short summaries are reported and dealt separately. It also means we wouldn't need to deprecate anything really.

bluetech:
I still would have preferred X (xpassed) to be included, however seems others disagree, due to some test suites abusing(?) it. So :+1: from me (I didn't review the code).

blueyed:
> I still would have preferred X (xpassed) to be included

Good point.

> however seems others disagree

I think it makes sense also.

> due to some test suites abusing(?) it.

IIRC I've found a reference to it somewhere and just mentioned it (https://github.com/pytest-dev/pytest/issues/6454#issuecomment-574648394).
AFAICS nobody disagrees?

nicoddemus:
>  So 👍 from me (I didn't review the code).

Feel free to approve the PR then. 😁 

> AFAICS nobody disagrees?

Perhaps we should be conservative and just stick with `fE` for now?

blueyed:
> > So +1 from me (I didn't review the code).
> 
> Feel free to approve the PR then. 

Without an actual review? :confused: 

> > AFAICS nobody disagrees?
> 
> Perhaps we should be conservative and just stick with `fE` for now?

Yeah, ok - needs another pass to address your feedback anyway, so I've thought it could go in still also.  But we can create a followup issue for it then.

nicoddemus:
> Without an actual review? 😕

Err you are right, my mistake. 👍

blueyed:
Amended for `N` to not reset `w`.
Moved handling of `w` to the end also, since "warnings and short summaries are reported and dealt separately" according to https://github.com/pytest-dev/pytest/pull/6524#discussion_r369854033.

This also changes the weird behavior that `-ra` and `-rA` would previously include warnings even though `--disable-warnings` was used.

blueyed:
> This also changes the weird behavior that `-ra` and `-rA` would previously include warnings even though `--disable-warnings` was used.

Might be worth a separate changelog entry?

nicoddemus:
> Might be worth a separate changelog entry?

Sounds good, but you might just use the same entry and just add another paragraph (your call). 👍

blueyed:
Added `changelog/6454.bugfix.rst`.

blueyed:
Will merge once CI passes.

PR #6525, opened 2020-01-21
typing: Session.__init__

blueyed:
Pulled out of https://github.com/pytest-dev/pytest/pull/6491.

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -6,19 +6,28 @@ import importlib
 import os
 import sys
 from typing import Dict
+from typing import FrozenSet
+from typing import List
 
 import attr
 import py
 
 import _pytest._code
 from _pytest import nodes
+from _pytest.compat import TYPE_CHECKING
+from _pytest.config import Config
 from _pytest.config import directory_arg
 from _pytest.config import hookimpl
 from _pytest.config import UsageError
 from _pytest.fixtures import FixtureManager
+from _pytest.nodes import Node
 from _pytest.outcomes import exit
 from _pytest.runner import collect_one_node
 from _pytest.runner import SetupState
+
+
+if TYPE_CHECKING:
+    from _pytest.python import Package
 
 
 class ExitCode(enum.IntEnum):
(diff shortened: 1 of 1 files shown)


src/_pytest/main.py
```diff
@@ -383,7 +391,7 @@ class Session(nodes.FSCollector):
     # Set on the session by fixtures.pytest_sessionstart.
     _fixturemanager = None  # type: FixtureManager
 
-    def __init__(self, config):
+    def __init__(self, config) -> None:
```
nicoddemus: Out of curiosity, is considered good practice to annotate the return value of all functions even if they return `None`?

blueyed: It has to be for the function to be considered to be typed.
I've amended it to also type `config` (the arg) explicitly, although done via base class.

nicoddemus: I see, thanks makes sense.

Now it brings the next question, is:

```python
def __init__(self, config: Config):
```

enough for it to be considered typed?

blueyed: Yes (AFAIK), but might fail with strict options etc?!  /cc @bluetech

blueyed: btw: `reveal_type(var)` is very useful in that regard.

src/_pytest/main.py
```diff
@@ -394,14 +403,16 @@ def __init__(self, config):
         self.trace = config.trace.root.get("collection")
         self._norecursepatterns = config.getini("norecursedirs")
         self.startdir = config.invocation_dir
-        self._initialpaths = frozenset()
+        self._initialpaths = frozenset()  # type: FrozenSet[py.path.local]
+
         # Keep track of any collected nodes in here, so we don't duplicate fixtures
-        self._node_cache = {}
+        self._node_cache = {}  # type: Dict[str, List[Node]]
```
bluetech: Apparently sometimes the key can be tuple so the type is not exactly correct:

https://github.com/pytest-dev/pytest/blob/ad02f6f8790df4252ba9af69aa76c58eb105a1db/src/_pytest/main.py#L574-L579

I am working on some annotations and will include a fix for this, hopefully tomorrow.

PR #6527, opened 2020-01-21
typing: fix/adjust _code.source.getfslineno

blueyed:


bluetech:
:+1: on the Optional. Curious that it wasn't caught.

Since the caller already needs to handle `None`, does it make sense to exchange the `return '', 1` with `return None, -1`? Then the type will be simpler. But I didn't check the callers.

blueyed:
> Since the caller already needs to handle `None`, does it make sense to exchange the `return '', 1` with `return None, -1`? Then the type will be simpler. But I didn't check the callers.

I'm reluctant to messing with code here, but just want to type what is used.

blueyed:
I think `None` might only happen in case `inspect` would return an empty string (a theoretical case I guess).
Also curious why it wasn't caught by mypy, but have not checked.

PR #6531, opened 2020-01-21
ci: GHA: remove unnecessary check for `success()`

blueyed:
Following jobs get aborted on failure.

Ref: https://github.com/pytest-dev/pytest/pull/6530


nicoddemus:
Oh right, thanks.

PR #6533, opened 2020-01-22
ci: move linting to separate job, cache for pre-commit

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/6528.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,7 +42,7 @@ jobs:
           "macos-py37",
           "macos-py38",
 
-          "linting",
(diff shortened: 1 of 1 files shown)


nicoddemus:
Please when closing PRs/issues, leave a note explaining why it was closed. The PR stays on the issue history, so it is interesting to know why this was discarded.

blueyed:
@nicoddemus 
Sure.
I think it is too complicated, and would rather use caching in a more generic way.
I think it makes sense to have `-coverage` tox envs also, and not a `run-with-coverage` flag.

PR #6534, opened 2020-01-22
tests: add test_plugin_loading_order

blueyed:
Ref: https://github.com/pytest-dev/pytest/pull/6443


blueyed:
Should we have this already while waiting on https://github.com/pytest-dev/pytest/pull/6443?
I can include it there otherwise (with changed order), but I think it is good having an explicit test for the current behavior.  Please ACK, or close.

nicoddemus:
Seperate is better 👍

PR #6544, opened 2020-01-23
doc: remove costlysetup example

blueyed:
It is not included with docs, and
`example/costlysetup/sub_a/test_quick.py::test_quick` sleeps for 5s,
slowing down `doctesting` unnecessarily.


blueyed:
@nicoddemus from what I can tell it was never in the toctree of the docs.

PR #6545, opened 2020-01-23
config: typing for create_terminal_writer, re-export TerminalWriter

blueyed:
This also imports `TerminalWriter` explicitly, allowing for easier
phasing out / replacing of it (via the import, which then could go
through the `_pytest.compat` module).

Note to self:
```
pick b34f07360 config: typing for create_terminal_writer
pick 443335426 Import TerminalWriter from `py.io`
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,7 @@ repos:
                 _code\.|
                 builtin\.|
                 code\.|
-                io\.(BytesIO|saferepr)|
+                io\.(BytesIO|saferepr|TerminalWriter)|
                 path\.local\.sysfind|
                 process\.|
                 std\.
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -29,6 +29,7 @@ import pluggy
 import py
 
 import _pytest
+from _pytest._io import TerminalWriter
 from _pytest._io.saferepr import safeformat
 from _pytest._io.saferepr import saferepr
 from _pytest.compat import overload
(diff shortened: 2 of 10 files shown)


src/_pytest/terminal.py
```diff
@@ -357,6 +357,9 @@ def rewrite(self, line, **markup):
 
     def write_sep(self, sep, title=None, **markup):
         self.ensure_newline()
+        if sep == "_":
```
bluetech: This seems unrelated?

blueyed: Fixed.  That's what I want to use it for, but it needs an adjusted TerminalWriter, of course.

I wonder if I should import it via `_pytest.compat` here already, so that there's really a single place where to extend it, and/or put it into `_pytest._io`?

blueyed:
I wonder if I should import it via `_pytest.compat` here already, so that there's really a single place where to extend/change it, and/or put it into `_pytest._io` in the first place - which would be just the import for now, but could later extend the class, i.e. to support "underline" markup?

bluetech:
Importing from `_pytest._io` makes sense to me.

blueyed:
> Importing from `_pytest._io` makes sense to me.

Moved it there.  Not sure if it should use `__all__`, explicit `# noqa: F401` etc though.

bluetech:
Mypy has a convention (a bit more than a convention actually, given the `--no-implicit-reexport` flag) where reexports are done like

```py
from py.io import TerminalWriter as TerminalWriter
```

I don't know how flake8 would like it, but I'd do it like this with a comment explaining anyway:

```py
# Reexport TerminalWriter from here instead of py, to make it easier to
# extend or swap our own implementation in the future.
from py.io import TerminalWriter as TerminalWriter  # noqa: F<something>
```

blueyed:
Thanks, amended:
```diff
diff --git c/src/_pytest/_io/__init__.py i/src/_pytest/_io/__init__.py
index 6265353d8..047bb179a 100644
--- c/src/_pytest/_io/__init__.py
+++ i/src/_pytest/_io/__init__.py
@@ -1 +1,3 @@
-from py.io import TerminalWriter
+# Reexport TerminalWriter from here instead of py, to make it easier to
+# extend or swap our own implementation in the future.
+from py.io import TerminalWriter as TerminalWriter  # noqa: F401
diff --git c/tox.ini i/tox.ini
index fae23b882..707f239d0 100644
--- c/tox.ini
+++ i/tox.ini
@@ -189,8 +189,6 @@ markers =
 [flake8]
 max-line-length = 120
 extend-ignore = E203
-per-file-ignores =
-  src/_pytest/_io/__init__.py: F401

 [isort]
 ; This config mimics what reorder-python-imports does.
```

PR #6547, opened 2020-01-23
Refactor Session._initialparts to have a more explicit type

bluetech:
Previously, `_initialparts` was a list whose first item was a
`py.path.local` and the rest were `str`s. This is not something that
mypy is capable of modeling. The type `List[Union[str, py.path.local]]`
is too broad and would require asserts for every access.

Instead, make each item a `Tuple[py.path.local, List[str]]`. This way
the structure is clear and the types are accurate.

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -8,6 +8,7 @@ import sys
 from typing import Dict
 from typing import FrozenSet
 from typing import List
+from typing import Tuple
 
 import attr
 import py
@@ -485,13 +486,13 @@ class Session(nodes.FSCollector):
         self.trace("perform_collect", self, args)
         self.trace.root.indent += 1
         self._notfound = []
-        initialpaths = []
-        self._initialparts = []
+        initialpaths = []  # type: List[py.path.local]
+        self._initial_parts = []  # type: List[Tuple[py.path.local, List[str]]]
         self.items = items = []
         for arg in args:
-            parts = self._parsearg(arg)
-            self._initialparts.append(parts)
-            initialpaths.append(parts[0])
+            fspath, parts = self._parsearg(arg)
+            self._initial_parts.append((fspath, parts))
+            initialpaths.append(fspath)
         self._initialpaths = frozenset(initialpaths)
         rep = collect_one_node(self)
         self.ihook.pytest_collectreport(report=rep)
@@ -511,13 +512,13 @@ class Session(nodes.FSCollector):
             return items
 
     def collect(self):
-        for initialpart in self._initialparts:
-            self.trace("processing argument", initialpart)
+        for fspath, parts in self._initial_parts:
+            self.trace("processing argument", (fspath, parts))
             self.trace.root.indent += 1
             try:
-                yield from self._collect(initialpart)
+                yield from self._collect(fspath, parts)
             except NoMatch:
-                report_arg = "::".join(map(str, initialpart))
+                report_arg = "::".join((str(fspath), *parts))
                 # we are inside a make_report hook so
                 # we cannot directly pass through the exception
                 self._notfound.append((report_arg, sys.exc_info()[1]))
@@ -526,11 +527,8 @@ class Session(nodes.FSCollector):
         self._collection_node_cache.clear()
         self._collection_pkg_roots.clear()
 
-    def _collect(self, arg):
+    def _collect(self, argpath, names):
         from _pytest.python import Package
-
-        names = arg[:]
-        argpath = names.pop(0)
 
         # Start with a Session root, and delve to argpath item (dir or file)
         # and stack all Packages found on the way.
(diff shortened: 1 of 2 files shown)


src/_pytest/main.py
```diff
@@ -660,19 +658,19 @@ def _tryconvertpyarg(self, x):
 
     def _parsearg(self, arg):
```
blueyed: Add typing for the signature?

src/_pytest/main.py
```diff
@@ -508,25 +509,22 @@ def _perform_collect(self, args, genitems):
...
-                report_arg = "::".join(map(str, initialpart))
+                report_arg = "::".join(part for part in [str(fspath), *parts])
                 # we are inside a make_report hook so
                 # we cannot directly pass through the exception
                 self._notfound.append((report_arg, sys.exc_info()[1]))
 
             self.trace.root.indent -= 1
 
-    def _collect(self, arg):
+    def _collect(self, argpath, names):
```
blueyed: Typing for signature?

src/_pytest/main.py
```diff
@@ -508,25 +509,22 @@ def _perform_collect(self, args, genitems):
             return items
 
     def collect(self):
```
blueyed: `-> Generator[…]`

bluetech:
@blueyed I forgot to mention, I'm working on these types you suggest, but I decided not to mix code changes with typing changes, so that they can be considered separately. I will send the typing changes in a separate PR.

src/_pytest/main.py
```diff
@@ -508,25 +509,22 @@ def _perform_collect(self, args, genitems):
...
-            self.trace("processing argument", initialpart)
+        for fspath, parts in self._initialparts:
+            self.trace("processing argument", (fspath, parts))
             self.trace.root.indent += 1
             try:
-                yield from self._collect(initialpart)
+                yield from self._collect(fspath, parts)
             except NoMatch:
-                report_arg = "::".join(map(str, initialpart))
+                report_arg = "::".join(part for part in [str(fspath), *parts])
```
blueyed:
```suggestion
                report_arg = "::".join(part for part in (str(fspath), *parts))
```

bluetech: Can be simplified further actually:

```diff
-                report_arg = "::".join(part for part in [str(fspath), *parts])
+                report_arg = "::".join((str(fspath), *parts))
```

bluetech:
Thanks for review @blueyed @nicoddemus.

> Perhaps we should target features instead though, even if this is renaming an internal attribute?

I prefer features too because I don't think these changes need to go on patch releases. But @blueyed sends typing changes to master so I started following suit. I must say I am still confused by this :smile:. I think I and @blueyed should be synchronized on this, WDYT @blueyed?

If I have time this weekend, I'll try to address your comments on going forward with the master/features workflow changes. Then we won't have this dilemma :)

nicoddemus:
> I prefer features too because I don't think these changes need to go on patch releases. But @blueyed sends typing changes to master so I started following suit.

I think adding type hints to `master` is fine, but in this case we are actually changing the attribute, from a list to a list of tuples, so it affects third party code using it (even though it shouldn't be accessing it).

Now that I mention this, perhaps we should *also* rename it? I think it is good practice to rename an attribute when the underlying data changes like this one, because I think an `AttributeError` would be better than the other kinds of errors this could cause. If people agree, I would suggest rename it to `_initial_parts` then.

bluetech:
> If people agree, I would suggest rename it to _initial_parts then.

Good idea, I'll do it.

> I think adding type hints to master is fine, but in this case we are actually changing the attribute

The tricky part here is that I only make this change so that further typing changes can depend on it. So they will need to go to features. As I said I actually prefer it, but I'm not sure @blueyed does.

blueyed:
> > Perhaps we should target features instead though, even if this is renaming an internal attribute?

I think we do not need to care about compatibility for internal attributes really, but I agree that this should be avoided in bugfix releases.

Currently this means that this should go into features, but given the situation described in https://github.com/pytest-dev/pytest/pull/6387#issuecomment-576871496 I actually think that we're about to drop features soon, and consider master what features is currently for us, using release branches for bugfix releases.

> If I have time this weekend, I'll try to address your comments on going forward with the master/features workflow changes. Then we won't have this dilemma :)

That would be great.  I've also started looking at the documented release process briefly (https://github.com/pytest-dev/pytest/pull/6549).  In my opinion we could start with a `release-5.3` branch based on ddfa41b (the merge of 5.3.4).  We should probably continue the discussion in #6387, and/or create a new issue to discuss/outline the process.  This could also be done in #6549 maybe.

Anyway, for this PR I think it is better to stay with master, which shouldn't really be a problem when we're creating the next 5.3.x release as suggested above (and like I've suggested for 5.3.4 already).

nicoddemus:
> using release branches for bugfix releases.

This is new to me... I thought we would just release from `master` every time, and what would decide if it was a bugfix release or a feature release would be the presence of new features on the branch, by looking at the changelog.

Or did you mean hotfixes?

(Anyway we should probably discuss this in another venue, sorry for the noise).

nicoddemus:
> I think we do not need to care about compatibility for internal attributes really, but I agree that this should be avoided in bugfix releases.

Yes that's the point. 👍 

Even if a plugin is wrongly accessing it, I would not like to "break" it for users in a bugfix release.

bluetech:
Rebased on features.

PR #6549, opened 2020-01-23
doc: minor fixes for the release process

blueyed:


scripts/release.patch.rst
```diff
@@ -3,7 +3,7 @@ pytest-{version}
 
 pytest {version} has just been released to PyPI.
 
-This is a bug-fix release, being a drop-in replacement. To upgrade::
+This is a bug fix release, being a drop-in replacement. To upgrade::
```
hugovk: I think this one was actually correct, see general principle 3 and compare "client-centered counseling" at https://apastyle.apa.org/learn/faqs/when-use-hyphen

blueyed: The last time it was said that "bug fix" should be preferred over "bugfix".
Google says "did you mean 'bugfix'?" for both of "bug-fix" and "bug fix".
So I think we should go with "bugfix" also.

hugovk: Computing examples: 
https://docs.microsoft.com/en-gb/style-guide/punctuation/dashes-hyphens/hyphens#noun-modifiers

hugovk: "Bug fix" as a standalone noun phrase, "bug-fix release" as a compound modifying another noun.

blueyed: @hugovk what about Google's "did you mean?", and more results in the form I am used to "bugfix" and "bugfix release"?
Anyway, I can revert this, but thought it would be consistent with "bug fix" from the other discussion.
Are you a native speaker btw? :)

hugovk: Yes, I am a native (UK) speaker :) And I've re-read the Microsoft style guide, which says to hyphenate here if "Confusion might result without the hyphen". APA style says do hyphenate, so it seems not a strict rule as such, more a matter of style. [This](https://sites.utexas.edu/legalwriting/2014/07/21/the-compound-modifier-hyphen-connects-and-clarifies/) says it can be easier to always hyphenate than have to judge case-by-case whether one is needed.

So I take back my original comment and approve all the changes here, sorry for any confusion!

As for Google's "did you mean?", I don't know where it comes from and isn't reliable language advice. I don't even see those suggestions when I search.

PR #6553, opened 2020-01-24
ci: GHA: use scripts/report-coverage.sh

blueyed:
This allows for more control, e.g. `-X fix` with codecov-bash, and it has all the coverage-calls we need already etc.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -154,21 +154,9 @@ jobs:
       run: |
         python scripts/append_codecov_token.py
 
-    - name: Combine coverage
+    - name: Report coverage
       if: (!matrix.skip_coverage)
-      run: |
-        python -m coverage combine
-        python -m coverage xml
-
-    - name: Codecov upload
-      if: (!matrix.skip_coverage)
-      uses: codecov/codecov-action@v1
-      with:
-        token: ${{ secrets.codecov }}
-        file: ./coverage.xml
-        flags: GHA
-        fail_ci_if_error: false
-        name: ${{ matrix.name }}
+      run: bash scripts/report-coverage.sh -F GHA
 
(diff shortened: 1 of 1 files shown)


nicoddemus:
TBH I prefer using the action instead of a custom script, so I'm -0 (but not completely opposed).

blueyed:
@nicoddemus 
Why do you prefer the action?  It wraps codecov-bash apparently also only (https://github.com/pytest-dev/pytest/pull/6463#issuecomment-578092851), while not handling e.g. retries which we have with our `curl` invocation already.
Also it seems to be basically a large, obfuscated code blob, while using the bash script ourselves is more transparent.
Additionally, it might result in differences with regard to what is reported between Travis and GHA.

blueyed:
Alternative: https://github.com/pytest-dev/pytest/pull/6560 - but failed on the first run already.

nicoddemus:
Actually it does retries automatically, I've seen it sleeping for 30s a number of times before trying to upload again before codecov doesn't respond.

> Also it seems to be basically a large, obfuscated code blob, while using the bash script ourselves is more transparent.

Not sure if you mean the yaml configuration or the action code itself.

> Additionally, it might result in differences with regard to what is reported between Travis and GHA.

But this is temporary, I hope. Which would be fine to narrow down why GH action coverage is less than Travis.

Like I said, I'm -0 on this, so please go ahead if think it is better. 👍

blueyed:
> Not sure if you mean the yaml configuration or the action code itself.

The JS code (the action).

> But this is temporary, I hope. Which would be fine to narrow down why GH action coverage is less than Travis.

Well, it also is helpful to have it failing.  As seen in #6560 it failed twice in a row, which likely is the reason for flaky coverage.

PR #6554, opened 2020-01-24
tests: add test_fixture_arg_ordering

blueyed:
This is a regression test for part of
https://github.com/pytest-dev/pytest/issues/6492, testing one of the
fixes in https://github.com/pytest-dev/pytest/pull/6551.

### DIFF
--- a/testing/python/fixtures.py
+++ b/testing/python/fixtures.py
@@ -4207,3 +4207,38 @@ def test_fixture_parametrization_nparray(testdir):
     )
     result = testdir.runpytest()
     result.assert_outcomes(passed=10)
+
+
+def test_fixture_arg_ordering(testdir):
+    """
+    This test describes how fixtures in the same scope but without explicit dependencies
+    between them are created. While users should make dependencies explicit, often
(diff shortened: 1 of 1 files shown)


testing/python/fixtures.py
```diff
@@ -4207,3 +4207,32 @@ def test_bug(value):
     )
     result = testdir.runpytest()
     result.assert_outcomes(passed=10)
+
+
+def test_fixture_arg_ordering(testdir):
```
nicoddemus:
```suggestion
def test_fixture_arg_ordering(testdir):
    """
    This test describes how fixtures in the same scope but without explicit dependencies
    between them are created. While users should make dependencies explicit, often
    they rely on this order, so this test exists to catch regressions in this regard.
    See #6540 and #6492.
    """ 
```

blueyed: @nicoddemus 
Thanks, please force-push your suggestion and merge then.

nicoddemus: Done. 👍

PR #6556, opened 2020-01-24
Type annotate main.py and some parts related to collection 

bluetech:
~~Depends on #6547 and #6555 (and includes them until merged).~~

This adds type annotations to main.py. To do this, I needed to feel in some blanks in other parts, mostly related to collection.

It took some effort to untangle but I think I got it mostly right...

### DIFF
--- a/src/_pytest/doctest.py
+++ b/src/_pytest/doctest.py
@@ -7,6 +7,7 @@ import traceback
 import warnings
 from contextlib import contextmanager
 from typing import Dict
+from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Sequence
@@ -108,13 +109,18 @@ def pytest_unconfigure():
     RUNNER_CLASS = None
 
 
-def pytest_collect_file(path, parent):
+def pytest_collect_file(
+    path: py.path.local, parent
+) -> Optional[Union["DoctestModule", "DoctestTextfile"]]:
     config = parent.config
     if path.ext == ".py":
         if config.option.doctestmodules and not _is_setup_py(config, path, parent):
-            return DoctestModule.from_parent(parent, fspath=path)
+            mod = DoctestModule.from_parent(parent, fspath=path)  # type: DoctestModule
+            return mod
     elif _is_doctest(config, path, parent):
-        return DoctestTextfile.from_parent(parent, fspath=path)
+        txt = DoctestTextfile.from_parent(parent, fspath=path)  # type: DoctestTextfile
+        return txt
+    return None
 
 
 def _is_setup_py(config, path, parent):
@@ -361,7 +367,7 @@ def _get_continue_on_failure(config):
 class DoctestTextfile(pytest.Module):
     obj = None
 
-    def collect(self):
+    def collect(self) -> Iterable[DoctestItem]:
         import doctest
 
         # inspired by doctest.testfile; ideally we would use it directly,
@@ -440,7 +446,7 @@ def _patch_unwrap_mock_aware():
 
 
 class DoctestModule(pytest.Module):
-    def collect(self):
+    def collect(self) -> Iterable[DoctestItem]:
         import doctest
 
         class MockAwareDocTestFinder(doctest.DocTestFinder):
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -1,13 +1,20 @@
 """ hook specifications for pytest plugins, invoked from main.py and builtin plugins.  """
 from typing import Any
+from typing import List
 from typing import Optional
+from typing import Union
 
+import py.path
 from pluggy import HookspecMarker
 
 from _pytest.compat import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from _pytest.main import Session
+    from _pytest.nodes import Collector
+    from _pytest.nodes import Item
+    from _pytest.python import Module
+    from _pytest.python import PyCollector
 
 
 hookspec = HookspecMarker("pytest")
@@ -215,7 +222,7 @@ def pytest_collect_directory(path, parent):
     """
 
 
-def pytest_collect_file(path, parent):
+def pytest_collect_file(path: py.path.local, parent) -> "Optional[Collector]":
     """ return collection Node or None for the given path. Any new node
     needs to have the specified ``parent`` as a parent.
 
@@ -255,7 +262,7 @@ def pytest_make_collect_report(collector):
 
 
 @hookspec(firstresult=True)
-def pytest_pycollect_makemodule(path, parent):
+def pytest_pycollect_makemodule(path: py.path.local, parent) -> "Optional[Module]":
     """ return a Module collector or None for the given path.
     This hook will be called for each matching test module path.
     The pytest_collect_file hook needs to be used if you want to
@@ -268,7 +275,9 @@ def pytest_pycollect_makemodule(path, parent):
 
 
 @hookspec(firstresult=True)
-def pytest_pycollect_makeitem(collector, name, obj):
+def pytest_pycollect_makeitem(
+    collector: "PyCollector", name: str, obj
+) -> "Union[None, Item, Collector, List[Union[Item, Collector]]]":
     """ return custom item/collector for a python object in a module, or None.
 
     Stops at first non-None result, see :ref:`firstresult` """
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -5,11 +5,14 @@ import importlib
 import os
 import sys
 from typing import Callable
+from typing import cast
 from typing import Dict
 from typing import FrozenSet
+from typing import Iterator
 from typing import List
 from typing import Optional
 from typing import Sequence
+from typing import Set
 from typing import Tuple
 from typing import Union
 
@@ -18,12 +21,14 @@ import py
 
 import _pytest._code
 from _pytest import nodes
+from _pytest.compat import overload
 from _pytest.compat import TYPE_CHECKING
 from _pytest.config import Config
 from _pytest.config import directory_arg
 from _pytest.config import ExitCode
 from _pytest.config import hookimpl
 from _pytest.config import UsageError
+from _pytest.config.argparsing import Parser
 from _pytest.fixtures import FixtureManager
 from _pytest.outcomes import Exit
 from _pytest.reports import CollectReport
@@ -33,11 +38,12 @@ from _pytest.runner import SetupState
 
 if TYPE_CHECKING:
     from typing import Type
+    from typing_extensions import Literal
 
     from _pytest.python import Package
 
 
-def pytest_addoption(parser):
+def pytest_addoption(parser: Parser) -> None:
     parser.addini(
         "norecursedirs",
         "directory patterns to avoid for recursion",
@@ -237,7 +243,7 @@ def wrap_session(
     return session.exitstatus
 
 
-def pytest_cmdline_main(config):
+def pytest_cmdline_main(config: Config) -> Union[int, ExitCode]:
     return wrap_session(config, _main)
 
 
@@ -254,11 +260,11 @@ def _main(config: Config, session: "Session") -> Optional[Union[int, ExitCode]]:
     return None
 
 
-def pytest_collection(session):
+def pytest_collection(session: "Session") -> Sequence[nodes.Item]:
     return session.perform_collect()
 
 
-def pytest_runtestloop(session):
+def pytest_runtestloop(session: "Session") -> bool:
     if session.testsfailed and not session.config.option.continue_on_collection_errors:
         raise session.Interrupted(
             "%d error%s during collection"
@@ -278,7 +284,7 @@ def pytest_runtestloop(session):
     return True
 
 
-def _in_venv(path):
+def _in_venv(path: py.path.local) -> bool:
     """Attempts to detect if ``path`` is the root of a Virtual Environment by
     checking for the existence of the appropriate activate script"""
     bindir = path.join("Scripts" if sys.platform.startswith("win") else "bin")
@@ -295,7 +301,7 @@ def _in_venv(path):
     return any([fname.basename in activates for fname in bindir.listdir()])
 
 
-def pytest_ignore_collect(path, config):
+def pytest_ignore_collect(path: py.path.local, config: Config) -> bool:
     ignore_paths = config._getconftest_pathlist("collect_ignore", path=path.dirpath())
     ignore_paths = ignore_paths or []
     excludeopt = config.getoption("ignore")
@@ -323,7 +329,7 @@ def pytest_ignore_collect(path, config):
     return False
 
 
-def pytest_collection_modifyitems(items, config):
+def pytest_collection_modifyitems(items, config: Config) -> None:
     deselect_prefixes = tuple(config.getoption("deselect") or [])
     if not deselect_prefixes:
         return
@@ -380,8 +386,8 @@ class Session(nodes.FSCollector):
         )
         self.testsfailed = 0
         self.testscollected = 0
-        self.shouldstop = False
-        self.shouldfail = False
+        self.shouldstop = False  # type: Union[bool, str]
+        self.shouldfail = False  # type: Union[bool, str]
         self.trace = config.trace.root.get("collection")
         self.startdir = config.invocation_dir
         self._initialpaths = frozenset()  # type: FrozenSet[py.path.local]
@@ -407,10 +413,11 @@ class Session(nodes.FSCollector):
         self.config.pluginmanager.register(self, name="session")
 
     @classmethod
-    def from_config(cls, config):
-        return cls._create(config)
+    def from_config(cls, config: Config) -> "Session":
+        session = cls._create(config)  # type: Session
+        return session
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "<%s %s exitstatus=%r testsfailed=%d testscollected=%d>" % (
             self.__class__.__name__,
             self.name,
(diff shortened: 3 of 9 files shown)


src/_pytest/main.py
```diff
@@ -572,29 +629,30 @@ def _collect(self, arg):
...
                 if col:
-                    self._node_cache[argpath] = col
+                    self._node_cache1[argpath] = col
             m = self.matchnodes(col, names)
             # If __init__.py was the only file requested, then the matched node will be
             # the corresponding Package, and the first yielded item will be the __init__
             # Module itself, so just use that. If this special case isn't taken, then all
             # the files in the package will be yielded.
             if argpath.basename == "__init__.py":
+                assert isinstance(m[0], nodes.Collector)
```
bluetech: Needed to add an assert here (can also be a `nodes.Item`), but seems safe given it should be a `Package` and `.collect()` is called just below.

src/_pytest/main.py
```diff
@@ -572,29 +629,30 @@ def _collect(self, arg):
...
                 try:
-                    yield next(m[0].collect())
+                    yield next(iter(m[0].collect()))
```
bluetech: Needed to add `iter()` here because I made the official type of `collect()` be `Iterable` and not `Iterator` (which isn't the case for all implementations). Calling `iter()` on an iterator should just be idempotent, so I think it's safe.

src/_pytest/main.py
```diff
@@ -651,30 +715,32 @@ def _tryconvertpyarg(self, x):
         # ValueError: not a module name
         except (AttributeError, ImportError, ValueError):
             return x
-        if spec is None or spec.origin in {None, "namespace"}:
+        if spec is None or spec.origin is None or spec.origin == "namespace":
```
bluetech: Had to split it because mypy doesn't handle `in {None}` for type narrowing.

src/_pytest/python.py
```diff
@@ -571,20 +587,20 @@ def setup(self):
...
             self.addfinalizer(func)
 
-    def _recurse(self, dirpath):
+    def _recurse(self, dirpath: py.path.local) -> bool:
         if dirpath.basename == "__pycache__":
             return False
         ihook = self.gethookproxy(dirpath.dirpath())
         if ihook.pytest_ignore_collect(path=dirpath, config=self.config):
-            return
+            return False
```
bluetech: Had to make it a proper False - shouldn't matter to `py.path.local` recurse as long as it's falsey.

src/_pytest/python.py
```diff
@@ -683,10 +704,11 @@ def _get_first_non_fixture_func(obj, names):
 class Class(PyCollector):
     """ Collector for test methods. """
 
-    def collect(self):
+    def collect(self) -> Iterable[Union[nodes.Item, nodes.Collector]]:
         if not safe_getattr(self.obj, "__test__", True):
             return []
         if hasinit(self.obj):
+            assert self.parent is not None
```
bluetech: This assert and below should be safe since dereferenced just below.

src/_pytest/main.py
```diff
@@ -391,22 +412,30 @@ class Session(nodes.FSCollector):
     _setupstate = None  # type: SetupState
     # Set on the session by fixtures.pytest_sessionstart.
     _fixturemanager = None  # type: FixtureManager
+    exitstatus = None  # type: Union[int, ExitCode]
```
bluetech: Didn't find any place which does `hasattr('exitstatus')` or such, so I think this is safe.

src/_pytest/main.py
```diff
@@ -507,26 +564,26 @@ def _perform_collect(self, args, genitems):
...
             try:
-                yield from self._collect(initialpart)
+                yield from self._collect(fspath, parts)
             except NoMatch:
-                report_arg = "::".join(map(str, initialpart))
+                report_arg = "::".join((str(fspath), *parts))
                 # we are inside a make_report hook so
                 # we cannot directly pass through the exception
-                self._notfound.append((report_arg, sys.exc_info()[1]))
+                exc = cast(NoMatch, sys.exc_info()[1])
```
bluetech: I don't really follow the comment above. I *think* this cast is correct...

src/_pytest/doctest.py
```diff
@@ -108,13 +109,16 @@ def pytest_unconfigure():
     RUNNER_CLASS = None
 
 
-def pytest_collect_file(path, parent):
+def pytest_collect_file(
+    path: py.path.local, parent
```
blueyed: Type `parent`?
You've left it out below also, so there's likely a reason?

src/_pytest/hookspec.py
```diff
@@ -1,6 +1,19 @@
 """ hook specifications for pytest plugins, invoked from main.py and builtin plugins.  """
+from typing import List
+from typing import Optional
+from typing import Union
+
+import py
```
blueyed:
```suggestion
import py.path
```
I've found this to be better with regard to seeing which parts of `py` are used in a file still.

src/_pytest/main.py
```diff
@@ -349,12 +368,14 @@ def pytest_collection_modifyitems(items, config):
...
-    def __init__(self, fspath, pm, remove_mods):
+    def __init__(
+        self, fspath: py.path.local, pm: PytestPluginManager, remove_mods
+    ) -> None:
         self.fspath = fspath
         self.pm = pm
         self.remove_mods = remove_mods
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str):
```
blueyed: Conflicts with https://github.com/pytest-dev/pytest/pull/6461.

src/_pytest/main.py
blueyed: It's still kind of an API change.
But I also think it is good to have it being set always.

src/_pytest/python.py
blueyed: Also conflicts with https://github.com/pytest-dev/pytest/pull/6461, where I've aligned that already.

src/_pytest/unittest.py
```diff
@@ -1,27 +1,37 @@
...
-def pytest_pycollect_makeitem(collector, name, obj):
+def pytest_pycollect_makeitem(
+    collector: PyCollector, name: str, obj
+) -> Optional["UnitTestCase"]:
     # has unittest been imported and is obj a subclass of its TestCase?
     try:
-        if not issubclass(obj, sys.modules["unittest"].TestCase):
-            return
+        ut = sys.modules["unittest"]
+        # Type ignored because ut is an opaque module.
```
blueyed:
```suggestion
        # Type ignored because it is an opaque module.
```

src/_pytest/unittest.py
```diff
@@ -1,27 +1,37 @@
...
+        if not issubclass(obj, ut.TestCase):  # type: ignore
```
blueyed: I'd still prefer to see the ignore codes here.

src/_pytest/unittest.py
```diff
@@ -58,7 +68,8 @@ def collect(self):
             runtest = getattr(self.obj, "runTest", None)
             if runtest is not None:
                 ut = sys.modules.get("twisted.trial.unittest", None)
-                if ut is None or runtest != ut.TestCase.runTest:
+                # Type ignored because ut is an opaque module.
```
blueyed:
```suggestion
                # Type ignored because it is an opaque module.
```

src/_pytest/main.py
```diff
@@ -391,22 +412,30 @@ class Session(nodes.FSCollector):
...
+        self.shouldfail = False  # type: Union[bool, str]
         self.trace = config.trace.root.get("collection")
         self._norecursepatterns = config.getini("norecursedirs")
         self.startdir = config.invocation_dir
         self._initialpaths = frozenset()  # type: FrozenSet[py.path.local]
 
         # Keep track of any collected nodes in here, so we don't duplicate fixtures
-        self._node_cache = {}  # type: Dict[str, List[Node]]
+        self._node_cache1 = {}  # type: Dict[py.path.local, Sequence[nodes.Collector]]
+        self._node_cache2 = (
```
nicoddemus: (We discussed this in another thread, but worth mentioning here)

I still think it is not a good approach to change/create attributes in bugfix releases, so this should go into `features`.

nicoddemus: Oh and this will also be a headache to merge with `features` later, because those names have changed there already. 😞

blueyed: @nicoddemus JFI: this is not really part of this PR (just includes it (https://github.com/pytest-dev/pytest/pull/6555) for now)
As for reviewing, you can look at the commits this is adding.

blueyed:
As mentioned earlier (conflicts), we should get https://github.com/pytest-dev/pytest/pull/6461 in first.

src/_pytest/doctest.py
bluetech: I just haven't looked into what it's supposed to be. Sometimes to not get stuck too much and make progress, I leave some unrelated parameters blank, to hopefully be filled in the future.

BTW, I think one of the good things to do is go over all of the hookspecs and type annotate them, and then backfill all of the implementations with that. But the types are not always clear...

src/_pytest/unittest.py
bluetech: This refers to the `ut` variable -- I'll add backticks to clarify.

src/_pytest/unittest.py
bluetech: I also think it's a good idea to add them. But let's maybe first add `show-error-codes` to the mypy config to make it "official" and show up by default in editor/CI/etc. WDYT?

bluetech:
Addressed (some) comments and rebased on features. This is still blocked.

src/_pytest/unittest.py
blueyed: > But let's maybe first add show-error-codes to the mypy config to make it "official" and show up by default in editor/CI/etc.

:+1: 
(FWIW I have it in my editor's config already, and I think we can use them without it in the default config already.  Anyway, should not hold this back in general, but the next time maybe.. :))

src/_pytest/doctest.py
```diff
@@ -108,13 +109,18 @@ def pytest_unconfigure():
...
     config = parent.config
     if path.ext == ".py":
         if config.option.doctestmodules and not _is_setup_py(config, path, parent):
-            return DoctestModule.from_parent(parent, fspath=path)
+            mod = DoctestModule.from_parent(parent, fspath=path)  # type: DoctestModule
+            return mod
     elif _is_doctest(config, path, parent):
-        return DoctestTextfile.from_parent(parent, fspath=path)
+        txt = DoctestTextfile.from_parent(parent, fspath=path)  # type: DoctestTextfile
+        return txt
```
Zac-HD: Instead of splitting these out so you can add a comment, annotate [the `from_parent` method](https://github.com/bluetech/pytest/blob/49f830d1337b4949ad3684322d78bf6de08376ee/src/_pytest/nodes.py#L148-L150)

src/_pytest/doctest.py
```diff
@@ -108,13 +109,18 @@ def pytest_unconfigure():
     RUNNER_CLASS = None
 
 
-def pytest_collect_file(path, parent):
+def pytest_collect_file(
+    path: py.path.local, parent
+) -> Optional[Union["DoctestModule", "DoctestTextfile"]]:
```
Zac-HD: Instead of `Optional[Union[...]]`, `Union[..., None]`.

src/_pytest/main.py
```diff
@@ -246,11 +262,11 @@ def wrap_session(config, doit):
     return session.exitstatus
 
 
-def pytest_cmdline_main(config):
+def pytest_cmdline_main(config: Config) -> Union[int, ExitCode]:
     return wrap_session(config, _main)
 
 
-def _main(config, session):
+def _main(config: Config, session: "Session") -> Optional[Union[int, ExitCode]]:
```
Zac-HD: `Union[int, ExitCode, None]`

src/_pytest/main.py
```diff
@@ -418,10 +448,11 @@ def __init__(self, config: Config) -> None:
         self.config.pluginmanager.register(self, name="session")
 
     @classmethod
-    def from_config(cls, config):
-        return cls._create(config)
+    def from_config(cls, config: Config) -> "Session":
+        session = cls._create(config)  # type: Session
+        return session
```
Zac-HD: Better to annotate the `_create` method here than to override type inference.

src/_pytest/main.py
```diff
@@ -643,10 +713,11 @@ def _recurse(self, dirpath):
         return True
 
     @staticmethod
-    def _visit_filter(f):
-        return f.check(file=1)
+    def _visit_filter(f: py.path.local) -> bool:
+        # TODO: Remove type: ignore once `py` is typed.
+        return f.check(file=1)  # type: ignore
```
Zac-HD: Better would be to add the [`warn_unused_ignores`](https://mypy.readthedocs.io/en/stable/config_file.html#configuring-warnings) option to our mypy config, so that the tools remind us :smile:

src/_pytest/unittest.py
```diff
@@ -2,37 +2,48 @@
...
+        # Type ignored because `ut` is an opaque module.
+        if not issubclass(obj, ut.TestCase):  # type: ignore
+            return None
     except Exception:
-        return
+        return None
     # yes, so let's collect it
-    return UnitTestCase.from_parent(collector, name=name, obj=obj)
+    item = UnitTestCase.from_parent(collector, name=name, obj=obj)  # type: UnitTestCase
+    return item
```
Zac-HD: Again, I'd prefer to annotate the `from_parent` method (a different one!) so the type-checker can tell what's happening without the comment.

src/_pytest/main.py
blueyed: @Zac-HD why would you prefer this?  AFAIK we're using the `Optional[…]` "style" in general.  (I know that it is the same, just curious).

src/_pytest/main.py
blueyed: Taken into https://github.com/pytest-dev/pytest/pull/6461 (https://github.com/pytest-dev/pytest/pull/6461/commits/ae5d16be10c139da6a21eab34f7decbf93af721b).

src/_pytest/main.py
Zac-HD: Nesting them just feels wrong to me... if we're using that style though, that's that.

src/_pytest/main.py
blueyed: @Zac-HD 
I think the plan is to run this manually from time to time - with that option in the config it might vary between mypy versions etc.
But it makes sense to not use a "TODO" comment then, of course.  (and I prefer having the error code with the ignore btw (created https://github.com/pytest-dev/pytest/pull/6587)).

src/_pytest/hookspec.py
```diff
@@ -260,7 +273,9 @@ def pytest_pycollect_makemodule(path, parent):
 
 
 @hookspec(firstresult=True)
-def pytest_pycollect_makeitem(collector, name, obj):
+def pytest_pycollect_makeitem(
+    collector: "PyCollector", name: str, obj
+) -> "Union[None, Item, Collector, List[Union[Item, Collector]]]":
     """ return custom item/collector for a python object in a module, or None.
 
     Stops at first non-None result, see :ref:`firstresult` """
```
blueyed: Despite what the docs say I think this actually returns only Items, not Collectors, doesn't it?

blueyed: Oh, it might be `Class(PyCollector)` for example..  (which however only returns a single `Instance` via `collect()` then, but that's for something else).  So this is likely good then.

src/_pytest/doctest.py
bluetech: `from_parent` methods can't be properly typed as-is, there is discussion on this in #6387, for now I just work around it.

src/_pytest/main.py
bluetech: `_create()` is in a similar situation to `from_parent`, at least changing it would conflict with #6387, so I work around it too.

src/_pytest/main.py
bluetech: While `Optional` and `Union[None]` mean exactly the same to mypy, I tend to assign different semantic meaning to them. If `None` is "just another possibility" among others, use Union (for example, `JsonPrimitive = Union[None, bool, float, string]`). If it indicates absence/optionality, use `Optional`.

I'm am probably not entirely consistent with that though...

blueyed: @bluetech makes sense.
(btw: using `None` often might avoid having to import `Optional`, but that's not really an argument.)

src/_pytest/main.py
blueyed: https://github.com/pytest-dev/pytest/pull/6461 was merged.
Should be rebased now after master has been merged into features.

src/_pytest/main.py
blueyed: https://github.com/pytest-dev/pytest/pull/6387 has been merged, so that might unblock this a bit.

bluetech:
Rebased.

The Travis 3.5.1 job [failed](https://travis-ci.org/pytest-dev/pytest/jobs/647684603) with `ImportError: No module named 'zipp'`. I don't think it's related to this PR.

bluetech:
Rebased on `master` :grin:

bluetech:
Maintaining all of these open PRs is too annoying and spammy. I'm going to close them and maintain just a single `typing` branch (with clean separate commits) in my repo, and occasionally add commits and rebase it on master. Maybe one day it can be merged, or cherry-picked from.

PR #6558, opened 2020-01-24
Make EncodedFile.write() return the return value from inner write()

### LINKED ISSUE #6557: In tests, `sys.stderr.write` does not return the number of characters written
Within tests, `sys.stderr.write` does not return the number of characters written as `write` should.
This originates in `EncodedFile.write` which returns None, missing a return statement.

A minimal test (needs to be run as a test by pytest):

```
def test_stderr():
    assert sys.stderr.write("A") > 0
```

I am using pytest 5.2.1 but the code in master is the same.

gavento:
Stream write() methods should/may return the number of characters or bytes written. Encoded file needlessly discarded the value returned from the wrapped stream write.
Fixes #6557

### DIFF
--- a/changelog/6557.bugfix.rst
+++ b/changelog/6557.bugfix.rst
@@ -1,0 +1,1 @@
+Make capture output streams ``.write()`` method return the same return value from original streams.
--- a/testing/test_capture.py
+++ b/testing/test_capture.py
@@ -1492,3 +1492,8 @@ def test_typeerror_encodedfile_write(testdir):
     result_with_capture.stdout.fnmatch_lines(
         ["E * TypeError: write() argument must be str, not bytes"]
     )
+
+
+def test_stderr_write_returns_len(capsys):
+    """Write on Encoded files, namely captured stderr, should return number of characters written."""
+    assert sys.stderr.write("Foo") == 3
--- a/AUTHORS
+++ b/AUTHORS
@@ -256,6 +256,7 @@ Tim Hoffmann
 Tim Strazny
 Tom Dalton
 Tom Viner
+Tomáš Gavenčiak
 Tomer Keren
 Trevor Bekolay
 Tyler Goodlet
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -424,7 +424,7 @@ class EncodedFile:
             raise TypeError(
                 "write() argument must be str, not {}".format(type(obj).__name__)
             )
-        self.buffer.write(obj)
+        return self.buffer.write(obj)
 
     def writelines(self, linelist):
         data = "".join(linelist)


testing/test_capture.py
```diff
@@ -1492,3 +1492,8 @@ def test_fails():
     result_with_capture.stdout.fnmatch_lines(
         ["E * TypeError: write() argument must be str, not bytes"]
     )
+
+
+def test_stderr_write_returns_len():
```
nicoddemus:
```suggestion
def test_stderr_write_returns_len(capsys):
```

This will ensure the test is testing what we expect even if the test suite is running with `-s`

changelog/6557.bugfix.rst
```diff
@@ -0,0 +1 @@
+Make EncodedFile.write() return the return value from inner write().
```
nicoddemus:
```suggestion
Make capture output streams ``.write()`` method return the same return value from original streams.
```

blueyed:
Should also have a `return` with `writelines` probably: https://github.com/pytest-dev/pytest/blob/1ef137c1348a08b04eefa60a7baf341111f2cfea/src/_pytest/capture.py#L429-L431

testing/test_capture.py
gavento: Great point, thanks! Glad to learn something new :)

gavento:
> Should also have a `return` with `writelines` probably:
> 
> https://github.com/pytest-dev/pytest/blob/1ef137c1348a08b04eefa60a7baf341111f2cfea/src/_pytest/capture.py#L429-L431

Thanks for the pointer, however `.writelines` in CPython returns None, so I would leave it as is. See the code at https://github.com/python/cpython/blob/master/Lib/_pyio.py#L583

blueyed:
> Thanks for the pointer, however `.writelines` in CPython returns None

I'd still make it into a `return`, in case that would ever change etc.
I think it is good practice in general, and do not think you would need a test.
Feel free to leave it out though, I might do it in a separate PR then.

gavento:
@blueyed I had another suggestion I would make into a separate PR if you think it useful, that would resolve `writelines` more generally:
Make `EncodedFIle` inherit from `io.TextIOBase`, or `io.IOBase` directly.
* That would make it recognizable as a stream, inheriting from `IOBase`. (I was surprised by that recently).
* `io.IOBase` has the canonical `writelines` implementation, so any change would happen there.
* There are many more methods in `io.TextIOBase` but we do not need to define them (they raise unsupported by default). The only ones that would be good - and filelike objects are sort of expected to have regradless of `IOBase` inheritance - are `flush()`, `writable()`,  and `close()` + `closed`.

blueyed:
@gavento makes sense.  I've quickly done https://github.com/pytest-dev/pytest/pull/6566, which hopefully will fail (since it apparently does not inherit from IOBase yet).
Feel free to give it a try yourself.

blueyed:
@gavento just for info: it redirects attribute lookup to the original buffer already: https://github.com/pytest-dev/pytest/pull/6566/files#diff-e4269384a671a462a4824e8b06d4644bR441-R442

gavento:
> @gavento just for info: it redirects attribute lookup to the original buffer already: https://github.com/pytest-dev/pytest/pull/6566/files#diff-e4269384a671a462a4824e8b06d4644bR441-R442

Right, :man_facepalming: I noticed that elegant piece earlier. In that case, adding inheritance from `IOBase` would break that and we would need to manually reimplement the relevant methods as redirects. This could be done by some meta-programming or `__getattribute__` to save trivial code repetition, but that would make it even more obscure ;)

Other file-like objects already do not inherit `IOBase` .. but having that inheritance in pytest would be sort of nice.

PR #6566, opened 2020-01-25
Fix `EncodedFile.writelines`

blueyed:
This is implemented by IOBase already, which additionally checks if the
stream is not closed, and calls `write` per line.

Ref/via: https://github.com/pytest-dev/pytest/pull/6558#issuecomment-578210807

TODO:

- [x] test: could check if it calls through to `self._checkClosed()` (https://github.com/cpython/cpython/blob/58f525a216e651f1e17a18d31f190db727ed3de5/Lib/_pyio.py#L589) now
- [x] changelog

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -9,6 +9,7 @@ import os
 import sys
 from io import UnsupportedOperation
 from tempfile import TemporaryFile
+from typing import List
 
 import pytest
 from _pytest.compat import CaptureIO
@@ -426,9 +427,8 @@ class EncodedFile:
             )
         return self.buffer.write(obj)
 
-    def writelines(self, linelist):
-        data = "".join(linelist)
-        self.write(data)
+    def writelines(self, linelist: List[str]) -> None:
+        self.buffer.writelines([x.encode(self.encoding, "replace") for x in linelist])
 
     @property
     def name(self):
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -9,6 +9,7 @@ import os
 import sys
 from io import UnsupportedOperation
 from tempfile import TemporaryFile
+from typing import BinaryIO
 from typing import List
 
 import pytest
@@ -414,29 +415,27 @@ def safe_text_dupfile(f, mode, default_encoding="UTF8"):
 class EncodedFile:
     errors = "strict"  # possibly needed by py3 code (issue555)
 
-    def __init__(self, buffer, encoding):
+    def __init__(self, buffer: BinaryIO, encoding: str) -> None:
         self.buffer = buffer
         self.encoding = encoding
 
-    def write(self, obj):
-        if isinstance(obj, str):
-            obj = obj.encode(self.encoding, "replace")
-        else:
+    def write(self, obj: str) -> int:
+        if not isinstance(obj, str):
             raise TypeError(
                 "write() argument must be str, not {}".format(type(obj).__name__)
             )
-        return self.buffer.write(obj)
+        return self.buffer.write(obj.encode(self.encoding, "replace"))
 
     def writelines(self, linelist: List[str]) -> None:
         self.buffer.writelines([x.encode(self.encoding, "replace") for x in linelist])
 
     @property
-    def name(self):
+    def name(self) -> str:
         """Ensure that file.name is a string."""
         return repr(self.buffer)
 
     @property
-    def mode(self):
+    def mode(self) -> str:
         return self.buffer.mode.replace("b", "")
 
     def __getattr__(self, name):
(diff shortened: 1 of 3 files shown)


blueyed:
Oh.. it is not even covered.

I've found that it appears to be broken currently even, since it assumes a list of strs, while it should be bytes: https://docs.python.org/3/library/asyncio-stream.html?highlight=writelines#asyncio.StreamWriter.writelines

blueyed:
Rebased on master since it appears to be a bugfix.

blueyed:
/cc @gavento 
Does this look good to you?

blueyed:
@bluetech please check if the typing things are proper here.

src/_pytest/capture.py
```diff
@@ -413,30 +415,27 @@ def safe_text_dupfile(f, mode, default_encoding="UTF8"):
...
             raise TypeError(
                 "write() argument must be str, not {}".format(type(obj).__name__)
             )
-        return self.buffer.write(obj)
+        return self.buffer.write(obj.encode(self.encoding, "replace"))
 
-    def writelines(self, linelist):
-        data = "".join(linelist)
-        self.write(data)
+    def writelines(self, linelist: List[str]) -> None:
```
bluetech: typeshed uses Iterable here: https://github.com/python/typeshed/blob/d5851eca6f448b648e1a7b3f3010cc3dde498535/stdlib/3/typing.pyi#L490

Would also change the argument name to just `lines`.

src/_pytest/capture.py
```diff
@@ -413,30 +415,27 @@ def safe_text_dupfile(f, mode, default_encoding="UTF8"):
...
+        self.buffer.writelines([x.encode(self.encoding, "replace") for x in linelist])
```
bluetech: Any reason to use a list comprehension and not generator expression (omit the `[]`)?

testing/test_capture.py
```diff
@@ -831,7 +833,7 @@ def test_dontreadfrominput():
 
 
 @pytest.fixture
-def tmpfile(testdir):
+def tmpfile(testdir) -> Generator[BinaryIO, None, None]:
```
bluetech: Generally better to use `Iterator[X]` over `Generator[X, None, None]` - shorter and stricter.

Hmm but now that I think about it, `pytest.fixture` actually would want a real generator, I think? If so, then yeah, better keep it.

src/_pytest/capture.py
```diff
@@ -413,30 +415,27 @@ def safe_text_dupfile(f, mode, default_encoding="UTF8"):
 class EncodedFile:
     errors = "strict"  # possibly needed by py3 code (issue555)
```
bluetech: A bit strange that it says `strict` here but actually goes to use `replace`.

src/_pytest/capture.py
```diff
@@ -413,30 +415,27 @@ def safe_text_dupfile(f, mode, default_encoding="UTF8"):
...
-    def __init__(self, buffer, encoding):
+    def __init__(self, buffer: BinaryIO, encoding: str) -> None:
         self.buffer = buffer
         self.encoding = encoding
 
-    def write(self, obj):
-        if isinstance(obj, str):
-            obj = obj.encode(self.encoding, "replace")
-        else:
+    def write(self, obj: str) -> int:
```
bluetech: Would change the argument name `obj` to `s` while at it, `obj` is a somewhat misleading name for it.

bluetech:
If we're looking for potential breakage, previously `EncodedFile` only required the buffer to implement `write`, now it also requires `writelines`. I didn't examine how `EncodedFile` is used, so this might be fine.

src/_pytest/capture.py
blueyed: I wanted to please the typing (which was `List`).

src/_pytest/capture.py
blueyed: Yeah, likely an oversight.  Ref to issue: https://bitbucket.org/pytest-dev/pytest/issues/555
Will create a followup issue.

blueyed:
> If we're looking for potential breakage, previously `EncodedFile` only required the buffer to implement `write`, now it also requires `writelines`. I didn't examine how `EncodedFile` is used, so this might be fine.

Will check this / add a test for this in the next iteration - could likely fall back to the old behavior of using `write` then.

blueyed:
> > If we're looking for potential breakage, previously `EncodedFile` only required the buffer to implement `write`, now it also requires `writelines`. I didn't examine how `EncodedFile` is used, so this might be fine.
> 
> Will check this / add a test for this in the next iteration - could likely fall back to the old behavior of using `write` then.

Should be safe, since we only use `EncodedFile` via `safe_text_dupfile` for the `tmpfile` in `FDCaptureBinary`.

blueyed:
Waiting for another approval / re-review.

nicoddemus:
Bummer, a "fixup!" commit bf5c76359cf9dce7a94989d7f9edd4e22e6ffa3a got merged.

PR #6567, opened 2020-01-25
ci: codecov: only use "comment: off"

blueyed:
The changes status is quite buggy, remove it for now.
This just uses "comment: off" then.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,1 @@
-coverage:
-  status:
-    project: true
-    patch: true
-    changes: true
-
 comment: off


bluetech:
This check seemingly always fails, looking selfishly at my own PRs, I would be to see it gone.

blueyed:
> This check seemingly always fails, looking selfishly at my own PRs, I would be happy to see it gone.

Not always, but often.  It seems like they have issues with finding the correct base to compare against.  (talked briefly with their support)

PR #6580, opened 2020-01-26
typing: Testdir.__init__

blueyed:


src/_pytest/pytester.py
```diff
@@ -534,13 +535,13 @@ class Testdir:
...
-    def __init__(self, request, tmpdir_factory):
+    def __init__(self, request: FixtureRequest, tmpdir_factory: TempdirFactory) -> None:
         self.request = request
-        self._mod_collections = WeakKeyDictionary()
+        self._mod_collections = WeakKeyDictionary()  # type: ignore[var-annotated] # noqa: F821
         name = request.function.__name__
         self.tmpdir = tmpdir_factory.mktemp(name, numbered=True)
         self.test_tmproot = tmpdir_factory.mktemp("tmp-" + name, numbered=True)
-        self.plugins = []
+        self.plugins = []  # type: ignore[var-annotated] # noqa: F821
```
Zac-HD: Mypy's [`allow_untyped_globals `](https://mypy.readthedocs.io/en/stable/config_file.html#miscellaneous-strictness-flags) option might be cleaner than adding these comments to each case?

blueyed: @Zac-HD yeah.  Usually we're typing them then also, but e.g. here I'd rather get a bit more typing than nothing.

@bluetech what do you think about using the option?

bluetech: I generally wouldn't want to turn off strictness flags, only add more :)

blueyed:
Updated typing.

src/_pytest/pytester.py
```diff
@@ -33,12 +33,18 @@
...
 from _pytest.pathlib import Path
+from _pytest.python import Module
 from _pytest.reports import TestReport
+from _pytest.tmpdir import TempdirFactory
 
 if TYPE_CHECKING:
     from typing import Type
 
+    _PluggyPlugin = object
+
```
blueyed: Also used in https://github.com/pytest-dev/pytest/pull/6449/files#r371502971 - where to put it?

src/_pytest/pytester.py
```diff
@@ -534,13 +539,15 @@ class Testdir:
...
         self.request = request
-        self._mod_collections = WeakKeyDictionary()
+        self._mod_collections = (
+            WeakKeyDictionary()
+        )  # type: WeakKeyDictionary[Module, List[Union[Item, Collector]]]
         name = request.function.__name__
         self.tmpdir = tmpdir_factory.mktemp(name, numbered=True)
         self.test_tmproot = tmpdir_factory.mktemp("tmp-" + name, numbered=True)
-        self.plugins = []
+        self.plugins = []  # type: Sequence[Union[str, _PluggyPlugin]]
```
bluetech: `Sequence` means it won't be possible to append to it, e.g.

```
testing/test_resultlog.py:        testdir.plugins.append("resultlog")
```

So it should be a `List`.

blueyed: Thanks!

Amended, also to cover it completely:
```diff
diff --git c/src/_pytest/pytester.py w/src/_pytest/pytester.py
index 6d1a60154..cfe1b9a6c 100644
--- c/src/_pytest/pytester.py
+++ w/src/_pytest/pytester.py
@@ -547,7 +547,7 @@ def __init__(self, request: FixtureRequest, tmpdir_factory: TempdirFactory) -> N
         name = request.function.__name__
         self.tmpdir = tmpdir_factory.mktemp(name, numbered=True)
         self.test_tmproot = tmpdir_factory.mktemp("tmp-" + name, numbered=True)
-        self.plugins = []  # type: Sequence[Union[str, _PluggyPlugin]]
+        self.plugins = []  # type: List[Union[str, _PluggyPlugin]]
         self._cwd_snapshot = CwdSnapshot()
         self._sys_path_snapshot = SysPathsSnapshot()
         self._sys_modules_snapshot = self.__take_sys_modules_snapshot()
@@ -1082,7 +1082,6 @@ def collect_by_name(
         :param modcol: a module collection node; see :py:meth:`getmodulecol`
 
         :param name: the name of the node to return
-
         """
         if modcol not in self._mod_collections:
             self._mod_collections[modcol] = list(modcol.collect())
diff --git c/testing/test_collection.py w/testing/test_collection.py
index 62de0b953..5073f675e 100644
--- c/testing/test_collection.py
+++ w/testing/test_collection.py
@@ -9,6 +9,7 @@
 from _pytest.main import _in_venv
 from _pytest.main import ExitCode
 from _pytest.main import Session
+from _pytest.pytester import Testdir
 
 
 class TestCollector:
@@ -18,7 +19,7 @@ def test_collect_versus_item(self):
         assert not issubclass(Collector, Item)
         assert not issubclass(Item, Collector)
 
-    def test_check_equality(self, testdir):
+    def test_check_equality(self, testdir: Testdir) -> None:
         modcol = testdir.getmodulecol(
             """
             def test_pass(): pass
@@ -40,12 +41,15 @@ def test_fail(): assert 0
         assert fn1 != fn3
 
         for fn in fn1, fn2, fn3:
-            assert fn != 3
+            assert isinstance(fn, pytest.Function)
+            assert fn != 3  # type: ignore[comparison-overlap]  # noqa: F821
             assert fn != modcol
-            assert fn != [1, 2, 3]
-            assert [1, 2, 3] != fn
+            assert fn != [1, 2, 3]  # type: ignore[comparison-overlap]  # noqa: F821
+            assert [1, 2, 3] != fn  # type: ignore[comparison-overlap]  # noqa: F821
             assert modcol != fn
 
+        assert testdir.collect_by_name(modcol, "doesnotexist") is None
+
     def test_getparent(self, testdir):
         modcol = testdir.getmodulecol(
             """
```

PR #6588, opened 2020-01-27
tests: enable test_expression_multiline

blueyed:
Added in pylib like that already (https://github.com/pytest-dev/py/commit/a8729403).


blueyed:
Appears to differ between Python versions, closing for now.

PR #6589, opened 2020-01-27
ci: GHA: run less jobs with coverage

blueyed:
This often might be causing for jobs to take longer than 10 minutes,
which is a timeout Codecov uses to wait for successful CI.

Also it is good in general to have CI finish faster, of course.


blueyed:
Thanks.. \o/

PR #6590, opened 2020-01-28
[WIP] typing: fix _pytest._code.source.getfslineno: returns str; assignment

blueyed:
Fixes:

> src/_pytest/_code/source.py:309: error: Incompatible types in
>   assignment (expression has type "Union[local, str]", variable has type
>   "Optional[local]")  [assignment]

### DIFF
--- a/src/_pytest/_code/source.py
+++ b/src/_pytest/_code/source.py
@@ -8,6 +8,7 @@ import warnings
 from bisect import bisect_right
 from types import CodeType
 from types import FrameType
+from typing import Any
 from typing import Iterator
 from typing import List
 from typing import Optional
@@ -282,7 +283,7 @@ def compile_(  # noqa: F811
     return s.compile(filename, mode, flags, _genframe=_genframe)
 
 
-def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int]:
+def getfslineno(obj: Any) -> Tuple[Optional[Union[str, py.path.local]], int]:
     """ Return source location (path, lineno) for the given object.
     If the source cannot be determined return ("", -1).
 
@@ -292,6 +293,7 @@ def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int
 
     try:
         code = Code(obj)
+        return code.path, code.firstlineno
     except TypeError:
         try:
             fn = inspect.getsourcefile(obj) or inspect.getfile(obj)
@@ -305,11 +307,7 @@ def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int
                 _, lineno = findsource(obj)
             except IOError:
                 pass
-    else:
-        fspath = code.path
-        lineno = code.firstlineno
-    assert isinstance(lineno, int)
-    return fspath, lineno
+        return fspath, lineno
 
 
 #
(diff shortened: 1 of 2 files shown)


src/_pytest/_code/source.py
```diff
@@ -305,11 +310,7 @@ def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int
                 _, lineno = findsource(obj)
             except IOError:
                 pass
```
blueyed: @bluetech 
We've discussed the `fspath = fn and py.path.local(fn) or None` line above recently.
I think it might actually never happen (and will try to write a test for it), but in general I think we should remove the `Optional` from the return type, which would simpify typing internally already.

src/_pytest/_code/source.py
```diff
@@ -282,7 +283,7 @@ def compile_(  # noqa: F811
     return s.compile(filename, mode, flags, _genframe=_genframe)
 
 
-def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int]:
+def getfslineno(obj: Any) -> Tuple[Optional[Union[str, py.path.local]], int]:
```
blueyed: @bluetech 
Not sure about `obj: Any` here.  I've tried `Callable[[Any], None]` inbetween, but that would e.g. not cover classes etc.
I think it makes sense to allow Any here explicitly, given that TypeErrors are handled etc anyway.

src/_pytest/_code/source.py
blueyed: This is relevant for https://github.com/pytest-dev/pytest/blob/0e777a6ef5ef28ac830116a16fbc047af202c86a/src/_pytest/compat.py#L310-L317

src/_pytest/_code/source.py
bluetech: Yes, looks like this really is supposed to handle everything. I'd make it `object` though, instead of `Any`. `object` means the code can assume *nothing* on the value, while `Any` means the code can assume *anything*.

src/_pytest/_code/source.py
```diff
@@ -292,6 +293,7 @@ def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int
 
     try:
         code = Code(obj)
+        return code.path, code.firstlineno
```
bluetech: I think it's better to keep this in the `else` branch of the `try`, to keep the `try` minimal (e.g. if `code.firstlineno` raises a `TypeError` for some reason, it should propagate, not be caught).

src/_pytest/_code/source.py
bluetech: It depends on whether `fn = inspect.getsourcefile(obj) or inspect.getfile(obj)` can end up being falsey. I didn't check.

src/_pytest/_code/source.py
blueyed: Then it complains with the `inspect.get{source,}file` calls though, which expects `"Union[Module, Type[Any], MethodType, FunctionType, TracebackType, FrameType, CodeType, Callable[..., Any]]"`.
Should we use this then maybe also, or `# type: ignore` it?

bluetech: Catching `TypeError` is basically a runtime `type: ignore`. So seems justifiable in this case. 

(Of course, eventually it will be good to get rid of this `except TypeError`, which indicates some type confusion)

blueyed:
Moved into https://github.com/pytest-dev/pytest/pull/6656.

PR #6592, opened 2020-01-28
tox/GHA: mypy with py typing stubs

blueyed:


blueyed:
@bluetech 
Do you think it makes sense?  Otherwise let's close it.

bluetech:
So it passes cleanly now? Nice!

> Do you think it makes sense? Otherwise let's close it.

I personally don't like using unofficial trees, but that's just me. If the py stubs will eventually be merged, I think we should just wait for that. If not, then we should consider alternatives, like maybe bringing `py`'s stubs into the `pytest` repo.

But as you already know, I don't see a reason why it *shouldn't* be merged into `py`.

blueyed:
Yes, it passes.. \o/
Ok, let's close it here - I'm using it myself (so should get aware of regressions).

PR #6595, opened 2020-01-28
typing: fix Code.path

blueyed:
Fixes:

> src/_pytest/_code/code.py:83: error: Incompatible types in assignment
>   (expression has type "str", variable has type "local")  [assignment]


blueyed:
@RonnyPfannschmidt FYI: I prefer to merge my own (ACKed) PRs myself.  So instead of merging you should / could have just approved it.

PR #6598, opened 2020-01-28
Fix crash when faulthandler starts initialized

### LINKED ISSUE #6575: AttributeError: 'Config' object has no attribute 'fault_handler_stderr' when enabling faulthandler via Python
When launching pytest 5.3.2 with `python3 -X faulthandler -m pytest` (or `PYTHONFAULTHANDLER=1`, or enabling [Development Mode](https://docs.python.org/dev/library/devmode.html)) and with a `faulthandler_timeout` set, pytest fails with this after collecting:

```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/_pytest/main.py", line 196, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/_pytest/main.py", line 246, in _main
INTERNALERROR>     config.hook.pytest_runtestloop(session=session)
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/hooks.py", line 286, in __call__
INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/manager.py", line 93, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/manager.py", line 87, in <lambda>
INTERNALERROR>     firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/callers.py", line 208, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/callers.py", line 187, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/_pytest/main.py", line 270, in pytest_runtestloop
INTERNALERROR>     item.config.hook.pytest_runtest_protocol(item=item, nextitem=nextitem)
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/hooks.py", line 286, in __call__
INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/manager.py", line 93, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/manager.py", line 87, in <lambda>
INTERNALERROR>     firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/callers.py", line 208, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/pluggy/callers.py", line 182, in _multicall
INTERNALERROR>     next(gen)  # first yield
INTERNALERROR>   File "/home/florian/proj/qutebrowser/git/.tox/py37-pyqt513/lib/python3.7/site-packages/_pytest/faulthandler.py", line 60, in pytest_runtest_protocol
INTERNALERROR>     stderr = item.config.fault_handler_stderr
INTERNALERROR> AttributeError: 'Config' object has no attribute 'fault_handler_stderr'
```

As a workaround, `-p no:faulthandler` can be used.

It looks like `pytest_configure` in `faulthandler.py` exits early if it's already enabled:

https://github.com/pytest-dev/pytest/blob/6a26ac41255a304eb3eec2da565101caf2112394/src/_pytest/faulthandler.py#L17-L26

Thus it doesn't set the `config.fault_handler_stderr` attribute, but later `pytest_runtest_protocol` assumes that it exists:

https://github.com/pytest-dev/pytest/blob/6a26ac41255a304eb3eec2da565101caf2112394/src/_pytest/faulthandler.py#L54-L60

nicoddemus:
Fix #6575

### DIFF
--- a/src/_pytest/faulthandler.py
+++ b/src/_pytest/faulthandler.py
@@ -17,70 +17,92 @@ def pytest_addoption(parser):
 def pytest_configure(config):
     import faulthandler
 
-    # avoid trying to dup sys.stderr if faulthandler is already enabled
-    if faulthandler.is_enabled():
-        return
+    if not faulthandler.is_enabled():
+        # faulthhandler is not enabled, so install plugin that does the actual work
+        # of enabling faulthandler before each test executes.
+        config.pluginmanager.register(FaultHandlerHooks(), "faulthandler-hooks")
+    else:
+        from _pytest.warnings import _issue_warning_captured
 
-    stderr_fd_copy = os.dup(_get_stderr_fileno())
-    config.fault_handler_stderr = os.fdopen(stderr_fd_copy, "w")
-    faulthandler.enable(file=config.fault_handler_stderr)
+        # Do not handle dumping to stderr if faulthandler is already enabled, so warn
+        # users that the option is being ignored.
+        timeout = FaultHandlerHooks.get_timeout_config_value(config)
+        if timeout > 0:
+            _issue_warning_captured(
+                pytest.PytestConfigWarning(
+                    "faulthandler module enabled before pytest configuration step, "
+                    "'faulthandler_timeout' option ignored"
+                ),
+                config.hook,
+                stacklevel=2,
+            )
 
 
-def _get_stderr_fileno():
-    try:
-        return sys.stderr.fileno()
-    except (AttributeError, io.UnsupportedOperation):
-        # python-xdist monkeypatches sys.stderr with an object that is not an actual file.
-        # https://docs.python.org/3/library/faulthandler.html#issue-with-file-descriptors
-        # This is potentially dangerous, but the best we can do.
-        return sys.__stderr__.fileno()
+class FaultHandlerHooks:
+    """Implements hooks that will actually install fault handler before tests execute,
+    as well as correctly handle pdb and internal errors."""
 
+    def pytest_configure(self, config):
+        import faulthandler
 
-def pytest_unconfigure(config):
-    import faulthandler
+        stderr_fd_copy = os.dup(self._get_stderr_fileno())
+        config.fault_handler_stderr = os.fdopen(stderr_fd_copy, "w")
+        faulthandler.enable(file=config.fault_handler_stderr)
 
-    faulthandler.disable()
-    # close our dup file installed during pytest_configure
-    f = getattr(config, "fault_handler_stderr", None)
-    if f is not None:
+    def pytest_unconfigure(self, config):
+        import faulthandler
+
+        faulthandler.disable()
+        # close our dup file installed during pytest_configure
         # re-enable the faulthandler, attaching it to the default sys.stderr
         # so we can see crashes after pytest has finished, usually during
         # garbage collection during interpreter shutdown
         config.fault_handler_stderr.close()
         del config.fault_handler_stderr
-        faulthandler.enable(file=_get_stderr_fileno())
+        faulthandler.enable(file=self._get_stderr_fileno())
 
+    @staticmethod
+    def _get_stderr_fileno():
+        try:
+            return sys.stderr.fileno()
+        except (AttributeError, io.UnsupportedOperation):
+            # pytest-xdist monkeypatches sys.stderr with an object that is not an actual file.
+            # https://docs.python.org/3/library/faulthandler.html#issue-with-file-descriptors
+            # This is potentially dangerous, but the best we can do.
+            return sys.__stderr__.fileno()
 
-@pytest.hookimpl(hookwrapper=True)
-def pytest_runtest_protocol(item):
-    timeout = float(item.config.getini("faulthandler_timeout") or 0.0)
-    if timeout > 0:
+    @staticmethod
+    def get_timeout_config_value(config):
+        return float(config.getini("faulthandler_timeout") or 0.0)
+
+    @pytest.hookimpl(hookwrapper=True)
+    def pytest_runtest_protocol(self, item):
+        timeout = self.get_timeout_config_value(item.config)
+        stderr = item.config.fault_handler_stderr
+        if timeout > 0 and stderr is not None:
+            import faulthandler
+
+            faulthandler.dump_traceback_later(timeout, file=stderr)
+            try:
+                yield
+            finally:
+                faulthandler.cancel_dump_traceback_later()
+        else:
+            yield
+
+    @pytest.hookimpl(tryfirst=True)
+    def pytest_enter_pdb(self):
+        """Cancel any traceback dumping due to timeout before entering pdb.
+        """
         import faulthandler
 
-        stderr = item.config.fault_handler_stderr
-        faulthandler.dump_traceback_later(timeout, file=stderr)
-        try:
-            yield
-        finally:
-            faulthandler.cancel_dump_traceback_later()
-    else:
-        yield
+        faulthandler.cancel_dump_traceback_later()
 
+    @pytest.hookimpl(tryfirst=True)
+    def pytest_exception_interact(self):
+        """Cancel any traceback dumping due to an interactive exception being
+        raised.
+        """
+        import faulthandler
 
-@pytest.hookimpl(tryfirst=True)
-def pytest_enter_pdb():
-    """Cancel any traceback dumping due to timeout before entering pdb.
-    """
-    import faulthandler
-
-    faulthandler.cancel_dump_traceback_later()
-
-
-@pytest.hookimpl(tryfirst=True)
-def pytest_exception_interact():
-    """Cancel any traceback dumping due to an interactive exception being
-    raised.
-    """
-    import faulthandler
-
-    faulthandler.cancel_dump_traceback_later()
+        faulthandler.cancel_dump_traceback_later()
--- a/testing/test_faulthandler.py
+++ b/testing/test_faulthandler.py
@@ -88,7 +88,7 @@ def test_cancel_timeout_on_hook(monkeypatch, hook_name):
     exception (pytest-dev/pytest-faulthandler#14).
     """
     import faulthandler
-    from _pytest import faulthandler as plugin_module
+    from _pytest.faulthandler import FaultHandlerHooks
 
     called = []
 
@@ -98,6 +98,35 @@ def test_cancel_timeout_on_hook(monkeypatch, hook_name):
 
     # call our hook explicitly, we can trust that pytest will call the hook
     # for us at the appropriate moment
-    hook_func = getattr(plugin_module, hook_name)
-    hook_func()
+    hook_func = getattr(FaultHandlerHooks, hook_name)
+    hook_func(self=None)
     assert called == [1]
+
+
+@pytest.mark.parametrize("faulthandler_timeout", [0, 2])
+def test_already_initialized(faulthandler_timeout, testdir):
+    """Test for faulthandler being initialized earlier than pytest (#6575)"""
+    testdir.makepyfile(
+        """
+        def test():
+            import faulthandler
+            assert faulthandler.is_enabled()
+    """
+    )
+    result = testdir.run(
+        sys.executable,
+        "-X",
+        "faulthandler",
+        "-mpytest",
+        testdir.tmpdir,
+        "-o",
+        "faulthandler_timeout={}".format(faulthandler_timeout),
+    )
+    # ensure warning is emitted if faulthandler_timeout is configured
+    warning_line = "*faulthandler.py*faulthandler module enabled before*"
+    if faulthandler_timeout > 0:
+        result.stdout.fnmatch_lines(warning_line)
+    else:
+        result.stdout.no_fnmatch_line(warning_line)
+    result.stdout.fnmatch_lines("*1 passed*")
+    assert result.ret == 0
--- a/changelog/6575.bugfix.rst
+++ b/changelog/6575.bugfix.rst
@@ -1,0 +1,3 @@
+Fix internal crash when ``faulthandler`` starts initialized
+(for example with ``PYTHONFAULTHANDLER=1`` environment variable set) and ``faulthandler_timeout`` defined
+in the configuration file.


blueyed:
The test could be hardened maybe, but is not really necessary:
```diff
diff --git i/testing/test_faulthandler.py w/testing/test_faulthandler.py
index 617fe9b21..0f040ebbe 100644
--- i/testing/test_faulthandler.py
+++ w/testing/test_faulthandler.py
@@ -105,7 +105,11 @@ def test_cancel_timeout_on_hook(monkeypatch, hook_name):

 def test_already_initialized(testdir):
     """AttributerError when faulthandler is initialized earlier than pytest (#6575)"""
-    testdir.makepyfile("def test(): pass")  # need 1 test to trigger the error
+    testdir.makepyfile("""
+        def test():
+            import faulthandler
+            assert faulthandler.is_enabled()
+    """)
     result = testdir.run(
         sys.executable,
         "-X",
```

The-Compiler:
Hmm, right now that means `faulthandler_timeout` will just be silently ignored in that case. Maybe there should be a warning or something?

blueyed:
It checks for the faulthandler to be enabled in `pytest_configure` already, and likely could have a warning there then, and also registering a subplugin then (if not already enabled) - basically behaving as with `-p no:faulthandler` then, but with the additional warning.

nicoddemus:
Emitting a warning and using a subplugin to only have the hooks active when they should.

testing/test_faulthandler.py
```diff
@@ -98,6 +98,33 @@ def test_cancel_timeout_on_hook(monkeypatch, hook_name):
...
     # for us at the appropriate moment
-    hook_func = getattr(plugin_module, hook_name)
-    hook_func()
+    hook_func = getattr(FaultHandlerHooks, hook_name)
+    hook_func(self=None)
     assert called == [1]
+
+
+def test_already_initialized(testdir):
+    """AttributerError when faulthandler is initialized earlier than pytest (#6575)"""
```
blueyed:
```suggestion
    """Test for faulthandler being initialized earlier than pytest (#6575)"""
```
Seen a typo, but think it makes sense to rephrase.

src/_pytest/faulthandler.py
```diff
@@ -17,70 +17,92 @@ def pytest_addoption(parser):
...
-        # This is potentially dangerous, but the best we can do.
-        return sys.__stderr__.fileno()
+    if not faulthandler.is_enabled():
+        # faulthhandler is not enabled, so install plugin that does the actual work
+        # of enabling faulthandler before each test executes
+        config.pluginmanager.register(FaultHandlerHooks(), "faulthandler-hooks")
+    else:
+        from _pytest.warnings import _issue_warning_captured
+
+        # do not handle dumping to stderr if faulthandler is already enabled, so warn
```
blueyed:
```suggestion
        # Do not handle dumping to stderr if faulthandler is already enabled, so warn
```

src/_pytest/faulthandler.py
```diff
@@ -17,70 +17,92 @@ def pytest_addoption(parser):
...
+        # users that the option is being ignored
```
blueyed:
```suggestion
        # users that the option is being ignored.
```

src/_pytest/faulthandler.py
```diff
@@ -17,70 +17,92 @@ def pytest_addoption(parser):
...
-    try:
-        return sys.stderr.fileno()
-    except (AttributeError, io.UnsupportedOperation):
-        # python-xdist monkeypatches sys.stderr with an object that is not an actual file.
-        # https://docs.python.org/3/library/faulthandler.html#issue-with-file-descriptors
-        # This is potentially dangerous, but the best we can do.
-        return sys.__stderr__.fileno()
+    if not faulthandler.is_enabled():
+        # faulthhandler is not enabled, so install plugin that does the actual work
+        # of enabling faulthandler before each test executes
```
blueyed:
```suggestion
        # of enabling faulthandler before each test executes.
```

src/_pytest/faulthandler.py
```diff
@@ -17,70 +17,92 @@ def pytest_addoption(parser):
...
+        faulthandler.enable(file=self._get_stderr_fileno())
 
-        stderr = item.config.fault_handler_stderr
-        faulthandler.dump_traceback_later(timeout, file=stderr)
+    @staticmethod
+    def _get_stderr_fileno():
         try:
+            return sys.stderr.fileno()
+        except (AttributeError, io.UnsupportedOperation):
+            # python-xdist monkeypatches sys.stderr with an object that is not an actual file.
```
blueyed:
```suggestion
            # pytest-xdist monkeypatches sys.stderr with an object that is not an actual file.
```

src/_pytest/faulthandler.py
```diff
@@ -17,70 +17,92 @@ def pytest_addoption(parser):
...
+            # https://docs.python.org/3/library/faulthandler.html#issue-with-file-descriptors
+            # This is potentially dangerous, but the best we can do.
+            return sys.__stderr__.fileno()
```
blueyed: Note: this is not covered (but also before).
Wonder if pytest-xdist has been fixed in this regard?
(Anyway, out of scope for this PR to cover it.)

nicoddemus: `pytest-xdist` can't really fix it while using `execnet` as backend, I'm afraid.

This code is covered in pytest-xdist's own suite, this can be covered using some mocking perhaps?

blueyed: I've thought we should have seen it being covered since we're testing with xdist also?

nicoddemus: Hmm yeah that is strange.

OK I found the PR which changed that to a try/except:

https://github.com/pytest-dev/pytest-faulthandler/commit/26cd5308ed41be2f800b0c0504e2e46baeb987ce

> If you try to test a pytest plugin using the pytester module, and you
also have pytest-faulthandler installed, then currently it crashes
with obscure errors. The problem is that by default, pytester runs the
sub-pytest in-process, and the sub-pytest has sys.stderr set to a
_pytest.capture.CaptureIO object, which has a fileno method (so the
old hasattr(sys.stderr, "fileno") check passes), but if you try to
call this method then it raises io.UnsupportedOperation.

> So this commit robustifies the sys.stderr.fileno() calling code to
handle both a missing fileno (as before), and it raising
UnsupportedOperation.

So it seems this is not really required anymore. I will push a change to remove this workaround, let's see.

blueyed: > So it seems this is not really required anymore. I will push a change to remove this workaround, let's see.

Better do it separately (in another PR against features later).

I think it is good to keep it in general (since there might be other "bad citizens" in this regard, but covering/triggering it would be nice of course).

blueyed: But the commit message gives more info, and it might not be required then really anymore:

> this change is no longer required now that `CaptureIO` subclasses
`io.TextIOWrapper` directly and we don't support Python 2.

However I'd still be conservative and do it in a separate PR (against features then likely).

nicoddemus: Fair enough, let's wait for the CI to finish and then I will do a separate PR.

nicoddemus: Will need to wait for the next merge of `master` into `features`. Keeping the hash here for reference so I can do it later (fcbfd89c30103ec3e871dec0e0c749ac55f7f6f8).

testing/test_faulthandler.py
```diff
@@ -98,6 +98,33 @@ def test_cancel_timeout_on_hook(monkeypatch, hook_name):
...
+        "faulthandler",
+        "-mpytest",
+        testdir.tmpdir,
+        "-o",
+        "faulthandler_timeout=20",
+    )
+    result.stdout.fnmatch_lines(
+        [
+            "* PytestConfigWarning: faulthandler module enabled before pytest configuration*",
+            "*1 passed*",
```
blueyed: Note that the test fails with `-o faulthandler_timeout=0` - not sure if that is good/expected.

nicoddemus: Thanks for double checking, but it is expected: the test only "fails" because we don't produce any warning, but there's no error. 

I will update the test to cover that case though.

nicoddemus: Hardened the test. 👍

nicoddemus:
@The-Compiler @blueyed ready for another review.

testing/test_faulthandler.py
blueyed: Ok.  I wasn't sure however if specifying `timeout=0` should still enable it though.  But it seems like this is also a way to disable it.

blueyed:
Given that it is a subplugin (instance) now, it can be simplified:
```diff
 src/_pytest/faulthandler.py | 90 +++++++++++++++++++++++----------------------
 1 file changed, 46 insertions(+), 44 deletions(-)

diff --git i/src/_pytest/faulthandler.py w/src/_pytest/faulthandler.py
index ed2dfd025..855ffaa55 100644
--- i/src/_pytest/faulthandler.py
+++ w/src/_pytest/faulthandler.py
@@ -1,8 +1,10 @@
 import io
 import os
 import sys
+from typing import Generator
 
 import pytest
+from _pytest.config import Config
 
 
 def pytest_addoption(parser):
@@ -14,42 +16,52 @@ def pytest_addoption(parser):
     parser.addini("faulthandler_timeout", help, default=0.0)
 
 
-def pytest_configure(config):
+def pytest_configure(config: Config) -> None:
+    """
+    Install subplugin doing the actual work only with timeout, and when not
+    enabled already.
+    """
+    timeout = float(config.getini("faulthandler_timeout") or 0.0)
+    if timeout <= 0:
+        return
+
     import faulthandler
 
-    if not faulthandler.is_enabled():
-        # faulthhandler is not enabled, so install plugin that does the actual work
-        # of enabling faulthandler before each test executes.
-        config.pluginmanager.register(FaultHandlerHooks(), "faulthandler-hooks")
-    else:
+    if faulthandler.is_enabled():
         from _pytest.warnings import _issue_warning_captured
 
-        # Do not handle dumping to stderr if faulthandler is already enabled, so warn
-        # users that the option is being ignored.
-        timeout = FaultHandlerHooks.get_timeout_config_value(config)
-        if timeout > 0:
-            _issue_warning_captured(
-                pytest.PytestConfigWarning(
-                    "faulthandler module enabled before pytest configuration step, "
-                    "'faulthandler_timeout' option ignored"
-                ),
-                config.hook,
-                stacklevel=2,
-            )
+        _issue_warning_captured(
+            pytest.PytestConfigWarning(
+                "faulthandler module enabled before pytest configuration step, "
+                "'faulthandler_timeout' option ignored"
+            ),
+            config.hook,
+            stacklevel=2,
+        )
+        return
+
+    config.pluginmanager.register(
+        FaultHandlerHooks(timeout=timeout), "faulthandler-hooks"
+    )
 
 
 class FaultHandlerHooks:
-    """Implements hooks that will actually install fault handler before tests execute,
-    as well as correctly handle pdb and internal errors."""
+    """
+    Implements hooks that will actually install the fault handler before tests
+    execute, as well as correctly handling of pdb, and internal errors.
+    """
 
-    def pytest_configure(self, config):
+    def __init__(self, timeout: float) -> None:
+        self.timeout = timeout
+
+    def pytest_configure(self) -> None:
         import faulthandler
 
         stderr_fd_copy = os.dup(self._get_stderr_fileno())
-        config.fault_handler_stderr = os.fdopen(stderr_fd_copy, "w")
-        faulthandler.enable(file=config.fault_handler_stderr)
+        self.fault_handler_stderr = os.fdopen(stderr_fd_copy, "w")
+        faulthandler.enable(file=self.fault_handler_stderr)
 
-    def pytest_unconfigure(self, config):
+    def pytest_unconfigure(self) -> None:
         import faulthandler
 
         faulthandler.disable()
@@ -57,12 +69,11 @@ def pytest_unconfigure(self, config):
         # re-enable the faulthandler, attaching it to the default sys.stderr
         # so we can see crashes after pytest has finished, usually during
         # garbage collection during interpreter shutdown
-        config.fault_handler_stderr.close()
-        del config.fault_handler_stderr
+        self.fault_handler_stderr.close()
         faulthandler.enable(file=self._get_stderr_fileno())
 
     @staticmethod
-    def _get_stderr_fileno():
+    def _get_stderr_fileno() -> int:
         try:
             return sys.stderr.fileno()
         except (AttributeError, io.UnsupportedOperation):
@@ -71,27 +82,18 @@ def _get_stderr_fileno():
             # This is potentially dangerous, but the best we can do.
             return sys.__stderr__.fileno()
 
-    @staticmethod
-    def get_timeout_config_value(config):
-        return float(config.getini("faulthandler_timeout") or 0.0)
-
     @pytest.hookimpl(hookwrapper=True)
-    def pytest_runtest_protocol(self, item):
-        timeout = self.get_timeout_config_value(item.config)
-        stderr = item.config.fault_handler_stderr
-        if timeout > 0 and stderr is not None:
-            import faulthandler
+    def pytest_runtest_protocol(self) -> Generator[None, None, None]:
+        import faulthandler
 
-            faulthandler.dump_traceback_later(timeout, file=stderr)
-            try:
-                yield
-            finally:
-                faulthandler.cancel_dump_traceback_later()
-        else:
+        faulthandler.dump_traceback_later(self.timeout, file=self.fault_handler_stderr)
+        try:
             yield
+        finally:
+            faulthandler.cancel_dump_traceback_later()
 
     @pytest.hookimpl(tryfirst=True)
-    def pytest_enter_pdb(self):
+    def pytest_enter_pdb(self) -> None:
         """Cancel any traceback dumping due to timeout before entering pdb.
         """
         import faulthandler
@@ -99,7 +101,7 @@ def pytest_enter_pdb(self):
         faulthandler.cancel_dump_traceback_later()
 
     @pytest.hookimpl(tryfirst=True)
-    def pytest_exception_interact(self):
+    def pytest_exception_interact(self) -> None:
         """Cancel any traceback dumping due to an interactive exception being
         raised.
         """
```

nicoddemus:
@blueyed oh definitely, thanks. Indeed it makes sense now to just store things in `self` now.

Could you push that commit to this PR please?

blueyed:
> Could you push that commit to this PR please?

It's not a commit yet, but can do so later.  Please ping me if I forget - or just `patch -p1` it yourself.. ;)

nicoddemus:
Done, thanks!

nicoddemus:
@The-Compiler @blueyed ready for another review.

blueyed:
@nicoddemus so are a lot of mine.. ;)
As for this one I'd appreciate a `co-authored-by`, and therefore wanted to wait for a more objective review from somebody else.

nicoddemus:
> As for this one I'd appreciate a co-authored-by

Sure, done (let me know if anything is amiss).

src/_pytest/faulthandler.py
```diff
@@ -17,70 +17,92 @@ def pytest_addoption(parser):
...
+        faulthandler.enable(file=config.fault_handler_stderr)
 
-def pytest_unconfigure(config):
-    import faulthandler
+    def pytest_unconfigure(self, config):
+        import faulthandler
 
-    faulthandler.disable()
-    # close our dup file installed during pytest_configure
-    f = getattr(config, "fault_handler_stderr", None)
```
bluetech: (Haven't read the discussion, just something I noticed)

@nicoddemus I've heard a rumor that it's possible for `pytest_unconfigure` to be called without `pytest_configure` having been called:

https://github.com/pytest-dev/pytest/blob/b7ad4c2bed3906e1a12d9bb1829328b4045a72de/src/_pytest/debugging.py#L61-L62

Dunno if it's relevant here, but maybe that's why it was written in this defensive style. cc @blueyed e61e81a7b53496ea53f1f9958a2be5468190b8ef.

nicoddemus: Thanks for pointing that out @bluetech.

We definitely should fix pytest though so `pytest_unconfigure` is only called for `pytest_configure`, although this might need changes in `pluggy`: otherwise we make things very tricky for plugin authors.

Cc @goodboy

goodboy: Not sure I follow @nicoddemus.
The calling of each is determined by an internal `pytest` protocol no?

FD I've only briefed this PR.

blueyed: IIRC it might happen if a plugin raises UsageError, that then `pytest_unconfigure` gets called although `pytest_configure` was not.  (there is a test for this somewhere)
Yes, that behavior is related to pytest, not pluggy I assume.

nicoddemus: To exemplify:

Suppose we have plugins `A` and `B`. pytest will call `pytest_configure` and `pytest_unconfigure` using something like this:

```python
hook.pytest_configure(config=config)
```

Now suppose `A.pytest_configure` raises `UsageError`. From the call above currently has no way to know that `B.pytest_configure` never got called due to the error being raised.

Later when pytest tries to call:

```python
hook.pytest_unconfigure(config=config)
```

There's now way for pytest to know that it should not call `B.pytest_unconfigure` (because `B.pytest_configure` was never called).

So unless I'm missing something, we need some support in pluggy to be able to handle this in pytest?

PR #6599, opened 2020-01-28
terminal: fix progress report with duplicate nodeids

blueyed:
Fixes https://github.com/pytest-dev/pytest/issues/6597.

TODO:

- [x] after https://github.com/pytest-dev/pytest/pull/6409
- [x] test with tests having an error during "setup" already.
- [x] adjust test to reliably use duplicate nodeids?  (https://github.com/pytest-dev/pytest/issues/6597#issuecomment-579280434)

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -480,7 +480,11 @@ class TerminalReporter:
     def pytest_runtestloop(self) -> Generator[None, None, None]:
         """Write final progress indicator."""
         yield
-        if self.verbosity <= 0 and self._show_progress_info:
+        if (
+            getattr(self, "_tests_ran", False)
+            and self.verbosity <= 0
+            and self._show_progress_info
+        ):
             self._write_progress_information_filling_space()
 
     def _get_progress_information_message(self) -> str:
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -12,6 +12,7 @@ from functools import partial
 from typing import Any
 from typing import Callable
 from typing import Dict
+from typing import Generator
 from typing import List
 from typing import Mapping
 from typing import Optional
@@ -257,7 +258,7 @@ class TerminalReporter:
         self.reportchars = getreportopt(config)
         self.hasmarkup = self._tw.hasmarkup
         self.isatty = file.isatty()
-        self._numreported = 0
+        self._progress_items_reported = 0
         self._show_progress_info = self._determine_show_progress_info()
         self._collect_report_last_write = None  # type: Optional[float]
 
@@ -414,6 +415,8 @@ class TerminalReporter:
         else:
             markup = None
         self.stats.setdefault(category, []).append(rep)
+        if rep.when == "setup":
+            self._progress_items_reported += 1
         if not letter and not word:
             # probably passed setup/teardown
             return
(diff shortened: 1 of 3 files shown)


src/_pytest/terminal.py
```diff
@@ -437,7 +436,8 @@ def pytest_runtest_logreport(self, report: TestReport) -> None:
             else:
                 self._tw.write(letter, **markup)
         else:
-            self._progress_nodeids_reported.add(rep.nodeid)
+            if rep.when == "call":
+                self._numreported += 1
```
blueyed: Feels fragile.
TODO: test with tests having an error during "setup" already.

blueyed:
https://github.com/pytest-dev/pytest/pull/6599/commits/be6a7cdb903546f8f439a259adff69d5a07e9c4c changes a lot, but is actually a bit of a revert, see the commit message.
I think it's better to just have a count, and not to use a `set()` as a "workaround", and consider this to be ready - assuming tests pass.

src/_pytest/terminal.py
```diff
@@ -483,43 +484,51 @@ def pytest_runtest_logreport(self, report: TestReport) -> None:
 
     @property
     def _is_last_item(self):
-        return len(self._progress_nodeids_reported) == self._session.testscollected
+        return self._progress_items_reported == self._session.testscollected
 
-    def pytest_runtest_logfinish(self, nodeid):
```
bluetech: Is the change from `pytest_runtest_logfinish` to `pytest_runtest_teardown` + `pytest_runtestloop` part of the fix or just a refactor?

bluetech:
According to [this comment](https://github.com/pytest-dev/pytest/issues/6597#issuecomment-580246207) by @nicoddemus, duplicate nodeids are no longer possible. The change (set -> counter) might still make sense, but maybe the test doesn't?

blueyed:
@bluetech nothing prevents you from creating duplicate nodeids.  The test here uses a private property, but you could also yield `Item`s with duplicate nodeids.

The test is for a corner case, yes, but makes sense.

It's also unfortunate that the patch is so big, but only because it was not done properly/better before and it has to undo previous changes (IIRC).

src/_pytest/terminal.py
blueyed: Part of the fix IIRC.

bluetech:
Some lazy questions that would help me understand the diff -- feel free to dismiss.

Considering each change I can detect separately:

1. Changing when progress is indicated from `if verbosity > 0` to `if rep.when == "call" or (rep.when == "setup" and rep.failed)`. I understand why you removed it from the `verbosity` check - incrementing a counter doesn't cost anything so no need for it. But what are these new conditions?

2. Changing from set -> counter. This seems to make sense regardless, why use a heavy data structure when it's not needed? But the question is why was a set used in the first place? What it because it was *assumed* that no duplicate IDs are possible, or because it explicitly wanted to *handle* duplicate IDs by deduplicating them?

3. Changing from `pytest_runtest_logfinish` to `pytest_runtest_teardown`. IIUC these two should be mostly equivalent for this use case -- is it not so? If not, why? I possible I think this change should be avoided, mostly because the "log" hooks seem to exist particularly for this use case (reporting), and the code already uses `logstart` and `logreport` and it seems nice to use the matching `logfinish` hook.

4. Change from checking `_is_last_item` in `pytest_runtest_logfinish` hook to unconditionally running in `pytest_runtestloop` hookwrapper. This seems nice, but I'm not sure it's worth it as long as `_is_last_item` still exists. If you can get rid of `_is_last_item` entirely in favor the hook, that would be a nice cleanup.

src/_pytest/terminal.py
```diff
@@ -440,6 +440,8 @@ def pytest_runtest_logreport(self, report: TestReport) -> None:
         else:
             markup = None
         self._add_stats(category, [rep])
+        if rep.when == "call" or (rep.when == "setup" and rep.failed):
```
nicoddemus: Shouldn't this come after `if not letter and not word:`? Because "probably passed setup/teardown" should not be reported again.

src/_pytest/terminal.py
```diff
@@ -483,43 +484,51 @@ def pytest_runtest_logreport(self, report: TestReport) -> None:
...
 
-    def pytest_runtest_logfinish(self, nodeid):
-        assert self._session
-        if self.verbosity <= 0 and self._show_progress_info:
-            if self._show_progress_info == "count":
-                num_tests = self._session.testscollected
-                progress_length = len(" [{}/{}]".format(str(num_tests), str(num_tests)))
-            else:
-                progress_length = len(" [100%]")
+    def pytest_runtest_teardown(self) -> None:
```
nicoddemus: I'm curious, why using `pytest_runtest_teardown` instead of `pytest_runtest_logfinish`? On pytest-xdist I think `pytest_runtest_teardown` is called only on worker nodes (because they actually run the tests), not on `master` (which only does reporting). The `pytest_runtest_log*` run on both `master` and `workers`, by design.

src/_pytest/terminal.py
```diff
@@ -483,43 +484,51 @@ def pytest_runtest_logreport(self, report: TestReport) -> None:
...
+            msg = self._get_progress_information_message()
+            main_color, _ = self._get_main_color()
+            self._tw.write(msg + "\n", **{main_color: True})
+
+    @pytest.hookimpl(hookwrapper=True)
+    def pytest_runtestloop(self) -> Generator[None, None, None]:
+        """Write final progress indicator."""
+        yield
+        if (
+            getattr(self, "_tests_ran", False)
```
nicoddemus: I think we should avoid attributes being set but not declared on `__init__` when that is possible; this has been a great source of confusion/bugs in the past.

src/_pytest/terminal.py
nicoddemus: Hmmm strange, I commented out this function from your branch, and I still get the exact same output both with xdist/without, and `-v` and normal modes.

---

With the patch as is:

```
λ pytest testing\test_nodes.py
collected 11 items

testing\test_nodes.py ...........                              [100%]

======================== 11 passed in 0.10s =========================
```

Commenting out `pytest_runtest_teardown`:

```
 λ pytest testing\test_nodes.py
collected 11 items

testing\test_nodes.py ...........                              [100%]

======================== 11 passed in 0.10s =========================
```

---

With the patch as is:

```
λ pytest testing\test_nodes.py -n2
gw0 [11] / gw1 [11]
...........                                                    [100%]
======================== 11 passed in 0.93s =========================
```

Commenting out `pytest_runtest_teardown`:

```
λ pytest testing\test_nodes.py -n2
gw0 [11] / gw1 [11]
...........                                                    [100%]
======================== 11 passed in 0.92s =========================
```

---

So I guess `pytest_runtest_teardown` is not being used?

I got curious, and when in `master` and `pytest_runtest_logfinish` commented out, the `[%]` progress indicator is not printed.

nicoddemus:
> But the question is why was a set used in the first place? What it because it was assumed that no duplicate IDs are possible, or because it explicitly wanted to handle duplicate IDs by deduplicating them?

IIRC the former, because it is easier/more solid to just use the node ids if they are unique. But indeed, there's nothing preventing plugins to create items with duplicated ids, so using a `set` here is not really the right call.

src/_pytest/terminal.py
blueyed: > The pytest_runtest_log* run on both master and workers, by design.

That might result in too much being reported then (if both master and nodes report the same item(s)).
Changed it to use pytest_runtest_logfinish again, let's see if it passes.

blueyed:
@nicoddemus @bluetech 
Please squash-merge or close it.

blueyed:
(btw: nodeids were generated automatically not too long ago, which is actually something I've came up with somewhere already: 5e5935759)

PR #6601, opened 2020-01-28
typing: pytest_collection

blueyed:
Not included in https://github.com/pytest-dev/pytest/pull/6556, but conflicts (trivially).

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -1,5 +1,13 @@
 """ hook specifications for pytest plugins, invoked from main.py and builtin plugins.  """
+from typing import Any
+from typing import Optional
+
 from pluggy import HookspecMarker
+
+from _pytest.compat import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from _pytest.main import Session
 
 
 hookspec = HookspecMarker("pytest")
@@ -158,7 +166,7 @@ def pytest_load_initial_conftests(early_config, parser, args):
 
 
 @hookspec(firstresult=True)
-def pytest_collection(session):
+def pytest_collection(session: "Session") -> Optional[Any]:
     """Perform the collection protocol for the given session.
 
     Stops at first non-None result, see :ref:`firstresult`.
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -8,6 +8,8 @@ import sys
 from typing import Dict
 from typing import FrozenSet
 from typing import List
+from typing import Optional
+from typing import Union
 
 import attr
 import py
@@ -249,7 +251,7 @@ def pytest_cmdline_main(config):
     return wrap_session(config, _main)
 
 
-def _main(config, session):
+def _main(config: Config, session: "Session") -> Optional[Union[int, ExitCode]]:
     """ default command line protocol for initialization, session,
     running tests and reporting. """
     config.hook.pytest_collection(session=session)
@@ -259,6 +261,7 @@ def _main(config, session):
         return ExitCode.TESTS_FAILED
     elif session.testscollected == 0:
         return ExitCode.NO_TESTS_COLLECTED
+    return None
 
 
 def pytest_collection(session):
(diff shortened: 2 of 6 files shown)


src/_pytest/hookspec.py
```diff
@@ -158,7 +166,7 @@ def pytest_load_initial_conftests(early_config, parser, args):
 
 
 @hookspec(firstresult=True)
-def pytest_collection(session):
+def pytest_collection(session: "Session") -> Optional[Any]:
```
blueyed: `Optional[Any]`: the return values are not used, this is meant to indicate that.

src/_pytest/main.py
```diff
@@ -249,7 +253,11 @@ def pytest_cmdline_main(config):
     return wrap_session(config, _main)
 
 
-def _main(config, session):
+def _main(
+    config: Config, session: "Session"
+) -> Optional[
+    Union["Literal"[ExitCode.TESTS_FAILED], "Literal"[ExitCode.NO_TESTS_COLLECTED]]
```
bluetech: Until pyflakes has a new release, it'd be better to put the entire thing in quotes, then the `noqa` on the `Literal` import is not needed.

You can also combine the literals.

So: `"Union[Literal[ExitCode.TESTS_FAILED, ExitCode.NO_TESTS_COLLECTED]]"`.

bluetech: Are you sure `_main` will only ever return these two error codes? If it is put in the type annotation, then once they're published, it becomes an API guarantee.

blueyed: Without `Union` then even?  `Optional["Literal[ExitCode.TESTS_FAILED, ExitCode.NO_TESTS_COLLECTED]"]`

But `noqa` is still required.

blueyed: That's what is used in that function.  Also `_main` is not really public API, is it?

bluetech: Right. I think for the noqa to go away you need to wrap the entire type in quotes, including the `Optional`. Really not important though.

blueyed: Also did not work IIRC.
The problem here is that pyflakes does not look at the strings to begin with.
I think it is fine like it is now.

bluetech: It's what used in the function but it's better to keep our options open IMO.

IIUC, eventually this return value is funneled to the `main` function in `_pytest/config/__init__.py` which has return type `Union[int, _pytest.main.ExitCode]`. My suggestion is to use the same type here.

blueyed: Sorry, I fail to see why.  If a function returns only `1`, or `2`, it should/could have `Literal[1, 2]`, even when used from another function that returns `int`, no?

bluetech: This is really not very important but I'll try to explain my view.

I think restricting the type in such a way is not beneficial. The function is an (indirect) implementation of a hook and the only restriction that needs to apply to it is that it implements the hook specification.

Unless this is an actual API guarantee that should not be changed without it being a breaking change, the fact that it currently returns these two values is just an incidental implementation detail. If someone wants to say add a new error code, then they 1) might be unsure if they're allowed to change the type, and 2) need to "uselessly" adjust the type.

blueyed: I've really seen it more like being able to see from the signature already, what it might "raise".
But I do not have a strong opinion on it, and certainly trust you more with regard to type hinting.. :)
Amended.

bluetech: Thanks for indulging me :)

blueyed: @bluetech just for understanding: would you type a function always returning True also with `bool` then?  (https://github.com/blueyed/pytest/blob/70481429f8236141e3a39098082b254f4a1cc017/src/_pytest/python.py#L172)

My general understanding/idea with that was that we could later tell that a certain hook, when used, would always skip later ones.  This is somehow indicated with `-> bool` also (not `Optional[bool]`, but following your argument it should be typed like the hookspec, which is `Optional[…]`/`Optional[Any]` then for `firstresult=True`.

bluetech: I'd tend to use the same type as the hookspec. I see that fact that the impl currently skips (returns True) is a runtime property, which might change. But if it's really fundamental, as in we want to "enforce" that this particular impl always skips, then making it `-> True` does make sense.

PR #6604, opened 2020-01-28
tests: cover absolute path handling in _compute_fixture_value

blueyed:


testing/python/fixtures.py
```diff
@@ -3653,22 +3653,72 @@ def test_foo(request):
...
+                "*1 failed*",
+            ]
+        )
+
+        # With non-overlapping rootdir, passing testfile.
+        result = testdir.runpytest("--rootdir", rootdir, testfile)
+        result.stdout.fnmatch_lines(
+            [
+                "The requested fixture has no parameter defined for test:",
+                # XXX: missing filename in funcitem.nodeid here!
```
blueyed: This appears to be a generic issue with nodeid handling it seems - will pull this out here.

blueyed:
@nicoddemus 
> Please rebase/add a descriptive commit message.

Not sure what you mean.. do you mean it should be squashed?
I've done it in 2 commits, since the 2nd is a partial cherry-pick, and AFAICS the commit messages are fine?

nicoddemus:
> Not sure what you mean.. do you mean it should be squashed?

Oops sorry, yes.

> I've done it in 2 commits, since the 2nd is a partial cherry-pick, and AFAICS the commit messages are fine?

The commit messages don't really explain the context of the changes, for example the first one is just "source_path: py.path.local directly", which explains what it is doing, but not "why". I always try to think of the POV of someone who needs to look at the history a few years down the road and needs to figure out why things are they way they are.

But if they do look fine to you, please go ahead and merge. And thanks for the coverage improvement. 👍

blueyed:
"source_path: py.path.local directly" could be "Assign source_path via py.path.local directly", but should really be clear from the change itself (https://github.com/pytest-dev/pytest/pull/6604/commits/7c878742773d50e3bc04d1a5b0a786ce69562847).  The reference there is where it is taken from (in `features`), and keeping it separate helps with conflicts when rebasing/merging.
Thanks for the suggestion / review though.. :+1:

PR #6609, opened 2020-01-28
[WIP] getfslineno: skip trying Code

blueyed:
It goes through `inspect` for e.g. classes already anyway, and
simplifies the code.

Pulled out of https://github.com/pytest-dev/pytest/pull/6590.


blueyed:
Does not really work out / make sense for now.

PR #6610, opened 2020-01-29
tests: move test_getfslineno

blueyed:
It should be in `test_code` when testing `_pytest._code.getfslineno`,
not to be confused with `_pytest._code.source.getfslineno`.

Adds an extra assert (via https://github.com/pytest-dev/pytest/pull/6590).


blueyed:
Got it wrong.. followup: https://github.com/pytest-dev/pytest/pull/6623

Issue #6596, opened 2020-01-28
5.3.4 causing internal errors, as well


oseiberts11:
#6517 is closed already, but I want to add some information. There is some finger-pointing to something called "nose", but I see the same problem and I don't use "nose", whatever it is. So a proper fix in the future which only takes "nose" into account may not be perfect.

Fortunately the same workaround (pinning pytest to 5.3.3) works for me.

You can see my test code for instance at https://github.com/syseleven/rebootmgr/commit/cf0ec0b56a3ecdd40712e43ebb4b1c930852a241 in branch https://github.com/syseleven/rebootmgr/commits/oseibert-os-5267-use-enabled-instead-of-disabled . The next commit adds the workaround.

The weird thing is that all tests succeed, except those in the file `tests/test_lock.py`. As soon as the testing gets there, it fails. This is the case if you run the file on its own, or when you just run all tests (and then this file gets run as second file).

You can run the tests in their proper environment with the command `docker-compose run --rm integration_tests_py35` (or 36 or 37).

blueyed:
@oseiberts11 
Thanks for a way to reproduce it (have not tried it though).
Please post a traceback yourself.

blueyed:
@oseiberts11 `docker-compose run --rm integration_tests_py35` works for me on https://github.com/syseleven/rebootmgr/commit/cf0ec0b?
So to get more insight a traceback would really be useful.

Keep in mind though that the issue with #6517 is known, and might e.g. get triggered when paths/nodeids are rewritten (which might happen especially with Docker, but would require `-v` (verbose mode)).

oseiberts11:
The output is short enough that I can paste all of it. I forced the pytest-5.3.4 version here.
```
$ docker-compose run --rm integration_tests_py35
Starting rebootmgr_consul3_1 ... done
Starting rebootmgr_consul3_1 ...
Starting rebootmgr_consul1_1 ... done
Starting rebootmgr_consul4_1 ... done
Starting rebootmgr_consul2_1 ... done
GLOB sdist-make: /src/setup.py
py35 recreate: /src/.tox/py35
py35 installdeps: coverage, pytest==5.3.4, pytest-mock, -rrequirements.txt
py35 inst: /src/.tox/.tmp/package/1/rebootmgr-0.0.22.zip
[...]

PR #6614, opened 2020-01-29
Preparing release version 5.3.5

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/6517

Fix: #6596

### DIFF
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -462,7 +462,6 @@ class Item(Node):
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
-        assert isinstance(location[0], py.path.local), location[0]
         fspath = self.session._node_location_to_relpath(location[0])
         assert type(location[2]) is str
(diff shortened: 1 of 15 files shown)


src/_pytest/nodes.py
```diff
@@ -462,7 +462,6 @@ def reportinfo(self) -> Tuple[Union[py.path.local, str], Optional[int], str]:
     @cached_property
     def location(self) -> Tuple[str, Optional[int], str]:
         location = self.reportinfo()
-        assert isinstance(location[0], py.path.local), location[0]
```
blueyed: Feels like it could have been only this line's removal (+ test), but I guess it is OK/better to be safe, and does not really add diff churn for the future (since it is a release branch.. \o/)

blueyed:
I think this is good to go.  Since I might not be available for hours, please proceed with merging and creating the tag.  See https://github.com/pytest-dev/pytest/blob/8b123446fd53e771c9894efbee6b987c0fee17ec/HOWTORELEASE.rst.

nicoddemus:
Thanks!

PR #6622, opened 2020-01-29
typing: fix getfslineno (assignment)

blueyed:
Fixes:

> src/_pytest/_code/source.py:309: error: Incompatible types in
> assignment (expression has type "Union[local, str]", variable has type
> "Optional[local]")  [assignment]

Taken out of https://github.com/pytest-dev/pytest/pull/6590.


blueyed:
Will keep it in https://github.com/pytest-dev/pytest/pull/6590.

PR #6627, opened 2020-01-30
Reword fixture docs for clarity

DanielArndt:
I stumbled over the wording here a bit, and I thought this would make it read a little easier.

I'm not sure if I need a `changelog` entry for this. Obviously, there is no issue to attribute it to. Let me know if that is necessary.


DanielArndt:
I'm not a linguist, but alternatively "Not only can fixtures be used in test functions, but fixture functions can use other fixtures themselves".

PR #6636, opened 2020-01-30
PyCollector._genfunctions: use already created fixtureinfo

blueyed:
`Function` creates a `_fixtureinfo` already:
https://github.com/pytest-dev/pytest/blob/fed535694/src/_pytest/python.py#L1392-L1395

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -387,7 +387,7 @@ class PyCollector(PyobjMixin, nodes.Collector):
         fm = self.session._fixturemanager
 
         definition = FunctionDefinition(name=name, parent=self, callobj=funcobj)
-        fixtureinfo = fm.getfixtureinfo(definition, funcobj, cls)
+        fixtureinfo = definition._fixtureinfo
 
         metafunc = Metafunc(
             definition, fixtureinfo, self.config, cls=cls, module=module


blueyed:
NOTE: I've not looked too closely at this, just noticed it via pdb'ing that it got twice.

nicoddemus:
I had checked it before approval, but decided to do a double check:

They seem to be identical calls:

https://github.com/pytest-dev/pytest/blob/1dc265e34a591ddabecd33b6629626df301dbe53/src/_pytest/python.py#L1396-L1398

and:

https://github.com/pytest-dev/pytest/blob/1dc265e34a591ddabecd33b6629626df301dbe53/src/_pytest/python.py#L393

(`funcargs` is True by default)

`FuncFixtureInfo` is mostly a data class:

https://github.com/pytest-dev/pytest/blob/1dc265e34a591ddabecd33b6629626df301dbe53/src/_pytest/fixtures.py#L303-L339

and `prune_dependency_tree` is called here:

https://github.com/pytest-dev/pytest/blob/1dc265e34a591ddabecd33b6629626df301dbe53/src/_pytest/python.py#L415

So I think it is OK, althought the `prune_dependency_tree` code could be improved (I'm not sure I like the same object being changed over like this).

blueyed:
@nicoddemus 
Thanks for looking into / confirming it.
Let's wait for another approval though.

PR #6640, opened 2020-01-31
Merge master to features - one last time

bluetech:
Part of #6571 transition plan. I am doing this merge so that the *reverse* merge (coming up shortly) is clean.

### DIFF
--- a/doc/en/fixture.rst
+++ b/doc/en/fixture.rst
@@ -849,7 +849,7 @@ Running this test will *skip* the invocation of ``data_set`` with value ``2``:
 Modularity: using fixtures from a fixture function
 ----------------------------------------------------------
 
-You can not only use fixtures in test functions but fixture functions
+In addition to using fixtures in test functions, fixture functions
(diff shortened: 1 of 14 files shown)


bluetech:
Hmm I guess this needs to be approved, I didn't think of that :face_with_head_bandage:  There were only trivial conflicts. To proceed with the other steps, if anyone sees this, please approve :)

blueyed:
> Hmm I guess this needs to be approved, I didn't think of that  There were only trivial conflicts.

Typically good to add them with the merge commit message (i.e. uncomment what `git merge` gives you).

> To proceed with the other steps, if anyone sees this, please approve :)

I'm not sure if / how you want to proceed, but I'd like to at least you, @nicoddemus and @RonnyPfannschmidt to also approve that features should get merged into master as-is really (https://github.com/pytest-dev/pytest/pull/6571#issuecomment-580668348).

bluetech:
Thanks @blueyed, I think I will need to ping you for a few more steps...

PR #6642, opened 2020-01-31
Merge the features branch to master, before deleting features

bluetech:
Part of #6571 transition plan. This step merges `features` into `master`, after which the `features` branch will be deleted.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,8 +53,8 @@ repos:
     -   id: changelogs-rst
         name: changelog filenames
         language: fail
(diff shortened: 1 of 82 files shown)


bluetech:
Pending decision on @blueyed's concern in https://github.com/pytest-dev/pytest/pull/6571#issuecomment-580668348.

blueyed:
@bluetech I think this is not really necessary for the transition anyway, but only before the next release, isn't it?
We need to decide also still if "features" or another branch gets used then, with node-from-parent removed/replaced.
See https://github.com/pytest-dev/pytest/pull/6571#issuecomment-580761091.
Let's close this here for now.

Issue #6651, opened 2020-02-01
A commandline flag to override `pytrace` in failure


GPery:
I use pytest to automatically check my students' exercises.
They receive the output from `pytest.fail`, and I don't want them seeing the entire stack trace, including local paths, potential edge cases they're supposed to find themselves, et cetera.
Currently, all my tests use `pytest.fail(..., pytrace=False)`, but this is far from ideal, as I have many assistants writing tests and not all of them are familiar with this necessity.

I made a fork that adds the --no-trace flag, opening the issue to submit a pull request.

jbbqqf:
Hi! Triaging older bug reports — I'd like to nudge this one toward closure unless it's still reproducible on a current pytest.

- Filed **2020-02-01**, no activity since **2020-02-01** (6+ years; 0 prior comments).

If you can still reproduce on the latest pytest release with a minimal example (versions, exact invocation, the full output), I'm happy to dig further. Otherwise it might be a candidate to close.

(Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the activity dates and any source-path references above were verified directly against current `main`.)

PR #6652, opened 2020-02-01
Add --no-trace option, overrides pytrace=True for pytest.fail

GPery:
Hey, I just opened [this issue](https://github.com/pytest-dev/pytest/issues/6651) to open a PR for the fork I'll be using for my automated exercise testing.
Hoping this change can be useful to others as well :smile:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -110,6 +110,7 @@ Grig Gheorghiu
 Grigorii Eremeev (budulianin)
 Guido Wesdorp
 Guoqiang Zhang
+Guy Perry
 Harald Armin Massa
 Henk-Jaap Wagenaar
 Holger Kohr
--- a/changelog/6651.feature.rst
+++ b/changelog/6651.feature.rst
@@ -1,0 +1,2 @@
+Add the --no-trace flag, allowing to override `pytrace=True` for `pytest.fail` calls.
+This is useful to avoid automated `pytest` runs outputting unwanted information.
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -180,6 +180,13 @@ def pytest_addoption(parser):
         default=False,
         help="Don't ignore tests in a local virtualenv directory",
     )
+    group.addoption(
+        "--notrace",
+        "--no-trace",
+        action="store_true",
+        default=False,
+        help="Do not output any traceback in case of failure, even when passing pytrace=True",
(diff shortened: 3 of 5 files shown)


blueyed:
Another idea that comes to mind is enforcing `pytrace=False` in `OutcomeException` always when the option is enabled.

GPery:
That's where I started, but the session isn't available there in a clear manner.
The `_repr_failure_py` is a relatively nice place for it, but perhaps it should be organised a bit more nicely.

nicoddemus:
> Note that _repr_failure_py is bad already (with regard to typing) to return a str here in some cases - this should not get extended.

Not sure I agree... I don't think the method is being extended; it used to return `str` in some instances, and continues to return `str` in an additional condition... is that what you mean by "extended"?

I agree that `_repr_failure_py` is a bit messy, but this is not making it worse than already is.

blueyed:
> > Note that _repr_failure_py is bad already (with regard to typing) to return a str here in some cases - this should not get extended.
> 
> Not sure I agree... I don't think the method is being extended; it used to return `str` in some instances, and continues to return `str` in an additional condition...

I've looked into making it not return `str` anymore, so this would make that a bit more difficult.


@GPery 
> btw: have you tried `--tb=line`?

GPery:
> @GPery
> 
> > btw: have you tried `--tb=line`?

I have not.
This will do nicely, thank you!

nicoddemus:
So I think `--tb=line` is sufficient and we can close this @GPery?

GPery:
No, I was mistaken on the last comment, `--tb=line` still outputs the file path, which doesn't work for me. `--tb=none` also wouldn't do, because it doesn't output the failure message. I'll keep using the fork, but I'm not sure if my use case is common enough that anyone would find it necessary.

I would appreciate the merge, it's a minor change that'll make things easier for me (keeping up to date with upstream).

bluetech:
My opinion is that this need is not common enough to deserve a configuration option, so I propose we close this PR. The issue can remain open in case someone comes up with another suggestion for satisfying this need.

bluetech:
Now that #7100 is merged, we have a better avenue for resolving this issue, as mentioned here: https://github.com/pytest-dev/pytest/pull/7100#pullrequestreview-415533577, so I'm closing this PR.

PR #6653, opened 2020-02-01
pytester: LineMatcher: typing, docs, consecutive line matching

blueyed:


src/_pytest/pytester.py
```diff
@@ -413,8 +413,8 @@ class RunResult:
     def __init__(
         self,
         ret: Union[int, ExitCode],
-        outlines: Sequence[str],
-        errlines: Sequence[str],
+        outlines: List[str],
+        errlines: List[str],
```
blueyed: This has to be a list, since `pop` is used.

src/_pytest/pytester.py
```diff
@@ -1327,49 +1327,32 @@ class LineMatcher:
...
 
-    def __init__(self, lines):
+    def __init__(self, lines: List[str]) -> None:
         self.lines = lines
-        self._log_output = []
+        self._log_output = []  # type: List[str]
 
-    def str(self):
-        """Return the entire original text."""
-        return "\n".join(self.lines)
```
blueyed: This was moved to the end, otherwise `builtins.str` would be required for/with mypy.

PR #6656, opened 2020-02-02
internal: clean up getfslineno

blueyed:
Everything was using `_pytest.compat.getfslineno` basically, which
wrapped `_pytest._code.source.getfslineno`.

This moves the extra code from there into it directly, and uses the
latter everywhere.

This helps to eventually remove the one in compat eventually, and also
causes less cyclic imports.

### DIFF
--- a/src/_pytest/_code/source.py
+++ b/src/_pytest/_code/source.py
@@ -17,6 +17,7 @@ from typing import Union
 
 import py
 
+from _pytest.compat import get_real_func
 from _pytest.compat import overload
 from _pytest.compat import TYPE_CHECKING
 
@@ -289,6 +290,13 @@ def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int
     The line number is 0-based.
     """
     from .code import Code
+
+    # xxx let decorators etc specify a sane ordering
+    # NOTE: this used to be done in _pytest.compat.getfslineno, initially added
+    #       in 6ec13a2b9.  It ("place_as") appears to be something very custom.
+    obj = get_real_func(obj)
+    if hasattr(obj, "place_as"):
+        obj = obj.place_as
 
     try:
         code = Code(obj)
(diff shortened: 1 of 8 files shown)


src/_pytest/_code/source.py
```diff
@@ -282,14 +284,21 @@ def compile_(  # noqa: F811
     return s.compile(filename, mode, flags, _genframe=_genframe)
 
 
-def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int]:
+def getfslineno(obj: Any) -> Tuple[Union[str, py.path.local], int]:
```
blueyed: @bluetech had to take `Any` here instead of `obj`, for the `place_as` attribute check.

src/_pytest/_code/source.py
```diff
@@ -298,18 +307,16 @@ def getfslineno(obj) -> Tuple[Optional[Union["Literal['']", py.path.local]], int
         except TypeError:
             return "", -1
 
-        fspath = fn and py.path.local(fn) or None
+        fspath = fn and py.path.local(fn) or ""
```
blueyed: This cannot really happen (becoming `""`, only if `co_filename` is empty on a code object, but that gets handled by `Code` then, which I've fixed to keep it (instead of turning it into cwd).
But I guess the `or ""` does not hurt - at least I'm confident that it never returned `None` actually.

src/_pytest/compat.py
```diff
@@ -308,13 +307,10 @@ def get_real_method(obj, holder):
...
-    obj = get_real_func(obj)
-    if hasattr(obj, "place_as"):
-        obj = obj.place_as
-    fslineno = _pytest._code.getfslineno(obj)
-    assert isinstance(fslineno[1], int), obj
-    return fslineno
+    """(**Deprecated**, use _pytest._code.source.getfslineno directly)"""
+    import _pytest._code.source
+
+    return _pytest._code.source.getfslineno(obj)
```
nicoddemus: Can't we remove this right away? Everything in `compat` is internal, so we are free to remove this function now.

blueyed: Sure.. :)

PR #6658, opened 2020-02-02
Use code highlighting if pygments is installed

nicoddemus:


src/_pytest/_code/code.py
```diff
@@ -1038,20 +1038,27 @@ def __init__(
         self.style = style
 
     def toterminal(self, tw: TerminalWriter) -> None:
+        error_prefix = "E   "
+        src = "\n".join(x for x in self.lines if not x.startswith(error_prefix))
+        error_lines = (x for x in self.lines if x.startswith(error_prefix))
```
blueyed: Highlighting should be done in a single block, with the source only.
Otherwise it does not work with e.g.:
```python
def test():
    print("""
    """); assert 0
```
You might want to check the pdbpp source I've linked you to already, which is similar in that regard (since it removes prefixes added by pdb also).

In general I think this should be factored out into a separate (private) method (using `_highlight` theren then), and then just using `tw.write` with the result.  Makes it also easier to unit test then.

testing/test_terminal.py
```diff
@@ -944,10 +937,7 @@ def test_this():
                 "=*= FAILURES =*=",
                 "{red}{bold}_*_ test_this _*_{reset}",
                 "{bold}{red}test_color_yes.py{reset}:5: in test_this",
-                "{bold}    fail(){reset}",
                 "{bold}{red}test_color_yes.py{reset}:2: in fail",
-                "{bold}    assert 0{reset}",
-                "{bold}{red}E   assert 0{reset}",
                 "{red}=*= {red}{bold}1 failed{reset}{red} in *s{reset}{red} =*={reset}",
             ]
         ]
```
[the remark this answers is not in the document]
nicoddemus: Yeah but unfortunately the highlighted parts are huge and I suspect a bit of pain to maintain. Let me see.

src/_pytest/_io/__init__.py
```diff
@@ -1,3 +1,22 @@
...
+        if not self.hasmarkup:
+            return source
+        try:
+            from pygments.formatters.terminal import TerminalFormatter
+            from pygments.lexers.python import PythonLexer
+            from pygments import highlight
+        except ImportError:
+            return source
+        else:
+            return highlight(source, PythonLexer(), TerminalFormatter())
```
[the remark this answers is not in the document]
nicoddemus: Not at the moment; I think we can ship it in a straightforward way and perhaps consider adding some options to it in the future.

src/_pytest/_code/code.py
nicoddemus: > You might want to check the pdbpp source I've linked you to already

Sorry it seems I've missed it (both here and in `py`'s PR). Would you mind linking it again?

> Highlighting should be done in a single block, with the source only.

Isn't it what's being done here? Note that `src` is the whole source string.

Here's what your example looks to me:

image

src/_pytest/_io/__init__.py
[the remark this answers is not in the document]
nicoddemus: Btw, but I'm fine using other default here.

src/_pytest/_code/code.py
nicoddemus: > In general I think this should be factored out into a separate (private) method (using _highlight theren then)

Great idea! I just disagree a bit about this function also doing highlight: I like the idea of other parts of the terminal writing highlighted source code. I will push in a bit a WIP of what I mean.

nicoddemus: Actually I'm not so sure anymore of what I came up with; I wanted something to help extracting the "parsing" logic of the traceback. I'm just pushing to have the commit online.

src/_pytest/_io/__init__.py
```diff
@@ -19,4 +23,44 @@ def _highlight(self, source):
...
+
+
+@attr.s(frozen=True)
+class ParsedTraceback:
+    sources = attr.ib()  # type: Tuple[str, ...]
+    errors = attr.ib()  # type: Tuple[str, ...]
+    explanations = attr.ib()  # type: Tuple[str, ...]
+
+    ERROR_PREFIX = ">    "
+    EXPLANATION_PREFIX = "E    "
```
blueyed: Instead of parsing it out of Repr's again (`ReprEntry` here) it might make sense to extend those instead?

testing/test_terminal.py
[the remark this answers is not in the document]
blueyed: We could use a simpler/custom formatter during tests / for this test.
(Requires this to be configurable in general then though.)

src/_pytest/_io/__init__.py
nicoddemus: You mean adding this function/utility to `ReprEntry`? Or do you mean something else?

Another thing your comment made me consider was to "follow the source" (no pun intended), in which case I can see that the markers are generated in `FormattedExcinfo.get_source`. 

Ideally I would like for it to return actually structured data instead of formatted lines, so we can re-use it later without re-parsing, unfortunately this might be considered a breaking change. Will think about it some more (meanwhile suggestions are welcome).

src/_pytest/_code/code.py
blueyed: ">" is highlighted as a comment.

See https://github.com/pdbpp/pdbpp/blob/cb56e1b6126bb87597684d01af6f18c873be85ee/src/pdbpp.py#L1071-L1123 - not sure if it helps, since it does much more.

src/_pytest/_io/__init__.py
blueyed: Yes, structured data would be much better.
This would also help when it comes to highlighting different parts differently (e.g. for https://github.com/pytest-dev/pytest/issues/522).

src/_pytest/_code/code.py
nicoddemus: Thanks, will take a look later. :+

nicoddemus: Indeed that does much more than we need now it seems, thanks for the link thought.

testing/code/test_terminal_writer.py
```diff
@@ -0,0 +1,45 @@
+import re
+from io import StringIO
+
+import pytest
+from _pytest._io import TerminalWriter
+
+
+# TODO: move this and the other two related attributes from test_terminal.py into conftest as a
```
nicoddemus: Thinking of moving this to a `fixture` and reuse on tests appropriately.

src/_pytest/_io/__init__.py
```diff
@@ -1,3 +1,39 @@
...
+        new_lines = self._highlight(source).splitlines()
+        for indent, new_line in zip(indents, new_lines):
+            self.line(indent + new_line)
+
+    def _highlight(self, source):
+        """Highlight the given source code according to the "code_highlight" option"""
+        if not self.hasmarkup:
+            return source
+        try:
+            from pygments.formatters.terminal import TerminalFormatter
```
nicoddemus: Should we require `pygments` always? Doesn't seem to be that big of a dependency given the gains in readability.

nicoddemus:
Screenshot of a real failure I was dealing with just now:

image

src/_pytest/_io/__init__.py
RonnyPfannschmidt: I'd suggest to add as extra in 5.4, call for feedback and turn regular based on the feedback in 5.5 or 6

testing/code/test_terminal_writer.py
RonnyPfannschmidt: Should have a look at the constant declaration in colorama

nicoddemus: Done all, except that I could not find the exact equivalent of 'hl-reset' code using colorama
constants though.

src/_pytest/_io/__init__.py
nicoddemus: I did on my local copy, but after so thought I'm not sure it is a good idea: the extra would be named "pygments" or "coloring", and will then become part of our "API" because we will need to keep having that extra in there, even when it is no longer an extra, so users installing with `pip install pytest[pygments]` don't break. Also it won't save much typing.

Perhaps it is simple to just suggest users to install it manually, and let people comment if this should part of the regular dependencies on 5.5 or 6.0.

What do you say?

nicoddemus:
I think this is ready for another round of reviews. 👍

testing/code/test_terminal_writer.py
nicoddemus: Oh `colorama` is not a dependency on Linux, so I've moved this back to constants; not sure it is worth the extra dependency just because of this.

src/_pytest/_io/__init__.py
```diff
@@ -1,3 +1,39 @@
...
+            )
+        if not indents:
+            indents = [""] * len(lines)
+        source = "\n".join(lines)
+        new_lines = self._highlight(source).splitlines()
+        for indent, new_line in zip(indents, new_lines):
+            self.line(indent + new_line)
+
+    def _highlight(self, source):
+        """Highlight the given source code according to the "code_highlight" option"""
```
blueyed: There is no "code_highlight" option yet.
Please fix the docstring.

blueyed:
@nicoddemus this is what I came up during the merge conflict: https://github.com/blueyed/pytest/pull/142/commits/3c66c7d9de0822ec3b60daec608ab996c8965bd0#diff-eb0ddbc8c0fe892bd0b8cce2282bdc1aR1067-R1093
In general I've started to play around with making the terminal writer handle a pygments Token stream, which allows for e.g. LineMatcherFailed exceptions to be colored / tokenized.

PR #6660, opened 2020-02-02
Handle `Exit` exception in `pytest_sessionfinish`

blueyed:
Similar to a7268aa (https://github.com/pytest-dev/pytest/pull/6258).


blueyed:
> In the long run there should be `_wrap_session` likely with
> `wrap_session` wrapping it with that exception handling.

Not really worth it / trivial, since exceptions there are not re-raised, and the processing is a bit different.  Will amend the message when squash-merging (and you agree).

blueyed:
Amended already and added typing (pulled out of https://github.com/pytest-dev/pytest/pull/6556), since I've merged it already, after noticing that #6556 has conflicts anyway already.
/cc @bluetech

PR #6662, opened 2020-02-02
Explicitly state on the PR template that we can squash commits

nicoddemus:
This way we don't need to ask everytime, and users who for some reason
would not like us to squash their commits can uncheck the box then.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,6 +7,7 @@ Here is a quick checklist that should be present in PRs.
 - [ ] Target the `features` branch for new features, improvements, and removals/deprecations.
 - [ ] Include documentation when adding new features.
 - [ ] Include new tests or update existing tests when applicable.
+- [X] Allow maintainers to push and squash when merging my commits. Please uncheck this if you prefer to squash the commits yourself.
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,4 +20,7 @@ Unless your change is trivial or a small documentation fix (e.g., a typo or rewo
   Also make sure to end the sentence with a `.`.
 
 - [ ] Add yourself to `AUTHORS` in alphabetical order.
+
+If you leave the *Allow edits from maintainers* checked, the mainteiners might squash your commits before merging. If unchecked, maintainers
+will probably ask you to squash yourself then.
```
blueyed: This is confusing.
squash-merging is not the same as (force-)pushing to the branch ("allow edits from maintainers").

> might squash your commits before merging

This sounds like it would mean "squashed, force-pushed and then merged" (not squash-merged).

(Squash-merging is only an edit in the sense of maybe changing the commit message(, and having a single commit only of course.).)

So while a sentence like that is good, and the checkbox could be used as an indicator for that, it should be clarified.
I can think of it and make a suggestion in case my point does not come across clearly.. :+1: 

(also there's a typo at least: mainteiners)

nicoddemus: Changed, thanks. Feel free to make further suggestions directly. 👍

Issue #6657, opened 2020-02-02
Error location / crash entry missing from non-native tracebacks (due to generator / early setup?)


blueyed:
With the following:
```diff
diff --git i/src/_pytest/logging.py w/src/_pytest/logging.py
index df0da3daa..2e5d0cfc0 100644
--- i/src/_pytest/logging.py
+++ w/src/_pytest/logging.py
@@ -639,6 +639,7 @@ def _runtest_for_main(self, item, when):
             if self.print_logs:
                 # Add a captured log section to the report.
                 log = log_handler.stream.getvalue().strip()
+                assert 0, log
                 item.add_report_section(when, "log", log)

     @pytest.hookimpl(hookwrapper=True)
```
And running `pytest testing/test_capture.py -k test_capture_with_live_logging` you will get:
```
============== test session starts ===============
platform linux -- Python 3.8.1, pytest-5.3.3.dev50+gca8a1e47b, py-1.8.2.dev4+g9d3c8be7.d20200128, pluggy-0.13.2.dev5+g76838b1
…
collected 1 item

testing/test_capture.py EE                 [100%]

===================== ERRORS =====================
_ ERROR at setup of test_capture_with_live_logging[capfd] _
/usr/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
/usr/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
E   AssertionError
_ ERROR at teardown of test_capture_with_live_logging[capfd] _
/usr/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
/usr/lib/python3.8/contextlib.py:120: in __exit__
    next(self.gen)
E   AssertionError
============ short test summary info =============
ERROR testing/test_capture.py::test_capture_with_live_logging[capfd]
ERROR testing/test_capture.py::test_capture_with_live_logging[capfd]
=============== 2 errors in 0.10s ================
```
I.e. the location for the assertion error is wrong / not included.

With `--tb=native`:
```
===================== ERRORS =====================
_ ERROR at setup of test_capture_with_live_logging[capfd] _
Traceback (most recent call last):
  File "/usr/lib/python3.8/contextlib.py", line 120, in __exit__
    next(self.gen)
  File "…/Vcs/pytest/src/_pytest/logging.py", line 613, in _runtest_for
    yield
  File "/usr/lib/python3.8/contextlib.py", line 120, in __exit__
    next(self.gen)
[...]

PR #6664, opened 2020-02-03
Inline FunctionMixin with Function

blueyed:
`Generator` was removed in 7eb28f9eb, and this pleases mypy to correctly
complain that `FunctionMixin` has no `config` (within
`_prunetraceback`).

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -773,45 +773,6 @@ class Instance(PyCollector):
         return self.obj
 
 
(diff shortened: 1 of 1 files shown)


blueyed:
> codecov/patch — 96.66% of diff hit (target 96.8%)

Will not cover that here likely, but likely later when fixing `_prunetraceback` (likely related to https://github.com/pytest-dev/pytest/issues/6657).

PR #6665, opened 2020-02-03
[WIP] docs: use sphinx-autodoc-typehints, minor typing improvements

blueyed:
Ref: https://github.com/agronholm/sphinx-autodoc-typehints

### DIFF
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -326,27 +326,23 @@ class LogCaptureFixture:
             logger.setLevel(level)
 
     @property
-    def handler(self):
-        """
-        :rtype: LogCaptureHandler
-        """
-        return self._item.catch_log_handler
+    def handler(self) -> LogCaptureHandler:
+        return self._item.catch_log_handler  # type: ignore[no-any-return]  # noqa: F723
 
-    def get_records(self, when):
+    def get_records(self, when: str) -> List[logging.LogRecord]:
         """
         Get the logging records for one of the possible test phases.
 
-        :param str when:
+        :param when:
             Which test phase to obtain the records from. Valid values are: "setup", "call" and "teardown".
 
-        :rtype: List[logging.LogRecord]
         :return: the list of captured records at the given stage
 
         .. versionadded:: 3.4
         """
         handler = self._item.catch_log_handlers.get(when)
         if handler:
-            return handler.records
+            return handler.records  # type: ignore[no-any-return]  # noqa: F723
         else:
             return []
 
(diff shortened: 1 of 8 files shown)


src/_pytest/logging.py
```diff
@@ -613,7 +609,7 @@ def _runtest_for(self, item, when):
                 yield
 
     @contextmanager
-    def _runtest_for_main(self, item, when):
+    def _runtest_for_main(self, item, when: str) -> Generator[None, None, None]:
```
blueyed: @bluetech 
`item` not typed because of `catch_log_handler{,s}` being set/unset there.
Ignoring any refactors I've wondered if a `ItemWithLogging` type would make sense in general, which would have that attribute allowed then?  (I do not think it is worth it, was just wondering)

blueyed:
Appears to not being able to handle the circular imports:
```
WARNING: Cannot resolve forward reference in type annotations of "_pytest.outcomes.fail": name 'NoReturn' is not defined
…
WARNING: Cannot resolve forward reference in type annotations of "_pytest.nodes.Node": name 'Session' is not defined
```
(With and without `set_type_checking_flag`)

src/_pytest/logging.py
bluetech: Might work, but because the attribute is added and deleted, it can't be entirely correct, and would require casts as well, so probably doesn't make sense.

As for refactor, seems like a good case for an "item-level storage" mechanism https://github.com/pytest-dev/pytest/issues/3740. Such storage could be typed in some appropriate manner, probably.

bluetech: I *would* type this as `Item`, and add gitignores to the custom attribute get/sets. For the caller `Item` is the correct type, and the ignores are also correct insofar as static typing goes.

src/_pytest/mark/structures.py
```diff
@@ -147,8 +149,6 @@ class Mark:
     def combined_with(self, other):
         """
         :param other: the mark to combine with
-        :type other: Mark
```
bluetech: Did you intentionally not convert these to type annotations?

src/_pytest/mark/structures.py
```diff
@@ -249,13 +249,8 @@ def get_unpacked_marks(obj):
...
 
 
-def normalize_mark_list(mark_list):
-    """
-    normalizes marker decorating helpers to mark objects
-
-    :type mark_list: List[Union[Mark, Markdecorator]]
-    :rtype: List[Mark]
-    """
+def normalize_mark_list(mark_list: List[Union[Mark, MarkDecorator]]) -> List[Mark]:
```
bluetech: `mark_list` can be `Iterable` (if you feel like it)

blueyed:
@bluetech applied your suggestion in https://github.com/pytest-dev/pytest/pull/6699.

Issue #6289, opened 2019-11-29
enhance pytest.xfail documentation


graingert:
> Just out of curiosity: Why do I get to pytest.xfail() and pytest.fail() progammatically? 

`pytest.fail()` and `pytest.skip()` and `pytest.xfail()` are there for situations where you don't want to check for a condition at *module level*, which is when the condition is evaluated for marks. 

Consider:

```python
import slow_module_to_import
@pytest.mark.xfail(slow_module_to_import.very_slow_function(), reason="...")
def test_foo():
    ...
```

Vs:

```python
def test_foo():
    import slow_module_to_import
    if slow_module_to_import.very_slow_function():
        pytest.xfail(reason="...")
```

Performance is not the only reason too, sometimes your condition needs to access things which are not available at module import time (access a database, read a config file, etc.)

> Will that be deprecated in the future in favour of marking?

No plans at all.

_Originally posted by @nicoddemus in https://github.com/pytest-dev/pytest/issues/6288#issuecomment-559765982_

graingert:
I think we should include the:

```python
def test_foo():
    import slow_module_to_import
    if slow_module_to_import.very_slow_function():
        pytest.xfail(reason="...")
```

example in the pytest.xfail reference material

shankarj67:
Can I work on this issue?

nicoddemus:
Sure, thanks!

PR #6685, opened 2020-02-07
#6289: Add new example to XFAIL documentation

nathanielcompton:
Based on the discussions linked in [open issue #6289 ](https://github.com/pytest-dev/pytest/issues/6289#issuecomment-559775183), I added an additional example to the `xfail` documentation.

### DIFF
--- a/doc/en/skipping.rst
+++ b/doc/en/skipping.rst
@@ -234,7 +234,7 @@ expect a test to fail:
     def test_function():
         ...
 
-This test will run but no traceback will be reported when it fails. Instead terminal
+This test will run but no traceback will be reported when it fails. Instead, terminal
 reporting will list it in the "expected to fail" (``XFAIL``) or "unexpectedly
 passing" (``XPASS``) sections.
 
--- a/doc/en/skipping.rst
+++ b/doc/en/skipping.rst
@@ -234,11 +234,11 @@ expect a test to fail:
     def test_function():
         ...
 
-This test will be run but no traceback will be reported
-when it fails. Instead terminal reporting will list it in the
-"expected to fail" (``XFAIL``) or "unexpectedly passing" (``XPASS``) sections.
+This test will run but no traceback will be reported when it fails. Instead terminal
+reporting will list it in the "expected to fail" (``XFAIL``) or "unexpectedly
+passing" (``XPASS``) sections.
 
-Alternatively, you can also mark a test as ``XFAIL`` from within a test or setup function
+Alternatively, you can mark tests as ``XFAIL`` from within a test or setup function
 imperatively:
 
 .. code-block:: python
@@ -247,9 +247,20 @@ imperatively:
         if not valid_config():
             pytest.xfail("failing configuration (but should work)")
 
-This will unconditionally make ``test_function`` ``XFAIL``. Note that no other code is executed
-after ``pytest.xfail`` call, differently from the marker. That's because it is implemented
-internally by raising a known exception.
+.. code-block:: python
+
+    def test_function2():
+        import slow_module
+
+        if slow_module.slow_function():
+            pytest.xfail("slow_module taking too long")
+
+These two examples illustrate situations where you don't want to check for a condition
+at the module level, which is when a condition would otherwise be evaluated for marks.
+
+They will unconditionally make the two examples ``XFAIL``. Note that no other code is executed
+after a ``pytest.xfail`` call, which is different from a marker. This is because current
+implementation raises an internally known exception.
 
 **Reference**: :ref:`pytest.mark.xfail ref`
 
@@ -261,7 +272,7 @@ internally by raising a known exception.
 
 
 
-Both ``XFAIL`` and ``XPASS`` don't fail the test suite, unless the ``strict`` keyword-only
+Both ``XFAIL`` and ``XPASS`` don't fail the test suite unless the ``strict`` keyword-only
 parameter is passed as ``True``:
 
 .. code-block:: python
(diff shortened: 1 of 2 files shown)


nathanielcompton:
@graingert or @nicoddemus would one of you please help me understand the failing codecov test? I've tried to grok the details/output on the codecov page. Unless I'm mistaken, I believe I've only edited files that aren't even tracked by codecov. Is this correct?

bluetech:
@nathanielcompton Don't worry about this check, it's a little touch-and-go. Since you don't change any code, only docs, you should ignore it in this PR.

doc/en/skipping.rst
```diff
@@ -234,11 +234,11 @@ expect a test to fail:
...
 
-This test will be run but no traceback will be reported
-when it fails. Instead terminal reporting will list it in the
-"expected to fail" (``XFAIL``) or "unexpectedly passing" (``XPASS``) sections.
+This test will run but no traceback will be reported when it fails. Instead terminal
+reporting will list it in the "expected to fail" (``XFAIL``) or "unexpectedly
+passing" (``XPASS``) sections.
 
-Alternatively, you can also mark a test as ``XFAIL`` from within a test or setup function
+Alternatively, you can mark tests as ``XFAIL`` from within a test or setup function
```
bluetech: I think the previous phrasing was slightly better, because the new one might imply that you can skip test A from within test B.

I would phrase it like this though:

``
 Alternatively, you can also mark a test as ``XFAIL`` from within the test or its setup function
imperatively:
```

File: doc/en/skipping.rst [hunk 2]
```diff
@@ -247,9 +247,20 @@ imperatively:
...
+    def test_function2():
+        import slow_module
+
+        if slow_module.slow_function():
+            pytest.xfail("slow_module taking too long")
+
+These two examples illustrate situations where you don't want to check for a condition
+at the module level, which is when a condition would otherwise be evaluated for marks.
+
+They will unconditionally make the two examples ``XFAIL``. Note that no other code is executed
```
bluetech: I know this is from before, but I'm confused by the "unconditionally" here -- the `pytets.xfail` is literally under an `if`. Maybe I'm misunderstanding?

doc/en/skipping.rst
```diff
@@ -247,9 +247,20 @@ imperatively:
...
+after a ``pytest.xfail`` call, which is different from a marker. This is because current
```
bluetech: I'd leave the last sentence as-is. I don't think there are any plans to change this? Nor can I think of any other possible implementation. And also it would be a breaking change to change anyway.

doc/en/skipping.rst
```diff
@@ -261,7 +272,7 @@ internally by raising a known exception.
 
 
 
-Both ``XFAIL`` and ``XPASS`` don't fail the test suite, unless the ``strict`` keyword-only
+Both ``XFAIL`` and ``XPASS`` don't fail the test suite unless the ``strict`` keyword-only
```
bluetech:
```suggestion
Both ``XFAIL`` and ``XPASS`` don't fail the test suite by default. You can change this by setting the ``strict`` keyword-only parameter to ``True``:
```

nicoddemus:
Great review @bluetech.

LGTM with @bluetech's suggestions applied.

doc/en/skipping.rst
nathanielcompton: I was confused by it as well, but didn't want to remove without proper justification. I'll update with the "unconditionally" removed, and leave it to you for approval.

doc/en/skipping.rst
nathanielcompton: Updated PR based on feedback

doc/en/skipping.rst
nathanielcompton: Updated PR based on feedback

doc/en/skipping.rst
nathanielcompton: Updated PR based on feedback

doc/en/skipping.rst
nathanielcompton: Updated PR based on feedback

nathanielcompton:
@nicoddemus and @bluetech : Ready for re-review when you get time. Thanks for the good feedback.

doc/en/skipping.rst
```diff
@@ -234,11 +234,11 @@ expect a test to fail:
     def test_function():
         ...
 
-This test will be run but no traceback will be reported
-when it fails. Instead terminal reporting will list it in the
-"expected to fail" (``XFAIL``) or "unexpectedly passing" (``XPASS``) sections.
+This test will run but no traceback will be reported when it fails. Instead terminal
```
bluetech:
```suggestion
This test will run but no traceback will be reported when it fails. Instead, terminal
```

doc/en/skipping.rst
```diff
@@ -247,8 +247,19 @@ imperatively:
...
+        import slow_module
+
+        if slow_module.slow_function():
+            pytest.xfail("slow_module taking too long")
+
+These two examples illustrate situations where you don't want to check for a condition
+at the module level, which is when a condition would otherwise be evaluated for marks.
+
+This will make ``test_function`` ``XFAIL``. Note that no other code is executed after
+``pytest.xfail`` call, differently from the marker. That's because it is implemented
```
bluetech:
```suggestion
the ``pytest.xfail`` call, differently from the marker. That's because it is implemented
```

nicoddemus:
Thanks @bluetech and @nathanielcompton!

PR #6686, opened 2020-02-07
doc: highlight difference between progress percentage & code coverage

katrinleinweber:
[Some people](https://stackoverflow.com/questions/49738081/) and me as well were a bit confused about the `[...%]` indicators in stout. Since I didn't find a specific mention that it's not the code/test _coverage_, how about to add it here?

I hope my rst syntax is correct. GitHub at least rendered it OK, but I didn't re-build the docu locally. Can do that in the next few days if needed.

### DIFF
--- a/doc/en/getting-started.rst
+++ b/doc/en/getting-started.rst
@@ -77,6 +77,13 @@ This test returns a failure report because ``func(3)`` does not return ``5``.
 
     You can use the ``assert`` statement to verify test expectations. pytest’s `Advanced assertion introspection <http://docs.python.org/reference/simple_stmts.html#the-assert-statement>`_ will intelligently report intermediate values of the assert expression so you can avoid the many names `of JUnit legacy methods <http://docs.python.org/library/unittest.html#test-cases>`_.
 
+.. note::
+
+    The ``[100%]`` (or other right-aligned percentages on the same line
+    as a ``test_….py`` file) indicate the `progress of running all test cases
+    <https://blog.pytest.org/2017/whats-new-in-pytest-33/>`_, not the code coverage.
+    To obtain the latter, use `pytest-cov <https://pypi.org/project/pytest-cov/>`_.
+
 Run multiple tests
 ----------------------------------------------------------
 
--- a/doc/en/getting-started.rst
+++ b/doc/en/getting-started.rst
@@ -73,11 +73,9 @@ That’s it. You can now execute the test function:
 
 The ``[100%]`` refer to the overall progress of running all test cases. After it finishes, pytest then shows a failure report because ``func(3)`` does not return ``5``.
 
-
 .. note::
 
     You can use the ``assert`` statement to verify test expectations. pytest’s `Advanced assertion introspection <http://docs.python.org/reference/simple_stmts.html#the-assert-statement>`_ will intelligently report intermediate values of the assert expression so you can avoid the many names `of JUnit legacy methods <http://docs.python.org/library/unittest.html#test-cases>`_.
-
 
 Run multiple tests
 ----------------------------------------------------------
--- a/doc/en/getting-started.rst
+++ b/doc/en/getting-started.rst
@@ -71,18 +71,13 @@ That’s it. You can now execute the test function:
     test_sample.py:6: AssertionError
     ============================ 1 failed in 0.12s =============================
 
-This test returns a failure report because ``func(3)`` does not return ``5``.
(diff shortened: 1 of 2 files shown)


nicoddemus:
Thanks!

The-Compiler:
Seems like a good idea to mention this, but I'm not sure if the capturing documentation is the right place. Wouldn't something like the [getting started page](https://docs.pytest.org/en/latest/getting-started.html) make more sense?

nicoddemus:
Ahh good point, I read the text and was a bit too quick to approve. 😁

katrinleinweber:
OK, good point. Will do!

I was further thinking about adding a `TEST RUN PROGRESS` header to the output a la "Don't document the problem, fix it". Will try to implement that afterwards.

blueyed:
Nonetheless, doc improvements are very much appreciated, of course - but I thin we should get them right.. :)

katrinleinweber:
Respectfully, I disagree. The fact that no 1st-party explanation besides that blog post was easily found (and not [only by me](https://stackoverflow.com/questions/49738081/), prompted this PR. I'll not remove its essentials now. "Allow edits..." is activated, if you disagree.

At least for the sake of Google indexing it & pointing searchers to it, the note is IMHO good enough. They can click forward to either more info about test coverage or the progress percentage. Additionally, if merged, this would also be a 1st-party place to point future StackOverflow questions to. Neither outcome is truly "need"ed as you say, but IMHO probably useful.

Alternatively, we can of course change the wording (I'm open to suggestions), or with a larger undertaking [add a `PROGRESS` header to the output](https://github.com/pytest-dev/pytest/pull/6686#issuecomment-583402959) (not sure whether or how fast I can do that) and then scuttle the note entirely.

Figuratively speaking, as long as documentation needs to exist, it should IMHO not be a tree full of leafs, but a network of paths.

bluetech:
It seems clear from your description and the number of upvotes on the StackOverflow question that there is confusion here, so I think it's a good idea to clarify.

I agree with @blueyed's comment, though -- I think just saying what the percentage means is sufficient, adding the link to the blog and mentioning unrelated things (code coverage) might just confuse people more.

I also think it can be a regular sentence and not a note, which grabs a lot of attention.

nicoddemus:
> I agree with @blueyed's comment, though -- I think just saying what the percentage means is sufficient, adding the link to the blog and mentioning unrelated things (code coverage) might just confuse people more.

After reading the stack overflow question, indeed there is value in clarifying what the percentage means, but I'm also leaning towards just describing it like suggested by @blueyed.

doc/en/getting-started.rst
```diff
@@ -77,6 +77,13 @@ This test returns a failure report because ``func(3)`` does not return ``5``.
 
     You can use the ``assert`` statement to verify test expectations. pytest’s `Advanced assertion introspection <http://docs.python.org/reference/simple_stmts.html#the-assert-statement>`_ will intelligently report intermediate values of the assert expression so you can avoid the many names `of JUnit legacy methods <http://docs.python.org/library/unittest.html#test-cases>`_.
 
+.. note::
```
nicoddemus: Following the discussion, how about we replace line 74 by:

```
The ``[100%]`` refer to the overall progress of running all test cases. After it finishes, pytest then shows a failure report because ``func(3)`` does not return ``5``.
```

And remove the note? What do you think @katrinleinweber?

katrinleinweber: Thanks for the suggestion! LGTM

nicoddemus:
Awesome @katrinleinweber, thanks for that! We appreciate the interest and patience on following up with the PR. 👍 

Small tip: I suggest to avoid mentioning people in commit messages, because every time someone forks or clones pytest I would get a notification. But don't worry, we will remove the mention when squashing/merging.

Thanks again!

doc/en/getting-started.rst
```diff
@@ -71,7 +71,7 @@ That’s it. You can now execute the test function:
     test_sample.py:6: AssertionError
     ============================ 1 failed in 0.12s =============================
 
-This test returns a failure report because ``func(3)`` does not return ``5``.
+The ``[100%]`` refer to the overall progress of running all test cases. After it finishes, pytest then shows a failure report because ``func(3)`` does not return ``5``.
```
hugovk:
```suggestion
The ``[100%]`` refers to the overall progress of running all test cases. After it finishes, pytest then shows a failure report because ``func(3)`` does not return ``5``.
```

nicoddemus: Thanks!

PR #6692, opened 2020-02-08
ci: remove Travis

blueyed:
3.5.1 fails there currently and I do not feel like we should cover it
here anymore.

Ref: https://github.com/pytest-dev/pytest/pull/6691#issuecomment-583702514
Ref: https://github.com/pytest-dev/pytest/pull/6463.


blueyed:
Feel free to pick it up yourself.

PR #6698, opened 2020-02-09
doc for _capturing_for_request

blueyed:
Follow-up to https://github.com/pytest-dev/pytest/pull/6663.

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -152,6 +152,13 @@ class CaptureManager:
     def _capturing_for_request(
         self, request: FixtureRequest
     ) -> Generator["CaptureFixture", None, None]:
+        """
+        Context manager that creates a ``CaptureFixture`` instance for the
+        given ``request``, ensuring there is only a single one being requested
+        at the same time.
+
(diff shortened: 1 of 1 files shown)


src/_pytest/capture.py
```diff
@@ -256,6 +234,36 @@ def pytest_internalerror(self, excinfo):
...
 
+@contextlib.contextmanager
+def _capturing_for_request(
+    request: FixtureRequest,
+) -> Generator["CaptureFixture", None, None]:
+    """
+    Context manager that creates a ``CaptureFixture`` instance for the
+    given ``request``, ensuring there is only a single one being requested.
+
+    This is used has a helper with ``capsys``, ``capfd`` etc.
```
bluetech:
```suggestion
    This is used as a helper in ``capsys``, ``capfd`` etc.
```

blueyed:
> An argument in favor of keeping it inside the `CaptureManager` class is that it sets and accesses internal details of `CaptureManager`, and is basically its "heart", so and it's nice to have it encapsulated.

Makes sense.  The main motivation was to have less (duplicated) boilerplate code mainly.

PR #6699, opened 2020-02-09
Typing around/from types in docs

blueyed:
Follow-up to https://github.com/pytest-dev/pytest/pull/6665, without `sphinx_autodoc_typehints`.


src/_pytest/mark/structures.py
```diff
@@ -144,7 +147,7 @@ class Mark:
     #: keyword arguments of the mark decorator
     kwargs = attr.ib()  # Dict[str, object]
 
-    def combined_with(self, other):
+    def combined_with(self, other: "Mark"):
```
bluetech: Can also be `-> "Mark"` here.

PR #6702, opened 2020-02-10
_diff_text: use repr with escape characters

blueyed:
This helps especially with comparing strings containing terminal escape
sequences, but also newlines already.

Ref: #3443

TODO:

- [ ] 8de826cb1

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -235,7 +235,7 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
 
     if any(
         wcwidth(ch) <= 0
-        for ch in itertools.chain.from_iterable(x for x in left_lines + right_lines)
+        for ch in [ch for lines in left_lines + right_lines for ch in lines]
     ):
         left_lines = [repr(x) for x in left_lines]
         right_lines = [repr(x) for x in right_lines]
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -2,6 +2,7 @@
 import collections.abc
 import itertools
 import pprint
+import re
 from typing import AbstractSet
 from typing import Any
 from typing import Callable
@@ -222,14 +223,15 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
                 ]
                 left = left[:-i]
                 right = right[:-i]
-    keepends = True
     if left.isspace() or right.isspace():
         left = repr(str(left))
         right = repr(str(right))
         explanation += ["Strings contain only whitespace, escaping them using repr()"]
 
-    left_lines = left.splitlines(keepends)
-    right_lines = right.splitlines(keepends)
+    left_split = len(left) and re.split("(\r?\n)", left) or []
+    left_lines = left_split[::2]
+    right_split = len(right) and re.split("(\r?\n)", right) or []
+    right_lines = right_split[::2]
 
     if any(
         wcwidth(ch) <= 0
@@ -240,6 +242,33 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
         explanation += [
             "NOTE: Strings contain non-printable characters. Escaping them using repr()."
         ]
+    else:
+        max_lines = max(len(left_lines), len(right_lines))
+        left_ends = left_split[1:max_lines:2]
+        right_ends = right_split[1:max_lines:2]
+        if left_ends != right_ends:
+            explanation += [
+                "NOTE: Strings contain different line-endings. Escaping them using repr()."
+            ]
+            left_lines = [line for line in left_lines]
+            right_lines = [line for line in right_lines]
+
+            for idx, (left_line, right_line) in enumerate(
+                itertools.zip_longest(left_lines, right_lines, fillvalue="")
+            ):
+                try:
+                    left_end = left_ends[idx]
+                except IndexError:
+                    left_end = ""
+                try:
+                    right_end = right_ends[idx]
+                except IndexError:
+                    right_end = ""
+                if left_end != right_end:
+                    left_lines[idx] += repr(left_end)[1:-1]
+                    right_lines[idx] += repr(right_end)[1:-1]
+                if not left_end or not right_end:
+                    break
 
     explanation += [line.strip("\n") for line in ndiff(left_lines, right_lines)]
     return explanation
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -243,32 +243,23 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
             "NOTE: Strings contain non-printable characters. Escaping them using repr()."
         ]
     else:
-        max_lines = max(len(left_lines), len(right_lines))
-        left_ends = left_split[1:max_lines:2]
-        right_ends = right_split[1:max_lines:2]
+        max_split = max(len(left_split), len(right_split))
+        left_ends = left_split[1:max_split:2]
+        right_ends = right_split[1:max_split:2]
         if left_ends != right_ends:
             explanation += [
                 "NOTE: Strings contain different line-endings. Escaping them using repr()."
             ]
-            left_lines = [line for line in left_lines]
-            right_lines = [line for line in right_lines]
-
-            for idx, (left_line, right_line) in enumerate(
-                itertools.zip_longest(left_lines, right_lines, fillvalue="")
+            for idx, (left_line, right_line, left_end, right_end) in enumerate(
+                itertools.zip_longest(
+                    left_lines, right_lines, left_ends, right_ends, fillvalue=None
+                )
             ):
-                try:
-                    left_end = left_ends[idx]
-                except IndexError:
-                    left_end = ""
-                try:
-                    right_end = right_ends[idx]
-                except IndexError:
-                    right_end = ""
                 if left_end != right_end:
-                    left_lines[idx] += repr(left_end)[1:-1]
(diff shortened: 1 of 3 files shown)


src/_pytest/assertion/util.py
```diff
@@ -225,10 +226,18 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
...
         explanation += ["Strings contain only whitespace, escaping them using repr()"]
-    explanation += [
-        line.strip("\n")
-        for line in ndiff(left.splitlines(keepends), right.splitlines(keepends))
-    ]
+
+    left_lines = left.splitlines(keepends)
+    right_lines = right.splitlines(keepends)
+
+    if any(wcswidth(x) == -1 for x in left_lines + right_lines):
```
blueyed: This can/should be optimized.

src/_pytest/assertion/util.py
```diff
@@ -225,10 +226,18 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
...
+        left_lines = [repr(x) for x in left_lines]
+        right_lines = [repr(x) for x in right_lines]
+        explanation += [
+            "Strings contain non-printable/escape characters, escaping them using repr()"
```
bluetech:
```suggestion
            "Strings contain non-printable/escape characters. Escaping them using repr()."
```

Also, maybe add a `NOTE: ` prefix? IIRC pytest does this in similar notes.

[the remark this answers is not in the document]
blueyed: Terminal escape codes for example.  Suggestions welcome.
Ref: https://github.com/jquast/wcwidth/blob/c71459ea91af86f3bbcdac2c8ed5e7773da2d848/wcwidth/wcwidth.py#L114-L139

src/_pytest/assertion/util.py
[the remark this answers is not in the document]
blueyed: Yeah.  So checking for 0 instead of -1 you mean?

src/_pytest/assertion/util.py
blueyed: I like the punctuation, but left it since others did not contain it.
We can start having it here.  Also the `NOTE:`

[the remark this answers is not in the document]
blueyed: Changed to `"NOTE: Strings contain non-printable/zero-width characters. Escaping them using repr()."` for now.

testing/test_assertion.py
```diff
@@ -334,8 +334,14 @@ def test_multiline_text_diff(self):
...
-        assert "- spam" in diff
-        assert "+ eggs" in diff
+        assert diff == [
+            r"'foo\nspam\nbar' == 'foo\neggs\nbar'",
+            r"NOTE: Strings contain non-printable/zero-width characters. Escaping them using repr().",
+            r"  'foo\n'",
+            r"- 'spam\n'",
+            r"+ 'eggs\n'",
+            r"  'bar'",
+        ]
```
The-Compiler: I'm a bit confused by this change (and others) - do you consider a newline a non-printable/zero-width character? Why? It seems quite confusing to me to see a multi-line output but also `\n`, i.e. with newlines represented twice.

blueyed: @The-Compiler 
> Why?

Otherwise it will not be visible.  Check https://github.com/pytest-dev/pytest/pull/6702/files/9318b236ff35fc6a0734ee91bff2011afb93971a#diff-14ad89cac128f1dd3aaa927c656035b3R1342.

This is for a line (text/str) diff.
They are there for lists already.

blueyed: It could be made smarter though maybe to not show them when not at the end of a line?

nicoddemus: For illustration, on `master`:

```
    def test_eq_similar_text():
        x ="foo\n1 bar"
>       assert x == "foo\n2 bar"
E       AssertionError: assert 'foo\n1 bar' == 'foo\n2 bar'
E           foo
E         - 1 bar
E         ? ^
E         + 2 bar
E         ? ^
```

On this branch:

```
    def test_eq_similar_text():
        x ="foo\n1 bar"
>       assert x == "foo\n2 bar"
E       AssertionError: assert 'foo\n1 bar' == 'foo\n2 bar'
E         NOTE: Strings contain non-printable/zero-width characters. Escaping them using repr().
E           'foo\n'
E         - '1 bar'
E         ?  ^
E         + '2 bar'
E         ?  ^
```

Is this what you would prefer @The-Compiler ?

```
    def test_eq_similar_text():
        x ="foo\n1 bar"
>       assert x == "foo\n2 bar"
E       AssertionError: assert 'foo\n1 bar' == 'foo\n2 bar'
E         NOTE: Strings contain non-printable/zero-width characters. Escaping them using repr().
E         - 'foo\n1 bar'
E         ?       ^
E         + 'foo\n2 bar'
E         ?       ^
```

blueyed: > For illustration, on `master`:
> 
> ```
>     def test_eq_similar_text():
>         x ="foo\n1 bar"
> >       assert x == "foo\n2 bar"
> E       AssertionError: assert 'foo\n1 bar' == 'foo\n2 bar'
> E           foo
> E         - 1 bar
> E         ? ^
> E         + 2 bar
> E         ? ^
> ```

On master it has a different order (which I find still wrong - it should be left-to-right, top-to-bottom, with only +/- swapped, but off topic here - maybe therefore edited manually?):
```
    def test_eq_similar_text(self):
        x = "foo\n1 bar"
>       assert x == "foo\n2 bar"
E       AssertionError: assert 'foo\n1 bar' == 'foo\n2 bar'
E           foo
E         - 2 bar
E         ? ^
E         + 1 bar
E         ? ^
```

How about?

```
    def test_eq_similar_text(self):
        x = "foo\n1 bar\n"
>       assert x == "foo\n2 bar"
E       AssertionError: assert 'foo\n1 bar\n' == 'foo\n2 bar'
E         NOTE: Strings contain different line-endings. Escaping them using repr().
E         - 'foo\n1 bar\n'
E         ?       ^    --
E         + 'foo\n2 bar'
E         ?       ^
```

However, with longer strings it is useful to split them on newlines, of course.

```
>       assert x == "foo\n1 bar"
E       AssertionError: assert 'foo\n1 bar\n' == 'foo\n1 bar'
E         NOTE: Strings contain different line-endings. Escaping them using repr().
E           foo
E         - 1 bar\n
E         ?      --
E         + 1 bar
E         -
``` 

FWIW it always looked a bit strange to me seeing:

```
- foo
+ foo
     ^
```

(it could also be a space etc)

> ```
> E         - 'foo\n1 bar'
> E         ?       ^
> E         + 'foo\n2 bar'
> E         ?       ^
> ```

This could also be triggered via some minimal length (related: https://github.com/blueyed/pytest/pull/218, where I split it onto separate lines with a certain length).

src/_pytest/assertion/util.py
```diff
@@ -225,10 +227,21 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
...
-        line.strip("\n")
-        for line in ndiff(left.splitlines(keepends), right.splitlines(keepends))
-    ]
+
+    left_lines = left.splitlines(keepends)
+    right_lines = right.splitlines(keepends)
+
+    if any(
+        wcwidth(ch) <= -1
+        for ch in itertools.chain.from_iterable([x for x in left_lines + right_lines])
```
bluetech: Maybe `for ch in line for line in left_lines + right_lines` (or it's the other way around - I always forget) instead of itertools big guns?

src/_pytest/assertion/util.py
```diff
@@ -225,10 +227,21 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
...
-    explanation += [
-        line.strip("\n")
-        for line in ndiff(left.splitlines(keepends), right.splitlines(keepends))
-    ]
+
+    left_lines = left.splitlines(keepends)
+    right_lines = right.splitlines(keepends)
+
+    if any(
+        wcwidth(ch) <= -1
```
bluetech: So is this meant to be `0` or `-1`?

testing/test_assertion.py
```diff
@@ -1068,6 +1081,17 @@ def test_reprcompare_whitespaces():
     ]
 
 
+def test_reprcompare_zerowidth_and_non_printable():
+    assert callequal("\x00\x1b[31mred", "\x1b[31mgreen") == [
+        "'\\x00\\x1b[31mred' == '\\x1b[31mgreen'",
```
bluetech: Can you convert these to raw strings also?

src/_pytest/assertion/util.py
```diff
@@ -225,10 +227,21 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
...
+        for ch in itertools.chain.from_iterable([x for x in left_lines + right_lines])
+    ):
+        left_lines = [repr(x) for x in left_lines]
+        right_lines = [repr(x) for x in right_lines]
+        explanation += [
+            "NOTE: Strings contain non-printable/zero-width characters. Escaping them using repr()."
```
bluetech: To keep the message short, I think the reference to zero-width characters can be removed. "Non-printable characters" does not seem to be a well-defined thing in Unicode, so we are free to include zero-width characters under that term. And it seems reasonable to do so to me...

src/_pytest/assertion/util.py
blueyed: Changed to `for ch in [ch for lines in left_lines + right_lines for ch in lines]`.

testing/test_assertion.py
The-Compiler: Hmm, I guess explicit is indeed better than implicit in this case.

PR #6703, opened 2020-02-10
test_argcomplete: remove usage of `distutils.spawn`

blueyed:
Fixes collection error with Python 3.5.3 (Travis):

    testing/test_parseopt.py:2: in <module>
        import distutils.spawn
    .tox/py35-coverage/lib/python3.5/distutils/__init__.py:4: in <module>
        import imp
    .tox/py35-coverage/lib/python3.5/imp.py:33: in <module>
        PendingDeprecationWarning, stacklevel=2)
    E   PendingDeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses

Build log: https://travis-ci.org/blueyed/pytest/builds/648305304

### DIFF
--- a/testing/test_parseopt.py
+++ b/testing/test_parseopt.py
@@ -1,7 +1,7 @@
 import argparse
-import distutils.spawn
 import os
 import shlex
+import shutil
 import sys
 
 import py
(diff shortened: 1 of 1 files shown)


bluetech:
Maybe better to just use `shutil.which` for this? https://docs.python.org/3/library/shutil.html#shutil.which

blueyed:
> Maybe better to just use `shutil.which` for this? https://docs.python.org/3/library/shutil.html#shutil.which

Sure, done.

Issue #6377, opened 2019-12-29
Also rewrite assertions in test modules ("helpers")


blueyed:
Pytest does not appear to rewrite assert statements in `tests/utils.py`, although it is is used/imported from `tests/test_foo.py`.

The-Compiler:
That seems like expected behavior to me, see https://docs.pytest.org/en/latest/writing_plugins.html#assertion-rewriting

blueyed:
@The-Compiler 
Thanks for the link to docs (which do not mention that this is based on `python_files` though).

Note that this issue was meant to enhance the "expected"/documented behavior however.
I've given a new setting ("assert_rewrite_fnpatterns") a try in https://github.com/blueyed/pytest/pull/145, but could also imagine to do this based on some automation (e.g. whitelist/rewrite the top-level package(s)/module(s) where tests are collected from).

PR #6705, opened 2020-02-10
doc: link to `python_files` from "Assertion Rewriting"

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/6377
Ref: https://github.com/blueyed/pytest/pull/145

### DIFF
--- a/doc/en/writing_plugins.rst
+++ b/doc/en/writing_plugins.rst
@@ -180,7 +180,8 @@ modifies the parsed AST before it gets compiled to bytecode.  This is
 done via a :pep:`302` import hook which gets installed early on when
 ``pytest`` starts up and will perform this rewriting when modules get
 imported.  However since we do not want to test different bytecode
-then you will run in production this hook only rewrites test modules
+then you will run in production this hook only rewrites
+:confval:`test modules <python_files>`
 themselves as well as any modules which are part of plugins.  Any
 other imported module will not be rewritten and normal assertion
 behaviour will happen.
--- a/doc/en/writing_plugins.rst
+++ b/doc/en/writing_plugins.rst
@@ -180,8 +180,8 @@ modifies the parsed AST before it gets compiled to bytecode.  This is
 done via a :pep:`302` import hook which gets installed early on when
 ``pytest`` starts up and will perform this rewriting when modules get
 imported.  However since we do not want to test different bytecode
-then you will run in production this hook only rewrites
-:confval:`test modules <python_files>`
+then you will run in production this hook only rewrites test modules
+themselves (as defined by the :confval:`python_files` configuration option)
 themselves as well as any modules which are part of plugins.  Any
 other imported module will not be rewritten and normal assertion
 behaviour will happen.


doc/en/writing_plugins.rst
```diff
@@ -180,7 +180,8 @@ modifies the parsed AST before it gets compiled to bytecode.  This is
 done via a :pep:`302` import hook which gets installed early on when
 ``pytest`` starts up and will perform this rewriting when modules get
 imported.  However since we do not want to test different bytecode
-then you will run in production this hook only rewrites test modules
+then you will run in production this hook only rewrites
+:confval:`test modules <python_files>`
```
bluetech:
```suggestion
test modules themselves (as defined by the :confval:`<python_files>` configuration option)
```

(not sure about the link syntax...)

blueyed: Thanks, applied.

doc/en/writing_plugins.rst
```diff
@@ -181,6 +181,7 @@ done via a :pep:`302` import hook which gets installed early on when
 ``pytest`` starts up and will perform this rewriting when modules get
 imported.  However since we do not want to test different bytecode
 then you will run in production this hook only rewrites test modules
+themselves (as defined by the :confval:`python_files` configuration option)
 themselves as well as any modules which are part of plugins.  Any
```
bluetech: Sorry for the misleading suggestion, now the word "themselves" appears twice!

blueyed: No worries, should have noticed myself.  How about this, also fixing the wrong "then" there:
```
diff --git i/doc/en/writing_plugins.rst w/doc/en/writing_plugins.rst
index 045e13e75..be7bd8410 100644
--- i/doc/en/writing_plugins.rst
+++ w/doc/en/writing_plugins.rst
@@ -180,11 +180,11 @@ modifies the parsed AST before it gets compiled to bytecode.  This is
 done via a :pep:`302` import hook which gets installed early on when
 ``pytest`` starts up and will perform this rewriting when modules get
 imported.  However since we do not want to test different bytecode
-then you will run in production this hook only rewrites test modules
-themselves (as defined by the :confval:`python_files` configuration option)
-themselves as well as any modules which are part of plugins.  Any
-other imported module will not be rewritten and normal assertion
-behaviour will happen.
+from what you will run in production this hook only rewrites test modules
+themselves (as defined by the :confval:`python_files` configuration option),
+and any modules which are part of plugins.
+Any other imported module will not be rewritten and normal assertion behaviour
+will happen.

 If you have assertion helpers in other modules where you would need
 assertion rewriting to be enabled you need to ask ``pytest``
```

```
imported.  However since we do not want to test different bytecode
from what you will run in production this hook only rewrites test modules
themselves (as defined by the :confval:`python_files` configuration option),
and any modules which are part of plugins.
Any other imported module will not be rewritten and normal assertion behaviour
will happen.
```

bluetech: Yes that reads well to me. I'd also add commas after the words "However" and "production".

blueyed: Ok: https://github.com/pytest-dev/pytest/pull/6741.

PR #6714, opened 2020-02-11
[on hold] Fix CaptureManager.read_global_capture after stop

blueyed:
There are several places that use `read_global_capture` after
`stop_global_capturing` might have been used, e.g. via
`pytest_internalerror` etc, where it shouldn't crash then more.

TODO:

- [ ] changelog
- [ ] test (where this might get triggered, but not really worth it probably)

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -141,6 +141,8 @@ class CaptureManager:
         self.resume_fixture(self._current_item)
 
     def read_global_capture(self):
+        if not self._global_capturing:
+            return CaptureResult("", "")
         return self._global_capturing.readouterr()
 
     # Fixture Control (it's just forwarding, think about removing this later)
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -141,9 +141,9 @@ class CaptureManager:
         self.resume_fixture(self._current_item)
 
     def read_global_capture(self):
-        if not self._global_capturing:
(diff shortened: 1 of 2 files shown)


bluetech:
Is this situation (read after stop) legitimate or a bug in the program?

blueyed:
> Is this situation (read after stop) legitimate or a bug in the program?

Both I'd say.  As for the capture plugin itself it stops it in some cases (https://github.com/pytest-dev/pytest/blob/d89b5057cadb8716e25b783fee8a1d670e043109/src/_pytest/capture.py#L257-L263), which might be wrapped with some outer capturing (by itself or others).

bluetech:
> Both I'd say

If it's generally illegitimate but sometimes is, it seems bad to just silently allow it? Is there another way to fix just specifically the recursive case?

blueyed:
@bluetech 
There are many similar / same checks with other methods here.. :)
I agree that it would be better to properly fix it, but the current state is like it is - and I'd rather not get annoyed by an additional crash on top of another.. ;)

blueyed:
I think I've found the bug in that regard, will hold this back for now.

Would you suggest that the test should be changed to "document" the crash / misuse then?

bluetech:
I guess such a test can serve as "if this failed, then you probably tried to make it work like this, but maybe you shouldn't".

nicoddemus:
Closing for now.

PR #6716, opened 2020-02-12
Merge the features branch into master, before stopping to use it

bluetech:
The features branch is no more. Development of features is now also done on master.
    
See https://github.com/pytest-dev/pytest/pull/6571.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,8 +53,8 @@ repos:
     -   id: changelogs-rst
         name: changelog filenames
         language: fail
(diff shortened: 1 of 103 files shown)


nicoddemus:
Just to make sure, was this a clean merge?

bluetech:
The only conflict was the `features` in the GitHub Action config:

https://github.com/pytest-dev/pytest/blob/4209ad6fcacb679d953f3dd6be96f330139006f0/.github/workflows/main.yml#L13

(and also one below). I resolved by removing it in the merge.

nicoddemus:
Awesome, let's get this in then! 😁

PR #6722, opened 2020-02-13
Remove unused CallSpec2 fields _globalid, _globalparam

bluetech:


blueyed:
Thanks.
Just for reference: added in ec0565fac initially, never used apparently.  Similar to 93e55ad2f.

PR #6724, opened 2020-02-13
Type annotate CallSpec2

bluetech:
Extracted from #6717.

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -4,13 +4,16 @@ import fnmatch
 import inspect
 import os
 import sys
+import typing
 import warnings
 from collections import Counter
 from collections import defaultdict
 from collections.abc import Sequence
 from functools import partial
 from typing import Dict
+from typing import Iterable
 from typing import List
+from typing import Mapping
 from typing import Optional
 from typing import Tuple
 from typing import Union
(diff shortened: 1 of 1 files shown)


src/_pytest/python.py
```diff
@@ -803,25 +812,39 @@ def copy(self):
...
+        valset: Iterable[object],
+        id: str,
+        marks: Iterable[Union[Mark, MarkDecorator]],
+        scopenum: int,
+        param_index: int,
+    ) -> None:
         for arg, val in zip(argnames, valset):
             self._checkargnotcontained(arg)
             valtype_for_arg = valtypes[arg]
-            getattr(self, valtype_for_arg)[arg] = val
```
bluetech: I had to tweak the code here a bit, mypy can't understand `getattr`. I also think it's better to avoid it in general.

To show that it is safe, I used `Literal` here and in the source below `Metafunc._resolve_arg_value_types`.

src/_pytest/python.py
```diff
@@ -803,25 +812,39 @@ def copy(self):
...
+            if valtype_for_arg == "params":
+                self.params[arg] = val
+            elif valtype_for_arg == "funcargs":
+                self.funcargs[arg] = val
+            else:  # pragma: no cover
```
nicoddemus: (Small tip, no need to change)

When I encounter these situations where we have a fixed number of if/elifs and an else clause as an error, I opt to have an explicit else: clause for the last choice:

```python
            if valtype_for_arg == "params":
                self.params[arg] = val
            elif valtype_for_arg == "funcargs":
                self.funcargs[arg] = val
            else:  # pragma: no cover
                assert False, "Unhandled valtype for arg: {}".format(valtype_for_arg)
```

Becomes:

```python
            if valtype_for_arg == "params":
                self.params[arg] = val
            else:
                assert valtype_for_arg == "funcargs", "Unhandled valtype for arg: {}".format(valtype_for_arg)
                self.funcargs[arg] = val
```

It is shorter and avoids having to write a `pragma: nocover` 😁

bluetech: Thanks for the suggestion!

There is actually a reason why I prefer the form I wrote -- it allows to perform exhaustiveness checking. Consider for example that a new value is added to the `valtype_for_arg` Literal type. Then ideally, mypy would raise an error if I didn't remember to add a case for it in the conditional.

For enums, this is already possible, see https://github.com/python/mypy/issues/5818. In my own code I rely on this greatly.

For Literals, this is not yet supported by mypy, but hopefully will be in the future: https://github.com/python/mypy/issues/6366

Regarding to `no cover`, WDYT about adding `assert False` to the coverage `exclude_lines` list?

blueyed: `raise NotimplementedError()`?

bluetech: I think an assert is more appropriate than `NotimplementedError` for this. assert means "internal assumption broken", while `NotimplementedError` might imply "not implemented *intentionally*" which is not the case.

blueyed: I see it more like "might be implemented later", like you've decribed it yourself ("a new value is added").
But then maybe lets add `^\s*raise AssertionError\b` to the exclude list?  I think it is good to be explicit here, to not ignore any `assert 0`, but your `assert False` might work for this also already.

PR #6725, opened 2020-02-13
Type annotate some hookspecs & impls

bluetech:
Extracted from #6717.

Annotate some "easy" arguments of hooks that repeat in a lot of internal plugins.

Not all of the arguments are annotated fully for now.

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -12,6 +12,7 @@ from typing import Generator
 from typing import List
 from typing import Optional
 from typing import Set
+from typing import Union
 
 import attr
 import py
@@ -24,6 +25,8 @@ from .reports import CollectReport
 from _pytest import nodes
 from _pytest._io import TerminalWriter
 from _pytest.config import Config
+from _pytest.config import ExitCode
+from _pytest.config.argparsing import Parser
 from _pytest.main import Session
 from _pytest.python import Module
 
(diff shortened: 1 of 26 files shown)


src/_pytest/cacheprovider.py
```diff
@@ -372,16 +376,18 @@ def pytest_addoption(parser):
...
 
         return wrap_session(config, cacheshow)
+    return None
 
 
 @pytest.hookimpl(tryfirst=True)
-def pytest_configure(config):
-    config.cache = Cache.for_config(config)
+def pytest_configure(config: Config) -> None:
+    # Type ignored: pending mechanism to store typed objects scoped to config.
```
bluetech: This repeats in several plugins. I have a simple plan for how to do it, but for now I added type-ignores.

src/_pytest/logging.py
```diff
@@ -510,13 +512,13 @@ def __init__(self, config):
         )
 
         log_file = get_option_ini(config, "log_file")
-        if log_file:
+        if not log_file:
```
bluetech: Swapped the order so mypy can infer the type without an annotation.

src/_pytest/hookspec.py
```diff
@@ -185,15 +198,15 @@ def pytest_collection_modifyitems(session, config, items):
...
+def pytest_collection_finish(session: "Session"):
     """ called after collection has been performed and modified.
 
     :param _pytest.main.Session session: the pytest session object
     """
 
 
 @hookspec(firstresult=True)
-def pytest_ignore_collect(path, config):
+def pytest_ignore_collect(path, config: "Config"):
```
blueyed: `path: py.path.local`

bluetech: I prefer leaving annotating `path` out of this particular PR.

PR #6726, opened 2020-02-13
Type annotate ParameterSet & param()

bluetech:
Extracted from #6717.

### DIFF
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -1,6 +1,7 @@
 """
 python version compatibility code
 """
+import enum
 import functools
 import inspect
 import io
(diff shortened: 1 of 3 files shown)


src/_pytest/compat.py
```diff
@@ -35,13 +36,20 @@
...
 if TYPE_CHECKING:
     from typing import Type  # noqa: F401 (used in type string)
+    from typing_extensions import Final
 
 
 _T = TypeVar("_T")
 _S = TypeVar("_S")
 
 
-NOTSET = object()
```
bluetech: `object()` doesn't produce a singleton on the type level, so I had to switch it to an idiom that does.

src/_pytest/mark/structures.py
```diff
@@ -49,16 +62,32 @@ def get_empty_parameterset_mark(config, argnames, func):
...
+    def param(
+        cls,
+        *values: object,
+        marks: "Union[MarkDecorator, typing.Collection[Union[MarkDecorator, Mark]]]" = (),
+        id: Optional[str] = None
+    ) -> "ParameterSet":
         if isinstance(marks, MarkDecorator):
             marks = (marks,)
         else:
-            assert isinstance(marks, (tuple, list, set))
```
bluetech: Generalized this a bit, otherwise type annotation is very ugly.

bluetech: But turns out py3.5 doesn't have `collections.abc.Collection`, so I changed the runtime check to `collections.abc.Sequence | set`, but left the type as `Collection`. So there will be a slight mismatch until we can drop py3.5.

src/_pytest/mark/__init__.py
```diff
@@ -14,7 +18,11 @@
 __all__ = ["Mark", "MarkDecorator", "MarkGenerator", "get_empty_parameterset_mark"]
 
 
-def param(*values, **kw):
```
bluetech: I expanded the kwargs so I can annotate them and so that the types show up in the docs.

Issue #6505, opened 2020-01-18
5.3 changes plurality of keys in testdir.run().parseoutcomes()


altendky:
While `testdir.run().parseoutcomes()` used to give results such as `{'warnings': 1}` (plural) it now gives `{'warning': 1}` (singular).  This broke the tests in `pytest-twisted`...  and while I setup nightly 'cron' builds I apparently didn't bother to notice when they broke awhile ago.  :[  I do not know what the proper answer is here especially now that we have multiple versions out with each approach.  Since `pytest-twisted` is still supporting py2, I don't get to just update to latest pytest functionality across the board.  Though yes, py2 support will at some point be dropped when the pain is sufficient.

I'll just normalize the `parseoutcomes()` dict keys myself but wanted to document the change and see what opinions there are on it.

When `pytest-twisted` broke:
last good `pytest==5.2.4`: https://travis-ci.org/pytest-dev/pytest-twisted/jobs/614201584
first bad `pytest==5.3.0`: https://travis-ci.org/pytest-dev/pytest-twisted/jobs/614718766

Introduced in #5989 and https://github.com/pytest-dev/pytest/commit/d863c30c743138ed29cdf786cb18c41c24983fe1.

There seems to be little testing being done on `parseoutcomes()`.
https://github.com/pytest-dev/pytest/blob/e98176cf50c5b2d3df965ec409d7f62fefcd9ebd/testing/test_pytester.py#L687-L706
https://github.com/pytest-dev/pytest/blob/e98176cf50c5b2d3df965ec409d7f62fefcd9ebd/testing/test_capture.py#L1161-L1191

Zac-HD:
#5990 makes sense for user-facing reporting, but IMO the `parseoutcomes` keys should not vary with their values, and I would therefore be happy to accept a pull request that ensures we always use plurals.

blueyed:
Sounds related: #6382

altendky:
Looks like #6382 covers error[s] so yeah, definitely related and a start.  Perhaps we need an explicit singular/plural table for the outcomes which can be used for both pleasant human readable output, and for undoing that human readable output..

MarcoGorelli:
Really sorry for the breakage I introduced with my issue / PR! Gonna be pretty busy for the next two weeks, but if this isn't addressed by then I'll take a look

koscinskic:
If we could, myself and a few others can look into ensuring a system that functions solely on plural-only keys while retaining proper singular / plural output.

[...]

PR #6730, opened 2020-02-13
Fixed plurality issue #6505

barrezuetai:
#6505

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -72,6 +72,8 @@ class Code:
         """ return a path object pointing to source code (or a str in case
         of OSError / non-existing file).
         """
+        if not self.raw.co_filename:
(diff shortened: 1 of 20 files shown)


blueyed:
Thanks, but it should be kept singular/plural in user facing places - maybe you have mistaken the comment in https://github.com/pytest-dev/pytest/issues/6505#issuecomment-575960724?

A fix would look more like what was done in https://github.com/pytest-dev/pytest/pull/6382.
See https://github.com/pytest-dev/pytest/pull/6382/files#diff-1284161c2c98c8aedd66545ff79a8af6R440-R442 specifically.

Zac-HD:
Closing this in favor of #6905 - thanks for your persistence @barrezuetai :slightly_smiling_face:

PR #6732, opened 2020-02-13
reports: use attr.asdict with serialize_repr_{crash,traceback}

blueyed:
This helps with skipping "private" properties there.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -975,18 +975,13 @@ class ReprExceptionInfo(ExceptionRepr):
         super().toterminal(tw)
 
 
+@attr.s
 class ReprTraceback(TerminalRepr):
-    entrysep = "_ "
+    reprentries = attr.ib(type=Sequence[Union["ReprEntry", "ReprEntryNative"]])
+    extraline = attr.ib(type=Optional[str])
+    style = attr.ib(type="_TracebackStyle")
 
-    def __init__(
-        self,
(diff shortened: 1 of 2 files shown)


src/_pytest/_code/code.py
```diff
@@ -975,18 +975,14 @@ def toterminal(self, tw: TerminalWriter) -> None:
         super().toterminal(tw)
 
 
+@attr.s
 class ReprTraceback(TerminalRepr):
-    entrysep = "_ "
+    reprentries = attr.ib(type=Sequence[Union["ReprEntry", "ReprEntryNative"]])
+    extraline = attr.ib(type=Optional[str])
+    style = attr.ib(type="_TracebackStyle")
+    style = style
```
```suggestion
```

bluetech:
LGTM, though I think it will be better to also convert `TerminalRepr` and all of its other subclasses to attrs at the same time. Just to stave off any weird interactions between attrs and non-attrs classes in the class hierarchy. I'm sure it works fine but it just seems like something that is better avoided.

nicoddemus:
> LGTM, though I think it will be better to also convert TerminalRepr and all of its other subclasses to attrs at the same time.

That would be ideal, but let's not hold this PR on that (I don't think here will be any weird interactions). Let's leave it up to @blueyed if he wants to tackle converting the other classes or not.

blueyed:
It's not trivial for all of them, but will create a WIP PR for it already.

PR #6733, opened 2020-02-13
tests: re-harden test_cmdline_python_package_symlink

blueyed:
This was done on purpose in 7268462b3, and then reverted in a non-merge
"Merge" commit (wrong conflict resolution?):
https://github.com/pytest-dev/pytest/pull/4158/commits/9646a1cd7#diff-1f1fc4eef6bf4c4da558e0dce9e74e9bR761-R762

### DIFF
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -1,3 +1,4 @@
+import os
 import re
 import sys
 from typing import List
@@ -134,6 +135,24 @@ def dummy_yaml_custom_test(testdir):
 def testdir(testdir: Testdir) -> Testdir:
     testdir.monkeypatch.setenv("PYTEST_DISABLE_PLUGIN_AUTOLOAD", "1")
     return testdir
+
+
+@pytest.fixture
+def symlink_or_skip():
+    """Return a function to create symlinks or skip the test.
+
+    On Windows `os.symlink` is available, but normal users require special
+    admin privileges to create symlinks.
+    """
+
+    def wrap_os_symlink(src, dst, *args, **kwargs):
+        try:
+            os.symlink(src, dst, *args, **kwargs)
+        except OSError as e:
+            pytest.skip("os.symlink({!r}) failed: {!r}".format((src, dst), e))
+        assert os.path.islink(dst)
+
+    return wrap_os_symlink
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


nicoddemus:
> wrong conflict resolution?

Probably, thanks!

blueyed:
The Windows behavior in py38 confused me - is it creating only a non-dir symlink there or something?

blueyed:
So my guess is that this somehow regressed before, but is/was hidden by un-hardening it in two different places (and the initial mixup of checking py.path's attribute, additionally to skipping it etc).
Needs someone with Windows to investigate I guess.

nicoddemus:
> Needs someone with Windows to investigate I guess.

I'm assigning it to myself, thanks again.

testing/conftest.py
```diff
@@ -136,6 +137,24 @@ def testdir(testdir: Testdir) -> Testdir:
     return testdir
 
 
+@pytest.fixture
+def symlink_or_skip():
```
blueyed: Slightly improved version used in https://github.com/pytest-dev/pytest/pull/6733.

nicoddemus:
Gave this another go, but no luck so far.

It seems that on Linux nodeids are constructed by resolving symlinks, hence we get this output:

```
lib/foo/bar/test_bar.py::test_bar <- local/lib/foo/bar/test_bar.py PASSED
```

That output means `(nodeid) <- (fspath)`, built from here:

https://github.com/pytest-dev/pytest/blob/f77d606d4eeba38352bd3ea960f44b171fe06c1f/src/_pytest/terminal.py#L772-L775 

On Windows, `nodeid` and `fspath` are the symlink:

```
local/lib/foo/bar/test_bar.py::test_bar PASSED
```

I have to find where `nodeid` turns into a symlink. Will need to investigate more some other time.

blueyed:
@nicoddemus 
The interesting part would be why / if symlinks always (seem) to work on Windows.
I've expected this to fail / be skipped there in general.
Also with regard to what py.path uses to determine if there is symlink support.

nicoddemus:
> The interesting part would be why / if symlinks always (seem) to work on Windows.
I've expected this to fail / be skipped there in general.

Oh sorry I missed that: Administrators and users with special permissions can create symlinks, normal users cannot. I suspect the user running in CI has admin privileges, that's why it works there.

> Also with regard to what py.path uses to determine if there is symlink support.

It does the right thing to check for symlinks (just call `os.path.islink`).
https://github.com/pytest-dev/py/blob/962ad137fcfc52a218fc8acc3d87708eccf00bbc/py/_path/local.py#L373-L374

But yeah I need to set aside more time to investigate further and figure out what's going on.

PR #6735, opened 2020-02-14
Type annotate Metafunc

bluetech:
Extracted from #6717.

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -2,14 +2,18 @@
 import enum
 import fnmatch
 import inspect
+import itertools
 import os
 import sys
+import typing
 import warnings
 from collections import Counter
 from collections import defaultdict
 from collections.abc import Sequence
 from functools import partial
+from typing import Callable
 from typing import Dict
+from typing import Iterable
 from typing import List
 from typing import Optional
 from typing import Tuple
@@ -23,6 +27,8 @@ from _pytest import nodes
 from _pytest._code import filter_traceback
 from _pytest._code.code import ExceptionInfo
 from _pytest._code.source import getfslineno
+from _pytest._io import TerminalWriter
+from _pytest._io.saferepr import saferepr
 from _pytest.compat import ascii_escaped
 from _pytest.compat import get_default_arg_names
 from _pytest.compat import get_real_func
@@ -35,8 +41,10 @@ from _pytest.compat import REGEX_TYPE
 from _pytest.compat import safe_getattr
 from _pytest.compat import safe_isclass
 from _pytest.compat import STRING_TYPES
+from _pytest.config import Config
 from _pytest.config import hookimpl
 from _pytest.deprecated import FUNCARGNAMES
+from _pytest.fixtures import FuncFixtureInfo
 from _pytest.mark import MARK_GEN
 from _pytest.mark import ParameterSet
 from _pytest.mark.structures import get_unpacked_marks
@@ -125,9 +133,10 @@ def pytest_cmdline_main(config):
         return 0
 
 
-def pytest_generate_tests(metafunc):
+def pytest_generate_tests(metafunc: "Metafunc") -> None:
     for marker in metafunc.definition.iter_markers(name="parametrize"):
-        metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker)
+        # TODO: Fix this type-ignore (overlapping kwargs).
+        metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker)  # type: ignore[misc] # noqa: F821
 
 
 def pytest_configure(config):
@@ -839,8 +848,8 @@ class Metafunc:
     def __init__(
         self,
         definition: "FunctionDefinition",
-        fixtureinfo,
-        config,
+        fixtureinfo: fixtures.FuncFixtureInfo,
+        config: Config,
         cls=None,
         module=None,
     ) -> None:
@@ -872,14 +881,19 @@ class Metafunc:
 
     def parametrize(
         self,
-        argnames,
-        argvalues,
-        indirect=False,
-        ids=None,
-        scope=None,
+        argnames: Union[str, List[str], Tuple[str, ...]],
+        argvalues: Iterable[Union[ParameterSet, typing.Sequence[object], object]],
+        indirect: Union[bool, typing.Sequence[str]] = False,
+        ids: Optional[
+            Union[
+                Iterable[Union[None, str, float, int, bool]],
+                Callable[[object], Optional[object]],
+            ]
+        ] = None,
+        scope: "Optional[str]" = None,
         *,
         _param_mark: Optional[Mark] = None
-    ):
+    ) -> None:
         """ Add new invocations to the underlying test function using the list
         of argvalues for the given argnames.  Parametrization is performed
         during the collection phase.  If you need to setup expensive resources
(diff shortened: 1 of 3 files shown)


src/_pytest/python.py
```diff
@@ -1000,49 +1023,59 @@ def _resolve_arg_ids(
         :rtype: List[str]
         :return: the list of ids for each argname given
         """
-        idfn = None
```
bluetech: I rearranged the code here to make it amenable to mypy. I also think it's a bit clearer.

src/_pytest/python.py
```diff
@@ -1000,49 +1023,59 @@ def _resolve_arg_ids(
...
+        return idmaker(argnames, parameters, idfn, ids_, self.config, item=item)
+
+    def _validate_ids(
+        self,
+        ids: Iterable[Union[None, str, float, int, bool]],
+        parameters: typing.Sequence[ParameterSet],
+        func_name: str,
+    ) -> List[Union[None, str]]:
         try:
-            len(ids)
```
bluetech: I had to refactor a bit to make it amenable to mypy. It's meant to be equivalent.

src/_pytest/python.py
```diff
@@ -1000,49 +1023,59 @@ def _resolve_arg_ids(
...
+            num_ids = len(ids)  # type: ignore[arg-type] # noqa: F821
         except TypeError:
             try:
-                it = iter(ids)
+                iter(ids)
             except TypeError:
-                raise TypeError("ids must be a callable, sequence or generator")
-            else:
-                import itertools
```
bluetech: I moved this import up -- don't see why it should be lazy.

src/_pytest/python.py
```diff
@@ -1095,7 +1132,7 @@ def _validate_if_using_arg_names(self, argnames, indirect):
                         pytrace=False,
                     )
                 else:
-                    if isinstance(indirect, (tuple, list)):
+                    if isinstance(indirect, Sequence):
```
bluetech: Generalized this a bit to make typing easier. It's also what other functions already check - was inconsistent.

src/_pytest/python.py
```diff
@@ -1113,17 +1154,15 @@ def _validate_explicit_parameters(self, argnames, indirect):
         :param indirect: same ``indirect`` parameter of ``parametrize()``.
         :raise ValueError: if validation fails
         """
-        if isinstance(indirect, bool) and indirect is True:
```
bluetech: Had to rearrange things here a bit.

src/_pytest/python.py
```diff
@@ -1204,13 +1253,22 @@ def _idval(val, argname, idx, idfn, item, config):
         return ascii_escaped(val.pattern)
     elif isinstance(val, enum.Enum):
         return str(val)
-    elif hasattr(val, "__name__") and isinstance(val.__name__, str):
+    elif isinstance(getattr(val, "__name__", None), str):
```
bluetech: Changed a bit to avoid mypy attr warnings.

src/_pytest/python.py
```diff
@@ -1220,32 +1278,41 @@ def _idvalset(idx, parameterset, argnames, idfn, ids, item, config):
...
+
+def idmaker(
+    argnames: Iterable[str],
+    parametersets: Iterable[ParameterSet],
+    idfn: Optional[Callable[[object], Optional[object]]] = None,
+    ids: Optional[List[Union[None, str]]] = None,
+    config: Optional[Config] = None,
+    item=None,
+) -> List[str]:
+    resolved_ids = [
```
bluetech: Renamed so can have a different type from the input `ids`.

src/_pytest/python.py
```diff
@@ -125,9 +133,10 @@ def pytest_cmdline_main(config):
         return 0
 
 
-def pytest_generate_tests(metafunc):
+def pytest_generate_tests(metafunc: "Metafunc") -> None:
     for marker in metafunc.definition.iter_markers(name="parametrize"):
-        metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker)
+        # TODO: Fix this type-ignore (overlapping kwargs).
```
blueyed: Do you have a plan for this?

src/_pytest/python.py
```diff
@@ -1000,49 +1023,59 @@ def _resolve_arg_ids(
...
-        if ids:
-            func_name = self.function.__name__
-            ids = self._validate_ids(ids, parameters, func_name)
-        ids = idmaker(argnames, parameters, idfn, ids, self.config, item=item)
-        return ids
-
-    def _validate_ids(self, ids, parameters, func_name):
+            ids_ = None
+        elif not ids:
+            # Special case: https://github.com/pytest-dev/pytest/issues/1849
```
blueyed: Might get handled also by _validate_ids?  (handling empty list/returing there).

It looks like a lot of "validation" happens before it already (also before, but a bit more now?).
Not too bad, but just a thought.

src/_pytest/python.py
```diff
@@ -1220,32 +1278,41 @@ def _idvalset(idx, parameterset, argnames, idfn, ids, item, config):
         ]
         return "-".join(this_id)
     else:
-        return _ascii_escaped_by_config(ids[idx], config)
-
-
-def idmaker(argnames, parametersets, idfn=None, ids=None, config=None, item=None):
-    ids = [
+        id = ids[idx]
+        assert id is not None  # Checked above.
```
blueyed: I wonder if this could also have a type comment so that it can be detected/removed when mypy recognizes it?

src/_pytest/python.py
blueyed: Maybe s/resolved_ids/ret/ then?

src/_pytest/python.py
bluetech: Not really. There's a workaround of creating `kwargs = {**marker.kwargs, '_param_marker': marker}` but I think it's worse than the disease.

src/_pytest/python.py
bluetech: The following works:

```diff
diff --git a/src/_pytest/python.py b/src/_pytest/python.py
index e4d7cbf86..802298641 100644
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1029,10 +1029,6 @@ class Metafunc:
         elif callable(ids):
             idfn = ids
             ids_ = None
-        elif not ids:
-            # Special case: https://github.com/pytest-dev/pytest/issues/1849
-            idfn = None
-            ids_ = []
         else:
             idfn = None
             ids_ = self._validate_ids(ids, parameters, self.function.__name__)
@@ -1053,7 +1049,8 @@ class Metafunc:
                 raise TypeError("ids must be a callable or an iterable")
             num_ids = len(parameters)
 
-        if num_ids != len(parameters):
+        # num_ids == 0 is a special case: https://github.com/pytest-dev/pytest/issues/1849
+        if num_ids != len(parameters) and num_ids != 0:
             msg = "In {}: {} parameter sets specified, with different number of ids: {}"
             fail(msg.format(func_name, len(parameters), num_ids), pytrace=False)
```

If you think it's better (I do!) I can include it.

src/_pytest/python.py
bluetech: Not that I know of.

This can be rewritten to avoid the need for the assert:

```diff
diff --git a/src/_pytest/python.py b/src/_pytest/python.py
index e4d7cbf86..5edc89322 100644
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1271,15 +1271,14 @@ def _idvalset(
 ):
     if parameterset.id is not None:
         return parameterset.id
-    if ids is None or (idx >= len(ids) or ids[idx] is None):
+    id = None if ids is None or idx >= len(ids) else ids[idx]
+    if id is None:
         this_id = [
             _idval(val, argname, idx, idfn, item=item, config=config)
             for val, argname in zip(parameterset.values, argnames)
         ]
         return "-".join(this_id)
     else:
-        id = ids[idx]
-        assert id is not None  # Checked above.
         return _ascii_escaped_by_config(id, config)
```

Let's me know if you think it's better.

src/_pytest/python.py
bluetech: I used this name mostly because of the terminology established by `_resolve_arg_ids`. Can change if you think `ret` is better.

src/_pytest/python.py
blueyed: Yes, I also think it is better.

src/_pytest/python.py
blueyed: This seems to be a bit better also (avoiding a comment/explanation also).

src/_pytest/python.py
blueyed: It's fine then.

bluetech:
Thanks @blueyed, I applied the changes.

PR #6737, opened 2020-02-14
Don't delete FixtureDef.cached_result, set it to None instead

bluetech:
Previously `cached_result` was either set or deleted. Type annotations cannot handle this, so use `None` for the non-set state instead.

### DIFF
--- a/changelog/6737.breaking.rst
+++ b/changelog/6737.breaking.rst
@@ -1,0 +1,7 @@
+The ``cached_result`` attribute of ``FixtureDef`` is now set to ``None`` when
+the result is unavailable, instead of being deleted.
+
+If your plugin perform checks like ``hasattr(fixturedef, 'cached_result')``,
+for example in a ``pytest_fixture_post_finalizer`` hook implementation, replace
+it with ``fixturedef.cached_result is not None``. If you ``del`` the attribute,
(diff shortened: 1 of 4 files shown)


bluetech:
This is minor breaking change. I was able to find two projects which use `hasattr` on this field:

- https://github.com/griddic/code_blocks/blob/79057ee6352df9e158f4a9deaa8d2faaf9910430/python/pytest/freeze%20fixtures/conftest.py
-https://github.com/devova/pytest-yield/blob/37234c08ba55504039f14868a47ffc6691e07ba6/pytest_yield/fixtures.py

nicoddemus:
About https://github.com/griddic/code_blocks, I think we can open a PR there just changing `if not hasattr(fixturedef, 'cached_result'):` to `if not hasattr(fixturedef, 'cached_result', None):` which will work in all versions.

`pytest-yield` I suspect is deprecated, given some of its code is copied from pytest.

nicoddemus:
Went ahead and opened https://github.com/griddic/code_blocks/pull/1. 👍

nicoddemus:
Thanks @bluetech for the PR and research involved! 😁

griddic:
@nicoddemus , thank you for caring about my code!

PR #6743, opened 2020-02-15
Avoid getattr when dispatching pytest_runtest_* hooks

bluetech:
Using getattr doesn't work with typing, and also breaks grep. It took me a while to find where these hooks are called.

### DIFF
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -25,6 +25,7 @@ from _pytest.outcomes import TEST_OUTCOME
 
 if TYPE_CHECKING:
     from typing import Type
+    from typing_extensions import Literal
 
 #
 # pytest plugin hooks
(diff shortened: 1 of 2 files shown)


src/_pytest/runner.py
```diff
@@ -200,9 +203,15 @@ def check_interactive_exception(call, report):
...
-    ihook = getattr(item.ihook, hookname)
+def call_runtest_hook(item, when: "Literal['setup', 'call', 'teardown']", **kwds):
+    if when == "setup":
+        ihook = item.ihook.pytest_runtest_setup
+    elif when == "call":
+        ihook = item.ihook.pytest_runtest_call
+    elif when == "teardown":
+        ihook = item.ihook.pytest_runtest_teardown
+    else:
+        assert False, "Unhandled runtest hook case: {}".format(when)
```
blueyed: Do you like to go with `assert False` in .coveragerc then?
Please add it here then maybe already.

bluetech: Sure, added a commit for this.

PR #6747, opened 2020-02-15
pytester: LineMatcher: use `_fail` with `_match_lines_random`

blueyed:


bluetech:
I am not very familiar with this but from what I can understand this makes sense.

`get_lines_after` also raises `ValueError`, should it be changed as well?

blueyed:
> `get_lines_after` also raises `ValueError`, should it be changed as well?

It's not really used for validation/asserting itself, so a ValueError there makes sense still.

PR #6749, opened 2020-02-16
capture: do not delete attributes

blueyed:
Especially for `_old`, which was still used unchecked in `__repr__`.

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -552,12 +552,12 @@ class FDCaptureBinary:
             self.tmpfile_fd = tmpfile.fileno()
 
     def __repr__(self):
-        return "<{} {} oldfd={} _state={!r} tmpfile={!r}>".format(
+        return "<{} {} oldfd={} _state={!r} tmpfile={}>".format(
             self.__class__.__name__,
             self.targetfd,
             getattr(self, "targetfd_save", "<UNSET>"),
             self._state,
-            self.tmpfile,
+            hasattr(self, "tmpfile") and repr(self.tmpfile) or "<UNSET>",
         )
 
     def _start(self):
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -552,8 +552,12 @@ class FDCaptureBinary:
             self.tmpfile_fd = tmpfile.fileno()
 
     def __repr__(self):
-        return "<FDCapture {} oldfd={} _state={!r}>".format(
-            self.targetfd, getattr(self, "targetfd_save", None), self._state
+        return "<{} {} oldfd={} _state={!r} tmpfile={!r}>".format(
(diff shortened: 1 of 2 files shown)


src/_pytest/capture.py
```diff
@@ -490,9 +487,9 @@ def resume_capturing(self):
             self.out.resume()
         if self.err:
             self.err.resume()
-        if hasattr(self, "_in_suspended"):
```
bluetech: :+1:  Down with `hasattr`!

src/_pytest/capture.py
```diff
@@ -653,7 +650,7 @@ def snap(self):
 
     def done(self):
         setattr(sys, self.name, self._old)
-        del self._old
+        self._old = None
```
bluetech: This means `self._old` will need to be `Optional` once this is typed, and then all accesses to it will need to be guarded. Not sure which is better.

blueyed: Ok.  But I am not sure why it needs to be guarded really, given that it is private?
Do you mean that we cannot rely on AttributeError everywhere anymore, or rather have to `assert` it in some places?
I've also found `targetfd_save = self.__dict__.pop("targetfd_save")`, which is not even private, and needs some test adjustments (that test e.g. for AttributeError) when keeping it as `None`.
I've thought keeping the attribute would be preferred with regard to mypy / internal typing.
What do you suggest?

bluetech: > But I am not sure why it needs to be guarded really, given that it is private?

Once we add typing, the type of `self._old` will need to be `Optional[something]` to have `self._old = None` work. And then, places like

```py
    def writeorg(self, data):
        self._old.write(data)
        self._old.flush()
```

will give type errors, which will need to be handled with an assert or an `if` or an exception etc.

The `del` "conveniently" avoids that. I'm not sure which is better, so I'm fine with whatever you decide. But if you device to remove the `del`, it would be nice to add the guards already, as they will needed when typing is added anyway.

blueyed: @bluetech 
That what I've thought, yes.

> The del "conveniently" avoids that.

But requires to have `# type: ignore` comments then (for missing attrib), no?
Given the other places and explicit tests for Attribute/ValueError I'll probably stick with fixing the `__repr__`s only (first).

bluetech: > But requires to have # type: ignore comments then (for missing attrib), no?

I don't think so. As far as I know, mypy doesn't know about `del` and treats it like it didn't happen. Since `_old` is defined in `__init__` mypy thinks it exists even after `del`, so accessing it after it's deleted would be purely a runtime error.

blueyed: Ok, cool then.  Amended, and improved/aligned a bit.

Issue #291, opened 2013-04-08
tmpdir names get too long on Windows (exceeding PATH_MAX)


pytestbot:
Originally reported by: **Torsten Landschoff (BitBucket: [bluehorn](http://bitbucket.org/bluehorn), GitHub: [bluehorn](http://github.com/bluehorn))**

----------------------------------------

Hi pytest Team,

we run into a lot of error in our release tests lately, which do not occur in nightly an on-commit tests. Looking into this, I discovered that path names get too long when we are using tmpdir:

    c:\users\dynamore\appdata\local\temp\pytest-21\test_change_own_private_components_public_incoming_conflict_pull_C__jenkins_workspace_loco2_win7_32_client_release_testdata_testdb_dump_0\vaultdir\zipped\00\005292e500e4e76be18e798eccee423dec2e24c57db4b030348e67a7

On (german) Windows this results into

  WindowsError: [Error 3] Das System kann den angegebenen Pfad nicht finden

which would point to a bug in our implementation.

It is unobvious to me that tmpdir includes the following info in the path:
* the full function name
* the parameters of the test call (full path to testdb.dump over here)

This can make the temporary path arbitrary long which causes problems on Windows. I think, py.test should shorten the path automatically (at least on Windows). We can work around this by passing relative filenames (less robust in case the code does a chdir, but that would break other tests anyway) and by shorting the names of our test functions. However, I would like to keep the test function names expressive.

----------------------------------------
- Bitbucket: https://bitbucket.org/pytest-dev/pytest/issue/291

nicoddemus:
This has been fixed for awhile now, tmpdir automatically truncates directory names to around `30` chars:

```python
def test_tmpdir_test_change_own_private_components_public_incoming_conflict_pull_C__jenkins_workspace_loco2_win7_32_client_release_testdata_testdb_dump(tmpdir):
    print(tmpdir)
    assert 0 
```

Generates this path:

```
C:\Users\Bruno\AppData\Local\Temp\pytest-of-Bruno\pytest-173\test_tmpdir_test_change_own_pr0
```

So I'm closing this. :smile:

cc @bluehorn

Bluehorn:
Agreed, looks fixed to me :-)

PR #6755, opened 2020-02-17
Fix removal of very long paths on Windows

torcolvin:
Allow tmpdir to be removed in cases where the creation of files does not exceed the length of MAX_PATH, but adding garbage + UUID does exceed the length of path. Fixes https://github.com/pytest-dev/pytest/issues/291

### DIFF
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -100,24 +100,41 @@ def on_rm_rf_error(func, path: str, exc, *, start_path: Path) -> bool:
     return True
 
 
-def create_long_path(path: Path) -> Path:
-    """Construct a path which will work on Windows if greater
-    than 260 characters. Resolves into a absolute path which
-    bypasses the typical MAX_PATH limitation:
-    https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation"""
+def ensure_extended_length_path(path: Path) -> Path:
+    """Get the extended-length version of a path (Windows).
+
+    On Windows, by default, the maximum length of a path (MAX_PATH) is 260
+    characters, and operations on paths longer than that fail. But it is possible
+    to overcome this by converting the path to "extended-length" form before
+    performing the operation:
+    https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation
+
+    On Windows, this function returns the extended-length absolute version of path.
+    On other platforms it returns path unchanged.
+    """
     if sys.platform.startswith("win32"):
         path = path.resolve()
-        # for the API
-        if not str(path).startswith("\\\\?\\"):
-            path = Path("\\\\?\\" + str(path))
+        path = Path(get_extended_length_path_str(path))
     return path
+
+
+def get_extended_length_path_str(path: str) -> str:
+    """Converts to extended length path as a str"""
+    long_path_prefix = "\\\\?\\"
+    unc_long_path_prefix = "\\\\?\\UNC\\"
+    if path.startswith((long_path_prefix, unc_long_path_prefix)):
+        return path
+    # UNC
+    if path.startswith("\\\\"):
+        return unc_long_path_prefix + path[2:]
+    return long_path_prefix + path
 
 
 def rm_rf(path: Path) -> None:
     """Remove the path contents recursively, even if some elements
     are read-only.
     """
-    path = create_long_path(path)
+    path = ensure_extended_length_path(path)
     onerror = partial(on_rm_rf_error, start_path=path)
     shutil.rmtree(str(path), onerror=onerror)
 
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -99,11 +99,23 @@ def on_rm_rf_error(func, path: str, exc, *, start_path: Path) -> bool:
     func(path)
     return True
 
+def create_long_path(path: Path) -> Path:
+    """Construct a path which will work on Windows if greater
+    than 260 characters. Resolves into a absolute path which
+    bypasses the typical MAX_PATH limitation:
+    https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation"""
+    if sys.platform.startswith("win32"):
+        path = path.resolve()
+        # for the API
+        if not str(path).startswith(r"\\?\"):
+            path = Path(r"\\?\" + str(path))
+    return path
 
 def rm_rf(path: Path) -> None:
     """Remove the path contents recursively, even if some elements
     are read-only.
     """
+    path = create_long_path(path)
     onerror = partial(on_rm_rf_error, start_path=path)
     shutil.rmtree(str(path), onerror=onerror)
 
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -99,6 +99,7 @@ def on_rm_rf_error(func, path: str, exc, *, start_path: Path) -> bool:
     func(path)
     return True
 
+
 def create_long_path(path: Path) -> Path:
     """Construct a path which will work on Windows if greater
     than 260 characters. Resolves into a absolute path which
@@ -110,6 +111,7 @@ def create_long_path(path: Path) -> Path:
         if not str(path).startswith("\\\\?\\"):
             path = Path("\\\\?\\" + str(path))
     return path
+
 
 def rm_rf(path: Path) -> None:
     """Remove the path contents recursively, even if some elements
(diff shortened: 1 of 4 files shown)


src/_pytest/pathlib.py
```diff
@@ -99,11 +99,18 @@ def chmod_rw(p: str) -> None:
     func(path)
     return True
 
+def create_long_path(path: Path) -> Path:
```
bluetech: This is a little obscure, so a docstring comment would be helpful IMO. It should explain what problem this solves (can just reference the issue if you want), and also how it solves it, since that's also a little obscure (does `\\?` magically makes it work with long paths? I'm not very familiar with these Windows paths).

testing/test_pathlib.py
```diff
@@ -89,3 +89,15 @@ def renamed_failed(*args):
     lock_path = get_lock_path(path)
     maybe_delete_a_numbered_dir(path)
     assert not lock_path.is_file()
+
+def test_long_path_during_cleanup(tmp_path):
+    """Ensure that deleting long path will be able to be deleted"""
```
bluetech:
```suggestion
    """Ensure that deleting long paths works (particularly on Windows)."""
```

testing/test_pathlib.py
```diff
@@ -89,3 +89,15 @@ def renamed_failed(*args):
...
+    path = tmp_path / ("a" * 200)
+    if sys.platform == "win32":
+        dirname = path.resolve()
```
bluetech: Can you add a comment here too?

testing/test_pathlib.py
```diff
@@ -89,3 +89,15 @@ def renamed_failed(*args):
...
+        dirname = "\\\\?\\" + str(path)
```
bluetech: A raw string would be nice here, like you have in the diff above.

src/_pytest/pathlib.py
torcolvin: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation is a pretty good documentation

testing/test_pathlib.py
```diff
@@ -89,3 +89,16 @@ def renamed_failed(*args):
     lock_path = get_lock_path(path)
     maybe_delete_a_numbered_dir(path)
     assert not lock_path.is_file()
+
+
+def test_long_path_during_cleanup(tmp_path):
+    """Ensure that deleting long path works (particularly on Windows (#6775))."""
+    path = tmp_path / ("a" * 200)
```
bluetech: If the MAX_PATH on Windows is 260, wouldn't it be better to use e.g. 300 here instead of 200? Or is the test about a path that *expands* to more than 260? If so, should there be an assert that `len(path) > 260`, as a precondition?

testing/test_pathlib.py
bluetech: I think I may have been confusing, I meant it'd be nicer to switch from `"\\\\?\\"` to `r"\\?\"` but you went the other way :smile: 

It's not very important though.

torcolvin: I forgot that you can't include a trailing blackslash in a string literal in python.

testing/test_pathlib.py
torcolvin: For the test, it can occur either way. In my failing case, I thought it was important to mention that the path I was using was <260 characters, but would expand to more with the garbage + uuid!

I think that creating a long filename is sufficient here because it covers both cases.

torcolvin: I forgot some cases that apply which is mainly that an individual path element, between slashes, can not be greater than 260, so I tried 300 but then switched to using the assert > 260.

src/_pytest/pathlib.py
```diff
@@ -100,10 +100,24 @@ def chmod_rw(p: str) -> None:
     return True
 
 
+def create_long_path(path: Path) -> Path:
```
bluetech: Looking again, I think the function name might be a little confusing, in that it might imply that the path is actually created.

From the link, the formal name for this seems to be "extended length path". So how about `ensure_extended_length_path`?

src/_pytest/pathlib.py
```diff
@@ -100,10 +100,24 @@ def chmod_rw(p: str) -> None:
...
+    """Construct a path which will work on Windows if greater
```
bluetech: I'd suggest to expand this a bit more:

```py
"""Get the extended-length version of a path (Windows).

On Windows, by default, the maximum length of a path (MAX_PATH) is 260
characters, and operations on paths longer than that fail. But it is possible
to overcome this by converting the path to "extended-length" form before
performing the operation:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation

On Windows, this function returns the extended-length absolute version of path.
On other platforms it returns path unchanged.
"""
```

src/_pytest/pathlib.py
```diff
@@ -100,10 +100,24 @@ def chmod_rw(p: str) -> None:
...
+    than 260 characters. Resolves into a absolute path which
+    bypasses the typical MAX_PATH limitation:
+    https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation"""
+    if sys.platform.startswith("win32"):
+        path = path.resolve()
+        # for the API
```
bluetech: This comment is not clear to me. I'd suggest to remove it.

bluetech:
Hi @torcolvin, I don't feel comfortable approving this myself because I am not fully familiar with the material and I suspect it has a high risk of causing breakage. And especially when an `rm -rf` is involved. I've added reading [this article](https://googleprojectzero.blogspot.com/2016/02/the-definitive-guide-on-win32-to-nt.html) to my TODO list, hopefully soon. Until then, maybe another reviewer who knows more about Windows feels comfortable to approve.

testing/test_pathlib.py
```diff
@@ -89,3 +90,25 @@ def renamed_failed(*args):
     lock_path = get_lock_path(path)
     maybe_delete_a_numbered_dir(path)
     assert not lock_path.is_file()
+
+
+def test_long_path_during_cleanup(tmp_path):
+    """Ensure that deleting long path works (particularly on Windows (#6775))."""
+    path = tmp_path / ("a" * 250)
```
symonk, line 97: what happens with you do "a" * 256?

torcolvin: There is a limitation of the API that you can't create with any element longer than 260 without using long path API, which i was trying not to do in a test. I had this in an earlier version of the PR.

nicoddemus:
Hi @torcolvin,

Really sorry about the delay here, completely missed the continuation of this PR.

First of all thanks a lot for the very clean and well written PR!

I tried `test_long_path_during_cleanup` on my machine (with long paths disabled) and it would pass against master, so I changed the strategy a bit, and while at that I noticed it was missing a call to `ensure_extended_length_path` in `maybe_delete_a_numbered_dir`, otherwise the lock file creation would fail and we would still see a warning. Also it was important to ensure the directory existed and then got properly deleted, as this was hiding this issue.

You can see my change in 163d36e51, let me know what you think.

torcolvin:
This looks good to me.

nicoddemus:
@bluetech want to take a second look?

nicoddemus:
> Overall this seems a bit risky. Really weird that each program needs to handle this on its own.

Indeed it is weird, but I don't think this is risky, it is a well documented part of the Windows API.

Btw, technically Python does this automatically, but users need to change their system to do so: https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation

So I think this is a good thing to have in pytest.

bluetech:
> Indeed it is weird, but I don't think this is risky, it is a well documented part of the Windows API.

My main concern was that due to the baroque nature of Windows paths, there are some edge cases where prepending `\\?\` might end up referring to something unexpected, which will get `rm -rf`'ed.

@torcolvin already fixed one case I noticed (the UNC thing), and I don't know of others. But since I don't know all of the details of Windows paths, I didn't feel comfortable approving it myself.

nicoddemus:
> My main concern was that due to the baroque nature of Windows paths, there are some edge cases where prepending \\?\ might end up referring to something unexpected, which will get rm -rf'ed.

That's understandable; but the same API states that it should always be an absolute path (which we do with `.resolve()`), to avoid these problems. So as far as I can see the code seems correct.

> But since I don't know all of the details of Windows paths, I didn't feel comfortable approving it myself.

No problem, that's fine. But do you think we should delay this PR further?

bluetech:
> No problem, that's fine. But do you think we should delay this PR further?

No, I removed my pending review now, forgot it would block from merging.

nicoddemus:
Thanks @bluetech!

nicoddemus:
Backport: https://github.com/pytest-dev/pytest/pull/7298

PR #6758, opened 2020-02-18
Use a hack to make typing of pytest.fail.Exception & co work

bluetech:
Mypy currently is unable to handle assigning attributes on function: https://github.com/python/mypy/issues/2087.
pytest uses this for the outcome exceptions -- `pytest.fail.Exception`, `pytest.exit.Exception` etc, and this is the canonical name by which they are referred.

Initially we started working around this with type: ignores, and later by switching e.g. `pytest.fail.Exception` with the direct exception `Failed`. But this causes a lot of churn and is not as nice. And I also found that some code relies on it, in skipping.py:

```py
    def pytest_configure(config):
        if config.option.runxfail:
            # yay a hack
            import pytest

            old = pytest.xfail
            config._cleanup.append(lambda: setattr(pytest, "xfail", old))

            def nop(*args, **kwargs):
                pass

            nop.Exception = xfail.Exception
            setattr(pytest, "xfail", nop)
        ...
```

So it seems better to support it. Use a hack to make it work. The rest of the commit rolls back all of the workarounds we added up to now.

`pytest.raises.Exception` also exists, but it's not used much so I kept it as-is for now.

Hopefully in the future mypy supports this and this ugliness can be removed.

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -25,7 +25,7 @@ from _pytest.config import ExitCode
 from _pytest.config import hookimpl
 from _pytest.config import UsageError
 from _pytest.fixtures import FixtureManager
-from _pytest.outcomes import Exit
+from _pytest.outcomes import exit
 from _pytest.reports import CollectReport
 from _pytest.runner import collect_one_node
 from _pytest.runner import SetupState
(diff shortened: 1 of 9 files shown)


bluetech:
Thanks @RonnyPfannschmidt.

I merged even though Travis failed, it seems unrelated.

PR #6759, opened 2020-02-19
Set `Session.shouldfail` from `Session.Failed` exception

blueyed:
TODO:

- [ ] adjust tests (different reporting than on my fork probably)

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -193,7 +193,11 @@ def wrap_session(
         except UsageError:
             session.exitstatus = ExitCode.USAGE_ERROR
             raise
-        except Failed:
+        except Failed as exc:
+            if not session.shouldfail:
+                # Typically Failed is raised when session.shouldfail has been
+                # set.  This sets it for when the exception was raised only.
+                session.shouldfail = exc.args[0]
             session.exitstatus = ExitCode.TESTS_FAILED
         except (KeyboardInterrupt, Exit):
             excinfo = _pytest._code.ExceptionInfo.from_current()
--- a/testing/test_main.py
+++ b/testing/test_main.py
@@ -75,3 +75,30 @@ def test_wrap_session_exit_sessionfinish(
         assert result.ret == ExitCode.NO_TESTS_COLLECTED
     assert result.stdout.lines[-1] == "collected 0 items"
     assert result.stderr.lines == ["Exit: exit_pytest_sessionfinish"]
+
+
+def test_session_shouldfail_from_failed(testdir: Testdir) -> None:
+    testdir.makeconftest(
+        """
+        import pytest
+        def pytest_runtestloop(session):
+            raise session.Failed("session_failed")
+    """
+    )
+    result = testdir.runpytest()
+    assert result.ret == 1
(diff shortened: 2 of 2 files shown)


bluetech:
Is there a particular reason to raise `session.Failed` directly and not however it is raised currently?

blueyed:
> Is there a particular reason to raise `session.Failed` directly and not however it is raised currently?

It's not necessarily raised currently, but `session.shouldfail` gets set (and then checked after hooks).

nicoddemus:
The current interactions of `shouldfail/Failed` and `shouldstop/Interrupted` are:

https://github.com/pytest-dev/pytest/blob/5819536f00ace91660c95ef7d65ff1c425a7b5b8/src/_pytest/main.py#L426-L431

https://github.com/pytest-dev/pytest/blob/5819536f00ace91660c95ef7d65ff1c425a7b5b8/src/_pytest/main.py#L273-L277

So as it stands this is not really *needed*, unless we want plugins to use `Failed/Interrupted` themselves. Can you please describe your use case/intentions with the change @blueyed?

blueyed:
@nicoddemus 
And where/how does `session.shouldfail` get set / get caught?

My use case is throwing it from a plugin, yes.
I find it more useful / idiomatic than setting `session.shouldfail` (and returning/breaking etc).

The outcome is the same: pytest catches it, but since it is not set as `session.shouldfail` it does not get displayed with the terminal plugin / is not available as a message to other plugins.

nicoddemus:
> @nicoddemus
And where/how does session.shouldfail get set / get caught?

https://github.com/pytest-dev/pytest/blob/master/src/_pytest/main.py#L439

https://github.com/pytest-dev/pytest/blob/master/src/_pytest/stepwise.py#L88

`Failed`/`Interrupted` (which is a subclass of `KeyboardError`) both seem to be caught by `wrap_session` only (which makes sense as the top of the stack).

> The outcome is the same: pytest catches it, but since it is not set as session.shouldfail it does not get displayed with the terminal plugin / is not available as a message to other plugins.

I see, thanks.

Perhaps we should do the same for `Interrupted` though?

https://github.com/pytest-dev/pytest/blob/769ffc32bf32aab1800d9c58458285db11696527/src/_pytest/main.py#L197-L198

Becomes:

```python
        except (KeyboardInterrupt, exit.Exception) as exc:
            if not session.shouldstop:
                session.shouldfail = exc.args[0]
            excinfo = _pytest._code.ExceptionInfo.from_current() 
```

src/_pytest/main.py
```diff
@@ -193,7 +193,11 @@ def wrap_session(
         except UsageError:
             session.exitstatus = ExitCode.USAGE_ERROR
             raise
-        except Failed:
+        except Failed as exc:
+            if not session.shouldfail:
+                # Typically Failed is raised when session.shouldfail has been
+                # set.  This sets it for when the exception was raised only.
+                session.shouldfail = exc.args[0]
```
nicoddemus: I usually use `str(exc)` instead of `exc.args[0]`, is that just style or is there a reason to prefer this way? (Just out of curiosity, it doesn't matter here of course).

PR #6760, opened 2020-02-19
Fix CLOSE_STDIN sentinel

blueyed:
This makes it / it's repr more meaningful in docs, and should have been
`object()` anyway.

TODO:

- [x] after https://github.com/pytest-dev/pytest/pull/6872

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -11,8 +11,11 @@ import time
 import traceback
 from fnmatch import fnmatch
 from io import StringIO
+from typing import AnyStr
 from typing import Callable
 from typing import Dict
+from typing import Generic
+from typing import IO
 from typing import Iterable
 from typing import List
 from typing import Optional
@@ -25,8 +28,11 @@ import py
 
 import pytest
 from _pytest._code import Source
+from _pytest.capture import CLOSE_STDIN
+from _pytest.capture import CloseStdinType
 from _pytest.capture import MultiCapture
 from _pytest.capture import SysCapture
+from _pytest.compat import overload
 from _pytest.compat import TYPE_CHECKING
 from _pytest.config import _PluggyPlugin
 from _pytest.config import ExitCode
@@ -513,7 +519,7 @@ class SysPathsSnapshot:
         sys.path[:], sys.meta_path[:] = self.__saved
 
 
-class Testdir:
+class Testdir(Generic[AnyStr]):
     """Temporary test directory with tools to test/run pytest itself.
 
     This is based on the ``tmpdir`` fixture but provides a number of methods
@@ -533,8 +539,8 @@ class Testdir:
 
     __test__ = False
 
-    class CLOSE_STDIN:
-        pass
+    CLOSE_STDIN = CLOSE_STDIN
+    """Sentinel to close stdin."""
 
     class TimeoutExpired(Exception):
         pass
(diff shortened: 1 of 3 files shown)


src/_pytest/pytester.py
```diff
@@ -534,7 +534,8 @@ class Testdir:
 
     __test__ = False
 
-    CLOSE_STDIN = object
+    class CLOSE_STDIN:
+        pass
```
bluetech:
```suggestion
        __slots__ = ()
```

bluetech: Hmm given that you use the class itself as the sentinel and not an instance of it, this suggestion doesn't matter much...

blueyed:
> maybe handrolling a version of `unittest.mock.sentinel`

I'm not opposed to it.

Given that this needs to be rebased anyway suggestions are welcome.
OTOH we're not using it that much currently, and therefore I think it's fine to go with it for now.

src/_pytest/pytester.py
graingert: an enum instance is usually pretty good

```python3
class CloseStdinType(enum.Enum):
    CLOST_STDIN = enum.auto()
```

it's great for type annotations too

graingert: and it's got that None/NoneType feel

graingert: @blueyed you can do stuff like:
```
def popen(self, ..., stdin=CLOSE_STDIN: Union[CloseStdinType,bytes,int]): -> ...
```

bluetech: @graingert Right, this is the best current way to do it in type-compatible way: https://github.com/python/typing/issues/689. Also what I opted for here: https://github.com/bluetech/pytest/commit/bcd19fac5b30882e106ef1e3dbad4bb7edb8b303#diff-4ec37b1ad5b7d41fdac56c907d08dde1R46-R52

blueyed: What's wrong with `stdin: Optional[Union[IO[str], "EncodedFile", "DontReadFromInput", Type[CLOSE_STDIN]]] = CLOSE_STDIN,`?
It might be not as strict really, but serves the purpose in general, and has less/no boilerplate.

bluetech: Nothing wrong really, just that using a `Type[...]` as a sentinel is odd/surprising. I'd go with @graingert's suggestion, which is the ["official"](https://www.python.org/dev/peps/pep-0484/#support-for-singleton-types-in-unions) way to do it.

That said if you prefer the Type that's fine.

src/_pytest/pytester.py
```diff
@@ -513,7 +519,7 @@ def restore(self) -> None:
         sys.path[:], sys.meta_path[:] = self.__saved
 
 
-class Testdir:
+class Testdir(Generic[AnyStr]):
```
bluetech: I don't see `AnyStr` mentioned elsewhere, why did you need this generic?

src/_pytest/pytester.py
```diff
@@ -1089,14 +1096,42 @@ def collect_by_name(
...
         return None
 
+    @overload
     def popen(
         self,
         cmdargs,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        stdin=CLOSE_STDIN,
+        stdout: Optional[Union[int, IO]] = subprocess.PIPE,
```
bluetech: Is it specifically `IO[str]` or `IO[bytes]`?

I see that typeshed punts on this: https://github.com/python/typeshed/blob/d8b081130d3e3a7fde1a49ac787d304f5c4c1966/stdlib/3/subprocess.pyi#L13-L27

src/_pytest/pytester.py
```diff
@@ -1089,14 +1096,42 @@ def collect_by_name(
...
+        stderr: Optional[Union[int, IO]] = subprocess.PIPE,
+        stdin: Optional[Union[CloseStdinType, bytes, int, IO]] = CLOSE_STDIN,
```
bluetech: Does `Literal[CLOSE_STDIN]` work? If yes, that'd be a little nicer. (I think it is planned to work, but I don't remember if it does already).

src/_pytest/pytester.py
blueyed: I've used `AnyStr` here initially (but forgot to clean it up).
I've figured `IO` is the same as `IO[Any]`.  Do you prefer the latter?
We basically have the `_FILE` from typeshed here: `Union[None, int, IO[Any]]`

bluetech: I'd leave it with the explicit `Any` (like you did), maybe can be filled with something precise in the future.

blueyed: It can be `str` or `bytes`.  typeshed binds it according to overloads.

nicoddemus:
Closing for now.

PR #6763, opened 2020-02-19
terminal: remove confusing comment about `_tw`

blueyed:
Ref: https://github.com/blueyed/pytest/commit/cf0cac3b7#r37373455

I think this makes it clear that `writer` is preferred (for internal use).
Not removing `_tw` for now unnecessarily.


nicoddemus:
Yep we failed to deprecate this. Should we create an issue for it and deprecate it officially?

PR #6765, opened 2020-02-19
Refactor Capture classes: move/rename CaptureIO classes

blueyed:
Pulled out of https://github.com/pytest-dev/pytest/pull/6671

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -11,10 +11,9 @@ from io import UnsupportedOperation
 from tempfile import TemporaryFile
 from typing import Generator
 from typing import Optional
+from typing import TextIO
 
 import pytest
-from _pytest.compat import CaptureAndPassthroughIO
-from _pytest.compat import CaptureIO
 from _pytest.compat import TYPE_CHECKING
 from _pytest.config import Config
 from _pytest.fixtures import FixtureRequest
@@ -318,6 +317,25 @@ def capfdbinary(request):
     capman = request.config.pluginmanager.getplugin("capturemanager")
     with capman._capturing_for_request(request) as fixture:
         yield fixture
+
+
+class CaptureIO(io.TextIOWrapper):
+    def __init__(self) -> None:
+        super().__init__(io.BytesIO(), encoding="UTF-8", newline="", write_through=True)
+
+    def getvalue(self) -> str:
+        assert isinstance(self.buffer, io.BytesIO)
+        return self.buffer.getvalue().decode("UTF-8")
+
+
+class TeeCaptureIO(CaptureIO):
+    def __init__(self, other: TextIO) -> None:
+        self._other = other
+        super().__init__()
+
+    def write(self, s: str) -> int:
+        super().write(s)
+        return self._other.write(s)
 
 
 class CaptureFixture:
@@ -673,7 +691,7 @@ class TeeSysCapture(SysCapture):
             if name == "stdin":
                 tmpfile = DontReadFromInput()
             else:
-                tmpfile = CaptureAndPassthroughIO(self._old)
+                tmpfile = TeeCaptureIO(self._old)
         self.tmpfile = tmpfile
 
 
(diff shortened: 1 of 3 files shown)


src/_pytest/capture.py
```diff
@@ -686,16 +704,13 @@ def snap(self):
...
-        name = patchsysdict[fd]
-        self._old = getattr(sys, name)
-        self.name = name
-        if tmpfile is None:
-            if name == "stdin":
-                tmpfile = DontReadFromInput()
-            else:
-                tmpfile = CaptureAndPassthroughIO(self._old)
-        self.tmpfile = tmpfile
+    def __init__(self, fd: int) -> None:
```
bluetech: I'm not sure this is really better, still duplicates some logic, and now it relies on `SysCaptureBinary` implementation instead of being independent.

Two possible alternatives, both are not ideal too but a bit better IMO:

1. Add a `passthrough: bool` argument to `SysCaptureBinary` and use `PassthroughCaptureIO` if true.
2. Add class attribute `CaptureIOClass: Type[CaptureIO] = CaptureIO` to `SysCaptureBinary` and use that in `__init__`, and override to `PassthroughCaptureIO` in `TeeSysCapture`.

src/_pytest/capture.py
```diff
@@ -323,6 +322,25 @@ def capfdbinary(request):
...
+class CaptureIO(io.TextIOWrapper):
+    def __init__(self) -> None:
+        super().__init__(io.BytesIO(), encoding="UTF-8", newline="", write_through=True)
+
+    def getvalue(self) -> str:
+        assert isinstance(self.buffer, io.BytesIO)
+        return self.buffer.getvalue().decode("UTF-8")
+
+
+class PassthroughCaptureIO(CaptureIO):
```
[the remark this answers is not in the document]
blueyed: `TeeCaptureIO` sounds good - I've mainly wanted to have `CaptureIO` at the end.

src/_pytest/capture.py
blueyed: Well, I think it is quite simple/good.
Keep in mind that the "old" is only used for an assert, and can be removed.
As for your 2nd suggestion: this is basically what passing in `tmpfile` does, isn't it?

bluetech: > Keep in mind that the "old" is only used for an assert, and can be removed.

You mean `_old` right?

>  As for your 2nd suggestion: this is basically what passing in tmpfile does, isn't it?

\<offtopic>

The `tmpfile` argument exists for by `FDCaptureBinary`. I was surprised initially to see that `FDCaptureBinary` uses `SysCapture` under the hood. Turns out `FDCaptureBinary` not only redirects the FDs to a tempfile, it also *additionally* patches the `sys` file objects to redirect to the tempfile (this is where the argument is needed). It seemed redundant to me -- if we already redirect the FDs, there is no need to patch `sys` since the data will reach the tempfile anyway. So I tried to remove the `sys` patching from `FDCaptureBinary`. It actually works fine, but the problem is the flushing - if we don't patch `sys`, then data that remains in the file object buffer doesn't end up captured, i.e. an explicit `flush()` is sometimes needed before the capture fixture can be examined.

If backward compatibility wasn't a concern, I think that this is actually the better behavior for `capfd`. The implicit `flush()`s prevent a test from being able to verify that data is actually present in the underlying file at a point of time.

But, we can't break it now... So maybe it makes sense to add a new "`fdonly`" mode, and make `fd` an alias for "`fd+sys`" mode (for clarity).

\</offtopic>

TL;DR: I think the `tmpfile` argument should go away eventually, so wouldn't want to add new uses for it...

bluetech:
So I think moving the Capture classes over to `capture.py` makes some obvious sense. The `Tee` change however makes things slightly worse IMO. I have some WIP work on capture which refactors it as well.

So to move forward on this PR, I'll remove the Tee changes, but keep the move.

PR #6768, opened 2020-02-19
capture: re-order classes

blueyed:
This better reflects the inheritance / smartness with regard to raw or
encoded.

- FDCaptureBinary
- FDCapture
- SysCaptureBinary
- SysCapture
- TeeSysCapture

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -620,9 +620,9 @@ class FDCapture(FDCaptureBinary):
         return res
 
 
-class SysCapture:
+class SysCaptureBinary:
 
-    EMPTY_BUFFER = str()
+    EMPTY_BUFFER = b""
     _state = None
 
     def __init__(self, fd, tmpfile=None):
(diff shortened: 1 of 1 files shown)


bluetech:
Side note: it seems like `SysCapture.snap` returns `str`, while `SysCaptureBinary` returns `bytes` - is this correct? If so, it's a little unfortunate that `SysCapture` inherits from `SysCaptureBinary`, because it breaks [Liskov Substitution](https://en.wikipedia.org/wiki/Liskov_substitution_principle) which means the subtyping is unsound. In particular,

```py
# Allowed because SysCapture is a SysCaptureBinary.
cap: SysCaptureBinary = SysCapture(...)

snap = cap.snap()
reveal_type(snap) # -> bytes
assert isinstance(snap, bytes) # -> fails!
```

Mypy would also complain about this once typing is added. So it would be better if they are not subclasses. Possible alternatives are to use composition, or to use a common *generic* base class.

PR #6774, opened 2020-02-20
cacheprovider: fix typing with Path(py.path.local)

blueyed:
Fixes:

> Argument 1 to "Path" has incompatible type "Union[local, Any]";
> expected "Union[str, _PathLike[str]]"  [arg-type]

Ref: https://github.com/pytest-dev/py/pull/232#pullrequestreview-349549522

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -181,13 +181,13 @@ class LFPluginCollWrapper:
             # Sort any lf-paths to the beginning.
             lf_paths = self.lfplugin._last_failed_paths
             res.result = sorted(
-                res.result, key=lambda x: 0 if Path(x.fspath) in lf_paths else 1,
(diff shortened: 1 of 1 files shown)


bluetech:
Is it possible you are not using the latest version of the `py` PR? The latest has

```py
_PathType = Union[bytes, Text, _PathLike[Any], local]

class local:
     def __init__(self, path: _PathType = ..., expanduser: bool = ...) -> None: ...
```

so this error should not happen.

blueyed:
Oh, great - confirmed.
(I was using d530227 still)
Sorry for the noise.

blueyed:
@bluetech 
No, it does not work - the issue is with `py.path.local` put into `Path()`.

Sorry for the extra noise, I've only tried `tox -e mypy` on master, which does not use the stubs.. ;)

PR #6775, opened 2020-02-20
tests: harden some UsageError tests (matching the error msg)

blueyed:


testing/test_config.py
```diff
@@ -408,11 +409,12 @@ def pytest_addoption(parser):
 
     def test_confcutdir_check_isdir(self, testdir):
         """Give an error if --confcutdir is not a valid directory (#2078)"""
-        with pytest.raises(pytest.UsageError):
+        exp_match = "^--confcutdir must be a directory, given: "
```
bluetech: TIOLI: I like to always use `r'...'` for regexes, even if not needed, as an indication "this is a regex pattern, not a string pattern".

PR #6776, opened 2020-02-20
Improve doc/typing for `ExceptionInfo.match`

blueyed:


blueyed:
I also think the message could be improved:
```diff
diff --git i/src/_pytest/_code/code.py w/src/_pytest/_code/code.py
index 6f2cd3480..fceb3ac8b 100644
--- i/src/_pytest/_code/code.py
+++ w/src/_pytest/_code/code.py
@@ -643,7 +643,7 @@ def match(self, regexp: "Union[str, Pattern]") -> "Literal[True]":
         """
         __tracebackhide__ = True
         if not re.search(regexp, str(self.value)):
-            assert 0, "Pattern {!r} not found in {!r}".format(regexp, str(self.value))
+            assert 0, "Pattern {!r} does not match {!r}".format(regexp, str(self.value))
         return True
```

src/_pytest/_code/code.py
```diff
@@ -633,13 +633,13 @@ def getrepr(
...
+    def match(self, regexp: "Union[str, Pattern]") -> "Literal[True]":
         """
-        Check whether the regular expression 'regexp' is found in the string
-        representation of the exception using ``re.search``. If it matches
-        then True is returned (so that it is possible to write
-        ``assert excinfo.match()``). If it doesn't match an AssertionError is
-        raised.
+        Check whether the regular expression `regexp` matches the string
+        representation of the exception using :func:`python:re.search`.
+        If it matches `True` is returned (so that it is possible to write
```
bluetech: Is there a  reason to do `assert excinfo.match()` as opposed to just `excinfo.match()`? If not, I would just omit the comment in the parenthesis (or move it to a code comment) so it doesn't sound like an endorsement.

src/_pytest/_code/code.py
```diff
@@ -633,13 +633,13 @@ def getrepr(
...
+        :code:`assert excinfo.match()`).
+        If it doesn't match an ``AssertionError`` is raised.
         """
         __tracebackhide__ = True
         if not re.search(regexp, str(self.value)):
```
bluetech: Will the assertion message improve if the condition is moved to the assert? I mean

```py
assert re.search(regexp, str(self.value)), "Pattern {!r} not found in {!r}".format(regexp, str(self.value))
```

blueyed: Shouldn't make a difference.
What I've meant in the comment above is that "pattern does not match" reads better than "pattern not found in" (which sounds like a "needle in stack" check).

bluetech: If it doesn't make a difference I would change it anyway, seems to me `assert condition` is preferable to `if not condition: assert 0`.

PR #6778, opened 2020-02-20
Fix pytest_ignore_collect hooks: do not return False

blueyed:
It should only return `True` when something is to be ignored, not
`False` otherwise typically.

This caused e.g. bad interaction with the cacheprovider (before
https://github.com/pytest-dev/pytest/pull/6448).

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -198,7 +198,8 @@ def pytest_ignore_collect(path, config):
     This hook is consulted for all files and directories prior to calling
     more specific hooks.
 
-    Stops at first non-None result, see :ref:`firstresult`
+    Stops at first non-None result, see :ref:`firstresult`, i.e. you should
+    only return `False` if the file should never get ignored (by other hooks).
 
     :param path: a :py:class:`py.path.local` - the path to analyze
     :param _pytest.config.Config config: pytest config object
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -198,8 +198,7 @@ def pytest_ignore_collect(path, config):
     This hook is consulted for all files and directories prior to calling
     more specific hooks.
 
-    Stops at first non-None result, see :ref:`firstresult`, i.e. you should
-    only return `False` if the file should never get ignored (by other hooks).
+    Stops at first non-None result, see :ref:`firstresult`
 
     :param path: a :py:class:`py.path.local` - the path to analyze
     :param _pytest.config.Config config: pytest config object
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -33,6 +33,7 @@ from _pytest.runner import SetupState
 
 if TYPE_CHECKING:
     from typing import Type
+    from typing_extensions import Literal
 
     from _pytest.python import Package
 
@@ -295,7 +296,9 @@ def _in_venv(path):
     return any([fname.basename in activates for fname in bindir.listdir()])
 
 
-def pytest_ignore_collect(path, config):
+def pytest_ignore_collect(
+    path: py.path.local, config: Config
+) -> "Optional[Literal[True]]":
     ignore_paths = config._getconftest_pathlist("collect_ignore", path=path.dirpath())
     ignore_paths = ignore_paths or []
     excludeopt = config.getoption("ignore")
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -320,8 +320,6 @@ def pytest_ignore_collect(path, config):
     if not allow_in_venv and _in_venv(path):
         return True
 
-    return False
-
 
 def pytest_collection_modifyitems(items, config):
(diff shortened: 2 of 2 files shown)


nicoddemus:
Can you elaborate a bit? I ask because I don't see anywhere in the code where we make a distinction between `None` and `False` when calling `pytest_ignore_collect`; they are all in the form:

```python
if ihook.pytest_ignore_collect(path=path, config=self.config):
    ...
```

blueyed:
It's a `firstresult` hook (and `None` is not a result there then).

blueyed:
@nicoddemus please suggest how to improve the doc then even more - since it appears to be still unclear.

src/_pytest/main.py
```diff
@@ -320,8 +320,6 @@ def pytest_ignore_collect(path, config):
     if not allow_in_venv and _in_venv(path):
         return True
 
-    return False
```
bluetech: Can you keep a `return None`?

nicoddemus:
Ahh got it, the key point is the fact the hook is `firstresult`, sorry my bad. 👍

src/_pytest/main.py
nicoddemus: Good idea. 👍

blueyed:
> Ahh got it, the key point is the fact the hook is `firstresult`, sorry my bad.

Can this be improved in the doc then to make it clearer?

src/_pytest/hookspec.py
```diff
@@ -198,7 +198,8 @@ def pytest_ignore_collect(path, config):
     This hook is consulted for all files and directories prior to calling
     more specific hooks.
 
-    Stops at first non-None result, see :ref:`firstresult`
+    Stops at first non-None result, see :ref:`firstresult`, i.e. you should
```
nicoddemus:
```suggestion
    As any other :ref:`firstresult` hook, this stops at first non-`None` result. This hook
    can return `False` explicitly to prevent a file to be ignored by other hooks.
```

blueyed: I've pulled out the doc change.
There are more places using the same sentence, and this should be fixed across all of them, maybe adding and linking to an intermediate doc section then.

src/_pytest/main.py
```diff
@@ -295,7 +296,9 @@ def _in_venv(path):
     return any([fname.basename in activates for fname in bindir.listdir()])
 
 
-def pytest_ignore_collect(path, config):
+def pytest_ignore_collect(
+    path: py.path.local, config: Config
+) -> "Optional[Literal[True]]":
```
blueyed: @bluetech I assume using Literal[True] here makes sense now?  (although not in the hookspec itself)
(I've quickly checked your other typing PRs, but it was not included there, is it?)

nicoddemus: General question: is the hook *implementation* allowed to declare a return type that is different, but still a subset, of the return type declared by the *hookspec*?

Example:

```python
@hookspec
def pytest_ignore_collect(
    path: py.path.local, config: Config
) -> "Optional[bool]":
    ...

@hookimpl
def pytest_ignore_collect(
    path: py.path.local, config: Config
) -> "Optional[Literal[True]]":
    ...
```

blueyed: @nicoddemus 
Yes, this is allowed in general:
```python
from typing import Literal


class A:
    def f(self) -> bool:
        pass


class B(A):
    def f(self) -> Literal[True]:
        return True
```

(keep in mind though that mypy currently does not know that hookimpls are subsets of hookspecs (as with subclasses))

bluetech: @blueyed 

> @bluetech I assume using Literal[True] here makes sense now? (although not in the hookspec itself)

There are arguments for and against.

For: the type provides a quick indication that this impl either only returns `None` or `True`. Can gain useful semantic information just from the type, namely that it doesn't return `False`.

Against: If sometime the circumstances change, and the function wants to start returning `False`, that's perfectly legit logically, but the type needs to change. This suggests the type is over restrictive.

Against: Might confuse the person who wants to make the change, if they don't realize they are *allowed* to extend the type. Might confuse copy/pasters. (Note: the argument is weakened by the fact hookimpls are already allowed to omit arguments they don't use, so might not match the hookspec already).

My inclination is to use the more general type and match the hookspec, but I can see both ways, and am probably not too consistent myself. So I'm fine with whatever you decide.

blueyed: @bluetech 
Thanks for elaborating on it again.  I think the benefit of gaining useful semantic information just from the type is useful here in general.

Will wait for any other feedback, and otherwise merge it as is.

nicoddemus: > Yes, this is allowed in general

Yes thanks, that's exactly what I meant to ask (not about hooks/impls, but in general).

> Will wait for any other feedback, and otherwise merge it as is.

I'm fine with what you guys decide, and the current form. 👍

PR #6781, opened 2020-02-20
tests: runner: improve/ignore coverage

blueyed:
TODO:

- [x] check codecov report


testing/test_runner.py
```diff
@@ -399,7 +399,7 @@ def test_func():
         except pytest.exit.Exception:
             pass
         else:
```
blueyed: Maybe `  # pragma: no cover` also here.

PR #6788, opened 2020-02-21
capture: factor out _get_multicapture

blueyed:
Ref: https://github.com/pytest-dev/pytest/pull/6671#issuecomment-588408992

TODO:

- [x] enable typing checks for test (after https://github.com/pytest-dev/pytest/pull/6753)

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -66,6 +66,18 @@ def pytest_load_initial_conftests(early_config: Config):
         sys.stderr.write(err)
 
 
+def _get_multicapture(method: str) -> "MultiCapture":
+    if method == "fd":
+        return MultiCapture(out=True, err=True, Capture=FDCapture)
+    elif method == "sys":
+        return MultiCapture(out=True, err=True, Capture=SysCapture)
+    elif method == "no":
+        return MultiCapture(out=False, err=False, in_=False)
+    elif method == "tee-sys":
+        return MultiCapture(out=True, err=True, in_=False, Capture=TeeSysCapture)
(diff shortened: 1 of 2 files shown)


src/_pytest/capture.py
```diff
@@ -66,6 +66,18 @@ def pytest_load_initial_conftests(early_config: Config):
         sys.stderr.write(err)
 
 
+def _get_multicapture(method: str) -> "MultiCapture":
```
bluetech: TIOLI: Use `Literal["fd", "sys", "no", "tee-sys"]` instead of `str`.

blueyed: Well, here I took your advice that it should not be used when the API (even if internal IIUC) is going to change (when adding new methods).
But I would prefer the Literal also, I'm just confused.. :)

bluetech: Hehe :) Well up to now we discussed return types, but this is an input type. Input types should be as restrictive as possible to rule out illegal inputs.

This is assuming the function is not a *validating* function (whose *purpose* is to take out general inputs and check them). In this case the input is already validated by the `--capture` arg parsing (which has `choices=["fd", "sys", "no", "tee-sys"]`), so any other kind of input is a not supposed to reach here.

If you choose to make this a `Literal`, it would make sense to also change `CaptureManager(method=...)` to have this type already.

blueyed: Ok.

But it might become a validating / more flexible method (https://github.com/pytest-dev/pytest/pull/6671/files#diff-e4269384a671a462a4824e8b06d4644bR97-R121), where `choices` then however would need to be factored out likely.  But until then I think it makes sense to use Literal here already.

PR #6795, opened 2020-02-22
Improve UsageError with invalid `-o` style

blueyed:
This started from fixing the test, where `"xdist_strict True"` was used
as a single argument, although you typically would see `["xdist_strict",
"True"]`.

Improves the error message to mention the option that caused the error.

### DIFF
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -1058,8 +1058,12 @@ class TestOverrideIniArgs:
             xdist_strict=False
         """
         )
-        result = testdir.runpytest("--override-ini", "xdist_strict True", "-s")
-        result.stderr.fnmatch_lines(["*ERROR* *expects option=value*"])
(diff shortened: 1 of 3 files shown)


testing/test_config.py
```diff
@@ -1058,8 +1058,12 @@ def test_override_ini_usage_error_bad_style(self, testdir):
             xdist_strict=False
         """
         )
-        result = testdir.runpytest("--override-ini", "xdist_strict True", "-s")
```
bluetech: Maybe keep this test also? Doesn't really hurt to test both cases.

blueyed: But it is also unnecessary - this just removes the confusion from the test.

It tested `pytest -o "xdist_strict True"` before, but was meant to test `pytest -o xdist_strict True` really.
In both cases there's no "=" in the arg, which is what needs to be tested.

PR #6800, opened 2020-02-23
Group terminal warning summary by "when"

blueyed:
TODO:

- [x] changelog

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -226,12 +226,15 @@ class WarningReport:
     Simple structure to hold warnings information captured by ``pytest_warning_captured``.
 
     :ivar str message: user friendly message about the warning
+    :ivar str when: when the warning was emitted, e.g. "config", "collect", or
+      "runtest" (see :func:`_pytest.hookspec.pytest_warning_captured`)
     :ivar str|None nodeid: node id that generated the warning (see ``get_location``).
     :ivar tuple|py.path.local fslocation:
         file system location of the source of the warning (see ``get_location``).
     """
 
     message = attr.ib(type=str)
+    when = attr.ib(type=str)
     nodeid = attr.ib(type=Optional[str], default=None)
     fslocation = attr.ib(default=None)
     count_towards_summary = True
@@ -394,7 +397,7 @@ class TerminalReporter:
             self.write_line("INTERNALERROR> " + line)
         return 1
 
-    def pytest_warning_captured(self, warning_message, item):
+    def pytest_warning_captured(self, when: str, warning_message, item) -> None:
         # from _pytest.nodes import get_fslocation_from_item
         from _pytest.warnings import warning_record_to_str
 
@@ -403,7 +406,7 @@ class TerminalReporter:
 
         nodeid = item.nodeid if item is not None else ""
         warning_report = WarningReport(
-            fslocation=fslocation, message=message, nodeid=nodeid
+            fslocation=fslocation, message=message, nodeid=nodeid, when=when,
         )
         self._add_stats("warnings", [warning_report])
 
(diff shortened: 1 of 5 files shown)


src/_pytest/terminal.py
```diff
@@ -803,31 +806,36 @@ def getreports(self, name):
...
 
-            final = hasattr(self, "_already_displayed_warnings")
-            if final:
-                warning_reports = all_warnings[self._already_displayed_warnings :]
-            else:
-                warning_reports = all_warnings
-            self._already_displayed_warnings = len(warning_reports)
-            if not warning_reports:
-                return
+        all_warnings = self.stats.get("warnings")  # type: Optional[List[WarningReport]]
```
Zac-HD: It looks like starting with `self.stats = defaultdict(list)` would allow us to simplify a lot of other code... the similar invocations all have `self.stats.get(..., [])` at the very least.  Happy to leave that to your judgement though.

src/_pytest/terminal.py
```diff
@@ -803,31 +806,36 @@ def getreports(self, name):
...
+        if not all_warnings:
+            return
 
-            reports_grouped_by_message = (
-                collections.OrderedDict()
-            )  # type: collections.OrderedDict[str, List[WarningReport]]
-            for wr in warning_reports:
-                reports_grouped_by_message.setdefault(wr.message, []).append(wr)
+        final = hasattr(self, "_already_displayed_warnings")
```
nicoddemus: I understand the code was already like this, but how about we initialize this with `0` during `__init__`? besides using dynamic attributes kind of a smell which has brought bugs in the past, this will also simplify the code because you won't need the `if` anymore:

```python
warning_reports = all_warnings[self._already_displayed_warnings :]
```

src/_pytest/terminal.py
```diff
@@ -803,31 +806,36 @@ def getreports(self, name):
...
+        else:
+            warning_reports = all_warnings
+        self._already_displayed_warnings = len(warning_reports)
+        if not warning_reports:
+            return
 
-            title = "warnings summary (final)" if final else "warnings summary"
+        grouped = (
+            collections.OrderedDict()
+        )  # type: collections.OrderedDict[str, List[WarningReport]]
```
nicoddemus: Should this be:

```suggestion
        )  # type: collections.OrderedDict[str, collections.OrderedDict[str, List[WarningReport]]]
```

Or is it OK to only type-hint the values?

src/_pytest/terminal.py
```diff
@@ -803,31 +806,36 @@ def getreports(self, name):
...
+        for wr in warning_reports:
+            if wr.when not in grouped:
+                grouped[wr.when] = collections.OrderedDict()
+            grouped[wr.when].setdefault(wr.message, []).append(wr)
+
+        for when, grouped_by_message in grouped.items():
+            title = (
+                "warnings summary (final)" if final else "warnings summary"
+            ) + " [{}]".format(when)
```
nicoddemus: I suspect most warnings come from runtest... how about only showing the suffix in case we have more than one category?

nicoddemus: Also I believe users might get confuse if they see only `warnings summary [config]` or `warnings summary [collect]` (while "runtest" will probably be OK).

How about adding some word to clarify, perhaps `warnings summary [captured during: config]`.

src/_pytest/terminal.py
blueyed: @nicoddemus 
Feel free to suggest an edit.

src/_pytest/terminal.py
blueyed: That is wrong here actually - fixed it in a followup in my fork already (so this needs an update).

src/_pytest/terminal.py
blueyed: @Zac-HD what do you suggest?

I do not thinkg "captured during: config" is much better than "config", is it?

Zac-HD: IMO it probably *is* better, because few users have a good mental model of pytest's operation and so being very descriptive is useful - most people don't read docs until they have a problem, so the message is our first chance to teach them. 

"Summary of warnings emitted during config stage" would be good for that, for example.

(note that I'm happy to merge anything better than the status quo though!)

blueyed: @Zac-HD 
Ok, understood.
OTOH headings should be short (to allow for more exclamation marks ;)).
The status quo for myself is that I have it like it is here already, and I am happy with it.
So I suggest somebody should pick it up from here, and maybe improve it with regard to what you and @nicoddemus are suggesting.  I'm not gonna do it myself.
Also given that there are conflicts it needs some work anyway (which I am also not doing anytime soon myself).
Therefore, either adopt it with improved headings, or just fix the conflicts, or close it.

Zac-HD: OK, let's merge as-is then once someone fixes merge conflicts. 

(I'll try to get to it eventually but have some much more urgent projects at the moment)

nicoddemus: > Therefore, either adopt it with improved headings, or just fix the conflicts, or close it.

To avoid having this lingering here, let's close it then. @Zac-HD feel free to re-open it if you want to get back to this eventually, and thanks!

PR #6802, opened 2020-02-23
Fix usage of pytester with doctests

blueyed:
Use `request.node.name` instead of `request.function.__name__`:
`request.function` is `None` with `DoctestItem`s.

TODO:

- [x] changelog (after approval/review)

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -545,7 +545,7 @@ class Testdir:
             WeakKeyDictionary()
         )  # type: WeakKeyDictionary[Module, List[Union[Item, Collector]]]
         if request.function:
-            name = request.function.__name__
+            name = request.function.__name__  # type: str
         else:
             name = request.node.name
         self._name = name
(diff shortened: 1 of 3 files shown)


src/_pytest/pytester.py
```diff
@@ -544,7 +544,11 @@ def __init__(self, request: FixtureRequest, tmpdir_factory: TempdirFactory) -> N
         self._mod_collections = (
             WeakKeyDictionary()
         )  # type: WeakKeyDictionary[Module, List[Union[Item, Collector]]]
-        name = request.function.__name__
+        if request.function:
+            name = request.function.__name__  # type: str
+        else:
+            name = request.node.name
+        self._name = name
```
blueyed: @RonnyPfannschmidt 
Would like to get your view on this: is a fallback to `request.node.name` sane, or is there something better?

blueyed:
Ping @RonnyPfannschmidt

PR #6804, opened 2020-02-24
Remove unused `setns`

blueyed:
Last usage of it was removed in ba17363d7.

TODO:

- [ ] after https://github.com/pytest-dev/pytest/pull/6803


blueyed:
@RonnyPfannschmidt can this be merged regardless of https://github.com/pytest-dev/pytest/pull/6803?  (not that `setns` will be required in the end still)

PR #6808, opened 2020-02-24
docs: generate index and add it to globaltoc sidebar

blueyed:
2020-02-24-160334_999x701_escrotum

### DIFF
--- a/doc/en/_templates/globaltoc.html
+++ b/doc/en/_templates/globaltoc.html
@@ -21,3 +21,7 @@
   <hr>
   {{ toc }}
 {%- endif %}
+
+<hr>
+<a href="{{ pathto('genindex') }}">Index</a>
+<hr>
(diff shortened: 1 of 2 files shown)


blueyed:
2020-02-24-160334_999x701_escrotum

blueyed:
@nicoddemus WDYT?

blueyed:
Might be worth checking why it was explicitly not enabled.

nicoddemus:
> @nicoddemus WDYT?

Looks good, thanks; I don't see myself using it (I use the search box), but I don't see how it can hurt either. 👍 

> Might be worth checking why it was explicitly not enabled.


Seems to have been added in e7c8fc7db9b02921889ddc8ec1b930ec95991661, so it might be related to PDF generation (although I don't think that's relevant anymore?).

PR #6815, opened 2020-02-25
Disallow Any on public interface types

nipunn1313:
This ensures that public interface methods must be typed
This should help with #3342

Per discussion there, ideally we shouldn't require full typing before we move forward, but this should give us a sense of the typing coverage which we have currently.

### DIFF
--- a/src/pytest/__init__.py
+++ b/src/pytest/__init__.py
@@ -46,8 +46,57 @@ from _pytest.warning_types import PytestUnhandledCoroutineWarning
 from _pytest.warning_types import PytestUnknownMarkWarning
 from _pytest.warning_types import PytestWarning
 
+# For mypy Any type checking purposes.
+# This file sets disallow_any_expr to ensure that the public API
+# does not have dynamic typing via Any. Manually using each public API
+# type as an expression to enforce this.
+__version__ = __version__
+register_assert_rewrite = register_assert_rewrite
+_setup_collect_fakemodule = _setup_collect_fakemodule
+cmdline = cmdline
+ExitCode = ExitCode
+# hookimpl = hookimpl
+# hookspec = hookspec
+main = main
+UsageError = UsageError
+__pytestPDB = __pytestPDB
+_fillfuncargs = _fillfuncargs
+fixture = fixture
+yield_fixture = yield_fixture
+freeze_includes = freeze_includes
+Session = Session
+mark = mark
+param = param
+Collector = Collector
+File = File
+Item = Item
+exit = exit
+fail = fail
+importorskip = importorskip
+skip = skip
+xfail = xfail
+Class = Class
+Function = Function
+Instance = Instance
+Module = Module
+Package = Package
+approx = approx
+raises = raises
+deprecated_call = deprecated_call
+warns = warns
+PytestAssertRewriteWarning = PytestAssertRewriteWarning
+PytestCacheWarning = PytestCacheWarning
+PytestCollectionWarning = PytestCollectionWarning
+PytestConfigWarning = PytestConfigWarning
+PytestDeprecationWarning = PytestDeprecationWarning
+PytestExperimentalApiWarning = PytestExperimentalApiWarning
+PytestUnhandledCoroutineWarning = PytestUnhandledCoroutineWarning
+PytestUnknownMarkWarning = PytestUnknownMarkWarning
+PytestWarning = PytestWarning
 
-set_trace = __pytestPDB.set_trace
+
+# Allow set_trace() to be typed with None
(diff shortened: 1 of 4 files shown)


src/pytest/__init__.py
```diff
@@ -46,8 +46,57 @@
...
+# This file sets disallow_any_expr to ensure that the public API
+# does not have dynamic typing via Any. Manually using each public API
+# type as an expression to enforce this.
+__version__ = __version__
+register_assert_rewrite = register_assert_rewrite
+_setup_collect_fakemodule = _setup_collect_fakemodule
+cmdline = cmdline
+ExitCode = ExitCode
+#hookimpl = hookimpl
+#hookspec = hookspec
```
nipunn1313: hookimpl and hookspec are the two which aren't very easy to type.
I believe we want to do something similar to #3342 for pluggy in order to make this possible.

src/pytest/__init__.py
```diff
@@ -46,8 +46,57 @@
 from _pytest.warning_types import PytestUnknownMarkWarning
 from _pytest.warning_types import PytestWarning
 
+# For mypy Any type checking purposes.
+# This file sets disallow_any_expr to ensure that the public API
+# does not have dynamic typing via Any. Manually using each public API
+# type as an expression to enforce this.
```
blueyed: Can this be moved to `__init__.pyi`?

src/pytest/__init__.py
blueyed: With regard to hookimpl/hookspec it would also be great to have some plugin or similar that would make mypy aware of hookimpls beloning to hookspecs of the same name.
(Maybe also / additionally separate types could be used, but that might be tedious)
(Just throwing in some food for thought)

src/pytest/__init__.py
nipunn1313: We can explicitly put types into an __init__.pyi, but that seems less desirable than having the code annotated inline. With the current setup, the code is annotated inline, and the 
 `disallow_any_expr` ensures that `Any` is not used as an annotation in this file.

`disallow_any_expr` was implemented into mypy here https://github.com/python/mypy/pull/3519
Notably, it requires that each item exists in the file as an expression. The import itself does not count as an expr here - so we need to have a series of `a = a` style expression assigments

src/pytest/__init__.py
nipunn1313: The hookimpl and hookspec decorator can be typed relatively easily - but in order to do so, we need to have type stubs or exported types from pluggy.

blueyed: I think it is more complex: how does mypy know that `@hookimpl(pytest_foo)` refers to `@hookspec(pytest_foo)`?
What I've meant here is that the former should be validated against the latter somehow.

nipunn1313: https://github.com/pytest-dev/pluggy/blob/master/src/pluggy/hooks.py#L22
HookSpec looks like it is a decorator which takes a handful of bools as arguments

I suppose the project_name is something which we could encode into the type system (`HookSpecMarker[_T]` where _T is the project name).

I'm not sure I follow how hookimpl and hookspec are tied together here.

src/pytest/__init__.py
Zac-HD: In that case I'd prefer to simply not use `disallow_any_expr` for now; it's a particularly un-pythonic change and we can easily revisit later.

blueyed: @nipunn1313 I've meant to move the "Manually using each public API type as an expression to enforce this." part of it.  It is only boilerplate, isn't it?

nipunn1313: `disallow_any_expr` is actually the most important piece - as it's the one that is guaranteeing that the public interface is typed.

The public interface is only the `__init__.py` file - so without this boilerplate - we're not actually adding much value.

src/pytest/__init__.py
```diff
@@ -46,8 +46,57 @@
...
+_setup_collect_fakemodule = _setup_collect_fakemodule
+cmdline = cmdline
+ExitCode = ExitCode
+# hookimpl = hookimpl
+# hookspec = hookspec
+main = main
+UsageError = UsageError
+__pytestPDB = __pytestPDB
+_fillfuncargs = _fillfuncargs
+fixture = fixture
```
bluetech: `fixture` and `yield_fixture` are currently untyped. Shouldn't it fail for them (for example)?

bluetech:
Hi @nipunn1313,

I think this is probably not the way we want to go right now, and I'm not sure it actually works as intended (see previous comment). It might make more sense once the typing is more thorough. I'll try to keep the issue updated as this progresses.

PR #6817, opened 2020-02-25
Keep explicit newlines with help texts

blueyed:
This makes a difference for e.g. pytest-xdist:

Before:
```
--dist=distmode       set mode for distributing tests to exec environments. …
                      available environment. loadscope: …
                      grouped by file to any available environment.
(default) no: …
```

After:
```
--dist=distmode       set mode for distributing tests to exec environments.
                      each: send each test to all available environments.
                      load: load balance by sending any pending test to …
                      …
                      (default) no: run tests inprocess, don't distribute.
```

This might also result in unexpected changes (hard wrapping), when line 
endings where used unintentionally, e.g. with:

``` help="""
  some long
  help text
  """
```

But the benefits from that are worth it, and it is easy to fix, as will be
done for the internal `assertmode` option.

TODO:

- [ ] changelog

### DIFF
--- a/changelog/6817.improvement.rst
+++ b/changelog/6817.improvement.rst
@@ -1,0 +1,2 @@
+Explicit new-lines in help texts of command-line options are preserved, allowing plugins better control
+of the help displayed to users.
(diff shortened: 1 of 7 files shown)


nicoddemus:
I like it. 👍

blueyed:
@nicoddemus please approve and (squash-)merge then, or close it.

nicoddemus:
I was waiting on the CHANGELOG entry I believe.

PR #6818, opened 2020-02-25
assertion: rewrite: only catch EnvironmentError

blueyed:
This was changed unintentionally in 45c4a8fb3 (pytest 5.3.0), but only
EnvironmentErrors might have `errno`.
Since that is not really guaranteed and it is good to have more
information this uses the string representation of the exc in the trace
message.

TODO:

- [ ] test


nicoddemus:
Thanks!

> Since that is not really guaranteed and it is good to have more
information this uses the string representation of the exc in the trace
message.

Yep in general the string representation already includes the error number anyway. 👍

blueyed:
Cool, thanks.
I'll change it for the Windows case then also - `OSError is EnvironmentError`, so it should be fine even in case atomicwrites does not match it.

blueyed:
I'm not going to add a test here anytime soon myself - so please (squash-)merge it if that's OK.

PR #6819, opened 2020-02-25
Remove reference to old konira plugin from docs

stefano-m:
The pytest-konira plugin has not seen an update since 2011, moreover
the "project description" on PyPI points to a dubious website that
does not actually correspond to the project and instead redirects to
advertising content.


blueyed:
Thank you!

PR #6823, opened 2020-02-26
Automate release by comment notifications

nicoddemus:
This PR introduces preparing releases by creating a new issue or commenting on existing issues.

The comment is in the format:

```
@pytestbot please prepare release from BRANCH
```

Where `BRANCH` can be `master` or one of the maintenance branches.

Here are two examples of issues that generating appropriate PRs:

* https://github.com/nicoddemus/pytest/issues/31
* https://github.com/nicoddemus/pytest/issues/32

### DIFF
--- a/.github/workflows/release-on-comment.yml
+++ b/.github/workflows/release-on-comment.yml
@@ -1,0 +1,26 @@
+# part of our release process, see `release-on-comment.py`
+name: release on comment
+
+on:
+  issues:
+    types: [opened, edited]
+  issue_comment:
+    types: [created, edited]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: "3.8"
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install --upgrade setuptools tox
+    - name: Prepare release
+      run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/release-on-comment.yml
```diff
@@ -0,0 +1,26 @@
+# part of our release process, see `release-on-comment.py`
+name: release on comment
+
+on:
+  issues:
+    types: [opened, edited]
+  issue_comment:
+    types: [created, edited]
```
blueyed: Does this mean the job is run for any issue/comment being created?
Can this be limited to e.g. only for when pytestbot is mentioned?

nicoddemus: Yep, managed to find a way so we skip right at the beginning, before even starting up a VM. 👍

.github/workflows/release-on-comment.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    if: (github.event.comment && contains(github.event.comment.body, '@pytestbot please')) || (github.event.issue && contains(github.event.issue.body, '@pytestbot please'))
```
nicoddemus: > Can this be limited to e.g. only for when pytestbot is mentioned?

👍

blueyed:
@nicoddemus 
I'm -1 on this: it is rather complex, adding overhead, and not really needed AFAICT.
But feel free to merge it / use it, given that you are doing releases mostly.. :+1:

nicoddemus:
Well, I like the automation, plus the previous method is still valid.

PR #6830, opened 2020-02-27
Move _collectfile to FSCollector

blueyed:
Previously this was implemented both on `Session` and `Package`, where
the extra code in `Package._collectfile` was not covered/used.

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -587,28 +587,6 @@ class Session(nodes.FSCollector):
                 return
             yield from m
 
-    def _collectfile(self, path, handle_dupes=True):
-        assert (
-            path.isfile()
-        ), "{!r} is not a file (isdir={!r}, exists={!r}, islink={!r})".format(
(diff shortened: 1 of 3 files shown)


bluetech:
The extra code if I see correctly is this:

```py
        if self.fspath == path:  # __init__.py
            return [self]
```

Did you investigate why it was added and why it's not used?

blueyed:
@bluetech good call.
I've added it in fa35f650b.  The test from there started working without the patch with a6ff5e6bfc117f9151c6506a5f6f4bba51a1676c also shortly after.

#4319 was WIP when it was merged, and a6ff5e6bfc117f9151c6506a5f6f4bba51a1676c is from its cleanup: #4369.  Codecov reported it there already, but it was missed / not looked at: https://codecov.io/gh/pytest-dev/pytest/compare/7ab3d818f03331ebf2765f6c4ea005b721b1c3fe...a6ff5e6bfc117f9151c6506a5f6f4bba51a1676c/changes

PR #6832, opened 2020-02-28
cacheprovider: --lf/--ff: use hookwrapper, respect --collect-only

blueyed:
Ref: https://github.com/pytest-dev/pytest-django/issues/819

Includes: 3abc13d27 (tests: simplify test_newfirst_usecase)

Also fixes `--lf` with regard to `-k`, where with `--lf -k "not foo"` with "foo" as known failure would not run anything.

TODO:

- [ ] fix tests for different repr
- [ ] changelog
- [ ] stepwise?

Ref: https://github.com/pytest-dev/pytest/commit/f75f7c192586fd856dad10673cd6d1f8cce03f75 (https://github.com/pytest-dev/pytest/pull/5145)

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -234,8 +234,8 @@ class LFPlugin:
         self.lastfailed = config.cache.get(
             "cache/lastfailed", {}
         )  # type: Dict[str, bool]
-        self._previously_failed_count = None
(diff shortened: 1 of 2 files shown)


blueyed:
Needs test adjuments (see TODO), closing for now.
Please let me know if you are interested in it in general.

PR #6842, opened 2020-02-28
docs: use doc/en/_build/html for html output

blueyed:
This will use the same directory as `make -C doc/en` then, and reflects
the default location that Sphinx uses.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -86,7 +86,7 @@ without using a local copy.  This can be convenient for small fixes.
 
         $ tox -e docs
 
-    The built documentation should be available in the ``doc/en/_build/``.
+    The built documentation should be available in ``doc/en/_build/html``.
 
     Where 'en' refers to the documentation language.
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.rst
```diff
@@ -86,7 +86,7 @@ without using a local copy.  This can be convenient for small fixes.
 
         $ tox -e docs
 
-    The built documentation should be available in the ``doc/en/_build/``.
+    The built documentation should be available in ``doc/en/_build/html``.
 
     Where 'en' refers to the documentation language.
```
blueyed: > Where 'en' refers to the documentation language.

I was about to remove this sentence, and then also thought / got reminded of removing the "en" subdir.
It was useful for when there where other languages ("jp" IIRC), but I do not think it is planned to have other languages again?
This might require to setup redirections on RTD though likely, so is nothing too trivial.

nicoddemus: Yeah this already occurred to me, but I also not thought it worth the effort (as I don't know how to do the redirect and would need to research it), so I never pursued it.

blueyed: https://readthedocs.org/dashboard/pytest/redirects/

PR #6845, opened 2020-02-29
When writing rewritten module's pyc, only ignore OSError

bluetech:
The trace error in particular assumes `e.errno` works. This means it
would have crashed for anything except `OSError` previously. So limit to
only these exceptions instead of blanket `BaseException`.


blueyed:
Duplicate of https://github.com/pytest-dev/pytest/pull/6818 - where I wanted to add a test to.

PR #6522, opened 2020-01-21
ci: pip: use installed version

blueyed:
Currently fails with:
```
    Uninstalling pip-19.3.1:
      Successfully uninstalled pip-19.3.1
Successfully installed pip-20.0
Traceback (most recent call last):
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/bin/pip", line 8, in <module>
    sys.exit(main())
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/site-packages/pip/_internal/cli/main.py", line 73, in main
    command = create_command(cmd_name, isolated=("--isolated" in cmd_args))
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/site-packages/pip/_internal/commands/__init__.py", line 96, in create_command
    module = importlib.import_module(module_path)
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/lib-python/3/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1003, in _gcd_import
  File "<frozen importlib._bootstrap>", line 980, in _find_and_load
  File "<frozen importlib._bootstrap>", line 964, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 674, in _load_unlocked
  File "<builtin>/frozen importlib._bootstrap_external", line 691, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/site-packages/pip/_internal/commands/install.py", line 24, in <module>
    from pip._internal.cli.req_command import RequirementCommand
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/site-packages/pip/_internal/cli/req_command.py", line 20, in <module>
    from pip._internal.operations.prepare import RequirementPreparer
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/site-packages/pip/_internal/operations/prepare.py", line 16, in <module>
    from pip._internal.distributions import (
  File "/opt/hostedtoolcache/PyPy/3.6.9/x64/site-packages/pip/_internal/distributions/__init__.py", line 1, in <module>
    from pip._internal.distributions.source import SourceDistribution
ImportError: cannot import name 'SourceDistribution'
```

Ref: https://github.com/pytest-dev/pytest/runs/400730404

PR #6851, opened 2020-03-01
ci: Travis: pin pip

blueyed:


blueyed:
Not merging it myself already since this was not wanted in https://github.com/pytest-dev/pytest/pull/6522 by 4 others.

blueyed:
Seems to be fixed by just another update probably.

nicoddemus:
I think the recommended practice is to update pip always, instead of pinning... why you think we should pin it? Please write a description when opening the PR.

nicoddemus:
As before, I believe this can be closed.

blueyed:
@nicoddemus 
It does not get upgraded since 58ef95ed4 anymore.

But since it usually fails within tox it is rather related to virtualenv anyway, and/or temporary hickups due to other unpinned dependencies (that last one was about zipp again).

PR #6853, opened 2020-03-03
docs: remove doc/en/links.inc

blueyed:
This is a bad anti-pattern from what I can tell.

The only benefit here would be that the link for "nose" could be managed centrally, but there is certainly not really a need for it as it stands now.


blueyed:
> (I assume you'll want to merge this yourself?)

Yes, thanks for the review! :)

PR #6854, opened 2020-03-03
Check invalid operations for -k

gdhameeja:
#6822  `KeywordMapping` returns a bool on lookup which when passed to eval
fail on certain operations such as index access and attribute access.
Check has been added to raise a UsageError if such operations are passed
to -k

### DIFF
--- a/src/_pytest/mark/legacy.py
+++ b/src/_pytest/mark/legacy.py
@@ -106,5 +106,5 @@ def matchkeyword(colitem, keywordexpr):
             )
     try:
         return eval(keywordexpr, {}, mapping)
-    except SyntaxError:
+    except Exception:
         raise UsageError("Wrong expression passed to '-k': {}".format(keywordexpr))


gdhameeja:
The change is very simple check to see if . or [ exists in the `keywordexpr`, a `UsageError` is raised.
Please let me know if there is a better way to do this. Also kindly guide me on how I can write a test for this and if I need to add this information to the documentation that attribute access and indexing is not allowed in -k inputs.

gdhameeja:
@nicoddemus I checked out the comment https://github.com/pytest-dev/pytest/issues/6822#issuecomment-594757523 and I understand why checking for '.' isn't such a great idea. I can reference a new PR for changing the expression so it passes eval.
As far as reporting is concerned, I did check for `TypeError` and raise the appropriate message. Can I just remove the checking for this PR and let the reporting be?

bluetech:
I think this is a good measure to fix the problem at least while we're still using `eval`.

I agree with @nicoddemus in the issue that we might as well do a full `except Exception` to catch all cases.

src/_pytest/mark/legacy.py
```diff
@@ -106,5 +106,5 @@ def matchkeyword(colitem, keywordexpr):
             )
     try:
         return eval(keywordexpr, {}, mapping)
-    except SyntaxError:
+    except:
```
bluetech: `except:` is probably *too* broad, it also catches `Ctrl-C` and such, which should propagate. I think `except Exception` is the most appropriate.

gdhameeja: Fixed.

bluetech: It's not fixed ATM -- maybe you forgot to push?

gdhameeja: yeah. my bad. fixed now.

gdhameeja:
> Anybody up for backporting this?

I'm up. Although I don't understand what you mean by backporting it and would need a little guidance.

bluetech:
> I'm up. Although I don't understand what you mean by backporting it and would need a little guidance.

Since this change just missed the `5.4.0` release, it is merged to master which means it will only be included in `5.5.0` which can be some time from now. "Backporting" means applying the same change to the `5.4.x` release series, which means it will be included also in the next `5.4.x` patch release (e.g. `5.4.2`).

In practice, you want to:

- `git fetch --all`
- `git checkout 5.4.x`
- `git checkout -B Fix-6822-backport`
- `git cherry-pick 599bf075dbf1c1afb1a59b521dca52155c157488`
- Submit this `Fix-6822-backport` branch as a PR against the `5.4.x` branch

We intend to automate this, but haven't yet...

PR #6859, opened 2020-03-04
Use attrs in KeywordMapping

nicoddemus:
Also added type hinting.

Noticed this while taking a look at https://github.com/pytest-dev/pytest/issues/6822.

### DIFF
--- a/src/_pytest/mark/legacy.py
+++ b/src/_pytest/mark/legacy.py
@@ -3,10 +3,15 @@ this is a place where we put datastructures used by legacy apis
 we hope to remove
 """
 import keyword
+from typing import Set
 
 import attr
 
+from _pytest.compat import TYPE_CHECKING
 from _pytest.config import UsageError
+
+if TYPE_CHECKING:
+    from _pytest.nodes import Item  # noqa: F401 (used in type string)
 
 
 @attr.s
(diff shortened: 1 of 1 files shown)


src/_pytest/mark/legacy.py
```diff
@@ -25,16 +32,16 @@ def __getitem__(self, name):
...
     Given a list of names, map any substring of one of these names to True.
     """
 
-    def __init__(self, names):
-        self._names = names
+    _names = attr.ib(type=Set[str])
 
     @classmethod
-    def from_item(cls, item):
+    def from_item(cls, item: Union["Node", "Function"]) -> "KeywordMapping":
```
bluetech: `Function` is a subtype of `Node`, so a union of them is redundant.

BTW, is this really `Node` (effectively `Union[Item, Collector]`) or just `Item`?

nicoddemus: This is because when using only `Node`, `mypy` complained that `Node` has no `function` attribute, that's why I have changed to use a `getattr` instead of `hasattr` in line 58 below.

(Got that advice from here: https://github.com/python/mypy/issues/1424)

Of course I'm open to suggestions! 😁

bluetech: I think the `getattr` is the fix, and the `Union` isn't needed and is redundant. I may be wrong though.

> BTW, is this really Node (effectively Union[Item, Collector]) or just Item?

I checked -- from what I can see, it is an `Item` specifically, not a `Node`. (tracing up, it comes from `pytest_collection_modifyitems` which only takes `Item` from my current understanding).

nicoddemus: Ahh indeed, thanks!

PR #6860, opened 2020-03-04
Add pytest-twisted to list of async def handling plugins

altendky:
https://github.com/pytest-dev/pytest-twisted/tree/v1.12#ensuredeferred

[pytest-twisted](https://github.com/pytest-dev/pytest-twisted) supports `async def` test functions and fixtures as well as `async def`/`yield` fixtures.

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -166,7 +166,8 @@ def async_warn(nodeid: str) -> None:
     )
     msg += "  - pytest-asyncio\n"
     msg += "  - pytest-trio\n"
-    msg += "  - pytest-tornasync"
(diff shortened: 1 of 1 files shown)


altendky:
I will note the caveat that `pytest-twisted` does expect the developer to decorate their tests and fixtures to explicitly notify `pytest-twisted` of them.  I don't know if that is relevant to this warning or not.

nicoddemus:
Thanks @altendky!

>  I don't know if that is relevant to this warning or not.

No, I don't think so. Either way the developer should install an async-related plugin, so it definitely is worth mentioning `pytest-twisted`. 👍

PR #6868, opened 2020-03-06
Simplify some exception handling code

bluetech:
Mostly avoid unnecessary usage of `sys.exc_info()`. Since Python3, exception objects themselves have all that's needed. They are also easier to type.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -862,19 +862,19 @@ class FixtureDef:
         self._finalizers.append(finalizer)
 
     def finish(self, request):
-        exceptions = []
+        exc = None
         try:
             while self._finalizers:
                 try:
                     func = self._finalizers.pop()
                     func()
-                except:  # noqa
-                    exceptions.append(sys.exc_info())
-            if exceptions:
-                _, val, tb = exceptions[0]
-                # Ensure to not keep frame references through traceback.
-                del exceptions
-                raise val.with_traceback(tb)
+                except BaseException as e:
+                    # XXX Only first exception will be seen by user,
+                    #     ideally all should be reported.
+                    if exc is None:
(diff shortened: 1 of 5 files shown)


src/_pytest/fixtures.py
```diff
@@ -862,19 +862,17 @@ def addfinalizer(self, finalizer):
...
-            if exceptions:
-                _, val, tb = exceptions[0]
-                # Ensure to not keep frame references through traceback.
-                del exceptions
-                raise val.with_traceback(tb)
+                except BaseException as e:
+                    if exc is None:
+                        exc = e
+            if exc:
+                raise exc
```
blueyed: Looks like the initial idea here was to collect them for reporting them all somehow?  (see also https://github.com/pytest-dev/pytest/pull/6868/files#diff-4d6d922c3b97a9eae157f8f653627b13R320-R321 - maybe add a TODO/XXX comment like there here also?)
Anyway, simplifying it like this is nice.

src/_pytest/runner.py
```diff
@@ -133,16 +133,14 @@ def pytest_runtest_call(item):
...
-        assert tb is not None
-        tb = tb.tb_next  # Skip *this* frame
-        sys.last_type = type
-        sys.last_value = value
-        sys.last_traceback = tb
-        del type, value, tb  # Get rid of these in this frame
-        raise
+        sys.last_type = type(e)
+        sys.last_value = e
+        # Skip *this* frame
```
blueyed: Maybe move the comment down, wrap the two statements for skipping , or leave the comment with the `= …tb_next` line?
Having it before the `assert` is a bit confusing.

bluetech:
Thanks @blueyed, I addressed your comments.

PR #6869, opened 2020-03-06
BACKPORT: Introduction of Config.invocation_args.

eb-fmezzabotta:
This change was first introduced in https://github.com/pytest-dev/pytest/pull/5564 but, logically, only in the `5.x.x` version. I'm back-porting it into the `4.6.x` branch since many people still use it in the meantime they migrate to Python 3.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -92,6 +92,7 @@ Evan Kepner
 Fabien Zarifian
 Fabio Zadrozny
 Feng Ma
+Fernando Mezzabotta Rey
 Florian Bruhin
 Floris Bruynooghe
 Gabriel Reis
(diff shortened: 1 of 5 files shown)


AUTHORS
```diff
@@ -92,6 +92,7 @@ Evan Kepner
 Fabien Zarifian
 Fabio Zadrozny
 Feng Ma
+Fernando Mezzabotta Rey
```
eb-fmezzabotta: I'm not sure if this the thing to do when back-porting things. If it's not, let me know! Just following the steps-to-contribute 🙂

eb-fmezzabotta:
ah! closing this. I'll create another PR.

Issue #6821, opened 2020-02-26
Travis-ci build fails trying to remove .pytest_cache


SeanXu1984:
- [x] a detailed description of the bug or suggestion

```
path = '/tmp/pytest-of-travis/garbage-7b2ebcac-468f-4ee5-9433-7967d6217af8'
ignore_errors = False, onerror = <functools.partial object at 0x7f5b3e727788>
    def rmtree(path, ignore_errors=False, onerror=None):
        """Recursively delete a directory tree.
    
        If ignore_errors is set, errors are ignored; otherwise, if onerror
        is set, it is called to handle the error with arguments (func,
        path, exc_info) where func is os.listdir, os.remove, or os.rmdir;
        path is the argument to that function that caused it to fail; and
        exc_info is a tuple returned by sys.exc_info().  If ignore_errors
        is false and onerror is None, an exception is raised.
    
        """
        if ignore_errors:
            def onerror(*args):
                pass
        elif onerror is None:
            def onerror(*args):
                raise
        try:
            if os.path.islink(path):
                # symlinks to directories are forbidden, see bug #1669
                raise OSError("Cannot call rmtree on a symbolic link")
        except OSError:
            onerror(os.path.islink, path, sys.exc_info())
            # can't continue even if onerror hook returns
            return
        names = []
        try:
            names = os.listdir(path)
        except os.error, err:
            onerror(os.listdir, path, sys.exc_info())
        for name in names:
            fullname = os.path.join(path, name)
            try:
                mode = os.lstat(fullname).st_mode
            except os.error:
                mode = 0
            if stat.S_ISDIR(mode):
>               rmtree(fullname, ignore_errors, onerror)
/usr/lib/python2.7/shutil.py:247: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python2.7/shutil.py:247: in rmtree
    rmtree(fullname, ignore_errors, onerror)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
path = '/tmp/pytest-of-travis/garbage-7b2ebcac-468f-4ee5-9433-7967d6217af8/test_cache_writefail_permissions0/.pytest_cache'
[...]

PR #6870, opened 2020-03-06
BACKPORT: Introduction of Config.invocation_args

fermezz:
This change was first introduced in https://github.com/pytest-dev/pytest/pull/5564 but, logically, only since the `5.1.x` version. I'm back-porting it into the `4.6.x` branch since many people still use it in the meantime they migrate to Python 3.

I was also able to test this backport to be working successfully locally.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -92,6 +92,7 @@ Evan Kepner
 Fabien Zarifian
 Fabio Zadrozny
 Feng Ma
+Fernando Mezzabotta Rey
 Florian Bruhin
 Floris Bruynooghe
 Gabriel Reis
--- a/changelog/6870.feature.rst
+++ b/changelog/6870.feature.rst
@@ -1,0 +1,1 @@
+New ``Config.invocation_args`` attribute containing the unchanged arguments passed to ``pytest.main()``.
--- a/doc/en/py27-py34-deprecation.rst
+++ b/doc/en/py27-py34-deprecation.rst
@@ -17,9 +17,9 @@ are available on PyPI.
 
 While pytest ``5.0`` will be the new mainstream and development version, until **January 2020**
 the pytest core team plans to make bug-fix releases of the pytest ``4.6`` series by
-back-porting patches to the ``4.6-maintenance`` branch that affect Python 2 users.
+back-porting patches to the ``4.6.x`` branch that affect Python 2 users.
 
-**After 2020**, the core team will no longer actively backport patches, but the ``4.6-maintenance``
+**After 2020**, the core team will no longer actively backport patches, but the ``4.6.x``
 branch will continue to exist so the community itself can contribute patches. The core team will
 be happy to accept those patches and make new ``4.6`` releases **until mid-2020**.
 
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -13,6 +13,7 @@ import sys
 import types
 import warnings
 
+import attr
 import py
 import six
 from packaging.version import Version
@@ -35,6 +36,7 @@ from _pytest.compat import lru_cache
 from _pytest.compat import safe_str
 from _pytest.outcomes import fail
 from _pytest.outcomes import Skipped
+from _pytest.pathlib import Path
 from _pytest.warning_types import PytestConfigWarning
 
 hookimpl = HookimplMarker("pytest")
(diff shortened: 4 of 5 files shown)


AUTHORS
```diff
@@ -92,6 +92,7 @@ Evan Kepner
 Fabien Zarifian
 Fabio Zadrozny
 Feng Ma
+Fernando Mezzabotta Rey
```
fermezz: I'm not sure if this the thing to do when back-porting things. If it's not, let me know! Just following the steps-to-contribute 🙂

doc/en/py27-py34-deprecation.rst
```diff
@@ -17,9 +17,9 @@ are available on PyPI.
 
 While pytest ``5.0`` will be the new mainstream and development version, until **January 2020**
 the pytest core team plans to make bug-fix releases of the pytest ``4.6`` series by
-back-porting patches to the ``4.6-maintenance`` branch that affect Python 2 users.
+back-porting patches to the ``4.6.x`` branch that affect Python 2 users.
```
fermezz: Since there is not an actual `4.6-maintenance` branch, I changed it for what I believe is correct.

eb-fmezzabotta: huh, this should be in master as well 🤔. I can create a PR for that too if I get confirmation that that's the right thing to do.

asottile:
imo this patch extends beyond "bugfix" and probably shouldn't be backported -- curious to hear others thoughts though

fermezz:
Yeah, the way I think about it is that this change, although feature, was specifically introduced to fix a bug in the interaction with `pytest-xdist`. So while technically it is a feature for `pytest`, it's a bugfix for the `pytest`-ecosystem.

That said, I'm open to suggestions. I can maybe find a workaround for my specific use case, but I'm probably no the only one still not being able to migrate to 5.1.

What do you think?

blueyed:
@fermezz 
FWIW I think it makes sense, given that it is a bugfix after all.

For the Travis config check https://github.com/blueyed/pytest/blob/my-4.6-maintenance/.travis.yml, where it still works.

fermezz:
Thanks @blueyed. I'll give it a shot at fixing that up, although I don't see a lot of differences beyond having fewer jobs to build :thinking:.

fermezz:
I have a separate PR to fix CI by migrating it to GitHub actions https://github.com/pytest-dev/pytest/pull/6884/files. I appreciate input there, especially since I'm going to need assistance on modifying the required checks. Also, see https://github.com/pytest-dev/pytest/issues/6821

AUTHORS
nicoddemus: It is definitely the right thing! 👍

doc/en/py27-py34-deprecation.rst
nicoddemus: No need to port to master, we have changed that significantly already: https://docs.pytest.org/en/latest/py27-py34-deprecation.html

nicoddemus:
> imo this patch extends beyond "bugfix" and probably shouldn't be backported -- curious to hear others thoughts though

While I agree that technically it is a (small) new feature, the main reason this was created was for a bug fix in xdist as @fermezz mentioned. Given that this allows to use a more recent pytest-xdist version which fixed a serious config problem, I think we can open an exception. 👍

changelog/6870.feature.rst
```diff
@@ -0,0 +1 @@
+New ``Config.invocation_args`` attribute containing the unchanged arguments passed to ``pytest.main()``.
```
nicoddemus: Let's add a remark here:

```suggestion
New ``Config.invocation_args`` attribute containing the unchanged arguments passed to ``pytest.main()``.

Remark: while this is technically a new feature and according to our `policy <https://docs.pytest.org/en/latest/py27-py34-deprecation.html#what-goes-into-4-6-x-releases>`_ it should not have been backported, we have opened an exception in this particular case because it fixes a serious interaction with ``pytest-xdist``, so it can also be considered a bugfix.
```

@asottile what do you think?

asottile: 👍 makes it much more clear that this is one-off and not a greenlight to resume features on 4.6

fermezz: Remark added, it makes a lot of sense.

fermezz:
I added a remark to be clear that this is a one time thing about backporting a feature. We should rebase this branch after https://github.com/pytest-dev/pytest/pull/6884 is merged so that we can have a successful travis run before merging it.

fermezz:
Thoughts about this one?

nicoddemus:
@fermezz sorry about the delay, we were resolving some internal issues.

Thanks a lot for the PR!

fermezz:
@nicoddemus I heard, sorry you had to figured that out, but I'm glad y'all are back! And thanks for assisting!!

changelog/6870.feature.rst
nicoddemus: Oh just noticed this was not added in the end.

No worries, I will update directly on #7199. 👍

fermezz: I committed it locally and never pushed 🤦. Sorry about that

nicoddemus: No problem 😁 

4.6.10 has been released btw! 👍

PR #5611, opened 2019-07-16
Remove language_version specification in black hooks

mxr:
Resolves #3840

- [x] Target the `master` branch for bug fixes, documentation updates and trivial changes.
- [ ] Target the `features` branch for new features and removals/deprecations.
- [ ] Include documentation when adding new features.
- [ ] Include new tests or update existing tests when applicable.

Unless your change is trivial or a small documentation fix (e.g.,  a typo or reword of a small section) please:

- [ ] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>.<TYPE>.rst`. See [changelog/README.rst](https://github.com/pytest-dev/pytest/blob/master/changelog/README.rst) for details.
- [ ] Add yourself to `AUTHORS` in alphabetical order;

PR #6872, opened 2020-03-07
pre-commit: black: --target-version py35

blueyed:
This is meant to avoid having trailing commas after `**kwargs`, which
 only works in Python 3.5.2+.

Ref: https://github.com/pytest-dev/pytest/commit/2e756d698, https://github.com/pytest-dev/pytest/pull/5611

Might make more sense to add config for it: https://github.com/blueyed/pytest/pull/275

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,3 +54,6 @@ template = "changelog/_template.rst"
   directory = "trivial"
   name = "Trivial/Internal Changes"
   showcontent = true
+
+[tool.black]
+target-version = ['py35']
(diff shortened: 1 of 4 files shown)


blueyed:
What about using a config instead?

> Might make more sense to add config for it: blueyed#275

blueyed:
I assume this will make it work better out of pre-commit also.

But likely I was just being "tricked" by having added trailing commas myself, and then black does not remove them (I guess).

bluetech:
Config file would be better, this way editors will be able to catch it as well.

PR #6873, opened 2020-03-07
doc/en/Makefile: use Sphinx's make mode

blueyed:


doc/en/Makefile
```diff
@@ -4,13 +4,20 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
```
bluetech:
```suggestion
SPHINXOPTS    ?=
```

doc/en/Makefile
```diff
@@ -4,13 +4,20 @@
...
 SPHINXBUILD   = sphinx-build
```
bluetech:
```suggestion
SPHINXBUILD   ?= sphinx-build
```

blueyed:
It was synced with/from Sphinx' own `Makefile`, will check/compare with a fresh one.

PR #6874, opened 2020-03-07
doc: use `envvar` directive for environment variables

blueyed:
This changes the link anchors in "reference.html", from e.g.
`reference.html#pytest-current-test` to
`reference.html#envvar-PYTEST_CURRENT_TEST`, but I think that is OK, and
not worth adding labels for the old anchors.

### DIFF
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -153,9 +153,9 @@ def pytest_runtest_teardown(item, nextitem):
 
 def _update_current_test_var(item, when):
     """
-    Update PYTEST_CURRENT_TEST to reflect the current item and stage.
+    Update :envvar:`PYTEST_CURRENT_TEST` to reflect the current item and stage.
 
-    If ``when`` is None, delete PYTEST_CURRENT_TEST from the environment.
+    If ``when`` is None, delete ``PYTEST_CURRENT_TEST`` from the environment.
     """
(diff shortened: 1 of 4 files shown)


src/_pytest/runner.py
```diff
@@ -153,9 +153,9 @@ def pytest_runtest_teardown(item, nextitem):
 
 def _update_current_test_var(item, when):
     """
-    Update PYTEST_CURRENT_TEST to reflect the current item and stage.
+    Update :envvar:`PYTEST_CURRENT_TEST` to reflect the current item and stage.
 
-    If ``when`` is None, delete PYTEST_CURRENT_TEST from the environment.
+    If ``when`` is None, delete ``PYTEST_CURRENT_TEST`` from the environment.
```
Zac-HD: `:envvar:` here too :wink:

blueyed: @Zac-HD 
Thanks for the review!
IIRC it is a matter of only linking the first occurrence in a paragraph/section, but not all of them.
Please (squash-)merge if that sounds OK to you.

Zac-HD: For Hypothesis we prefer to use the role syntax for every occurrence so that they're all styled and linked, but happy for you to merge now if you disagree.

blueyed: I do not really disagree, but rather take it as-is for now, thanks!

Issue #6871, opened 2020-03-06
capsysbinary fixture breaks print()


Feuermurmel:
It seems that the replacement for `sys.stdout` installed by the `capsysbinary` does not work correctly. See the exception in the captured output in the example run below.

The setup is incredibly simple. The test file I'm using has the following content:

```
def test_capture(capsysbinary):
    print('hello')
```

There is no pytest.ini or anything else I've locally configured that should influence pytest.

```
(venv) ~/Desktop/pytest-issue$ pip list
Package        Version
-------------- -------
attrs          19.3.0 
more-itertools 8.2.0  
packaging      20.3   
pip            19.2.3 
pluggy         0.13.1 
py             1.8.1  
pyparsing      2.4.6  
pytest         5.3.5  
setuptools     41.2.0 
six            1.14.0 
wcwidth        0.1.8  
WARNING: You are using pip version 19.2.3, however version 20.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
(venv) ~/Desktop/pytest-issue$ ls
test_capture.py  venv
(venv) ~/Desktop/pytest-issue$ cat test_capture.py 
def test_capture(capsysbinary):
    print('hello')
(venv) ~/Desktop/pytest-issue$ pytest 
============================= test session starts ==============================
platform darwin -- Python 3.8.2, pytest-5.3.5, py-1.8.1, pluggy-0.13.1
rootdir: /Users/michi/Desktop/pytest-issue
collected 1 item                                                               

test_capture.py                                                         [100%]

=================================== FAILURES ===================================
_________________________________ test_capture _________________________________

self = <contextlib._GeneratorContextManager object at 0x105da09d0>, type = None
value = None, traceback = None

    def __exit__(self, type, value, traceback):
        if type is None:
            try:
>               next(self.gen)
E               TypeError: write() argument must be str, not bytes

/opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/contextlib.py:120: TypeError
--------------------------- Captured stdout teardown ---------------------------
F
============================== 1 failed in 0.06s ===============================
```

I've also tried `sys.stdout.write()` with the same result. Is this a bug or am I doing something incredibly stupid?

blueyed:
[...]

Issue #555, opened 2014-08-01
AttributeError: '_io.FileIO' object has no attribute 'errors'


pytestbot:
Originally reported by: **Jason R. Coombs (BitBucket: [jaraco](http://bitbucket.org/jaraco), GitHub: [jaraco](http://github.com/jaraco))**

---

New in pytest 2.6.0, many setuptools tests are failing thus: https://travis-ci.org/jaraco/setuptools/jobs/31453603

In particular, the error occurs with this traceback:

```
c:\python\lib\distutils\log.py:44: in info
    self._log(INFO, msg, args)
c:\python\lib\distutils\log.py:30: in _log
    if stream.errors == 'strict':
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <_pytest.capture.EncodedFile object at 0x0000000005C87128>
name = 'errors'

    def __getattr__(self, name):
>       return getattr(self.buffer, name)
E       AttributeError: '_io.FileIO' object has no attribute 'errors'

c:\python\lib\site-packages\pytest-2.6.0-py3.4.egg\_pytest\capture.py:242: AttributeError
```

Using Python 2.x or pytest 2.5.x works around the issue.

---
- Bitbucket: https://bitbucket.org/pytest-dev/pytest/issue/555

PR #6875, opened 2020-03-07
Simplify away safe_text_dupfile/EncodedFile

bluetech:
I tried to understand what the `safe_text_dupfile()` function and `EncodedFile` class do. Outside tests, `EncodedFile` is only used by `safe_text_dupfile`, and `safe_text_dupfile` is only used by `FDCaptureBinary.__init__()`. I then started to eliminate always-true conditions based on the single call site, and in the end nothing was left.

The old code had some workarounds:

- Ensure `errors` property is set: https://github.com/pytest-dev/pytest/issues/555. It is set now, although now it will return `replace`, which is the actual method used also before, instead of `strict`. AFAIU, the issue just wanted it to be set to something, not `strict` specifically.

- Infinite recursion trying to pickle `EncodedFile` (f7282b84bd2e3ba). Not sure, I don't think this object should be pickle-able at all (and it isn't...).

- A bug in `logging.Handler.__repr__` with int `file.name`: https://github.com/pytest-dev/pytest/issues/2555
  Was fixed in Python>=3.7.4: https://bugs.python.org/issue36015.
  TODO: This workaround is probably still needed?

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -9,9 +9,7 @@ import os
 import sys
 from io import UnsupportedOperation
 from tempfile import TemporaryFile
-from typing import BinaryIO
 from typing import Generator
-from typing import Iterable
 from typing import Optional
 
 import pytest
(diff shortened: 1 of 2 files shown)


src/_pytest/capture.py
```diff
@@ -579,7 +526,7 @@ def _start(self):
 
     def snap(self):
         self.tmpfile.seek(0)
-        res = self.tmpfile.read()
+        res = self.tmpfile.buffer.read()
```
bluetech: This I think is a bug fix anyway?

src/_pytest/capture.py
```diff
@@ -621,10 +568,10 @@ class FDCapture(FDCaptureBinary):
     EMPTY_BUFFER = str()  # type: ignore
 
     def snap(self):
-        res = super().snap()
```
bluetech: Since `tmpfile` is now text-mode for reading also (not just for writing like the weird `EncodedFile` did), I switched to read directly. Also symmetric with the `Sys` variants which do the same.

bluetech:
Hmm, seems like `TemporaryFiles(errors)` was only added in py3.8. Will try to find a fix for that.

src/_pytest/capture.py
blueyed: Maybe fixes https://github.com/pytest-dev/pytest/issues/6871 ?
Might make sense to pull it out then already.

blueyed: But would also avoid the indirection via `__getattr__`, no?

bluetech:
Replaced by #6875.

PR #6882, opened 2020-03-09
Fix one issue for compat::getfuncargnames

SeanXu1984:
Fix so the self of an instancemethod will not be removed again in getfuncargnames because it is already removed from signature(function)

I will deal with changelog later.

### DIFF
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -177,7 +177,9 @@ def getfuncargnames(
     # it's passed as an unbound method or function, remove the first
     # parameter name.
     if is_method or (
-        cls and not isinstance(cls.__dict__.get(name, None), staticmethod)
+        cls
+        and (not hasattr(function, "__self__") or not function.__self__)  # type: ignore
+        and not isinstance(cls.__dict__.get(function.__name__, None), staticmethod)
     ):
         arg_names = arg_names[1:]
     # Remove any names that will be replaced with mocks.


bluetech:
Hi @SeanXu1984,

Your PR title says `[pytest-5.3.x]` but you target the `master` branch. If the problem exists in `master`, it should indeed be targeted at `master` first, and afterwards can be backported separately as needed. In this case `[pytest-5.3.x]` should be removed from this PR.

To understand the issue a little better, would you be able to add a test which demonstrates the issue? This will also prevent it from regressing.

SeanXu1984:
Hi, the issue description can be found from https://github.com/pytest-dev/pytest/issues/6750
It was found in pytest-4.6.x but I believe it should be a issue for pytest 4.6.x as well.

bluetech:
OK, so I've edited the title now as this PR should be the main one.

To proceed with this, please add a test case to the PR which fails before your fix and passes after. It is a bit difficult to assess otherwise.

nicoddemus:
Hi @SeanXu198,

First of all we would like to thank you for your time and effort on working on this, the pytest team deeply appreciates it.

We noticed it has been awhile since you have updated this PR, however. pytest is a high activity project, with many issues/PRs being opened daily, so it is hard for us maintainers to track which PRs are ready for merging, for review, or need more attention.

So for those reasons we think it is best to close the PR for now, but with the only intention to cleanup our queue, it is by no means a rejection of your changes. We still encourage you to re-open this PR (it is just a click of a button away) when you are ready to get back to it.

Again we appreciate your time for working on this, and hope you might get back to this at a later time!

Cheers,

PR #6884, opened 2020-03-09
Fix travis config and coverage report.

eb-fmezzabotta:
This PR solves https://github.com/pytest-dev/pytest/issues/6821 by fixing the Travis CI which was borked, and adding GitHub actions to test some of the other environments and operating systems not being tested in Travis anymore. Also makes sure coverage is 💚.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -119,6 +119,7 @@ changedir = testing/freeze
 # Disable PEP 517 with pip, which does not work with PyInstaller currently.
 deps =
     pyinstaller
+    setuptools < 45.0.0
 commands =
     {envpython} create_executable.py
     {envpython} tox_run.py
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,191 @@
+name: main
+
+on:
+  push:
+    branches:
+      - 4.6.x
+    tags:
+      - "*"
+
+  pull_request:
+    branches:
+      - 4.6.x
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [
+          "windows-py27",
+          "windows-py35",
+          "windows-py36",
+          "windows-py37",
+          "windows-py37-pluggy",
+          "windows-py38",
+
+          "ubuntu-py27-pluggy",
+          "ubuntu-py27-nobyte",
+          "ubuntu-py37",
+          "ubuntu-py37-pluggy",
+          "ubuntu-py37-pexpect-py37-twisted",
+          "ubuntu-py37-freeze",
+          "ubuntu-pypy",
+          "ubuntu-pypy3",
+
+          "macos-py27",
+          "macos-py38",
+
+        ]
+
+        include:
+          # Windows jobs
+          - name: "windows-py27"
+            python: "2.7"
+            os: windows-latest
+            tox_env: "py27-xdist"
+            use_coverage: true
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -119,6 +119,7 @@ changedir = testing/freeze
 # Disable PEP 517 with pip, which does not work with PyInstaller currently.
 deps =
     pyinstaller
+    setuptools < 45.0.0
```
fermezz: Needed to make the installation of `pyinstaller` successful. Check out https://github.com/pypa/setuptools/issues/1963

fermezz:
I'll appreciate input on the coverage checks since I'm having a hard time understanding if those numbers look correct or not.

nicoddemus:
Thanks a lot @eb-fmezzabotta! 

> I'll appreciate input on the coverage checks since I'm having a hard time understanding if those numbers look correct or not.

The coverage overall was tweaked a lot by @blueyed, with specific tweaks like running some envs with `-v` to cover specific parts of the code for example.

Did you try to instead copy over the version from @blueyed fork for Travis as suggested in https://github.com/pytest-dev/pytest/pull/6870#issuecomment-596145333? That's probably easier and will have complete coverage.

nicoddemus:
@eb-fmezzabotta I've ported @blueyed's changes over at https://github.com/pytest-dev/pytest/pull/6889, and it seems to be working. I think we should add that configuration over your changes here because I believe we will fix coverage. 👍 

(I would do that myself unfortunately I can't push to your branch)

fermezz:
I will do that tomorrow, @nicoddemus. Sorry, forgot to allow you guys to push.

fermezz:
@nicoddemus alright, updated! I had seen that recommendation, but seemed weird to me to remove so many test-envs so I went with the other one, mentioned in https://github.com/pytest-dev/pytest/issues/6821#issuecomment-596145435.

fermezz:
Coverage report seems somewhat odd after fixing the travis config and actually reporting coverage, but seems to me that it's because of having had CI for some time :thinking:.

nicoddemus:
Hmm problem is now that we don't have Windows coverage anymore. Before moving officially to GitHub actions, we used AppVeyor, but since then have dropped its configuration and disabled it entirely (otherwise it sill tries to build IIRC).

How did it fair when you just updated the `.travis.yml` file, together with GitHub actions you have ported before? I was OK with that approach (GH Actions + Travis).

fermezz:
@nicoddemus oh, you meant for me to use both travis and GitHub actions! Did not get that. So do you believe we should have all environments in GH Actions apart from the travis ones? Or we should have only-windows/macos environments there?

nicoddemus:
> @nicoddemus oh, you meant for me to use both travis and GitHub actions! Did not get that. 

Oh sorry I wasn't clearer!

> So do you believe we should have all environments in GH Actions apart from the travis ones? Or we should have only-windows/macos environments there?

The Travis configuration contains only a few environments, so I'm OK to leave it together with the others, even if they overlap a bit. 👍

fermezz:
Alright, @nicoddemus! What do you think?

fermezz:
@nicoddemus bump! (No rush, just making sure you saw the latest update)

nicoddemus:
Thanks! Been a bit swamped/busy later, appreciate the ping.

Great work!

fermezz:
No issue whatsoever, thanks for taking the time! I have now updated https://github.com/pytest-dev/pytest/pull/6870 :).

PR #6903, opened 2020-03-12
Assume os.dup is always available

bluetech:
The commit which added the checks for os.dup a15afb5e4879d68033a723129e6 suggests it was done for Jython. But pytest doesn't support Jython anymore (Jython is Python 2 only).

Furthermore, it looks like the faulthandler plugin (bundled in pytest and enabled by default) uses os.dup() unprotected and there have not been any complaints.

So seems better to just remove these checks, and only add if someone with a legitimate use case complains.


bluetech:
Let's see what CI says. If it passes I will add a changelog.

PR #6905, opened 2020-03-12
Changing key "error" to "errors" to maintain plurality

barrezuetai:
This pull request was created to try and fix issue #6505
"error" has been changed to "errors" without affecting the user-facing text.
When testing the changes, there was one module that wouldn't pass (testing/test_junitxml.py).
I've tried to look deeper into this and saw that when an xmlschema is created, there is an <error /> node created instead of <errors />. It looks like the "s" is being dropped from "errors" in the junit-10.xsd file which ends up in creating that node.

I'm not sure where to look further in order to fix the failed test cases. If you know which direction I should take in order to fix this I'd gladly move in that direction. @blueyed

### DIFF
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -469,7 +469,7 @@ class LogXML:
         self.log_passing_tests = log_passing_tests
         self.report_duration = report_duration
         self.family = family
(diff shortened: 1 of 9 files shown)


nicoddemus:
Hi @barrezuetai,

Sorry for the long response time on this, but in the end we took a different approach in #7364. 👍 

Hope we see other contributions from you in the future!

PR #6916, opened 2020-03-13
Skip link checks when doing releases through the bot

nicoddemus:
Unfortunately this is really getting in the way of the releases not because
of broken links, but because it is very flaky.

Related: #6894


nicoddemus:
> as a follow-up we might want to consider schedule based link-checks +

The idea is good, but I'm not sure if the scheduled based won't just be ignored somehow

PR #6914, opened 2020-03-13
Revert "[parametrize] enforce explicit argnames declaration (#6330)"

nicoddemus:
This reverts commit 9e262038c84a99d1353551e8cbb32f46362b58b4.

Fix #6909

PR #6919, opened 2020-03-13
[Backport #6914] Revert "[parametrize] enforce explicit argnames declaration (#6330)"

nicoddemus:
I realize now that I should have done the other way around: revert the original PR on `master`, and then backport to `5.4.x`.

Either way, what should we do with the CHANGELOG on backports like this one?

### DIFF
--- a/changelog/6909.bugfix.rst
+++ b/changelog/6909.bugfix.rst
@@ -1,0 +1,3 @@
+Revert the change introduced by `#6330 <https://github.com/pytest-dev/pytest/pull/6330>`_, which required all arguments to ``@pytest.mark.parametrize`` to be explicitly defined in the function signature.
+
+The intention of the original change was to remove what was expected to be an unintended/surprising behavior, but it turns out many people relied on it, so the restriction has been reverted.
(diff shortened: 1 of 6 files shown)


nicoddemus:
> I guess it's more of a "forward" port :)

Heh yep. 😅 

> Also this PR is not a good test-case, because it went in the wrong way (first to release branch and then to master).

Definitely, my bad, I realized my mistake too late.

> Remove changelog from this PR & merge it.
> Once the patch release containing this is done, cherry-pick the release PR to master (which includes the doc updates & changelog).

5.4.1 has been released already, so aren't those two suggestions the same thing? Both would end with the change on `master` (unless I'm missing something).

bluetech:
> 5.4.1 has been released already, so aren't those two suggestions the same thing? Both would end with the change on master (unless I'm missing something).

My thinking is, that the changelog will be included in 5.4.2, and we will cherry-pick the 5.4.2 changelog to master (like 68d4b17a5f2fecc06806adaf335a6151d9df62d4), then if this PR changelog is also included in master, then it will eventually be duplicated in 5.5.0. But I may be missing something too :)

PR #6920, opened 2020-03-13
Cherry pick CHANGELOG from 5.4.1

nicoddemus:
Not sure, but I think we don't need to wait for approval for this kind of PR?


nicoddemus:
Linting passed, merging so I can make the proper announcement on Twitter.

bluetech:
> Not sure, but I think we don't need to wait for approval for this kind of PR? @bluetech 

Agreed, I don't think mechanical things like this need approval.

PR #6934, opened 2020-03-17
Fix-6906: Added code-highlight option to disable highlighting optionally

### LINKED ISSUE #6906: Ability to disable pygments highlighting even if pygments is installed
the [docstring mentions a `code_highlight` setting](https://github.com/pytest-dev/pytest/blob/93aa988e0152392c233acac1e24bf6db8a2772c7/src/_pytest/_io/__init__.py#L29) but it seems it was not implemented

I'd like to be able to disable the syntax highlighting for a couple reasons:
- the colors can be distracting
- I'd like to be able to disable them for the tests in [pygments-pytest](https://github.com/asottile/pygments-pytest) (currently failing)

gdhameeja:
Closes #6906 
code-highlight has default value to True, this value is used in `TerminalWriter`
in order to control highlighting of the code.

### DIFF
--- a/src/_pytest/_io/terminalwriter.py
+++ b/src/_pytest/_io/terminalwriter.py
@@ -74,6 +74,7 @@ class TerminalWriter:
         self.hasmarkup = should_do_markup(file)
         self._current_line = ""
         self._terminal_width = None  # type: Optional[int]
+        self.code_highlight = True
 
     @property
     def fullwidth(self) -> int:
@@ -180,7 +181,7 @@ class TerminalWriter:
 
     def _highlight(self, source: str) -> str:
         """Highlight the given source code if we have markup support."""
-        if not self.hasmarkup:
+        if not self.hasmarkup or not self.code_highlight:
             return source
         try:
             from pygments.formatters.terminal import TerminalFormatter
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1389,8 +1389,13 @@ def create_terminal_writer(
     tw = TerminalWriter(file=file)
     if config.option.color == "yes":
         tw.hasmarkup = True
-    if config.option.color == "no":
+    elif config.option.color == "no":
         tw.hasmarkup = False
+
+    if config.option.code_highlight == "yes":
+        tw.code_highlight = True
+    elif config.option.code_highlight == "no":
+        tw.code_highlight = False
     return tw
 
 
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -208,6 +208,12 @@ def pytest_addoption(parser: Parser) -> None:
         choices=["yes", "no", "auto"],
         help="color terminal output (yes/no/auto).",
     )
+    group._addoption(
+        "--code-highlight",
+        default="yes",
+        choices=["yes", "no"],
+        help="Whether code should be highlighted (only if --color is also enabled)",
+    )
 
     parser.addini(
         "console_output_style",
--- a/testing/io/test_terminalwriter.py
+++ b/testing/io/test_terminalwriter.py
@@ -213,19 +213,32 @@ class TestTerminalWriterLineWidth:
 
 
 @pytest.mark.parametrize(
-    "has_markup, expected",
+    ("has_markup", "code_highlight", "expected"),
     [
         pytest.param(
-            True, "{kw}assert{hl-reset} {number}0{hl-reset}\n", id="with markup"
+            True,
+            True,
+            "{kw}assert{hl-reset} {number}0{hl-reset}\n",
+            id="with markup and code_highlight",
         ),
-        pytest.param(False, "assert 0\n", id="no markup"),
+        pytest.param(
+            True, False, "assert 0\n", id="with markup but no code_highlight",
+        ),
+        pytest.param(
+            False, True, "assert 0\n", id="without markup but with code_highlight",
+        ),
+        pytest.param(
+            False, False, "assert 0\n", id="neither markup nor code_highlight",
+        ),
     ],
 )
-def test_code_highlight(has_markup, expected, color_mapping):
+def test_code_highlight(has_markup, code_highlight, expected, color_mapping):
     f = io.StringIO()
     tw = terminalwriter.TerminalWriter(f)
     tw.hasmarkup = has_markup
+    tw.code_highlight = code_highlight
     tw._write_source(["assert 0"])
+
     assert f.getvalue().splitlines(keepends=True) == color_mapping.format([expected])
 
     with pytest.raises(
--- a/changelog/6906.feature.rst
+++ b/changelog/6906.feature.rst
@@ -1,0 +1,1 @@
+Added `--code-highlight` command line option to enable/disable code highlighting in terminal output.


testing/test_terminal.py
```diff
@@ -2138,3 +2138,24 @@ def test_foo():
...
+        result = testdir.runpytest("--code-highlight=no")
+        color_mapping.requires_ordered_markup(result)
+        result.stdout.fnmatch_lines(
+            color_mapping.format_for_fnmatch(
+                [
+                    "    {kw}def{hl-reset} {function}test_foo{hl-reset}():",
+                    "        {print}print{hl-reset}({str}'''{hl-reset}{str}{hl-reset}",
+                    ">   {str}    {hl-reset}{str}'''{hl-reset}); {kw}assert{hl-reset} {number}0{hl-reset}",
+                    "assert 0{reset}",
+                ]
```
gdhameeja: I believe the test does not work, I'm not familiar with this format. I shall fix it.

testing/code/test_terminal_writer.py
```diff
@@ -18,6 +18,7 @@ def test_code_highlight(has_markup, expected, color_mapping):
     f = StringIO()
     tw = TerminalWriter(f)
     tw.hasmarkup = has_markup
+    tw.code_highlight = True
```
gdhameeja: Added this with the assumption that this test assumes code is to be highlighted.

src/_pytest/_io/__init__.py
```diff
@@ -36,4 +36,5 @@ def _highlight(self, source):
         except ImportError:
             return source
         else:
-            return highlight(source, PythonLexer(), TerminalFormatter(bg="dark"))
+            if self.code_highlight:
```
gdhameeja: should we return source in else condition?

gdhameeja:
@nicoddemus review please?

blueyed:
It would be better to have a config setting that would allow for specifying the formatter/args for it also.  See https://github.com/pytest-dev/pytest/pull/6658 for some feedback I've given there already.

(it is rather bad that is uses the 16 colors formatter currently by default, and even goes through all the imports all the time)

nicoddemus:
> It would be better to have a config setting that would allow for specifying the formatter/args for it also.

Agree, but out of scope for this PR.

blueyed:
> > It would be better to have a config setting that would allow for specifying the formatter/args for it also.
> 
> Agree, but out of scope for this PR.

Well, you are about to add an option to control this behavior: it would be better to do it "properly" already then, and not add another option afterwards.

nicoddemus:
An option to turn it on/off and selecting styles are separate options.

blueyed:
IMHO it would/could select the formatter, and "no" would use none, i.e. disable it then.
"auto" (which is used here, but unclear IIRC) would then e.g. use the best option (true colors, 256 colors, or 16 (what is used currently)).

blueyed:
(To specify the style the same option _could_ be used, but e.g. the 16 color formatter has no style (i.e. it makes no sense / is not possible to use a "style" currently anyway), but a colorscheme, i.e. this requires something like "formatter_kwargs" really - which is not easy to add via an ini/config option.)
Ref: https://github.com/pygments/pygments/issues/1380

nicoddemus:
Not sure we will ever have a single option for color highlighting, we might grow some other options rather than just number of colors. I think having an explicit option to turn everything highlight related is more future proof.

src/_pytest/terminal.py
```diff
@@ -161,6 +161,15 @@ def pytest_addoption(parser):
         choices=["yes", "no", "auto"],
         help="color terminal output (yes/no/auto).",
     )
+    group._addoption(
+        "--code-highlight",
+        metavar="code_highlight",
+        action="store",
+        dest="code_highlight",
+        default="yes",
+        choices=["yes", "no", "auto"],
```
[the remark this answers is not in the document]
gdhameeja: Saw it added in other options as well, removed it now since default is "yes" in `code-highlight`

src/_pytest/terminal.py
```diff
@@ -161,6 +161,15 @@ def pytest_addoption(parser):
         choices=["yes", "no", "auto"],
         help="color terminal output (yes/no/auto).",
     )
+    group._addoption(
+        "--code-highlight",
+        metavar="code_highlight",
```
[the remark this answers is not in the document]
gdhameeja: removed.

src/_pytest/_io/__init__.py
gdhameeja: fixed.

gdhameeja:
@nicoddemus  @bluetech  fixed the comments. Can you please review it again. Thanks.

The-Compiler:
If we might want to have some more related options in the future, shouldn't this be an option in the .ini file rather than a commandline switch?

src/_pytest/_io/__init__.py
```diff
@@ -36,4 +36,6 @@ def _highlight(self, source):
...
```
bluetech, line 39: I didn't notice earlier, but it would be better to move this condition to above, i.e. `if not self.hasmarkup or not self.code_highlight`. This way, if `code_highlight` is false, the imports are not even attempted.

src/_pytest/config/__init__.py
```diff
@@ -1215,6 +1215,11 @@ def create_terminal_writer(config: Config, *args, **kwargs) -> TerminalWriter:
         tw.hasmarkup = True
     if config.option.color == "no":
         tw.hasmarkup = False
+
+    if config.option.code_highlight == "no":
+        tw.code_highlight = False
+    else:
```
bluetech, line 1221: Let's change this to `config.option.code_highlight == "yes"`, to be consistent with the `hasmarkup` code above. I'd also switch the order such that `"yes"` is checked first, for the same reason.

bluetech:
@The-Compiler

> If we might want to have some more related options in the future, shouldn't this be an option in the .ini file rather than a commandline switch?

Is there a general principle to commandline vs. ini? Looking at `pytest --help` I can't really discern one I must say ;)

The-Compiler:
> Is there a general principle to commandline vs. ini? Looking at `pytest --help` I can't really discern one I must say ;)

Fair point. My personal take:

- Something that someone would probably set for individual invocations rather than always (e.g. `-x`, `-k`, `-m`, `--fixtures`, `--lf`, ...): command-line option
- Something that someone would probably want to always set (e.g. `testpaths`, `xfail_strict`, but also `--strict`): config option
- Something requiring a lot of data (e.g. `markers`, `filterwarnings`): config option
- Something where there could be multiple related options: config options

But I agree that's not really how things look at the moment, we also have lots of `--doctest=*` or `--log-*` options, so I don't have strong feelings on this.

src/_pytest/config/__init__.py
```diff
@@ -1260,6 +1260,7 @@ def create_terminal_writer(config: Config, *args, **kwargs) -> TerminalWriter:
     tw = TerminalWriter(*args, **kwargs)
     if config.option.color == "yes":
         tw.hasmarkup = True
+    tw.code_highlight = False if config.option.code_highlight == "yes" else True
```
asottile, line 1263:
```suggestion
    tw.code_highlight = config.option.code_highlight == "yes"
```

(I have a [wiki page](https://github.com/asottile/scratch/wiki/PythonHintsForNoobs#useless-branches) about this)

src/_pytest/_io/terminalwriter.py
```diff
@@ -180,7 +180,7 @@ def _write_source(self, lines: Sequence[str], indents: Sequence[str] = ()) -> No
 
     def _highlight(self, source: str) -> str:
         """Highlight the given source code if we have markup support."""
-        if not self.hasmarkup:
+        if not self.hasmarkup and not self.code_highlight:
```
nicoddemus, line 183: `self.code_highlight` should be initialized in `__init__` I think (both for correctness and clarity).

src/_pytest/config/__init__.py
gdhameeja: Fixed this.

src/_pytest/_io/terminalwriter.py
gdhameeja: Fixed.

doc/en/changelog.rst
```diff
@@ -504,6 +504,8 @@ Features
   such as default values or which set of command line options to add.
 
 
+- `#6906 <https://github.com/pytest-dev/pytest/issues/6906>`_: ``--code-highlight`` decides whethere code should be highlighted in terminal output.
+
```
asottile, line 508: we manage changelogs for new features by adding a changelog fragment.  You can see how these are set up by looking at the `./changelog` directory (there's a bunch of files in there already that you can base yours on).  The [CONTRIBUTING](https://github.com/pytest-dev/pytest/blob/master/CONTRIBUTING.rst) guide also covers changelogs as well

src/_pytest/_code/code.py
```diff
@@ -984,6 +984,7 @@ class ReprTraceback(TerminalRepr):
     entrysep = "_ "
 
     def toterminal(self, tw: TerminalWriter) -> None:
+        tw.code_highlight = False
```
asottile, line 987: was this left over from debugging? this doesn't quite feel right here

src/_pytest/config/__init__.py
asottile: looks like this didn't get fixed?

src/_pytest/terminal.py
```diff
@@ -164,7 +164,15 @@ def pytest_addoption(parser):
         choices=["yes", "no", "auto"],
         help="color terminal output (yes/no/auto).",
     )
-
+    group._addoption(
+        "--code-highlight",
+        action="store",
+        dest="code_highlight",
```
asottile, line 170: can leave these two lines out, they are the default when using argparse

testing/io/test_terminalwriter.py
```diff
@@ -218,18 +217,13 @@ def test_combining(self) -> None:
...
     tw = terminalwriter.TerminalWriter(f)
     tw.hasmarkup = has_markup
     tw._write_source(["assert 0"])
-    assert f.getvalue().splitlines(keepends=True) == color_mapping.format([expected])
 
-    with pytest.raises(
-        ValueError,
-        match=re.escape("indents size (2) should have same size as lines (1)"),
-    ):
-        tw._write_source(["assert 0"], [" ", " "])
```
asottile, line 235: why is this missing now?

src/_pytest/_code/code.py
gdhameeja: No this was added purposely. The tests fail if code_highlight is set to true (which by default is it). I assumed all tests were written asserting to non highlighted code, so I put it here. Please guide me on this.

asottile: I suspect this shouldn't happen here, and maybe we'll need to adjust some of the tests.  I'm not 100% certain though

src/_pytest/config/__init__.py
gdhameeja: Fixed.

src/_pytest/_code/code.py
gdhameeja:
```
FAILED testing/code/test_excinfo.py::TestFormattedExcinfo::test_repr_traceback_with_invalid_cwd - Failed: nomatch: 'def entry():'
FAILED testing/code/test_excinfo.py::TestFormattedExcinfo::test_exc_chain_repr_without_traceback[cause] - Failed: exact match: 'ValueError: invalid value'
FAILED testing/code/test_excinfo.py::TestFormattedExcinfo::test_exc_chain_repr_without_traceback[context] - Failed: exact match: 'ValueError: invalid value'
FAILED testing/io/test_terminalwriter.py::test_code_highlight[no markup] - AssertionError: assert ['\x1b[94mass...[39;49;00m\n'] == ['assert 0\n']
FAILED testing/acceptance_test.py::TestGeneralUsage::test_root_conftest_syntax_error - Failed: nomatch: '*raise SyntaxError*'
FAILED testing/acceptance_test.py::TestGeneralUsage::test_early_hook_error_issue38_1 - Failed: nomatch: '*INTERNALERROR*def pytest_sessionstart():*'
FAILED testing/acceptance_test.py::TestGeneralUsage::test_better_reporting_on_conftest_load_failure - assert ["ImportError...med 'qwerty'"] == ["ImportError...med 'qwerty'"]
FAILED testing/test_conftest.py::test_conftest_exception_handling - assert 'raise ValueError()' in ["ImportError while loading conftest '/tmp/pytest-of-melvault/pytest-0/test_conftest_exception_h...
FAILED testing/test_junitxml.py::TestPython::test_xfailure_xpass_strict[xunit1] - AssertionError: assert {'message': '...1B[39;49;00m'} == {'message': '...eds to fail!'}
FAILED testing/test_junitxml.py::TestPython::test_xfailure_xpass_strict[xunit2] - AssertionError: assert {'message': '...1B[39;49;00m'} == {'message': '...eds to fail!'}
FAILED testing/test_reports.py::TestReportSerialization::test_chained_exceptions[TestReport] - assert "ValueError('value error')" in "\x1b[94mdef\x1b[39;49;00m \x1b[92mtest_a\x1b[39;49;00m():\n  ...
FAILED testing/test_reports.py::TestReportSerialization::test_chained_exceptions[CollectReport] - assert "ValueError('value error')" in "test_chained_exceptions.py:5: in test_a\n    foo()\ntest_c...
FAILED testing/test_resultlog.py::TestWithFunctionIntegration::test_log_test_outcomes - assert -1 != -1
FAILED testing/python/collect.py::TestModule::test_show_traceback_import_error[0] - Failed: nomatch: 'ImportError while importing test module*'
FAILED testing/python/collect.py::TestModule::test_show_traceback_import_error[1] - Failed: nomatch: 'ImportError while importing test module*'
FAILED testing/python/collect.py::TestModule::test_show_traceback_import_error[2] - Failed: nomatch: 'ImportError while importing test module*'
FAILED testing/python/collect.py::TestModule::test_show_traceback_import_error_unicode - Failed: nomatch: 'ImportError while importing test module*'
```
These are all the tests which failed with code_highlight set to `True`. I'm lost on what is the right way to do this. Please guide me here.

@asottile  @nicoddemus

nicoddemus: Hi @gdhameeja,

The problem was the conditional in `_highlight`: that should be an `or`, not an `and`. I've fixed it and pushed to your branch, hope you don't mind.

src/_pytest/terminal.py
```diff
@@ -164,7 +164,15 @@ def pytest_addoption(parser):
         choices=["yes", "no", "auto"],
         help="color terminal output (yes/no/auto).",
     )
-
+    group._addoption(
```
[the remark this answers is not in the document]
nicoddemus: Another question: how should this interact with `--colors`? Should `--colors=no` also affect code highlighting?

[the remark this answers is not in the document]
asottile: yeah I agree, or we just make `--colors` `and` with this option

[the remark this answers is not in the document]
asottile: I don't think we need an `auto` if we combine the condition with `color` (looks like it already works that way)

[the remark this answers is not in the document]
nicoddemus: Agreed!

src/_pytest/terminal.py
```diff
@@ -208,6 +208,12 @@ def pytest_addoption(parser: Parser) -> None:
         choices=["yes", "no", "auto"],
         help="color terminal output (yes/no/auto).",
     )
+    group._addoption(
+        "--code-highlight",
+        default="yes",
+        choices=["yes", "no"],
+        help="Whether code should be highlighted on the terminal or not",
```
nicoddemus, line 215:
```suggestion
        help="Whether code should be highlighted (only if --color is also enabled)",
```

nicoddemus:
@bluetech would you like to take another look?

(Note to selves: squash on merge)

src/_pytest/_io/terminalwriter.py
```diff
@@ -74,6 +74,8 @@ def __init__(self, file: Optional[TextIO] = None) -> None:
         self.hasmarkup = should_do_markup(file)
         self._current_line = ""
         self._terminal_width = None  # type: Optional[int]
+        # default to true
```
bluetech, line 77: This comment I think is redundant:

```suggestion
```

src/_pytest/config/__init__.py
```diff
@@ -1339,6 +1339,7 @@ def create_terminal_writer(config: Config, *args, **kwargs) -> TerminalWriter:
     tw = TerminalWriter(*args, **kwargs)
     if config.option.color == "yes":
         tw.hasmarkup = True
+    tw.code_highlight = config.option.code_highlight == "yes" and tw.hasmarkup
     if config.option.color == "no":
         tw.hasmarkup = False
```
bluetech, line 1344: I suggest a few changes:

- Move the `code_highlight` below the last two lines otherwise it cuts in the middle of the `color` yes/no logic
- Make `code_highlight` boolean itself independent of `hasmarkup` -- the highlight code already checks `if not self.hasmarkup or not self.code_highlight` so it is better to let it standalone just in case.
- If no an explicit yes/no isn't given, don't override the default (same logic as `color` uses)

```suggestion

    if config.option.color == "yes":
        tw.hasmarkup = True
    elif config.option.color == "no":
        tw.hasmarkup = False

    if config.option.code_highlight == "yes":
        tw.code_highlight =  True
    elif config.option.code_highlight == "no":
        tw.code_highlight =  False
```

testing/io/test_terminalwriter.py
```diff
@@ -213,19 +213,32 @@ def test_combining(self) -> None:
...
+            True,
+            True,
+            "{kw}assert{hl-reset} {number}0{hl-reset}\n",
+            id="with markup and code_highlight",
+        ),
+        pytest.param(
+            True, False, "assert 0\n", id="with markup but no code_highlight",
+        ),
+        pytest.param(
+            False, True, "assert 0\n", id="with markup but no code_highlight",
```
bluetech, line 228:
```suggestion
            False, True, "assert 0\n", id="without markup but with code_highlight",
```

bluetech:
Also please see the discussion in #7132, where the ability to control the pygments style is discussed. If we want to support that, it would presumably use the same cmd line argument, e.g. `--code-highlight=<style>`. If that doesn't look right, we should probably come up with something that would be...

bluetech:
There's also a `Terminal256Formatter`: https://pygments.org/docs/formatters/?highlight=terminalformatter#Terminal256Formatter

gnikonorov:
Hi @gdhameeja. If this PR is meant to close an issue, could you please update the description as described [here](https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords) so that the linked issue may autoclose?

nicoddemus:
> Also please see the discussion in #7132, where the ability to control the pygments style is discussed. If we want to support that, it would presumably use the same cmd line argument, e.g. --code-highlight=<style>.

I think we can postpone the discussion of styles to a separate PR, as I think it warrants more discussion ( https://github.com/pytest-dev/pytest/issues/7132#issuecomment-650155320) and either way whatever we decide later will probably be backward compatible. 👍

nicoddemus:
Thanks @asottile for finishing it up. 👍

PR #6944, opened 2020-03-20
Issue 1120 Cleaning up tmpdir's

piotrhm:
Hi,
I combined work done by @bitmuster and @caitelatte which close [#1120](https://github.com/pytest-dev/pytest/issues/1120). Solution is a proposition which addresses @RonnyPfannschmidt comments [available here](https://github.com/pytest-dev/pytest/pull/6253).


src/_pytest/pathlib.py
```diff
@@ -305,11 +305,13 @@ def make_numbered_dir_with_cleanup(
             e = exc
         else:
             consider_lock_dead_if_created_before = p.stat().st_mtime - lock_timeout
-            cleanup_numbered_dir(
-                root=root,
-                prefix=prefix,
-                keep=keep,
-                consider_lock_dead_if_created_before=consider_lock_dead_if_created_before,
+            # Register a cleanup for program exit
+            atexit.register(
```
nicoddemus: This might still fail because some pytest process will finish before others so they might still step in each others toes, I guess?

piotrhm: Do you mean one of the race condition cases mentioned in maybe_delete_a_numbered_dir()? I thought try_cleanup() invoked from cleanup_numbered_dir() is preventing from such case or at least we have some guarantees - based on comments from maybe_delete_a_numbered_dir().

nicoddemus:
@RonnyPfannschmidt gentle ping.

nicoddemus:
I guess this is very hard to add a test for right?

Rebased.

PR #6950, opened 2020-03-21
ci: twisted and oldattrs tox envs are now incompatible, don't run the…

bluetech:
…m together

twisted started to use `attr.s(eq)` argument which was added recently,
so it fails with oldattrs. One of the CI jobs ran twisted and oldattrs
together, so it started to fail.

Make the twisted code be already covered by another job, and remove it from
the job with the oldattrs.


bluetech:
OK, seems to do the trick, and coverage seems fine too.

I'm merging without approval to fix the CI. But if there are any comments do let me know.

PR #6953, opened 2020-03-21
Issue 4677 - always relative path in skip report

### LINKED ISSUE #4677: SKIPPED module path is sometimes resolved, sometimes not
The module path when a test is skipped is sometimes relative, sometimes absolute.  Whether a test was skipped with a `pytest.skip()` call or with a `@pytest.mark.skip` decoration seems to be one possible cause for discrepancy.

It would be nice to make a consistent decision here, but which is the right choice - absolute path or relative?

<details><summary>Click me to see an example</summary>
<p>

```
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/deprecated_test.py:69: xdist workers disable the terminal reporter plugin
SKIPPED [1] testing/test_assertion.py:1158: This bug does not exist on PY3
SKIPPED [1] testing/test_assertrewrite.py:938: condition: sys.version_info[0] >= 3
SKIPPED [1] testing/test_assertrewrite.py:958: condition: sys.version_info[0] >= 3
SKIPPED [1] testing/test_assertrewrite.py:889: python 2 only
SKIPPED [1] testing/test_assertrewrite.py:947: condition: sys.version_info[0] >= 3
SKIPPED [1] testing/test_assertrewrite.py:969: condition: sys.version_info[0] >= 3
SKIPPED [1] testing/test_capture.py:554: only have capsysbinary in python 3
SKIPPED [1] testing/test_capture.py:883: python2 has no buffer
SKIPPED [2] testing/test_capture.py:1107: text output different for bytes on python3
SKIPPED [1] testing/test_capture.py:1371: only py3.6+ on windows
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/test_doctest.py:1214: could not import 'mock'
SKIPPED [1] testing/test_monkeypatch.py:211: Python 2 only test
SKIPPED [1] testing/test_monkeypatch.py:219: Python 2 only test
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/test_parseopt.py:306: need to be run with pytest executable, not -c
SKIPPED [31] /home/travis/build/pytest-dev/pytest/.tox/py37-xdist/lib/python3.7/site-packages/_pytest/pytester.py:1209: could not import 'pexpect'
SKIPPED [11] /home/travis/build/pytest-dev/pytest/testing/test_runner.py:429: could not import 'xdist.boxed'
SKIPPED [1] testing/test_skipping.py:52: condition: sys.version_info[0] >= 3
SKIPPED [1] testing/code/test_code.py:99: python 2 only issue
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/code/test_excinfo.py:281: could not import 'decorator'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/code/test_excinfo.py:386: could not import 'jinja2'
SKIPPED [1] testing/test_tmpdir.py:223: win only
SKIPPED [8] /home/travis/build/pytest-dev/pytest/testing/test_unittest.py:398: could not import 'twisted.trial.unittest'
SKIPPED [5] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:76: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:334: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:358: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:382: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:400: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:414: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:480: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/approx.py:491: could not import 'numpy'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/test_unittest.py:940: could not import 'unittest2'
SKIPPED [1] testing/test_warnings.py:159: warnings message is broken as it is not str instance
SKIPPED [1] testing/test_warnings.py:571: Python 2 only issue
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/integration.py:169: could not import 'mock'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/integration.py:192: could not import 'mock'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/integration.py:225: could not import 'mock'
SKIPPED [1] /home/travis/build/pytest-dev/pytest/testing/python/integration.py:249: could not import 'mock'
```

</p>
</details>

CarycaKatarzyna:
It refers to [issue #4677](https://github.com/pytest-dev/pytest/issues/4677) and changes absolute path to relative in skip report.

Fixes #4677.


bluetech:
Hi @CarycaKatarzyna,

Thanks for working on this.

The bug report says the skip reports sometimes have relative paths and sometimes absolute paths. Your patch fixes this by changing the reporting to check if the path is absolute, and making it relative if so. But I think it will be better to fix it at the source, such that absolute paths do not reach the report in the first place. At least, it will be good to understand why that happens, and then decide on the proper solution.

So, it will be helpful to have an example where an absolute path is reported, and then find where it comes from in the pytest code. And maybe compare and contrast with a report which comes in relative.

bluetech:
Also, please note there are some CI failures, you can see the logs in the "Checks" tab.

CarycaKatarzyna:
Thank you @bluetech for looking at my code.

I have done little investigation. As @wimglenn already seen, absolute or relative path in report depends on using ``pytest.skip()`` or ``@pytest.mark.skip``. The line which makes difference is here.

https://github.com/pytest-dev/pytest/blob/95fadd5740d05fda9c525e1e7d8e75b60ad1b143/src/_pytest/skipping.py#L171

Before this line both paths are absolute. But in case with ``pytest.skip()``, the path is ``$PYTEST-PATH/pytest/src/_pytest/skipping.py``. So it is changed to location of file - ``item.location`` is relative.

The another option to make consistent report is keep path to skipping place also relative. Have you any other proposition?

bluetech:
> The another option to make consistent report is keep path to skipping place also relative. Have you any other proposition?

I looked at the code you pointed to, and also dug a bit more. The approach I preferred, of having all paths that reach the reporting be relative in the first place, does seem unworkable. So your approach of post-processing the path seems to be the way to go.

You chose to modify the path in `_pytest.runner`. But from what I can see, the actual reporting code in `_pytest.terminal` actually does does something similar in a lot of cases (see `_get_pos()` function), so it seems more appropriate to do it there. The `show_skipped` code in that file is different from the others and doesn't use `_get_pos()`.

So, what I think we should do is to make `show_skipped` relativize the path. Here is a diff against `master` which seems fix the test and all the other tests pass as well:

```diff
diff --git a/src/_pytest/terminal.py b/src/_pytest/terminal.py
index a99463fe8..8559f07eb 100644
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -1027,7 +1027,7 @@ class TerminalReporter:
 
         def show_skipped(lines: List[str]) -> None:
             skipped = self.stats.get("skipped", [])
-            fskips = _folded_skips(skipped) if skipped else []
+            fskips = _folded_skips(self.startdir, skipped) if skipped else []
             if not fskips:
                 return
             verbose_word = skipped[0]._get_verbose_word(self.config)
@@ -1153,11 +1153,13 @@ def _get_line_with_reprcrash_message(config, rep, termwidth):
     return line
 
 
-def _folded_skips(skipped):
+def _folded_skips(startdir, skipped):
     d = {}
     for event in skipped:
-        key = event.longrepr
-        assert len(key) == 3, (event, key)
+        assert len(event.longrepr) == 3, (event, event.longrepr)
+        fspath, lineno, reason = event.longrepr
+        # For consistency, report all fspaths in relative form.
+        fspath = startdir.bestrelpath(fspath)
         keywords = getattr(event, "keywords", {})
         # folding reports with global pytestmark variable
         # this is workaround, because for now we cannot identify the scope of a skip marker
@@ -1167,7 +1169,9 @@ def _folded_skips(skipped):
             and "skip" in keywords
             and "pytestmark" not in keywords
         ):
-            key = (key[0], None, key[2])
+            key = (fspath, None, reason)
+        else:
+            key = (fspath, lineno, reason)
         d.setdefault(key, []).append(event)
     values = []
     for key, events in d.items():
diff --git a/testing/test_terminal.py b/testing/test_terminal.py
index 88d564519..d2857ca77 100644
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -2001,7 +2001,7 @@ def test_skip_reasons_folding():
     ev3.longrepr = longrepr
     ev3.skipped = True
 
-    values = _folded_skips([ev1, ev2, ev3])
+    values = _folded_skips(py.path.local(), [ev1, ev2, ev3])
     assert len(values) == 1
     num, fspath, lineno, reason = values[0]
     assert num == 3
```

@blueyed, I bet you know about this more than I do, WDYT?

CarycaKatarzyna:
@bluetech,  your solution looks better than mine because you use the same functions as in other reports. And the place of fixes is probably more consistent with another types of reports. But in your proposition is little mistake. Use:
```bash
fspath = startdir.bestrelpath(py.path.local(fspath))
```
instead
```bash
fspath = startdir.bestrelpath(fspath)
```
Note that `test_skipped_reasons_functional` in `testing/test_skipping.py` shoudn't pass in unchanged form. That test should be changed like in my commits.
 You should open your own pull request and close this one. It's a pity that you have proposed the whole solution instead of giving me a second chance.

bluetech:
> But in your proposition is little mistake

Yes, this should be fixed.

> Note that test_skipped_reasons_functional in testing/test_skipping.py shoudn't pass in unchanged form. That test should be changed like in my commits.

Right.

> You should open your own pull request and close this one.

I prefer if it were in this one, since it has good context and discussion. Can you update this PR?

> It's a pity that you have proposed the whole solution instead of giving me a second chance.

It is sometimes difficult to distinguish between people who just want to get some problem they've hit fixed, and ones who want to contribute to pytest and dig into the code more themselves. I hope you contribute more and I'll try to remember for the next time.

CarycaKatarzyna:
> Can you update this PR?

Yes, I've already added it. And incidentally other commits...

bluetech:
> Yes, I've already added it. And incidentally other commits...

Thanks!

I think what happened is that you merged `master` into this branch, which causes all of these commits to show up.

The way to fix this is to perform a [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) on the latest `master` branch instead. It should go something like this:

1. `git checkout master` - switch to the `master` branch.
2. `git pull origin` - get the latest version of `master` from pytest's repo.
3. `git checkout issue4677` - switch to the branch of this PR.
4. `git rebase -i master` - perform an interactive rebase of the `issue4677` branch on `master`.

This will open an editor with a list of commits. You should keep the lines that refer to your own commits for this PR, and delete all others. Then you can update the PR with `git push --force-with-lease`.

CarycaKatarzyna:
It works. Thanks!

changelog/4677.trivial.rst
```diff
@@ -0,0 +1 @@
+Path in report skip reasons is always relative.
```
bluetech, line 1:
```suggestion
The path shown in the summary report for SKIPPED tests is now always relative. Previously it was sometimes absolute.
```

I also think marking the change as "trivial" is selling it short -- I would use "bugfix" instead.

PR #6957, opened 2020-03-22
Better document xfail(condition)

### LINKED ISSUE #6952: Conditional xfail
I'd like to see something like `pytest.mark.xfailif(condition)`

### Use-case:
Our lib is built on multiple operating systems. Currently one test fails on OSX due to issues outside of the library. If it would fail always, I could add `@pytest.mark.xfail()`, but it's only happening on OSX. Alternatively, I could `@pytest.mark.skipif()`. But the advantage of `xflail` is that it would notify me when the external issue is solved and I can reactivate the test. With `skipif` I would not know when the test starts working again and probably forget it.

Alternatively, is there an existing pattern I could use to get a conditional xfail.

timhoffm:
Closes #6952.

### DIFF
--- a/doc/en/skipping.rst
+++ b/doc/en/skipping.rst
@@ -280,35 +280,6 @@ that condition as the first parameter:
 Note that you have to pass a reason as well (see the parameter description at
 :ref:`pytest.mark.xfail ref`).
 
-
-.. _`xfail strict tutorial`:
-
-``strict`` parameter
-~~~~~~~~~~~~~~~~~~~~
-
-
-
-Both ``XFAIL`` and ``XPASS`` don't fail the test suite by default.
-You can change this by setting the ``strict`` keyword-only parameter to ``True``:
-
-.. code-block:: python
-
-    @pytest.mark.xfail(strict=True)
-    def test_function():
-        ...
-
-
-This will make ``XPASS`` ("unexpectedly passing") results from this test to fail the test suite.
-
-You can change the default value of the ``strict`` parameter using the
-``xfail_strict`` ini option:
-
-.. code-block:: ini
-
-    [pytest]
-    xfail_strict=true
-
-
 ``reason`` parameter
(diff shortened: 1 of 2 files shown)


doc/en/skipping.rst
```diff
@@ -265,6 +265,22 @@ internally by raising a known exception.
 **Reference**: :ref:`pytest.mark.xfail ref`
 
 
+``condition`` parameter
+~~~~~~~~~~~~~~~~~~~~~~~
+
+If a test is only expected to fail under certain conditions, you can pass
+that condition as first parameter:
```
bluetech:
```suggestion
If a test is only expected to fail under a certain condition, you can pass
that condition as the first parameter:
```

doc/en/skipping.rst
```diff
@@ -265,6 +265,22 @@ internally by raising a known exception.
...
+
+.. code-block:: python
+
+    @pytest.mark.xfail(sys.platform == "windows", reason="bug in a 3rd party library")
```
bluetech:
```suggestion
    @pytest.mark.xfail(sys.platform == "win32", reason="bug in a 3rd party library")
```

timhoffm:
I've moved the `strict` section to the end. The order now corresponds to the order of the parameters in the signature, and it also brings the sections `condition` and `reason` next to each other as desired.

bluetech:
Thanks!

doc/en/skipping.rst
```diff
@@ -301,7 +288,7 @@ on a particular platform:
 
 .. code-block:: python
 
-    @pytest.mark.xfail(sys.version_info >= (3, 6), reason="python3.6 api changes")
+    @pytest.mark.xfail(reason="known parser issue")
```
blueyed: This example does not match the previous sentence anymore.

nicoddemus: Thanks, https://github.com/pytest-dev/pytest/pull/6980

PR #6972, opened 2020-03-26
unittest: fix the `_GetOutOf_testPartExecutor` exception kind. Fixes …

qbey:
…https://github.com/pytest-dev/pytest/issues/6947

`_GetOutOf_testPartExecutor` was introduced in
https://github.com/pytest-dev/pytest/commit/04f27d4eb469fb9c76fd2d100564a0f7d30028df
and allow to catch Exception from a unittest TestCase, but as it
inheritate from `KeyboardInterrupt` it is managed differently by
unittest (see.
https://github.com/python/cpython/blob/032de7324e30c6b44ef272cea3be205a3d768759/Lib/unittest/case.py#L61-L62)
as the `testPartExecutor` reraise it directly. This prevent unittest to
run the cleanup functions properly.

Since this change enable the teardown methods to be ran properly bu
unittest, the explicit teardown is not required anymore.

Checklist:
- [X] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>.<TYPE>.rst`. See [changelog/README.rst](https://github.com/pytest-dev/pytest/blob/master/changelog/README.rst) for details.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -220,6 +220,7 @@ Pieter Mulder
 Piotr Banaszkiewicz
 Pulkit Goyal
 Punyashloka Biswal
+Quentin Bey
 Quentin Pradet
 Ralf Schmitt
 Ralph Giles
--- a/changelog/6947.bugfix.rst
+++ b/changelog/6947.bugfix.rst
@@ -1,0 +1,1 @@
+Unittest cleanup functions are called even if a test fails.
--- a/src/_pytest/unittest.py
+++ b/src/_pytest/unittest.py
@@ -113,15 +113,12 @@ class TestCaseFunction(Function):
     _testcase = None
 
     def setup(self):
-        self._needs_explicit_tearDown = False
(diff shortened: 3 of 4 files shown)


blueyed:
The point of `_GetOutOf_testPartExecutor` is to get out of there, using `KeyboardInterrupt`.

I still think that it would be better to revert all that stuff, and require to use `pdb.set_trace()` for now, FWIW.

qbey:
Thank you Daniel.
Alright, this fix does not preserve the `--pdb` so it can't be accepted. I close it. 
Anyway, the fixup for the cleanup part can't be solved without also considering the async issue, so as said it might be better to revert the original commit for now.

qbey:
@blueyed You may still want to add the tests (about cleanup and teardown counter), `test_cleanup_called_issue6947` fails with pytest 5.4.1 and should not fail with version <5.4.0. So feel free to copy/paste/rewrite them if you rollback your commit :)

PR #6974, opened 2020-03-27
Fix ci: black is beautiful

blueyed:


nicoddemus:
Sorry did not see this, fixed in https://github.com/pytest-dev/pytest/pull/6976

blueyed:
@nicoddemus thanks for fixing it!

Issue #6925, opened 2020-03-14
pytest fails to run with pytest-xdist


MuslimBeibytuly:
base docker image: python:3.8.1-alpine3.11

amqp                          2.5.2     
apipkg                        1.5       
apns2                         0.7.1     
asgiref                       3.2.5     
attrs                         19.3.0    
autopep8                      1.5       
backcall                      0.1.0     
bandit                        1.6.2     
billiard                      3.6.3.0   
celery                        4.4.1     
certifi                       2019.11.28
cffi                          1.14.0    
chardet                       3.0.4     
click                         7.1.1     
coreapi                       2.3.3     
coreschema                    0.0.4     
coverage                      5.0.3     
cryptography                  2.8       
decorator                     4.4.2     
Django                        3.0.4     
django-cors-headers           3.2.1     
django-debug-toolbar          2.2       
django-environ                0.4.5     
django-extensions             2.2.8     
django-filter                 2.2.0     
django-fsm                    2.7.0     
django-guardian               2.2.0     
django-push-notifications     2.0.0     
django-redis-cache            2.1.0     
django-silk                   4.0.1     
django-solo                   1.1.3     
djangorestframework           3.11.0    
djangorestframework-simplejwt 4.4.0     
dparse                        0.4.1     
drf-yasg                      1.17.1    
entrypoints                   0.3       
eradicate                     1.0       
execnet                       1.7.1     
flake8                        3.7.9     
flake8-broken-line            0.1.1     
flake8-bugbear                20.1.4    
flake8-builtins               1.4.2     
flake8-comprehensions         3.2.2     
flake8-eradicate              0.3.0     
flake8-executable             2.0.3     
gevent                        1.4.0     
gitdb                         4.0.2     
GitPython                     3.1.0     
glob2                         0.7       
gprof2dot                     2019.11.30
greenlet                      0.4.15    
gunicorn                      20.0.4    
h2                            2.6.2     
hiredis                       1.0.1     
hpack                         3.0.0     
http-ece                      1.1.0     
hyper                         0.7.0     
hyperframe                    3.2.0     
idna                          2.9       
inflection                    0.3.1     
ipython                       7.13.0    
ipython-genutils              0.2.0     
itypes                        1.1.0     
jedi                          0.16.0    
Jinja2                        2.11.1    
kombu                         4.6.8     
Mako                          1.1.2     
MarkupSafe                    1.1.1     
mccabe                        0.6.1     
more-itertools                8.2.0     
[...]

PR #6988, opened 2020-03-30
Fix TerminalRepr instances to be hashable

blueyed:
pytest-xdist assumes `ExceptionChainRepr` is hashable.

Fixes https://github.com/pytest-dev/pytest/issues/6925.
Fixes https://github.com/pytest-dev/pytest-xdist/issues/515.

### DIFF
--- a/changelog/6925.bugfix.rst
+++ b/changelog/6925.bugfix.rst
@@ -1,0 +1,1 @@
+Fix TerminalRepr instances to be hashable again.
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -32,6 +32,7 @@ import _pytest
 from _pytest._io import TerminalWriter
 from _pytest._io.saferepr import safeformat
 from _pytest._io.saferepr import saferepr
+from _pytest.compat import ATTRS_EQ_FIELD
 from _pytest.compat import overload
 from _pytest.compat import TYPE_CHECKING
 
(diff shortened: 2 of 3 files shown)


bluetech:
Hmm but again I see this is targeted at `5.4.x`. The flow I (think) we agreed on is to always target `master`, and backport to the `MAJOR.MINOR.x` branch.

blueyed:
I would find it easier to just merge 5.4.x into master, instead of cherry-picking things to the release branch.  I've looked quickly for docs, found nothing and thought this would be easier.

bluetech:
> I would find it easier to just merge 5.4.x into master, instead of cherry-picking things to the release branch. I've looked quickly for docs, found nothing and thought this would be easier.

Yes we should document it somewhere, but the idea is that all new development is done on master and then backported as needed, without merging branches.

nicoddemus:
> Yes we should document it somewhere, but the idea is that all new development is done on master and then backported as needed, without merging branches.

Indeed, I understand that the main aspect of the change was for contributors to always target `master`, instead of having to decide on which branch to target when opening a PR.

blueyed:
Sure, I'm not saying they need to decide and that we should not do "backports" in general.
But if you are knowing that it is going to be backported it could go there directly in theory.  Avoids the overhead of "needs backport" labeling and PRs for the backports.
Not going to argue over it though - JFI.

PR #6991, opened 2020-03-31
Fix regressions with --lf plugin

blueyed:


bluetech:
Is this fixed in master already?

changelog/6991.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix --lf with explicit cmdline args.
```
nicoddemus, line 1:
```suggestion
Fix `--lf` for files explicitly given in the command-line.
```

blueyed: s/in/on ?

Anyway, I will add the other changes here also, and reword that then.

PR #6993, opened 2020-03-31
[WIP] doc: pytest_collection: has to set `session.items`

blueyed:
Would make sense to use its return value etc, but this helps for now.

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -170,7 +170,26 @@ def pytest_load_initial_conftests(early_config, parser, args):
 def pytest_collection(session: "Session") -> Optional[Any]:
     """Perform the collection protocol for the given session.
 
-    The hook has to set `session.items` to a sequence of items.
+    Usually plugins will implement this hook only to perform some action before
+    collection, for example the terminal plugin will use this to start displaying
+    the collection counter, so usually plugins return `None` from this hook after
+    performing the desired operation.
+    
+    However a plugin might decide to override the collection completely,
+    in which case it should return `True`, but then it is usually expected
+    that this hook will also need to perform the following operations 
+    that are usually part of the collection process:
+    
+      1. Call the pytest_collectstart hook.
+      2. Call the pytest_collectreport hook.
+      3. Call the pytest_collection_modifyitems hook.
+      4. Call the pytest_collection_finish hook.
+      5. Set session.testscollected to the amount of collect items.
+      6. Set `session.items` to a list of items.
+      
+    If a plugin just wants to skip collection entirely, like `pytest-xdist`
+    does for master nodes, then it is OK to not do anything other than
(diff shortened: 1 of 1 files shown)


bluetech:
This is usually done by the `pytest_collection` impl in the  `_pytest.main` plugin (which always runs last because registered first). Is the idea to document what the hook needs to do when *replacing* this impl (returning a value, rather than wrapping it or returning `None`)?

If so, it seems like such an implementation should do several other things (reading `Session.perform_collect()`):

1. Call the `pytest_collectstart` hook.
1. Call the `pytest_collectreport` hook.
1. Call the `pytest_collection_modifyitems` hook.
2. Call the `pytest_collection_finish` hook.
3. Set `session.testscollected` to the amount of collect items.

and other internal(?) things. Maybe it's OK to bypass all of that though? I'm curious if there is a plugin which does this?

nicoddemus:
Great points @bluetech.

> I'm curious if there is a plugin which does this?

I don't think any plugin out there actually overrides the hook completely, because as you noticed, it needs to do a lot of internal book keeping and call other hooks for things to keep working.

However, xdist does override this completely to prevent collection at all:

https://github.com/pytest-dev/pytest-xdist/blob/453cbd88ddb1ac4296b31df78b3c0f5c283d50b5/src/xdist/dsession.py#L89-L91 

So this use case seems to be more reasonable.

src/_pytest/hookspec.py
```diff
@@ -170,6 +170,8 @@ def pytest_load_initial_conftests(early_config, parser, args):
 def pytest_collection(session: "Session") -> Optional[Any]:
     """Perform the collection protocol for the given session.
 
+    The hook has to set `session.items` to a sequence of items.
```
nicoddemus, line 173: Following suggestion from https://github.com/pytest-dev/pytest/pull/6993#issuecomment-606882170:

```suggestion
    Usually plugins will implement this hook only to perform some action before
    collection, for example the terminal plugin will use this to start displaying
    the collection counter, so usually plugins return `None` from this hook after
    performing the desired operation.
    
    However a plugin might decide to override the collection completely,
    in which case it should return `True`, but then it is usually expected
    that this hook will also need to perform the following operations 
    that are usually part of the collection process:
    
      1. Call the pytest_collectstart hook.
      2. Call the pytest_collectreport hook.
      3. Call the pytest_collection_modifyitems hook.
      4. Call the pytest_collection_finish hook.
      5. Set session.testscollected to the amount of collect items.
      6. Set `session.items` to a list of items.
      
    If a plugin just wants to skip collection entirely, like `pytest-xdist`
    does for master nodes, then it is OK to not do anything other than
    returning `True` from here.
```

blueyed:
Will reword it myself again.
FWIW I've used it with some test IIRC to get a set of items in a straight way.

PR #6999, opened 2020-04-01
Simplify positional arguments compatibility code in pytest.fixture()

bluetech:
The dynamic scope feature added in 10bf6aac76d5060a0db4a94871d6dcf0a1 necessitated some wrangling of arguments in pytest.fixture(). In particular, it deprecated positional arguments in favor of keyword-only arguments, while keeping backward compatibility.

The way it did this avoided some code duplication but ended up being quite hard to follow and to annotate with types.

Replace it with some straightforward code, which is not very DRY but is simple and easy to remove when the time comes.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -1031,50 +1031,8 @@ class FixtureFunctionMarker:
         return function
 
 
-FIXTURE_ARGS_ORDER = ("scope", "params", "autouse", "ids", "name")
-
-
-def _parse_fixture_args(callable_or_scope, *args, **kwargs):
-    arguments = {
-        "scope": "function",
-        "params": None,
-        "autouse": False,
-        "ids": None,
-        "name": None,
-    }
-    kwargs = {
-        key: value for key, value in kwargs.items() if arguments.get(key) != value
-    }
-
-    fixture_function = None
-    if isinstance(callable_or_scope, str):
-        args = list(args)
-        args.insert(0, callable_or_scope)
-    else:
-        fixture_function = callable_or_scope
-
-    positionals = set()
-    for positional, argument_name in zip(args, FIXTURE_ARGS_ORDER):
-        arguments[argument_name] = positional
-        positionals.add(argument_name)
-
-    duplicated_kwargs = {kwarg for kwarg in kwargs.keys() if kwarg in positionals}
-    if duplicated_kwargs:
-        raise TypeError(
-            "The fixture arguments are defined as positional and keyword: {}. "
-            "Use only keyword arguments.".format(", ".join(duplicated_kwargs))
-        )
-
-    if positionals:
-        warnings.warn(FIXTURE_POSITIONAL_ARGUMENTS, stacklevel=2)
-
-    arguments.update(kwargs)
-
-    return fixture_function, arguments
-
-
 def fixture(
-    callable_or_scope=None,
+    fixture_function=None,
     *args,
     scope="function",
     params=None,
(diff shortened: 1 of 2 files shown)


src/_pytest/fixtures.py
```diff
@@ -1134,20 +1092,54 @@ def fixture(
...
+    if duplicated_args:
+        raise TypeError(
+            "The fixture arguments are defined as positional and keyword: {}. "
+            "Use only keyword arguments.".format(", ".join(duplicated_args))
+        )
+    if args:
+        warnings.warn(FIXTURE_POSITIONAL_ARGUMENTS, stacklevel=2)
+    # End backward compatiblity.
 
     if fixture_function and params is None and autouse is False:
```
bluetech, line 1144: BTW, I'm interested to know why the `params is None and autouse is False` condition exists here, if anyone knows... Seems weird to just ignore `fixture_function` if `params` or `autouse` are specified?

The-Compiler: Judging from the `# direct decoration` comment, that's to support

```python
@pytest.fixture
def func():
    ....
```

rather than

```python
@pytest.fixture(...)
def func():
    ...
```

so there's nothing being ignored, because there are no arguments given at all in that case.

bluetech: Right, the `if fixture_function` part I understand, what I'm having trouble figuring out is the `and params is None and autouse is False` part.

The reason I care about it is that I am working on adding type-annotating `pytest.fixture()`. Without this extra `params` condition, it is straightforward to distinguish the "direct decoration" case, but with it it is more difficult.

src/_pytest/fixtures.py
```diff
@@ -1134,20 +1092,54 @@ def fixture(
...
         # direct decoration
```
bluetech, line 1145: Also curious why `id` is omitted here (in the line below - GitHub doesn't let me comment on it).

src/_pytest/fixtures.py
```diff
@@ -1031,48 +1031,6 @@ def __call__(self, function):
         return function
 
 
-FIXTURE_ARGS_ORDER = ("scope", "params", "autouse", "ids", "name")
-
-
-def _parse_fixture_args(callable_or_scope, *args, **kwargs):
```
nicoddemus, line 1037: TBH I'm OK with this implementation; I like that it is general enough that we can reuse it eventually when needed.

Having said that if others prefer the new version I'm OK with that too.

bluetech: I don't think we would ever want to reuse it, because it's deprecated and new code should be keyword-only from the start. Any new parameters to `pytest.fixture()` would be keyword-only too. So this whole code block is "read-only" and I think it's better to have it straightforward and localized (and shorter).

bluetech:
After looking again I've made some further changes:

- Renamed `callable_or_scope` to `fixture_function`. The previous name is really internal-facing, let's use the proper name we would use without compat/deprecation. Also moved its compat check to inside the block marked with the comments.

- Moved `if params is not None: params = list(params)` to *after* the compat check -- was a slight bug in the previous code.

- Added tests to fix the coverage hopefully.

src/_pytest/fixtures.py
nicoddemus: Yeah I have that feeling like I wanted to reuse this somewhere, but can't really remember it right now. 😁

bluetech:
> It kinda hurts my eyes seeing so much if statements and makes me want to immediately revert to the previous code (as I'm the original author). But I need to agree, that this version is much easier to read and remove in the future.

:smile: Definitely ugly, but it's localized and easier to handle when adding type annotations, so I think the advantages outweigh the ugliness.

src/_pytest/fixtures.py
bluetech: Right. Still seems better to me not to ignore it, so any inconsistencies (like passing `ids` but not `params`) will be detected.

PR #7006, opened 2020-04-02
[5.4.x] Fix TerminalRepr instances to be hashable (#6988)

blueyed:
pytest-xdist assumes `ExceptionChainRepr` is hashable.

Fixes https://github.com/pytest-dev/pytest/issues/6925.
Fixes https://github.com/pytest-dev/pytest-xdist/issues/515.

(cherry picked from commit 20f6331afd3cf5af1d708f77aede08ad1a9d8661)

### DIFF
--- a/changelog/6925.bugfix.rst
+++ b/changelog/6925.bugfix.rst
@@ -1,0 +1,1 @@
+Fix TerminalRepr instances to be hashable again.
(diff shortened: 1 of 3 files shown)


bluetech:
FWIW, I think backports commits shouldn't require approval if they are simple cherry-picks, can just merge them since they have been approved already after all.

PR #7008, opened 2020-04-03
Fix tests: use explicit syspathinsert

blueyed:
Use `testdir.syspathinsert()` with multiprocessing tests:

- `test_chained_exceptions_no_reprcrash`
- `test_exception_handling_no_traceback`

This only works currently because `_importtestmodule` changes `sys.path`
as a side-effect.

It appears to be only required on Windows though - likely due to the
multiprocessing method used there.

### DIFF
--- a/testing/test_assertion.py
+++ b/testing/test_assertion.py
@@ -1264,6 +1264,7 @@ def test_exception_handling_no_traceback(testdir):
             multitask_job()
     """
     )
+    testdir.syspathinsert()
     result = testdir.runpytest(p1, "--tb=long")
     result.stdout.fnmatch_lines(
         [
--- a/testing/test_reports.py
+++ b/testing/test_reports.py
@@ -362,6 +362,7 @@ class TestReportSerialization:
             """
             )
 
+        testdir.syspathinsert()
         reprec = testdir.inline_run()
 
         reports = reprec.getreports("pytest_runtest_logreport")


bluetech:
Can you explain a bit more what problem this solves? I don't quite follow. The test passes the tmpfile path explicitly to `runpytest`, why does it need to be in the sys path?

blueyed:
The tests are crashing on Windows when not leaving "sys.path" mangled (via the side effect).
IIRC multiprocessing requires the module to be importable.
(I've noticed this with https://github.com/blueyed/pytest/pull/338)
It might also be a problem with using "importlib" import mode.
It is not strictly required for pytest as of now, but since I've had to debug it using pytest-timeout on CI I've thought it makes sense to do it "properly" already anyway.

bluetech:
Thanks for the explanation. Reading a bit, this is mentioned in the multiprocessing docs, for example the note at the bottom [here](https://docs.python.org/3/library/multiprocessing.html#using-a-pool-of-workers) (IIUC).

So LGTM, though would be nice to have comments above the new lines, like "multiprocessing on Windows requires the main module to be importable.".

blueyed:
It's more related to the method ("spawn") though, which can also be used on Linux (in theory): https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

Anyway, I think by now that `runpytest_inprocess` should default to `syspathinsert=True` - which brings it in line with `runpytest_subprocess`, which changes `PYTHONPATH` by default.
(A lot of tests currently only work due to the "pollution" of `sys.path` via `pyimport`)

PR #7015, opened 2020-04-04
Cleanup/move imports with tmpdir tests

blueyed:


bluetech:
The `test_timeout` faulthandler test is very flaky recently...

blueyed:
Yeah.
Maybe since 2cc3227f6?

bluetech:
Maybe...

Should we increase the extended sleep added in 81a9df6ed1ceda1a019323d10ff0bad2e9a2239c some more?

blueyed:
It's crashing - not being slow (AFAICT).

blueyed:
@bluetech 
An alternative would be to pin twisted for the oldattrs job (https://github.com/blueyed/pytest/commit/66d28b9279d59097457e3d7a7df16c8e03ee688a#diff-b91f3d5bd63fcd17221b267e851608e8R45) - I guess the current combination triggers some bug somewhere.

bluetech:
Hmm right, it seems like the timeout (very confusing name btw, it doesn't timeout anything...) works, but when it's running something in coverage crashes:

```
File "c:\hostedtoolcache\windows\python\3.8.2\x64\lib\concurrent\futures\_base.py"Windows fatal exception: access violation
```

I'm not sure about 2cc3227f6a36a0628b84301043ecf887973acb47 - why would twisted start to trigger it?

Curiously the `help` for the `faulthandler_timeout` option says "Not available on Windows.", but seems like support [was added](https://docs.python.org/3/library/faulthandler.html#faulthandler.dump_traceback_later) in Python 3.7.

My inclination is to allow it to fail on Windows...

PR #7016, opened 2020-04-04
doc: document inversed lines with terminal report hooks

blueyed:
It was surprising that `tryfirst=True` would not result in lines being
added to the beginning with `pytest_report_header`.
This is due to lines being reversed, and therefore the same applies to
`pytest_report_collectionfinish`.

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -523,10 +523,12 @@ def pytest_report_header(config, startdir):
     :param _pytest.config.Config config: pytest config object
     :param startdir: py.path object with the starting dir
 
-    .. note:
+    .. note::
 
-        The returned (set of) lines is reversed, i.e. `trylast=True` has to
-        be used when you want to have your line(s) at the beginning.
+        Lines returned by a plugin are displayed before those of plugins which
+        ran before it.
+        If you want to have your line(s) displayed first, use
+        :ref:`trylast=True <plugin-hookorder>`.
 
     .. note::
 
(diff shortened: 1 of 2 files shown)


src/_pytest/hookspec.py
```diff
@@ -523,6 +523,11 @@ def pytest_report_header(config, startdir):
     :param _pytest.config.Config config: pytest config object
     :param startdir: py.path object with the starting dir
 
+    .. note:
+
+        The returned (set of) lines is reversed, i.e. `trylast=True` has to
```
bluetech, line 528: Say Plugin1 returns Lines1 and Plugin2 returns Lines2. Which of the following describes what happens?

1. `"\n".join(reversed([Lines1, Lines2]))`
2. `"\n".join(reversed(Lines1 + Lines2))`
3. `"\n".join(reversed(Lines1) + reversed(Lines2))`

blueyed: `1.`

Ref: https://github.com/pytest-dev/pytest/blob/48c9f556ef17d67b820ac36944f5ee234dd624c6/src/_pytest/terminal.py#L640

bluetech: Got it. So I would rephrase slightly such that the first part of the sentence cannot be misconstrued. Suggestion:

Lines returned by a plugin are displayed *before* those of plugins which ran before it.
If you want to have your line(s) displayed first, use `trylast=True`.

PR #7017, opened 2020-04-04
Document the pytest_report_teststatus hook better and test uncovered functionality

bluetech:
This hook has some functionality to provide explicit markup for the test status. It seemed unused and wasn't tested, so I was tempted to remove it, but I found that the pytest-rerunfailures plugin uses it, so document it and add a test instead.

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -1,6 +1,9 @@
 """ hook specifications for pytest plugins, invoked from main.py and builtin plugins.  """
 from typing import Any
+from typing import Mapping
 from typing import Optional
+from typing import Tuple
+from typing import Union
 
 from pluggy import HookspecMarker
 
@@ -8,7 +11,9 @@ from .deprecated import COLLECT_DIRECTORY_HOOK
 from _pytest.compat import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from _pytest.config import Config
     from _pytest.main import Session
+    from _pytest.reports import BaseReport
 
 
 hookspec = HookspecMarker("pytest")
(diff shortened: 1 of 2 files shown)


src/_pytest/hookspec.py
```diff
@@ -546,12 +551,32 @@ def pytest_report_collectionfinish(config, startdir, items):
...
+    report: "BaseReport", config: "Config"
+) -> Tuple[
+    str, str, Union[str, Mapping[str, bool]],
+]:
+    """Return result-category, shortletter and verbose word for status
+    reporting.
 
-    :param _pytest.config.Config config: pytest config object
+    The result-category is a category in which to count the result, for
+    example "passed", "skipped", "error" or empty string.
```
blueyed, line 563:
```suggestion
    example "passed", "skipped", "error" or an empty string.
```

testing/test_terminal.py
```diff
@@ -306,6 +306,29 @@ def test_rewrite(self, testdir, monkeypatch):
         tr.rewrite("hey", erase=True)
         assert f.getvalue() == "hello" + "\r" + "hey" + (6 * " ")
 
+    def test_report_teststatus_explicit_markup(
+        self, testdir: Testdir, color_mapping
+    ) -> None:
+        """Test that TerminalReported handles markup explicitly provided by
```
blueyed, line 312:
```suggestion
        """Test that TerminalReporter handles markup explicitly provided by
```

src/_pytest/hookspec.py
blueyed: Also below, if it makes sense.

src/_pytest/hookspec.py
```diff
@@ -546,12 +551,32 @@ def pytest_report_collectionfinish(config, startdir, items):
...
+
+    The verbose word is shown as testing progresses in verbose mode, for
+    example "PASSED", "SKIPPED", "ERROR" or empty string.
+
+    pytest may style these implicitly according to the report outcome.
+    To provide explicit styling, return a tuple for the verbose word,
+    for example ``"rerun", "R", ("RERUN", {"yellow": True})``.
+
+    :param report: The report object whose status is to be returned.
+    :param config: The pytest config object.
```
blueyed, line 576: Types here would link it in docs.

bluetech:
Thanks @blueyed, I addressed your comments (though I used "the empty string").

PR #7019, opened 2020-04-05
Remove pyobj_property helper, inline it instead

bluetech:
It doesn't save much code but adds indirection which makes it a bit harder to follow and to type.


bluetech:
I have cloned all repos under the pytest-dev organization, and also the popular plugins with the "pytest" tag on GitHub (overall about a 100 repos). Couldn't find any external users of this.

PR #7023, opened 2020-04-05
Skip flaky test_faulthandler.py::test_timeout[True] test on Windows

### LINKED ISSUE #7022: test_faulthandler.py::test_timeout[True] occasionally crashes
This tests the `faulthandler_timeout` functionality which dumps stacktraces of all threads of tests that run for more than a specified amount of times.

The test sometimes crashes with `Windows fatal exception: access violation` while dumping the threads. This is probably a cpython bug.

Noticed on both `windows-py35` and `windows-py37` jobs at least. (Update: also on `pypy3-ubuntu`).

For now I think we should just skip it on Windows, unless anyone has better ideas or ability to investigate. I'll send a PR doing that.

bluetech:
Fixes #7022.

### DIFF
--- a/testing/test_faulthandler.py
+++ b/testing/test_faulthandler.py
@@ -2,6 +2,8 @@ import sys
 
 import pytest
 
+from _pytest.pytester import Testdir
+
 
 def test_enabled(testdir):
     """Test single crashing test displays a traceback."""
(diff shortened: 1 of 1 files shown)


testing/test_faulthandler.py
```diff
@@ -71,8 +84,6 @@ def test_timeout():
 
     result = testdir.runpytest_subprocess(*args)
     tb_output = "most recent call first"
-    if sys.version_info[:2] == (3, 3):
```
bluetech, line 74: (Also removed this, we don't support Python 3.3 so not needed anymore)

blueyed:
It also just failed with pypy in https://github.com/pytest-dev/pytest/runs/562142405.

I think it might be worth trying to bump the timeout - although the behavior on/with timeout appears to be buggy at least on pypy (no real/full stacktrace).

testing/test_faulthandler.py
```diff
@@ -49,8 +51,19 @@ def test_disabled():
...
 
-@pytest.mark.parametrize("enabled", [True, False])
-def test_timeout(testdir, enabled):
+@pytest.mark.parametrize(
+    "enabled",
+    [
+        pytest.param(
+            True,
+            marks=pytest.mark.skipif(
+                sys.platform.startswith("win"), reason="issue #7022",
```
blueyed, line 60: Please provide a "real" reason.  Issue numbers have to be looked up etc.  You can sill have it in parenthesis there then.

bluetech:
Not Windows specific, so closing.

PR #7029, opened 2020-04-06
Fix `test_popen_default_stdin_stderr_and_stdin_None` when run with `-s`

blueyed:


testing/test_pytester.py
```diff
@@ -679,19 +679,30 @@ def test_popen_default_stdin_stderr_and_stdin_None(testdir) -> None:
     # stdout, stderr default to pipes,
     # stdin can be None to not close the pipe, avoiding
     # "ValueError: flush of closed file" with `communicate()`.
+    #
+    # Wraps the test to not make it hang when run with "-s".
```
bluetech, line 683:
```suggestion
    # Wraps the test to make it not hang when run with "-s".
```

blueyed:
> Would an alternative to wrapping be to request an explicit `capfd` fixture in `test_popen_default_stdin_stderr_and_stdin_None`? The wrapping is too meta :)

That does not work/help.

bluetech:
> That does not work/help.

Curious! I thought it would achieve the same effect. Not worth the bother to understand why not though.

PR #7030, opened 2020-04-06
Increase test_faulthandler.py::test_timeout sleep duration on CI

### LINKED ISSUE #7022: test_faulthandler.py::test_timeout[True] occasionally crashes

bluetech:
This might help fix some flakiness.

Let's say it Fixes #7022 and reopen if not.

### DIFF
--- a/testing/test_faulthandler.py
+++ b/testing/test_faulthandler.py
@@ -58,7 +58,7 @@ def test_timeout(testdir, enabled):
         """
     import os, time
     def test_timeout():
-        time.sleep(1 if "CI" in os.environ else 0.1)
+        time.sleep(10 if "CI" in os.environ else 0.1)
     """
     )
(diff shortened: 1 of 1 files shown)


blueyed:
Should be reverted probably (https://github.com/pytest-dev/pytest/issues/7022#issuecomment-610215187).

PR #7034, opened 2020-04-06
Fix flaky TestDurations test

### LINKED ISSUE #7024: testing/acceptance_test.py::TestDurations::test_calls is flaky
It occasionally fails on pypy3-ubuntu, e.g. https://github.com/pytest-dev/pytest/runs/562181659.

We should try to avoid relying on actual timing there, as that is bound to be flaky.

bluetech:
Fixes #7024.

`TestDurations` tests the `--durations=N` functionality which reports N slowest tests, with durations <= 0.005s not shown by default.

The test relies on real `time.sleep()` (in addition to the code which uses `time.perf_counter()`) which makes it flaky and inconsistent between platforms.

Instead of trying to tweak it more, make it use fake time instead. The way it is done is a little hacky but seems to work.

### DIFF
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -2,8 +2,8 @@
 import bdb
 import os
 import sys
-from time import perf_counter
-from time import time
+from time import perf_counter  # Intentionally not `import time` to avoid being
+from time import time  # affected by tests which monkeypatch `time` (issue #185).
 from typing import Callable
 from typing import Dict
 from typing import List
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -6,9 +6,11 @@ import types
 import attr
 import py
 
+import _pytest.runner
 import pytest
 from _pytest.compat import importlib_metadata
 from _pytest.config import ExitCode
+from _pytest.monkeypatch import MonkeyPatch
 
 
 def prepend_pythonpath(*dirs):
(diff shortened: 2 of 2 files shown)


src/_pytest/runner.py
```diff
@@ -2,8 +2,8 @@
 import bdb
 import os
 import sys
-from time import perf_counter
-from time import time
+from time import perf_counter  # Intentionally not `import time` to avoid being
```
bluetech, line 5: Weird comment style to fool black into something reasonable. Note that this comment is worth having regardless.

testing/acceptance_test.py
```diff
@@ -893,61 +895,64 @@ def test_has_plugin(self, request):
...
+
+    current_time = 1586202699.9859412
+
+    def sleep(seconds: float) -> None:
+        nonlocal current_time
+        current_time += seconds
+
+    monkeypatch.setattr(time, "sleep", sleep)
+    monkeypatch.setattr(_pytest.runner, "time", lambda: current_time)
+    monkeypatch.setattr(_pytest.runner, "perf_counter", lambda: current_time)
```
blueyed, line 910: Have you considered using e.g. https://github.com/spulec/freezegun ?
(I think it is fine probably to not have an extra (test) dependency for now, but I assume there are other places where manipulating time programmatically would be good)

bluetech: No, didn't know about it. Doesn't look like it fits the exact use case though.

In my own project I never call date/time functions directly, only through some passed in object, so that it may be faked/mocked easily. But it can't be done here really.

BTW, Initially I went for doing

```py
class FakeTime:
    """A fake `time` module for TestDurations to make it not rely on actual time."""

    def __init__(self) -> None:
        self.current_time = 1586202699.9859412

    def sleep(self, seconds: float) -> None:
        self.current_time += seconds

    def time(self) -> float:
        return self.current_time

    def perf_counter(self) -> float:
        return self.current_time
```

and then

```py
testdir.monkeypatch.setattr(_pytest.runner, "time", FakeTime())
```

and that worked, but runs afoul of the "issue 185" thing.

PR #7036, opened 2020-04-07
Revert "Increase test_faulthandler.py::test_timeout sleep duration on CI"

bluetech:
Reverts pytest-dev/pytest#7030 - didn't fix the problem.


bluetech:
I was lazy and used "Revert" button from GitHub UI, but forgot it will create the branch on the main repo.

Anyway, this is a straight revert except it keeps the removal of the Python 3.3 check.

Since the CI fails due to the same issue it tried to fix, I will force-merge it and delete the branch.

PR #7047, opened 2020-04-08
Remove some no-longer-needed compat code in test_assertion

bluetech:


bluetech:
(CI failure is unrelated #7022)

PR #7048, opened 2020-04-08
Remove broken _reprcompare disabling fixture in test_assertrewrite.py

bluetech:
The `_pytest._code._reprcompare` that was referred to previously doesn't
exist -- it was moved to other places but wasn't update.

This regressed in f423ce9c016aff0d84c4a68f6d972833d032181e.

### DIFF
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -25,16 +25,6 @@ from _pytest.config import ExitCode
 from _pytest.pathlib import Path
 
 
-def setup_module(mod):
-    mod._old_reprcompare = util._reprcompare
-    _pytest._code._reprcompare = None
-
-
-def teardown_module(mod):
-    util._reprcompare = mod._old_reprcompare
-    del mod._old_reprcompare
-
-
 def rewrite(src):
     tree = ast.parse(src)
     rewrite_asserts(tree, src.encode())


bluetech:
TBH I didn't investigate what is the purpose of these fixtures -- if previous (non-working) behavior is what we want, then the `setup_module`/`teardown_module` should just be removed, instead of this PR.

cc @blueyed: you added the single test that seems affected by it, maybe you can tell which is the right fix.

bluetech:
Hmm curiously the changed test now fails but only on the macos jobs. I wonder with is different there...

blueyed:
> Hmm curiously the changed test now fails but only on the macos jobs. I wonder with is different there...

Likely due to tests being affected by if they are run with what verbosity..

blueyed:
> TBH I didn't investigate what is the purpose of these fixtures

Likely meant to save/restore the "state" for outer/other tests.

bluetech:
> Likely due to tests being affected by if they are run with what verbosity..

Right. The macos jobs are not configured any differently from the others as far as I can tell so that's still a bit strange.

> Likely meant to save/restore the "state" for outer/other tests.

So do you think we should keep the fix or just remove the fixture?

If we keep, looks like I would also need to change the `else` branch, but then the tests won't test at all what you wanted to test (in cb481a354a2e0c7585caf5d2b72ae036f3d39f1b).

blueyed:
Sorry, hadn't looked at the failure - always good to paste it for reference and the lazy ones.. ;)

I can reproduce the failing test using xdist (`-n 4`), what is what the macos job uses.

The problem is that with your patch it actually disables the rewriting now (partly) - the test you are adjusting is not how it should look after all / what I was meant to test there.

It also fails (still) when running it directly `testing/test_assertrewrite.py::TestAssertionRewrite::test_assertrepr_compare_same_width`.

Normally the hook is (re)stored for each test (see aa0328782f9c92d7497ff28f77972afe3cb5b8e2).
The `setup_module` however happens after that - i.e. it will be `None` typically only for the first test.  So running the test directly simulates what happens when using xdist, where some worker gets that test first.

I have some stash to improve/revisit those tests there, to actually parametrize them with different verbosity etc, but for now I think we should just remove the `setup/teardown_module` hooks there.

Test failure for reference:
```
     def test_assertrepr_compare_same_width(self, request):
         """Should use same width/truncation with same initial width."""
     
         def f():
             assert "1234567890" * 5 + "A" == "1234567890" * 5 + "B"
     
         msg = getmsg(f).splitlines()[0]
         if request.config.getoption("verbose") > 1:
             assert msg == (
                 "assert (('1234567890' * 5) + 'A') == (('1234567890' * 5) + 'B')"
             )
     
         else:
 >           assert msg == (
                 "assert '123456789012...901234567890A' "
                 "== '123456789012...901234567890B'"
             )
 E           assert "assert (('1234567890' * 5) + 'A') == (('1234567890' * 5) + 'B')" == "assert '123456789012...901234567890A' == '123456789012...901234567890B'"
```

bluetech:
Thanks @blueyed, I updated to remove the `setup_module`/`teardown_module` instead.

PR #7057, opened 2020-04-09
Skip flaky test test_faulthandler.py::test_timeout[True]

bluetech:
It occasionally crashes on CI, the reason seems out of our control, or at least we can't figure it out.

Refs #7022.


blueyed:
Note that there is an issue which appears to be related to it: https://github.com/pytest-dev/pytest/issues/7059

bluetech:
@blueyed I left a comment there, hopefully it's related.

For now I am going to merge this so we can have green CI again...

PR #7069, opened 2020-04-10
Remove note saying faulthandler_timeout is not available on Windows

bluetech:
I think it is available in all Python versions we support, but at least
since Python 3.7 the docs[0] say:

    Changed in version 3.7: This function is now always available.

so let's just remove the note.

[0] https://docs.python.org/3/library/faulthandler.html#faulthandler.dump_traceback_later


symonk:
LGTM

PR #7072, opened 2020-04-10
Remove asottile from TIDELIFT.rst

asottile:


nicoddemus:
I've also contacted Tidelift about this change, thanks! 👍

PR #7078, opened 2020-04-12
Fixes #7077: Added & improved docs for repr_failure() in Node & Collector

### LINKED ISSUE #7077: Node.repr_failure() not in the docs
I have implemented a collector for certain yaml files and was recently trying to look up the arguments of `pytest.Item.repr_failure()` that I'm overloading, because Pylint reported an argument mismatch.

It turned out that that `repr_failure()` does not show up in the pytest docs, also not in the base class `pytest.Node` where it is implemented. Looking at the source code, I found that the method does not show up in the docs because it does not have a docstring.

I suggest to add a docstring.

andy-maier:

### DIFF
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -364,6 +364,14 @@ class Node(metaclass=NodeMeta):
     def repr_failure(
         self, excinfo, style=None
     ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]:
+        """
+        Return a string representation of a collection failure.
+
+        :param excinfo: Exception information for the failure as a tuple
+            (type, value, traceback), see sys.exc_info().
+        :param style: Style of the representation ('long', 'short', 'auto',
+            None).
+        """
         return self._repr_failure_py(excinfo, style)
 
 
@@ -403,7 +411,12 @@ class Collector(Node):
         raise NotImplementedError("abstract")
 
     def repr_failure(self, excinfo):
-        """ represent a collection failure. """
+        """
+        Return a string representation of a collection failure.
+
+        :param excinfo: Exception information for the failure as a tuple
+            (type, value, traceback), see sys.exc_info().
+        """
         if excinfo.errisinstance(self.CollectError) and not self.config.getoption(
             "fulltrace", False
         ):
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -362,15 +362,12 @@ class Node(metaclass=NodeMeta):
         )
 
     def repr_failure(
-        self, excinfo, style=None
+        self, excinfo: ExceptionInfo[Union[Failed, FixtureLookupError]], style=None
     ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]:
         """
-        Return a string representation of a collection failure.
+        Return a representation of a collection or test failure.
 
-        :param excinfo: Exception information for the failure as a tuple
-            (type, value, traceback), see sys.exc_info().
-        :param style: Style of the representation ('long', 'short', 'auto',
-            None).
+        :param excinfo: Exception information for the failure.
         """
         return self._repr_failure_py(excinfo, style)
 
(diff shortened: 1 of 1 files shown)


src/_pytest/nodes.py
```diff
@@ -364,6 +364,14 @@ def _repr_failure_py(
     def repr_failure(
         self, excinfo, style=None
     ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]:
+        """
+        Return a string representation of a collection failure.
```
bluetech, line 368: Not only for collection failures, it is also called when tests themselves raise an exception.

src/_pytest/nodes.py
```diff
@@ -364,6 +364,14 @@ def _repr_failure_py(
...
+
+        :param excinfo: Exception information for the failure as a tuple
```
bluetech, line 370: The type of `excinfo` here is not the usual tuple, it's `ExceptionInfo[BaseException]`, which is a pytest type which wraps the tuple.

src/_pytest/nodes.py
```diff
@@ -364,6 +364,14 @@ def _repr_failure_py(
...
+            (type, value, traceback), see sys.exc_info().
+        :param style: Style of the representation ('long', 'short', 'auto',
```
bluetech, line 372: I think this parameter is a bit suspicious and I'm not really it's supposed to be used directly.

Do you use it? If not, I'll err on the side of conservatism and not document it, for now.

src/_pytest/nodes.py
bluetech: Additionally, "string representation" is not fully accurate - it can be a string or a terminal representation.

I would let the type annotation do the talking (it does appear in the documentation), and just omit the word "string".

src/_pytest/nodes.py
andy-maier: I don't use the `style` parameter in my override function. However, it was this very parameter that I first did not have in my override function and that was reported by Pylint. I'm fine with not documenting it in the docstring.

src/_pytest/nodes.py
andy-maier: Updated the description text to "Return a representation of a collection or test failure."

src/_pytest/nodes.py
andy-maier: The `_repr_failure_py()` method declares the type of `excinfo` as `ExceptionInfo[Union[Failed, FixtureLookupError]]`. Should that not be used for `repr_failure()` as well?

andy-maier:
At this point, mypy reports the following:
```
src/_pytest/nodes.py:410: error: Signature of "repr_failure" incompatible with supertype "Node"  [override]
src/_pytest/runner.py:310: error: Incompatible types in assignment (expression has type "CollectErrorRepr", variable has type "Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]")  [assignment]
src/_pytest/runner.py:311: error: Incompatible types in assignment (expression has type "Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]", variable has type "Optional[Tuple[str, Any, Any]]")  [assignment]
Found 3 errors in 2 files (checked 178 source files)
```

I suppose the first one is caused by the `style` parameter being in `Node.repr_failure()` but not in the overriding `Collector.repr_failure()`. This is a rightful warning because someone could invoke the base class interface against a derived class object and provide the `style` parameter. Not sure why this comes up now, though.

Guidance on what to do about these errors would be welcome.

src/_pytest/nodes.py
```diff
@@ -402,8 +407,14 @@ def collect(self):
         """
         raise NotImplementedError("abstract")
 
-    def repr_failure(self, excinfo):
-        """ represent a collection failure. """
+    def repr_failure(
+        self, excinfo: ExceptionInfo[Union[Failed, FixtureLookupError]]
+    ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]:
+        """
+        Return a representation of a collection or test failure.
```
bluetech, line 414: Here I would write only "a collection failure".

andy-maier: DONE

andy-maier:
In the latest commit, I removed the type annotations again because commit bluetech@ae2be82 i introduces them nicely.

PR #7079, opened 2020-04-12
Trying to find misspells in the source code and docs

sobolevn:
This is proof-of-concept PR with misspell finder. It might be helpful for docs and source code.
Let's see what it can find.


sobolevn:
Ok, I see. Actions are not run for this configuration.

PR #7081, opened 2020-04-12
Initial attempt at #7012; strip falsey values from strargs

symonk:
An attempt at fixing #7012, honestly I am not sure if this is the correct place to be doing this, the code base is completely new to me and pretty vast, feedback here is greatly appreciated by someone who knows the code base nicely.

I was able to recreate the issue of duplicate collection and my newly added tests confirms the fixed behaviour, regression suite remains passing.

Thanks for your time, have a great day and I look forward to your feedback.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -246,6 +246,7 @@ Segev Finer
 Serhii Mozghovyi
 Seth Junot
 Simon Gomizelj
+Simon Kerr
 Skylar Downes
 Srinivas Reddy Thatiparthy
 Stefan Farmbauer
--- a/changelog/7012.bugfix.rst
+++ b/changelog/7012.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed an issue where duplicate tests where collected when CLI passed empty quotes
--- a/src/_pytest/config/argparsing.py
+++ b/src/_pytest/config/argparsing.py
@@ -99,6 +99,7 @@ class Parser:
         namespace: Optional[argparse.Namespace] = None,
     ) -> argparse.Namespace:
         from _pytest._argcomplete import try_argcomplete
+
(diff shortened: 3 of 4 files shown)


symonk:
ok looks not so simple, codecov is not very happy with the drop :) I'll await some feedback on the overall approach, likely its in the completely wrong place

symonk:
"args" seems to have some immutability (provided by @attr (new to me)), should we do this check somewhere much sooner?

symonk:
@bluetech excellent thanks for the feedback, I will go back and investigate :) any recommendations on the best place for handling it?

bluetech:
> any recommendations on the best place for handling it?

Not out of hand, though I can look if needed.

> I am not so sure its truly a bug, but rather expected behaviour, for example:

Right, good find! According to #1187, this is expected behavior, though it does seem a little odd to me. On the other hand, if you don't want to run a test more than once, don't specify it more than once :smiley:

Perhaps the right move for now is for the reporter to move to reopen the issue.

bluetech:
Actually, it seems like that decision was overturned in #1609 and the behavior was changed to remove duplicate tests by default, with a `--keep-duplicates` flag to disable it. So it does seem like a bug.

My suggestion would be to look at `--keep-duplicates` (or rather what it *disables*) and find the bug there.

symonk:
Looking at something else right now then i will have a look at this again, would there be any merit in converting strargs to a set() ? will spend some time digging after.  I am having terrible trouble being able to debug the codebase using pycharm.  pdb is OK for now, but not ideal! it complains 'pytester' plugin is already registered with the pluginmanager when running from pycharm for me

symonk:
Free up again, will investigate a better solution. did some reading on the links provided - thanks again

Issue #6985, opened 2020-03-30
pytest.approx: Please improve repr readability


henzef:
It seems that pytest.approx repr is optimized for very high or very low tolerances, but is not as easily readable as it could be for values in between.
Some examples:

- `pytest.approx(42, abs=1)` repr is `42 ± 1.0e+00`, which could be just `42 ± 1`.
- `pytest.approx(5, rel=0.01)` repr is `5 ± 5.0e-02`, but could be `5 ± 0.05` or maybe even `5 ± 1%`
- `pytest.approx(24000, abs=500)` repr is `24000 ± 5.0e+02` and could be `24000 ± 500`
- removing accuracy example: `pytest.approx(1500, abs=555)` repr is `1500 ± 5.6e+02` and could be `1500 ± 555`

In these number ranges using the exact number instead of the exponential notation is both more accurate **and** easier readable for humans. Please adapt pytest.approx to switch it's repr style according to the input value ranges.

hodevin:
Hey, my team and I would like to look over this issue and resolve it if that's okay.

kalekundert:
Of course, PRs are always welcome!  Hopefully this one won't be too hard, either.

Zac-HD:
See https://docs.python.org/3/library/string.html#format-specification-mini-language - you probably want to use the `n` format specifier.

henzef:
Any news on this?

fazeelghafoor:
@Zac-HD is this issue still open, I would like to work on it

rishid:
I think this change has a regression on line 409. I am catching a FloatOperation exception as this change is comparing a float to a Decimal. 

```python
self = <[FloatOperation([<class 'decimal.FloatOperation'>]) raised in repr()] ApproxDecimal object at 0xffffa0403e00>

    def __repr__(self) -> str:
        """Return a string communicating both the expected value and the
        tolerance for the comparison being made.

        For example, ``1.0 ± 1e-6``, ``(3+4j) ± 5e-6 ∠ ±180°``.
        """
        # Don't show a tolerance for values that aren't compared using
        # tolerances, i.e. non-numerics and infinities. Need to call abs to
        # handle complex numbers, e.g. (inf + 1j).
        if (
            isinstance(self.expected, bool)
            or (not isinstance(self.expected, (Complex, Decimal)))
            or math.isinf(abs(self.expected) or isinstance(self.expected, bool))
        ):
            return str(self.expected)

[...]

PR #7123, opened 2020-04-25
Simplified Fix for Repr

wilsonj24:
Here is the link to the issue we had been working on. :https://github.com/pytest-dev/pytest/issues/6985.

It is a proposed fix for the repr readability

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -244,7 +244,7 @@ class ApproxScalar(ApproxBase):
             #     vetted_tolerance = "{:.1e}".format(self.tolerance)
             if isinstance(self.tolerance, int):
                 vetted_tolerance = "{:n}".format(self.tolerance)
-            elif self.rel is not None and not math.isinf(self.tolerance) and "e" not in str(self.rel):
+            elif self.rel is not None and not math.isinf(self.tolerance):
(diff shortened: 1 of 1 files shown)


bluetech:
Hi @wilsonj24,

Can you describe this PR a bit more? If it's a bug fix, what is the issue? If it's an improvement, in what way?

I suggest also to add tests to pytest's testing/ which cover all of the branches. That would help to demonstrate the changes as well.

bluetech:
I did not see your edit (edits doesn't send notifications).

I think the next step is to add some tests to demonstrate what this changes, and notify the issue that you have a proposed fix.

nicoddemus:
Hi @wilsonj24,

First of all we would like to thank you for your time and effort on working on this, the pytest team deeply appreciates it.

We noticed it has been awhile since you have updated this PR, however. pytest is a high activity project, with many issues/PRs being opened daily, so it is hard for us maintainers to track which PRs are ready for merging, for review, or need more attention.

So for those reasons we think it is best to close the PR for now, but with the only intention to cleanup our queue, it is by no means a rejection of your changes. We still encourage you to re-open this PR (it is just a click of a button away) when you are ready to get back to it.

Again we appreciate your time for working on this, and hope you might get back to this at a later time!

Cheers!

PR #7130, opened 2020-04-28
cacheprovider: speed up NFPlugin when --nf is not enabled

bluetech:
The code used an O(n^2) loop. Replace list with set to make it O(n).

For backward compatibility the filesystem cache still remains a list.

On this test:

    import pytest
    def test_foo(x): pass

run with `pytest --collect-only`:

Before: 0m1.251s
After:  0m0.921s

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -332,13 +332,13 @@ class NFPlugin:
     def __init__(self, config):
         self.config = config
         self.active = config.option.newfirst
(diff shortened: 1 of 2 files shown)


bluetech:
Due to the O(n^2), as N grows the slowness becomes more pronounced, e.g. changing 5000 -> 50000, the difference becomes 9.5s vs. 35.5s.

cc @boxed -- thought I'd let you know about performance fixes, let me know if you don't want me to.

boxed:
Thanks for the notification. I absolutely want you to!

PR #7134, opened 2020-04-29
config: replace usage of py.io.dupfile

bluetech:
As part of the effort to remove uses of `py`.


bluetech:
Thanks for reviewing @RonnyPfannschmidt.

The code isn't covered (it wasn't previously either). I'm not sure how much we want to cover `PYTEST_DEBUG`. I tested it manually anyway.

PR #7139, opened 2020-04-30
Allow tests to override "global" `log_level`

### LINKED ISSUE #7133: `caplog.set_level()` doesn't override `log_level`
A "global" log level set by `log_level` cannot be overridden at the test level.

Minimal Example:

```python
# setup.cfg
[tool:pytest]
log_level = CRITICAL

# test_example.py
import logging


def function_that_logs():
    logging.debug('message')


def test_logger(caplog):
    caplog.set_level(logging.DEBUG)
    function_that_logs()
    assert caplog.text
```

Running `pytest-master@077d1c3`, Python 3.7.6.

It looks like this is because `log_level` adjusts the level of `LogCaptureHandler()` in `_runtest_for_main()` whereas `caplog.set_level()` adjusts the level of a specific `Logger` (likely `Root`).

If I add `self.handler.setLevel(level)` to the end of `LogCaptureFixture.set_level()` the above example passes. Is this a suitable fix (assuming the corresponding cleanup is added too)? Or is this a misuse of `log_level`?

ruaridhw:
- Setting `log_level` via the CLI or `.ini` will control the "global" handler's log level
- This means any tests that rely on a lower level, cannot access those logs even if they wanted to
- It seems reasonable to set a "default" via the `.ini` but that the final say is given to `caplog`
- Fixes #7133

#### PR Checklist

### DIFF
--- a/changelog/7133.improvement.rst
+++ b/changelog/7133.improvement.rst
@@ -1,0 +1,1 @@
+``caplog.set_level()`` will now override any :confval:`log_level` set via the CLI or ``.ini``.
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -321,7 +321,8 @@ class LogCaptureFixture:
         """Creates a new funcarg."""
         self._item = item
         # dict of log name -> log level
-        self._initial_log_levels = {}  # type: Dict[str, int]
+        self._initial_logger_levels = {}  # type: Dict[str, int]
+        self._initial_handler_level = None  # type: Optional[int]
 
     def _finalize(self) -> None:
         """Finalizes the fixture.
@@ -329,7 +330,10 @@ class LogCaptureFixture:
         This restores the log levels changed by :meth:`set_level`.
         """
         # restore log levels
-        for logger_name, level in self._initial_log_levels.items():
+        if self._initial_handler_level is not None:
+            self.handler.setLevel(self._initial_handler_level)
+
+        for logger_name, level in self._initial_logger_levels.items():
             logger = logging.getLogger(logger_name)
             logger.setLevel(level)
 
@@ -413,8 +417,10 @@ class LogCaptureFixture:
         logger_name = logger
         logger = logging.getLogger(logger_name)
         # save the original log-level to restore it during teardown
-        self._initial_log_levels.setdefault(logger_name, logger.level)
+        self._initial_logger_levels.setdefault(logger_name, logger.level)
         logger.setLevel(level)
+        self._initial_handler_level = self.handler.level
+        self.handler.setLevel(level)
 
     @contextmanager
     def at_level(self, level, logger=None):
@@ -427,10 +433,13 @@ class LogCaptureFixture:
         logger = logging.getLogger(logger)
         orig_level = logger.level
         logger.setLevel(level)
+        handler_orig_level = self.handler.level
+        self.handler.setLevel(level)
         try:
             yield
         finally:
             logger.setLevel(orig_level)
+            self.handler.setLevel(handler_orig_level)
 
 
 @pytest.fixture
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -322,7 +322,6 @@ class LogCaptureFixture:
         self._item = item
         # dict of log name -> log level
         self._initial_logger_levels = {}  # type: Dict[str, int]
-        self._initial_handler_level = None  # type: Optional[int]
 
     def _finalize(self) -> None:
         """Finalizes the fixture.
@@ -330,9 +329,6 @@ class LogCaptureFixture:
         This restores the log levels changed by :meth:`set_level`.
         """
         # restore log levels
-        if self._initial_handler_level is not None:
-            self.handler.setLevel(self._initial_handler_level)
-
         for logger_name, level in self._initial_logger_levels.items():
             logger = logging.getLogger(logger_name)
             logger.setLevel(level)
@@ -419,7 +415,6 @@ class LogCaptureFixture:
         # save the original log-level to restore it during teardown
         self._initial_logger_levels.setdefault(logger_name, logger.level)
         logger.setLevel(level)
-        self._initial_handler_level = self.handler.level
         self.handler.setLevel(level)
 
     @contextmanager
--- a/testing/logging/test_fixture.py
+++ b/testing/logging/test_fixture.py
@@ -147,8 +147,9 @@ def test_ini_controls_global_log_level(testdir):
         def test_log_level_override(request, caplog):
             plugin = request.config.pluginmanager.getplugin('logging-plugin')
             assert plugin.log_level == logging.ERROR
-            logging.getLogger('catchlog').warning("WARNING message won't be shown")
-            logging.getLogger('catchlog').error("ERROR message will be shown")
+            logger = logging.getLogger('catchlog')
+            logger.warning("WARNING message won't be shown")
+            logger.error("ERROR message will be shown")
             assert 'WARNING' not in caplog.text
             assert 'ERROR' in caplog.text
     """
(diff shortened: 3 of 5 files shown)


testing/logging/test_fixture.py
```diff
@@ -137,3 +137,68 @@ def test_caplog_captures_for_all_stages(caplog, logging_during_setup_and_teardow
...
+            logging.getLogger('catchlog').debug("DEBUG message won't be shown")
+            logging.getLogger('catchlog').warning("WARNING message will be shown")
+            assert 'DEBUG' not in caplog.text
+            assert 'WARNING' in caplog.text
+    """
+    )
+    testdir.makeini(
+        """
+        [pytest]
+        log_level=INFO
```
ruaridhw, line 159: Perhaps a better test here would be to set this to `ERROR` and assert that a `WARNING` message is not shown? Now that I think about it, this test will probably pass without the `ini` file at all which kinda defeats the point.

testing/logging/test_fixture.py
```diff
@@ -137,3 +137,68 @@ def test_caplog_captures_for_all_stages(caplog, logging_during_setup_and_teardow
...
+    testdir.makepyfile(
+        """
+        import pytest
+        import logging
+        def test_log_level_override(request, caplog):
+            logger = 'catchlog'
+            plugin = request.config.pluginmanager.getplugin('logging-plugin')
+            assert plugin.log_level == logging.WARNING
+
+            logging.getLogger(logger).info("INFO message won't be shown")
```
thisch, line 178: Is there any reason why you don't assign the return value of `logging.getlogger('catchlog')` to the variable `logger`?

testing/logging/test_fixture.py
```diff
@@ -137,3 +137,68 @@ def test_caplog_captures_for_all_stages(caplog, logging_during_setup_and_teardow
...
+
+            caplog.set_level(logging.INFO, logger)
+
+            with caplog.at_level(logging.DEBUG, logger):
+                logging.getLogger(logger).debug("DEBUG message will be shown")
```
thisch, line 183: What happens If you raise after this line? Is the DEBUG message shown in the captured log output?

testing/logging/test_fixture.py
ruaridhw: No reason. I've updated accordingly

testing/logging/test_fixture.py
ruaridhw: Yes it will be. I've added another test to this effect

nicoddemus:
Thanks @ruaridhw!

testing/logging/test_fixture.py
thisch: Thx

testing/logging/test_fixture.py
thisch: IMO this is bug, but it is negligible. I think it would be better, if a new logging handler were used for caplog, whose level is modified in set_level/at_level.

src/_pytest/logging.py
```diff
@@ -321,15 +321,19 @@ def __init__(self, item) -> None:
         """Creates a new funcarg."""
         self._item = item
         # dict of log name -> log level
-        self._initial_log_levels = {}  # type: Dict[str, int]
+        self._initial_logger_levels = {}  # type: Dict[str, int]
+        self._initial_handler_level = None  # type: Optional[int]
```
thisch, line 325: I think it is not needed to reset the handler level, because each test gets a new instance of a LogCaptureHandler.

thisch: Sry I was wrong. It it indeed needed if the `at_level` ctx manager is used. However, in the `_finalize`  method it is not needed.

testing/logging/test_fixture.py
ruaridhw: > the levels passed to `caplog.set/at_level` may have an impact on the output log messages in the case of a failing test

@thisch, could you explain a little more how this is a bug? To me, this is what you would want. If you temporarily set the level to `DEBUG` to run a function and that function raises an exception, I would want to see the `DEBUG` messages output. Perhaps it could be a new feature to separate out the various handlers since I'm unsure this issue has been created by this PR.

src/_pytest/logging.py
ruaridhw: It's not used in the context manager though? If it's not needed in `_finalize` then surely it's not needed anywhere?

thisch: You are right. Then it is not needed anywhere.

changelog/7133.improvement.rst
```diff
@@ -0,0 +1 @@
+``caplog.set_level()`` will now override any :confval:`log_level` set via the CLI or ``.ini``.
```
thisch, line 1: Do you agree that tests must be written in a way that always succeed no matter what logging plugin settings are used on the cmd line or in `pytest.ini`? E.g. if a test passes when the user passed `--log-level=INFO` to `pytest` but fails if `--log-level=DEBUG` is used, then sth is wrong with the test.

In the current version of pytest writing tests that are independent of the logging settings (pytest.ini / cmd line) is not possible, but I plan to change that.

testing/logging/test_fixture.py
thisch: I have a different opinion about this. I think that a user who specifies `--log-level INFO`, would be surprised if the output (stdout) captured logs of a failing test, that uses `caplog.set_level(logging.DEBUG)`, contained `DEBUG` log messages.

As I said in https://github.com/pytest-dev/pytest/pull/7139#discussion_r419072623, I think that manually specified pytest logging settings on the cmd line or in `pytest.ini` must not lead to failing tests.

changelog/7133.improvement.rst
ruaridhw: > Do you agree that tests must be written in a way that always succeed no matter what logging plugin settings are used on the cmd line or in `pytest.ini`?

Yes, very much so, hence the need for this PR in the first place 😉 We need to override what `caplog` sees at the test level such that no matter the CLI or `.ini`, the test still passes.

testing/logging/test_fixture.py
ruaridhw: Ok, if i understand correctly, this is about separating out the logs that are printed to the log report upon failure from the logs that `caplog` sees? This feels like a new piece of work so I would be happy to tackle this but in a separate PR as I don't think this was previously possible with `pytest` anyway?

src/_pytest/logging.py
ruaridhw: Removed

testing/logging/test_fixture.py
ruaridhw: Let me know if this is what you had in mind and I can either include it here or raise a separate PR.

https://github.com/flexciton/pytest/commit/844259ac58817e740e0817ff4e7796d440b68e85

thisch: > Ok, if i understand correctly, this is about separating out the logs that are printed to the log report upon failure from the logs that `caplog` sees?

Exactly!

> This feels like a new piece of work so I would be happy to tackle this but in a separate PR as I don't think this was previously possible with pytest anyway?

Tackling this in a separate PR is fine with me.

thisch: > Let me know if this is what you had in mind and I can either include it here or raise a separate PR.
> 
> [flexciton@844259a](https://github.com/flexciton/pytest/commit/844259ac58817e740e0817ff4e7796d440b68e85)

Great. Thx for starting working on this! As I said before this should be moved to a separate PR.

changelog/7133.improvement.rst
thisch: FYI I have a WIP branch, where I try to get rid of the setLevel code for the root logger:
https://github.com/pytest-dev/pytest/compare/master...thisch:root_logger_refactoring

testing/logging/test_fixture.py
ruaridhw: FYI raised #7159 to continue this work

thisch:
@nicoddemus Can this be merged?

PR #7144, opened 2020-05-01
Let unittest frameworks deal with async functions

### LINKED ISSUE #6924: pytest 5.4+ doesn't handle IsolatedAsyncioTestCase from CPython core.

### LINKED ISSUE #7110: pytest now skips async tests
Here's a simple test case:

```
import asyncio
import asynctest


async def my_async_function(a, b):
    asyncio.sleep(1)
    return a + b


class TestMyFunctions(asynctest.TestCase):
    async def test_my_async_function(self):
        expected_value = 8
        value = await my_async_function(4, 4)
        self.assertEqual(expected_value, value)

```

Everything works using an older version of pytest (pytest-5.3.5).

```
(virtenv) me@me-Latitude-7490:~/src/quicktests/pytest_regression$ python -m pytest -vx
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.6.9, pytest-5.3.5, py-1.8.1, pluggy-0.13.1 -- /home/me/.virtualenvs/virtenv/bin/python
cachedir: .pytest_cache
rootdir: /home/me/src/quicktests/tests_pytest_regression
collected 1 item                                                                                                                                                                                           

test_simple.py::TestMyFunctions::test_my_async_function PASSED                                                                                                                                       [100%]

============================================================================================ 1 passed in 0.04s =============================================================================================
```

For some reason, the test is skipped in newer versions of pytest.

```
(virtenv) me@me-Latitude-7490:~/src/quicktests/pytest_regression$ pip install pytest --upgrade
(virtenv) me@me-Latitude-7490:~/src/quicktests/pytest_regression$ python -m pytest -vx
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.6.9, pytest-5.4.1, py-1.8.1, pluggy-0.13.1 -- /home/me/.virtualenvs/virtenv/bin/python
cachedir: .pytest_cache
rootdir: /home/me/src/quicktests/tests_pytest_regression
collected 1 item                                                                                                                                                                                           

test_simple.py::TestMyFunctions::test_my_async_function SKIPPED                                                                                                                                      [100%]

============================================================================================= warnings summary =============================================================================================
test_simple.py::TestMyFunctions::test_my_async_function
  /home/me/.virtualenvs/virtenv/lib/python3.6/site-packages/_pytest/python.py:171: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
    - pytest-asyncio
    - pytest-trio
    - pytest-tornasync
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))

```

Also, it seems like AioHTTPTestCase tests are NOT skipped.

nicoddemus:
Instead of trying to handle unittest-async functions in `pytest_pyfunc_call`,
let the `unittest` framework handle them instead.

This lets us remove the hack in `pytest_pyfunc_call`, with the upside that
we should support any unittest-async based framework.

Also included `asynctest` as test dependency for `py37-twisted`, and renamed
`twisted` to `unittestextras` to better reflect that we install `twisted` and
`asynctest` now instead of only `twisted`.

This also fixes the problem of cleanUp functions not being properly called for async functions.

Fix #7110
Fix #6924

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,7 +70,7 @@ jobs:
           - name: "windows-py38"
             python: "3.8"
             os: windows-latest
-            tox_env: "py38-twisted"
+            tox_env: "py38-unittestextras"
             use_coverage: true
 
           - name: "ubuntu-py35"
--- a/changelog/7110.bugfix.rst
+++ b/changelog/7110.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed regression: ``asyncbase.TestCase`` tests are executed correctly again.
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -93,6 +93,13 @@ def iscoroutinefunction(func: object) -> bool:
     return inspect.iscoroutinefunction(func) or getattr(func, "_is_coroutine", False)
 
 
+def is_async_function(func: object) -> bool:
+    """Return True if the given function seems to be an async function or async generator"""
+    return iscoroutinefunction(func) or (
+        sys.version_info >= (3, 6) and inspect.isasyncgenfunction(func)
+    )
+
+
 def getlocation(function, curdir=None) -> str:
     function = get_real_func(function)
     fn = py.path.local(inspect.getfile(function))
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -34,8 +34,8 @@ from _pytest.compat import get_default_arg_names
 from _pytest.compat import get_real_func
 from _pytest.compat import getimfunc
 from _pytest.compat import getlocation
+from _pytest.compat import is_async_function
 from _pytest.compat import is_generator
-from _pytest.compat import iscoroutinefunction
 from _pytest.compat import NOTSET
 from _pytest.compat import REGEX_TYPE
 from _pytest.compat import safe_getattr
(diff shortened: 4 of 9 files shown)


nicoddemus:
> I wonder if we perhaps should remove wrappers alltogether, are there exception handling hooks in the reporting for pdb to be hooked in

Definitely, but I prefer to do that in a separate PR (which I intend to tackle next).

Btw noticed that this also fixes #6924 regarding clean up functions not being called, so I updated both tests to also ensure that.

kasunsewwandika (commented):
-]()

PR #7153, opened 2020-05-02
Improve docs about junit_family warning message

nicoddemus:
From discussion in #6178

cc @bittner

### DIFF
--- a/doc/en/deprecations.rst
+++ b/doc/en/deprecations.rst
@@ -80,21 +80,39 @@ Note that ``from_parent`` should only be called with keyword arguments for the p
 
 .. deprecated:: 5.2
 
-The default value of ``junit_family`` option will change to ``xunit2`` in pytest 6.0, given
-that this is the version supported by default in modern tools that manipulate this type of file.
+The default value of ``junit_family`` option will change to ``xunit2`` in pytest 6.0, which
+is an update of the old ``xunit1`` format and is supported by default in modern tools
+that manipulate this type of file (for example, Jenkins, Azure Pipelines, etc.).
 
-In order to smooth the transition, pytest will issue a warning in case the ``--junitxml`` option
-is given in the command line but ``junit_family`` is not explicitly configured in ``pytest.ini``::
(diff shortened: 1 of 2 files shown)


doc/en/deprecations.rst
```diff
@@ -80,22 +80,35 @@ Note that ``from_parent`` should only be called with keyword arguments for the p
...
+that manipulate this type of file (for example, Jenkins, Azure Pipelines, etc.).
 
-In order to smooth the transition, pytest will issue a warning in case the ``--junitxml`` option
-is given in the command line but ``junit_family`` is not explicitly configured in ``pytest.ini``::
+Users are recommended to try the new ``xunit2`` format and see if their tooling that consumes the JUnit
+XML file supports it.
 
-    PytestDeprecationWarning: The 'junit_family' default value will change to 'xunit2' in pytest 6.0.
-      Add 'junit_family=legacy' to your pytest.ini file to silence this warning and make your suite compatible.
+To use the new format, update your ``pytest.ini``:
```
bittner, line 90: That suggests that by not adding `junit_family=xunit2` the modern format will not be used. That's not true IIUC, is it?

nicoddemus: It is True as per pytest <6.0, only in 6.0 the default will be `xunit2`.

bittner: Maybe mentioning this fact may be helpful, e.g.
```suggestion
To use the new format in pytest <6.0, update your ``pytest.ini``:
```

nicoddemus: Good point, thanks!

nicoddemus:
@bittner thanks for the review!

jaraco:
This looks good to me. Perhaps the docs should keep a list of environments that are known to work with the new format rather than to ask each user to perform the experiment. I’ve already validated Azure Pipelines.

nicoddemus:
> Perhaps the docs should keep a list of environments that are known to work with the new format

I will add Jenkins and Azure then, then users are welcome to contribute to update the list.

nicoddemus:
Thanks @jaraco and @bittner for the review.

PR #7157, opened 2020-05-03
Remove blueyed from TIDELIFT

nicoddemus:
@blueyed is no longer a member of the pytest-dev organization.


flub:
It appears I can't approve once it's merged.  But for the avoidance of doubt, I'm approving this.

PR #7167, opened 2020-05-05
testing: fix lint after merge of old branch

bluetech:

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -854,9 +854,12 @@ class TerminalReporter:
                 if len(locations) < 10:
                     return "\n".join(map(str, locations))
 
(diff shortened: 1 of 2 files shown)


bluetech:
PR #7020 added a test which incidentally also checks the order in which lines in the warning summary are printed. But turns out that order relied on dictionary order which is not guaranteed in Python 3.5, so it failed after merge in master.

The fix is straightforward - replace `collections.Counter` approach with `collections.OrderedDict`. Since I don't want to leave master broken, I'm merging without approval, but let me know if you have any comments.

nicoddemus:
> I'm merging without approval, but let me know if you have any comments.

Sure, for trivial changes like this, please go ahead! And thanks! 👍

PR #7172, opened 2020-05-06
Some easy micro optimizations to collection

bluetech:
Collectively, on this benchmark run with `--collect-only`:

```py
import pytest
@pytest.mark.parametrize("x", range(50000))
def test_foo(x): pass
```

Before:

```
============================ no tests ran in 11.37s ============================
         23775445 function calls (22920574 primitive calls) in 11.648 seconds
```

After:

```
============================ no tests ran in 10.51s ============================
         21725509 function calls (20870638 primitive calls) in 10.931 seconds
```

The first commit utilizes the fact that dicts are insertion-order preserving on Python>=3.7 to avoid  OrderedDict overhead when possible.

The second and third commits apply some micro optimizations to Node.

The fourth applies some minor optimizations to `PythonCollector.collect`.

### DIFF
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -381,3 +381,15 @@ else:
                 return self
             value = instance.__dict__[self.func.__name__] = self.func(instance)
             return value
+
+
+# Sometimes an algorithm needs a dict which yields items in the order in which
+# they were inserted when iterated. Since Python 3.7, `dict` preserves
+# insertion order. Since `dict` is faster and uses less memory than
+# `OrderedDict`, prefer to use it if possible.
+if sys.version_info >= (3, 7):
+    order_preserving_dict = dict
+else:
+    from collections import OrderedDict
+
+    order_preserving_dict = OrderedDict
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -6,7 +6,6 @@ ignores the external pytest-cache
 """
 import json
 import os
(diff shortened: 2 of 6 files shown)


src/_pytest/compat.py
```diff
@@ -381,3 +381,14 @@ def __get__(self, instance, owner=None):  # noqa: F811
                 return self
             value = instance.__dict__[self.func.__name__] = self.func(instance)
             return value
+
+
+# Sometimes an alogirthm needs an order-preserving dict. Since Python 3.7,
```
hugovk, line 386:
```suggestion
# Sometimes an algorithm needs an order-preserving dict. Since Python 3.7,
```

bluetech: Thanks; I also clarified the comment a bit.

bluetech:
I scaled back the last commit a bit after the CI kindly informed me that I had missed some details :smile: Doesn't affect the benchmark results however.

bluetech:
cc @boxed, #5516

boxed:
I had some time to check this and the other changes that have landed in master. I am comparing 5.4.1 and master:

before
after

It's slightly better but it's still very bad.

boxed:
I tried pytest 3.10.1 and it's at 7.8s for the parametrize_huge benchmark. Again still very bad, but a lot better than pytest 5. I even checked 3.6 because I have a memory of this being faster on our suite at work: 7.5s so basically the same.

PR #7176, opened 2020-05-07
warnings: speed up work done in catch_warnings_for_item()

bluetech:
When setting up the warnings capture, filter strings (with the general form `action:message:category:module:line`) are collected from the cmdline, ini and item and applied. This happens for every test and other cases.

To apply a string it needs to be parsed into a tuple, and it turns out this is slow. Since we already vendor the parsing code from Python's warnings.py, we can speed it up by caching the result. After splitting the parsing part from the applying part, the parsing is pure and is straightforward to cache.

An alternative is to parse ahead of time and reuse the result, however the caching solution turns out cleaner and more general in this case.

On this benchmark:

```py
import pytest
def test_foo(x): pass
```

Before:

```
============================ 5000 passed in 14.11s =============================
         14365646 function calls (13450775 primitive calls) in 14.536 seconds
```

After:

```
============================ 5000 passed in 13.61s =============================
         13290372 function calls (12375498 primitive calls) in 14.034 seconds
```

### DIFF
--- a/src/_pytest/warnings.py
+++ b/src/_pytest/warnings.py
@@ -1,34 +1,52 @@
+import re
 import sys
 import warnings
 from contextlib import contextmanager
+from functools import lru_cache
 from typing import Generator
+from typing import Tuple
 
 import pytest
+from _pytest.compat import TYPE_CHECKING
 from _pytest.main import Session
 
+if TYPE_CHECKING:
+    from typing_extensions import Type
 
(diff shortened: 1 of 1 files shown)


bluetech:
> lovely, at first glance the osx failure seems unrelated

Thanks!

Yeah, the failure happens also without this PR, I am debugging it.

PR #7179, opened 2020-05-07
Fix tests for python3.9

asottile:
Resolves #7161

- the text of `SyntaxError`s changed as part of the PEG parser rewrite ([bpo-40334](https://bugs.python.org/issue40334))
- `inspect.getsource` improved in python3.9 (utilizing the `ast` and `__qualname__`) ([bpo-35113](https://bugs.python.org/issue35113))

PR #7181, opened 2020-05-07
[4.6] Fix test_getfslineno for Python 3.9

hroncok:
`inspect.getsource` improved in Python 3.9 (utilizing the `ast` and `__qualname__`)
See https://bugs.python.org/issue35113

Resolves https://github.com/pytest-dev/pytest/issues/7161

### DIFF
--- a/testing/code/test_source.py
+++ b/testing/code/test_source.py
@@ -501,7 +501,7 @@ def test_getfslineno():
     class B(object):
         pass
 
-    B.__name__ = "B2"
+    B.__name__ = B.__qualname__ = "B2"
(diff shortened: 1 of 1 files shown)


nicoddemus:
Thanks a lot for the backport @hroncok!

We however have a documented procedure for the backport, see https://docs.pytest.org/en/latest/py27-py34-deprecation.html in the section "Backporting changes into 4.6". Could you attempt that procedure after #7179 gets merged? Thanks!

hroncok:
Sure.

hroncok:
See https://github.com/pytest-dev/pytest/pull/7190

PR #7184, opened 2020-05-07
testing: clean up parametrization in test_mark.py

bluetech:
Mostly trying to debug an odd CI failure in `test_keyword_option_custom` with `1 or 2` expression.


bluetech:
I am flabbergasted as to why it failed. But after adding the "Does the reverse also fail?" commit, the failure wasn't reproduced for 4 times. Could be luck, but hopefully not.

bluetech:
Of course it failed the fifth time... (after merge, in master). Guess I still need to figure it out.

PR #7188, opened 2020-05-07
[5.4] Merge pull request #7179 from asottile/py39

asottile:


asottile:
this one required a manual resolution where I dropped the `pypy` line in `tox.ini` since 5.4.x does not support python 2

PR #7197, opened 2020-05-08
[Port from 5.4.x] Revert "[parametrize] enforce explicit argnames declaration (#6330)"

nicoddemus:
Port from 5.4.x.

This was done by mistake directly into `5.4.x`, instead of the correct procedure which is to fix on `master` and *then* backporting it into `5.4.x`.


nicoddemus:
No longer needed, this change already landed on `master`.

PR #7201, opened 2020-05-08
Add myself to TIDELIFT

nicoddemus:
@asottile want to add yourself back as well?

@RonnyPfannschmidt @bluetech please also consider adding yourselves.

I plan to contact Tidelift next week, so it would be nice if we everyone who wants to join do so until then. 👍


asottile:
> @asottile want to add yourself back as well?

yes :) -- would you be ok if I edit your PR directly?

nicoddemus:
Sure go ahead! 👍

PR #7204, opened 2020-05-08
Issue 6866 - reason is optional in skipif marker

### LINKED ISSUE #6866: Not enforcing reason= with skipif/xfail markers
Currently, when e.g. `@pytest.mark.skipif(sys.platform != 'win32')` is used rather than `@pytest.mark.skipif("sys.platform != 'win32'")` (condition as a string), pytest enforces that a `reason="..."` is given.

As far as I can see, the only place that reason is used is with `-rs` /`-rx` which isn't used by default anyways. In other places where pytest can show additional information (like docstrings in `--fixtures` or descriptions in `--markers`), we show some default value instead of enforcing the user to specify something.

I was wondering: Should pytest do the same for `reason=` as well? Why not make it optional, and just report the location (without any description) in the skipped test summary for those cases?

CarycaKatarzyna:
It is implementation of proposition from #6866. The default reason is the same like in ``@pytest.mark.skipif`` with no argument.

Fixes #6866.

### DIFF
--- a/changelog/6866.breaking.rst
+++ b/changelog/6866.breaking.rst
@@ -1,0 +1,1 @@
+Reason is now an optional argument in ``@pytest.mark.skipif``.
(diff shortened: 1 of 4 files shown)


CarycaKatarzyna:
Could someone look at this implementation? Please let me know if something is missing or wrong.

PR #7207, opened 2020-05-10
WIP: implement custom arguments for pytester.pytest_spawn

Mattwmaster58:
One usecase for this is testing pytest-timeout. In order to test against the detection of `ipdb`, it's necessary to disable input capturing to even get ipdb up and running.

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1,6 +1,7 @@
 """(disabled by default) support for testing pytest and pytest plugins."""
 import collections.abc
 import gc
+import itertools
 import importlib
 import os
 import platform
@@ -1278,7 +1279,9 @@ class Testdir:
 
         """
         basetemp = self.tmpdir.mkdir("temp-pexpect")
-        invoke = " ".join([*map(str, self._getpytestargs()), *(args or [])])
+        if args is None:
+            args = []
+        invoke = " ".join(itertools.chain(map(str, self._getpytestargs()), args)
         cmd = "{} --basetemp={} {}".format(invoke, basetemp, string)
         return self.spawn(cmd, expect_timeout=expect_timeout)
 
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1267,7 +1267,7 @@ class Testdir:
         return self.run(*args, timeout=timeout)
 
     def spawn_pytest(
-        self, string: str, expect_timeout: float = 10.0, args: List[str]
+        self, string: str, expect_timeout: float = 10.0, args: List[str] = None
     ) -> "pexpect.spawn":
         """Run pytest with specified command line arguments using pexpect.
 
@@ -1278,7 +1278,7 @@ class Testdir:
 
         """
         basetemp = self.tmpdir.mkdir("temp-pexpect")
-        invoke = " ".join([*map(str, self._getpytestargs()), *args])
+        invoke = " ".join([*map(str, self._getpytestargs()), *(args or [])])
         cmd = "{} --basetemp={} {}".format(invoke, basetemp, string)
         return self.spawn(cmd, expect_timeout=expect_timeout)
 
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1267,9 +1267,9 @@ class Testdir:
         return self.run(*args, timeout=timeout)
 
     def spawn_pytest(
-        self, string: str, expect_timeout: float = 10.0
+        self, string: str, expect_timeout: float = 10.0, args: List[str]
     ) -> "pexpect.spawn":
-        """Run pytest using pexpect.
+        """Run pytest with specified command line arguments using pexpect.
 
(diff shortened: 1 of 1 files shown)


flub:
btw, it's not yet clear to me if this is needed for pytest-timeout.  So not to introduce unnecessary features which then need to be maintained let's not rush this and see how the need works out first.

nicoddemus:
Thanks @Mattwmaster58 and @flub.

> btw, it's not yet clear to me if this is needed for pytest-timeout. So not to introduce unnecessary features which then need to be maintained let's not rush this and see how the need works out first.

Sure, thanks. But how can we be sure? I mean, is there further testing being planned to see if the feature pans out?

Mattwmaster58:
> I mean, is there further testing being planned to see if the feature pans out?

I'm not quite sure what you mean here. Are you asking if I'm planning on adding tests for this capability if it becomes apparent this feature is necessary/useful enough?

flub:
> Sure, thanks. But how can we be sure? I mean, is there further testing being planned to see if the feature pans out?

The feature/test in pytest-timeout why this was added is still under development.

nicoddemus:
> I'm not quite sure what you mean here. Are you asking if I'm planning on adding tests for this capability if it becomes apparent this feature is necessary/useful enough?

Yeah flub nailed it what I meant: it was just not clear what was the plan for this PR at the time. 👍

src/_pytest/pytester.py
```diff
@@ -1278,7 +1278,7 @@ def spawn_pytest(
 
         """
         basetemp = self.tmpdir.mkdir("temp-pexpect")
-        invoke = " ".join(map(str, self._getpytestargs()))
+        invoke = " ".join([*map(str, self._getpytestargs()), *(args or [])])
```
flub, line 1281: I'd have written this using `itertools.chain` instead of the `*`-syntax and I also tend to do explicit `is None` instead of the `or` trick.  But that's not to say there's anything wrong with this.

Mattwmaster58:
```
if args is None:
    args = []

invoke = " ".join([*map(str, self._getpytestargs()), itertools.chain(args)])
```
Something like this?

flub:
```
if args is None:
    args = []
" ".join(itertools.chain(map(str, self._getpytestargs()), args)
```
is what I had in mind, otherwise you still have the (to me) weird `[*map()]` thing.

But honestly, that's me and may not represent the current internal pytest style.  It all achieves the same thing really.

Mattwmaster58:
I looked for other tests for pytest_spawn, but didn't find any IIRC

src/_pytest/pytester.py
Mattwmaster58: Woops, missed the second unpacking there

flub:
> I looked for other tests for pytest_spawn, but didn't find any IIRC

Never to late to improve things :)

And given the current unclosed parenthesis bug caught in a very strange way there clearly is a need for a test!

Mattwmaster58:
> And given the current unclosed parenthesis bug caught in a very strange way there clearly is a need for a test!

Well, it was caught without an explicit test, so I'd dispute that logic, but point taken otherwise. I guess that's what I get for editing through the web editor 😉 . I'll try my best to come up with some tests.

flub:
The test that springs to my mind is a test with a conftest which adds a
command line flag and a single test which check the flag was set.  If
the flag is passed the entire suite passes, otherwise it fails.  But
that are probably other reasonable approaches.

src/_pytest/pytester.py
```diff
@@ -1267,9 +1268,9 @@ def runpytest_subprocess(self, *args, timeout=None) -> RunResult:
         return self.run(*args, timeout=timeout)
 
     def spawn_pytest(
-        self, string: str, expect_timeout: float = 10.0
+        self, string: str, expect_timeout: float = 10.0, args: List[str] = None
```
bluetech, line 1271:
```suggestion
        self, string: str, expect_timeout: float = 10.0, args: Optional[Sequence[str]] = None
```

Alternatively, accepts `*args: str`. That would be more consistent with `runpytest` for example, and can replace `string`.

src/_pytest/pytester.py
```diff
@@ -1278,7 +1279,9 @@ def spawn_pytest(
 
         """
         basetemp = self.tmpdir.mkdir("temp-pexpect")
-        invoke = " ".join(map(str, self._getpytestargs()))
+        if args is None:
```
bluetech, line 1282: Building a command string by joining arguments like this is unsafe, should change `spawn` to accept args as well, and pass them to `pexpect.spawn` as separate arguments. According to the pexpect [doc](https://pexpect.readthedocs.io/en/stable/api/pexpect.html#pexpect.spawn.__init__) this is supported.

Backward compatibility might be a little tricky, but should be doable.

bluetech:
I'll close this for now to clear up the queue a bit, but feel free to reopen if you are planning to work on this again.

PR #7210, opened 2020-05-11
mark: deprecate a couple undocumented -k syntaxes

bluetech:
The `-k '-expr'` syntax is an old alias to `-k 'not expr'`. It's also not very convenient to have syntax that start with `-` on the CLI. Deprecate it and suggest replacing with `not`.

---

The `-k 'expr:'` syntax discards all items until the first match and keeps all subsequent, e.g. `-k foo` with

    test_bar
    test_foo
    test_baz

results in `test_foo`, `test_baz`. That's a bit weird, so deprecate it without a replacement. If someone complains we can reconsider or devise a better alternative.

### DIFF
--- a/changelog/7210.deprecation.rst
+++ b/changelog/7210.deprecation.rst
@@ -1,0 +1,5 @@
+The special ``-k '-expr'`` syntax to ``-k`` is deprecated. Use ``-k 'not expr'``
+instead.
+
+The special ``-k 'expr:'`` syntax to ``-k`` is deprecated. Please open an issue
+if you use this and want a replacement.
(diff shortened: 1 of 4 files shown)


bluetech:
> I don't even recall the use case for the : suffix

I can't think of a reasonable use case either. It interferes with some refactorings so hopefully we'll be able to remove it eventually.

NicWeiss:
I use it to run a single test case several times in a row without deselect tests
just an example: Python-m pytest-K-slow-K-long tests/my_test
This was convenient enough to avoid changing global settings for the entire test environment.

bluetech:
@NicWeiss I slightly edited your command to what I think you meant, as written the command wasn't legal.

First, you shouldn't pass `-k` multiple times because only the last one "wins".

Second, you can replace `-k-slow` and `-k-long` with `-k 'not slow'` and `-k 'not long'`, which is the non-deprecated syntax.

jonassmedegaard:
So how to skip two patterns?

This worked previously: `-k'-slow -long'` but it fails now to replace that with `-k'not slow not long'`...

jonassmedegaard:
Ah, I see now that is is clearly documented in `pytest --help` - for others looking here for answers, this works: `-k'not slow and not long'`

bluetech:
@jonassmedegaard right :+1: I'm actually surprised `-slow -long` without an `and` in the middle used to work.

Issue #7162, opened 2020-05-05
caplog fixture should not be affected by global log level


nicoddemus:
#7159 made me realize something: I think `caplog` by default should not be affected by the global log level.

For example:

```
import logging

def test(caplog):
    logging.info("info")
    logging.critical("critical")
    print(caplog.messages)
    assert 0
```

Running:

```
 λ pytest .tmp\test-caplevel.py
======================== test session starts ========================
collected 1 item

.tmp\test-caplevel.py F                                        [100%]

============================= FAILURES ==============================
_______________________________ test ________________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x000001827FB75358>

    def test(caplog):
        logging.info("info")
        logging.critical("critical")
        logging.debug("debug")
        print(caplog.messages)
>       assert 0
E       assert 0

.tmp\test-caplevel.py:8: AssertionError
----------------------- Captured stdout call ------------------------
['critical']
------------------------- Captured log call -------------------------
CRITICAL root:test-caplevel.py:5 critical
========================= 1 failed in 0.25s =========================
```

```
λ pytest .tmp\test-caplevel.py -o log_level=INFO
======================== test session starts ========================
platform win32 -- Python 3.6.6, pytest-5.4.1.dev154+gbe6849644.d20200501, py-1.8.1, pluggy-0.13.0
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('d:\\projects\\pytest\\.hypothesis\\examples')
rootdir: d:\projects\pytest\.tmp, inifile: pytest.ini
plugins: hypothesis-4.36.0, forked-1.1.1, xdist-1.31.0
collected 1 item

.tmp\test-caplevel.py F                                        [100%]

============================= FAILURES ==============================
_______________________________ test ________________________________

caplog = <_pytest.logging.LogCaptureFixture object at 0x0000014A1CCD32E8>

    def test(caplog):
        logging.info("info")
        logging.critical("critical")
        logging.debug("debug")
        print(caplog.messages)
>       assert 0
E       assert 0

.tmp\test-caplevel.py:8: AssertionError
----------------------- Captured stdout call ------------------------
['info', 'critical']
------------------------- Captured log call -------------------------
INFO     root:test-caplevel.py:4 info
CRITICAL root:test-caplevel.py:5 critical
====================== short test summary info ======================
FAILED .tmp\test-caplevel.py::test - assert 0
========================= 1 failed in 0.24s =========================
```

[...]

PR #7214, opened 2020-05-12
logging: some optimizations

bluetech:
Please see the commit messages for details.

On this benchmark:

```py
import pytest
def test_foo(x): pass
```

Before (only relevant entries shown):

```
============================ 5000 passed in 14.13s =============================
         13465818 function calls (12550967 primitive calls) in 14.571 seconds

   Ordered by: cumulative time

140003/60003 0.109    0.000    2.435    0.000 contextlib.py:108(__enter__)
    80000    0.098    0.000    2.224    0.000 logging.py:618(_runtest_for)
140003/60003 0.143    0.000    2.194    0.000 contextlib.py:117(__exit__)
    80000    0.257    0.000    1.721    0.000 logging.py:627(_runtest_for_main)
    30000    0.046    0.000    0.972    0.000 logging.py:685(pytest_runtest_logreport)
    40000    0.096    0.000    0.711    0.000 logging.py:302(__init__)
    10000    0.014    0.000    0.456    0.000 logging.py:668(pytest_runtest_teardown)
    80000    0.137    0.000    0.434    0.000 logging.py:271(catching_logs)
    10000    0.017    0.000    0.323    0.000 logging.py:673(pytest_runtest_logstart)
    10000    0.014    0.000    0.301    0.000 logging.py:680(pytest_runtest_logfinish)
    10000    0.013    0.000    0.285    0.000 logging.py:658(pytest_runtest_setup)
    10000    0.012    0.000    0.258    0.000 logging.py:663(pytest_runtest_call)
```

After:

```
============================ 5000 passed in 11.86s =============================
         10476806 function calls (9908460 primitive calls) in 12.164 seconds

   Ordered by: cumulative time

    35003    0.053    0.000    1.430    0.000 contextlib.py:117(__exit__)
    35003    0.037    0.000    1.155    0.000 contextlib.py:108(__enter__)
    30000    0.143    0.000    0.563    0.000 logging.py:622(_runtest_for)
    30000    0.067    0.000    0.223    0.000 logging.py:275(catching_logs)
    10000    0.020    0.000    0.222    0.000 logging.py:647(pytest_runtest_setup)
    10000    0.024    0.000    0.222    0.000 logging.py:657(pytest_runtest_teardown)
    10000    0.011    0.000    0.184    0.000 logging.py:653(pytest_runtest_call)
    15000    0.006    0.000    0.006    0.000 logging.py:675(pytest_runtest_logreport)
     5000    0.003    0.000    0.003    0.000 logging.py:663(pytest_runtest_logstart)
     5000    0.003    0.000    0.003    0.000 logging.py:670(pytest_runtest_logfinish)
```

### DIFF
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -619,46 +619,35 @@ class LoggingPlugin:
             else:
                 yield
 
-    def _runtest_for(
-        self, item: Optional[nodes.Item], when: str
-    ) -> Generator[None, None, None]:
+    def _runtest_for(self, item: nodes.Item, when: str) -> Generator[None, None, None]:
         """Implements the internals of pytest_runtest_xxx() hook."""
+        if self.log_cli_handler is not None:
+            self.log_cli_handler.set_when(when)
+
         with catching_logs(
             handler=self._runtest_log_capture_handler,
             formatter=self.formatter,
             level=self.log_level,
         ) as log_handler:
             log_handler.reset()
+            item._store[catch_log_handler_key] = log_handler
+            item._store[catch_log_records_key][when] = log_handler.records
 
-            if self.log_cli_handler:
-                self.log_cli_handler.set_when(when)
-
-            if item is not None:
-                empty = {}  # type: Dict[str, List[logging.LogRecord]]
-                item._store.setdefault(catch_log_records_key, empty)[
-                    when
-                ] = log_handler.records
-                item._store[catch_log_handler_key] = log_handler
-            try:
-                if self.log_file_handler is not None:
-                    with catching_logs(
-                        self.log_file_handler, level=self.log_file_level
-                    ):
-                        yield
-                else:
+            if self.log_file_handler is not None:
+                with catching_logs(self.log_file_handler, level=self.log_file_level):
                     yield
-            finally:
-                if item is not None and when == "teardown":
-                    del item._store[catch_log_records_key]
-                    del item._store[catch_log_handler_key]
+            else:
+                yield
 
-            if item is not None and self.print_logs:
+            if self.print_logs:
                 # Add a captured log section to the report.
                 log = log_handler.stream.getvalue().strip()
                 item.add_report_section(when, "log", log)
 
     @pytest.hookimpl(hookwrapper=True)
     def pytest_runtest_setup(self, item):
+        empty = {}  # type: Dict[str, List[logging.LogRecord]]
+        item._store[catch_log_records_key] = empty
         yield from self._runtest_for(item, "setup")
 
     @pytest.hookimpl(hookwrapper=True)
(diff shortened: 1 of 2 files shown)


bluetech:
The test failure is legitimate, so this needs some more work.

bluetech:
I fixed the test failure, and now it's also faster than the previous version (updated PR description), so it's ready for review.

testing/logging/test_fixture.py
```diff
@@ -136,4 +137,4 @@ def test_caplog_captures_for_all_stages(caplog, logging_during_setup_and_teardow
     assert [x.message for x in caplog.get_records("setup")] == ["a_setup_log"]
 
     # This reaches into private API, don't use this type of thing in real tests!
-    assert set(caplog._item.catch_log_handlers.keys()) == {"setup", "call"}
+    assert set(caplog._item._store[catch_log_records_key].keys()) == {"setup", "call"}
```
Zac-HD, line 140:
```suggestion
    assert set(caplog._item._store[catch_log_records_key]) == {"setup", "call"}
```

src/_pytest/logging.py
```diff
@@ -615,77 +619,63 @@ def pytest_collection(self) -> Generator[None, None, None]:
...
+        ) as log_handler:
+            log_handler.reset()
+            item._store[catch_log_handler_key] = log_handler
+            item._store[catch_log_records_key][when] = log_handler.records
+
             if self.log_file_handler is not None:
                 with catching_logs(self.log_file_handler, level=self.log_file_level):
                     yield
             else:
                 yield
```
thisch, line 640: @bluetech I like the usage of `yield-from` in this PR! WDYT about converting this `if self.log_file_handler` block (and also the block in `pytest_collection`) also into a generator from which you yield? Unfortunately I can't come up with a good name for this generator. What about `yield from self.forward_logs_to_logfile()`?

src/_pytest/logging.py
```diff
@@ -615,77 +619,63 @@ def pytest_collection(self) -> Generator[None, None, None]:
...
             if self.print_logs:
                 # Add a captured log section to the report.
                 log = log_handler.stream.getvalue().strip()
                 item.add_report_section(when, "log", log)
 
     @pytest.hookimpl(hookwrapper=True)
     def pytest_runtest_setup(self, item):
-        with self._runtest_for(item, "setup"):
-            yield
+        empty = {}  # type: Dict[str, List[logging.LogRecord]]
```
thisch, line 649: Maybe you can split up the type definiton of `catch_log_records_key` into two:
* `catch_log_records_map`
* `catch_log_records_key`

where the latter is defined as `Store[catch_log_records_key]`. Thus you can use `empty = {} # type catch_log_records_map`, here. TBH I don't know if this really works

bluetech:
Thanks for the reviews @Zac-HD and @thisch!

Upon looking at logging a bit more, and looking at the open issues/PRs, I think that the main change here ("reuse LoggingCaptureHandler") might have some interactions with them, so I decided to defer it for a bit. Instead, I took the other changes from here (and some additional ones) into a new PR #7224. Once/if it's merged, I will submit a few more changes that I have in mind (probably including the reuse optimization, and some changes to the fixture-plugin interaction).

testing/logging/test_fixture.py
bluetech: Thanks, I applied this in the other PR.

src/_pytest/logging.py
bluetech: I think we can make it mostly go away! But if not, I'll try your suggestion.

src/_pytest/logging.py
bluetech: These I think will also go away eventually. They are used for the plugin<->fixture interaction, but I think we should refactor this to be more like what the capture plugin does (for #7162).

Zac-HD:
Thanks @bluetech - I really appreciate the care you're putting into this, from design to code to PR writeups :heart_eyes:

Please \@-mention me in the new PR once tests are passing and I'll give it a priority review!

PR #7224, opened 2020-05-17
logging: some simplifications/cleanups

bluetech:
While looking at optimizing the logging plugin (since it shows up in profiles), and also at some issues and open PRs, I collected some simplifications and cleanups. I think they make the file a little cleaner and easier to understand and modify, and also a little faster, though this isn't the goal of this PR.

<details>
<summary>Benchmark</summary>

On this benchmark

```py
import pytest
def test_foo(x): pass
```

Before:

    ============================ 5000 passed in 13.66s =============================
            13345882 function calls (12431103 primitive calls) in 14.054 seconds

After:

    ============================ 5000 passed in 12.53s =============================
            11271238 function calls (10716458 primitive calls) in 12.938 seconds
</details>

It is better to review each commit separately -- one of the commits reorders some methods so the PR diff will be unreadable.

Includes a few breaking changes to private attributes, and one deprecation removal (`--no-print-logs`) that was planned for 6.0.

This will cause conflicts with some open PRs. This is intentional, since I think they will be easier to evaluate after these changes (and some others). I will comment on them separately once it becomes relevant.

### DIFF
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -273,12 +273,10 @@ def pytest_addoption(parser):
 
 
 @contextmanager
-def catching_logs(handler, formatter=None, level=None):
+def catching_logs(handler, level=None):
     """Context manager that prepares the whole logging machinery properly."""
     root_logger = logging.getLogger()
 
-    if formatter is not None:
-        handler.setFormatter(formatter)
     if level is not None:
         handler.setLevel(level)
 
@@ -303,15 +301,17 @@ def catching_logs(handler, formatter=None, level=None):
 class LogCaptureHandler(logging.StreamHandler):
     """A logging handler that stores log records and the log text."""
 
+    stream = None  # type: StringIO
+
     def __init__(self) -> None:
         """Creates a new log handler."""
-        logging.StreamHandler.__init__(self, StringIO())
+        super().__init__(StringIO())
         self.records = []  # type: List[logging.LogRecord]
 
     def emit(self, record: logging.LogRecord) -> None:
         """Keep the log records in a list in addition to the log text."""
         self.records.append(record)
-        logging.StreamHandler.emit(self, record)
+        super().emit(record)
 
     def reset(self) -> None:
         self.records = []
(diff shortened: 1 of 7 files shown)


bluetech:
Eek, looks like there is some issue on Python<=3.6. I'll take a look later.

src/_pytest/logging.py
```diff
@@ -614,27 +614,17 @@ def pytest_collection(self) -> Generator[None, None, None]:
...
-            try:
-                yield
-            finally:
-                if item is not None and when == "teardown":
-                    del item._store[catch_log_handlers_key]
-                    del item._store[catch_log_handler_key]
+            item._store[catch_log_handlers_key][when] = log_handler
+            item._store[catch_log_handler_key] = log_handler
+
+            yield
```
nicoddemus, line 638: `try/finally` probably required here?

changelog/TBD.breaking.rst
```diff
@@ -0,0 +1,2 @@
+The `item.catch_log_handler` and `item.catch_log_handlers` attributes, set by the
+logging plugin, are no longer available.
```
nicoddemus, line 2:
```suggestion
logging plugin and never meant to be public, are no longer available.
```

src/_pytest/logging.py
bluetech: The `yield` can't throw (pluggy only `send`s a Result in), so it's not really needed.

changelog/TBD2.breaking.rst
```diff
@@ -0,0 +1 @@
+The deprecated ``--no-print-logs`` option is removed. Use ``--show-capture`` instead.
```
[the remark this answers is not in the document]
bluetech: The README.rst the `changelog` directory says to use `breaking` for removals, and doesn't mention `removal`. Should I use `removal` nonetheless?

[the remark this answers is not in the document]
nicoddemus: Ahh sorry, brainfart on my part: we had "removal" in the past, but then changed to "breaking" afterwards. 👍 

So nevermind. 😬

PR #7227, opened 2020-05-18
logging: reuse LoggingCaptureHandler instance since it's expensive to create

bluetech:
Previously, a LoggingCaptureHandler was instantiated for each test's setup/call/teardown which turns out to be expensive.

Instead, only keep one instance and reset it between runs.

(Previously in #7214. Now I'm pretty sure it will interact well with the other PRs).

<details>
<summary>Benchmark</summary>

```py
import pytest
def test_foo(x): pass
```

Before:
```
============================ 5000 passed in 11.87s =============================
         10866808 function calls (10313462 primitive calls) in 12.163 seconds
```

After:

```
============================ 5000 passed in 11.25s =============================
         10431870 function calls (9878523 primitive calls) in 11.541 seconds
````
</details>


bluetech:
Thanks for the review @thisch.

> It would have been helpful for the review, if you mentioned why you changed the catch_log related code in the commit msg.

The reason is that there is no longer a different handler for each `_runtest_for`, so the previous solution of keeping references to the handlers of each phase no longer works. So instead, now it only keeps a reference to the *records* of each phase, which get replaces each phase and stay in tact.

PR #7238, opened 2020-05-22
A few tiny micro-optimizations/simplifications

bluetech:

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -443,8 +443,7 @@ class TerminalReporter:
             self.write_ensure_prefix(line, "")
             self.flush()
         elif self.showfspath:
-            fsid = nodeid.split("::")[0]
-            self.write_fspath_result(fsid, "")
+            self.write_fspath_result(nodeid, "")
             self.flush()
 
     def pytest_runtest_logreport(self, report: TestReport) -> None:
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -473,10 +473,7 @@ class TerminalReporter:
             else:
                 markup = {}
         if self.verbosity <= 0:
-            if not running_xdist and self.showfspath:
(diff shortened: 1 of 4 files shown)


src/_pytest/terminal.py
```diff
@@ -474,10 +473,7 @@ def pytest_runtest_logreport(self, report: TestReport) -> None:
             else:
                 markup = {}
         if self.verbosity <= 0:
-            if not running_xdist and self.showfspath:
-                self.write_fspath_result(rep.nodeid, letter, **markup)
```
nicoddemus, line 478: This is never reached?

src/_pytest/terminal.py
```diff
@@ -443,8 +443,7 @@ def pytest_runtest_logstart(self, nodeid, location):
             self.write_ensure_prefix(line, "")
             self.flush()
         elif self.showfspath:
-            fsid = nodeid.split("::")[0]
-            self.write_fspath_result(fsid, "")
```
nicoddemus, line 447: This is now different, correct? Or does is it have the same outcome?

src/_pytest/terminal.py
bluetech: It is reached, but has the same outcome as just the `write`.

What `write_fspath_result` does is check whether the path (extracted from the nodeid) is different than the last call to `write_fspath_result`, and if so, starts a new line and prints the path, before printing the argument.

The reason it's not needed here is that `pytest_runtest_logstart` already calls `write_fspath_result`, so the call here (`logreport`) doesn't.

Because the check it performs is mildly expensive, it's good to avoid it when possible.

src/_pytest/terminal.py
bluetech: `write_fspath_result` does this:

```py
fspath = self.config.rootdir.join(nodeid.split("::")[0])
```

(and doesn't use `nodeid` otherwise) so the split is redundant here.

PR #7249, opened 2020-05-23
Use deadsnakes/action@v1.0.0 to install python3.9 nightly

asottile:


.github/workflows/main.yml
```diff
@@ -132,14 +132,14 @@ jobs:
     - run: git fetch --prune --unshallow
     - name: Set up Python ${{ matrix.python }}
       uses: actions/setup-python@v2
+      if: matrix.python != '3.9-dev'
+      with:
+        python-version: ${{ matrix.python }}
+    - name: install python ${{ matrix.python }}
```
hugovk, line 138: Great to see an action for this and remove a bunch of config!

---

Perhaps keep this consistent with the non-3.9 name:
```suggestion
    - name: Set up Python ${{ matrix.python }}
```

nicoddemus, line 138: Minor nitpick, feel free to ignore:

The name of the standard action is `Set up Python {version}`, but for deadsnakes it is `install python {version}`. Maybe use something similar between them, say `Set up Python {version} (deadsnakes)` for the new action?

nicoddemus:
Thanks, great to see more standard actions for this 😁

Issue #6956, opened 2020-03-22
ConftestImportFailure: internal traceback when non-root conftest raises error


asottile:
an internal traceback is exposed when a nested conftest raises an error (doesn't happen for top level conftest failure)

- [x] a detailed description of the bug or suggestion
- [x] output of `pip list` from the virtual environment you are using
- [x] pytest and operating system versions
- [x] minimal example if possible

```console
$ pip freeze --all
attrs==19.3.0
importlib-metadata==1.5.0
more-itertools==8.2.0
packaging==20.3
pip==20.0.2
pluggy==0.13.1
py==1.8.1
pyparsing==2.4.6
pytest==5.4.1
setuptools==45.2.0
six==1.14.0
wcwidth==0.1.8
wheel==0.34.2
zipp==3.1.0
$ tree tests
tests
└── wat
    └── conftest.py

1 directory, 1 file
$ cat tests/wat/conftest.py 
import wat
```

```console
$ pytest tests
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /tmp/y
collected 0 items / 1 error                                                    

==================================== ERRORS ====================================
________________________ ERROR collecting test session _________________________
venv/lib/python3.6/site-packages/_pytest/config/__init__.py:495: in _importconftest
    return self._conftestpath2mod[key]
E   KeyError: PosixPath('/tmp/y/tests/wat/conftest.py')

During handling of the above exception, another exception occurred:
venv/lib/python3.6/site-packages/_pytest/config/__init__.py:501: in _importconftest
    mod = conftestpath.pyimport()
venv/lib/python3.6/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
venv/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:152: in exec_module
    exec(co, module.__dict__)
tests/wat/conftest.py:1: in <module>
    import wat
E   ModuleNotFoundError: No module named 'wat'

During handling of the above exception, another exception occurred:
venv/lib/python3.6/site-packages/py/_path/common.py:383: in visit
    for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
venv/lib/python3.6/site-packages/py/_path/common.py:424: in gen
    dirs = self.optsort([p for p in entries
venv/lib/python3.6/site-packages/py/_path/common.py:425: in <listcomp>
[...]

PR #7257, opened 2020-05-25
Prevent pytest from printing ConftestImportFailure traceback

DahlitzFlorian:
Closes: #6956 

@asottile I was not able to satisfy mypy so I added `# type: ignore` to the two lines. Let me know what you think :)

- [X] Create a new changelog file in the `changelog` folder, with a name like `<ISSUE NUMBER>.<TYPE>.rst`

### DIFF
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -29,7 +29,6 @@ from _pytest.mark.structures import Mark
 from _pytest.mark.structures import MarkDecorator
 from _pytest.mark.structures import NodeKeywords
 from _pytest.outcomes import fail
-from _pytest.outcomes import Failed
 from _pytest.store import Store
 
 if TYPE_CHECKING:
@@ -332,19 +331,15 @@ class Node(metaclass=NodeMeta):
         pass
 
     def _repr_failure_py(
-        self,
-        excinfo: ExceptionInfo[
-            Union[Failed, FixtureLookupError, ConftestImportFailure]
-        ],
-        style=None,
+        self, excinfo: ExceptionInfo[BaseException], style=None,
     ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]:
+        if isinstance(excinfo.value, ConftestImportFailure):
+            excinfo = ExceptionInfo(excinfo.value.excinfo)
         if isinstance(excinfo.value, fail.Exception):
             if not excinfo.value.pytrace:
                 return str(excinfo.value)
         if isinstance(excinfo.value, FixtureLookupError):
             return excinfo.value.formatrepr()
-        if isinstance(excinfo.value, ConftestImportFailure):
-            excinfo = ExceptionInfo(excinfo.value.excinfo)  # type: ignore
         if self.config.getoption("fulltrace", False):
             style = "long"
         else:
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -19,6 +19,7 @@ from _pytest._code.code import ReprExceptionInfo
 from _pytest.compat import cached_property
 from _pytest.compat import TYPE_CHECKING
 from _pytest.config import Config
+from _pytest.config import ConftestImportFailure
 from _pytest.config import PytestPluginManager
 from _pytest.deprecated import NODE_USE_FROM_PARENT
 from _pytest.fixtures import FixtureDef
@@ -340,6 +341,8 @@ class Node(metaclass=NodeMeta):
             return excinfo.value.formatrepr()
         if self.config.getoption("fulltrace", False):
             style = "long"
+        if excinfo.type is ConftestImportFailure:  # type: ignore
+            excinfo = ExceptionInfo.from_exc_info(excinfo.value.excinfo)  # type: ignore
         else:
             tb = _pytest._code.Traceback([excinfo.traceback[-1]])
             self._prunetraceback(excinfo)
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -332,17 +332,21 @@ class Node(metaclass=NodeMeta):
         pass
 
     def _repr_failure_py(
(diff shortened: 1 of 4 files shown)


src/_pytest/nodes.py
```diff
@@ -340,6 +341,8 @@ def _repr_failure_py(
             return excinfo.value.formatrepr()
         if self.config.getoption("fulltrace", False):
             style = "long"
+        if excinfo.type is ConftestImportFailure:  # type: ignore
+            excinfo = ExceptionInfo.from_exc_info(excinfo.value.excinfo)  # type: ignore
```
asottile, line 345: I believe the `excinfo` parameter should be changed to also include ConftestImportFailure in the Union and that'll make mypy happy

DahlitzFlorian:
I moved the check to the correct position in the method and add `ConftestImportFailure` to the Union. However, I could not remove the second `#type: ignore` as `ExceptionInfo` does not accept a tuple of the form `Tuple[Type[Exception], Exception, TracebackType]`, which `excinfo.value.excinfo` is.

src/_pytest/nodes.py
```diff
@@ -331,13 +332,19 @@ def _prunetraceback(self, excinfo):
...
+            Union[Failed, FixtureLookupError, ConftestImportFailure]
+        ],
+        style=None,
     ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]:
         if isinstance(excinfo.value, fail.Exception):
             if not excinfo.value.pytrace:
                 return str(excinfo.value)
         if isinstance(excinfo.value, FixtureLookupError):
             return excinfo.value.formatrepr()
+        if isinstance(excinfo.value, ConftestImportFailure):
```
bluetech, line 346: This should probably come first, in case the wrapped exception is a `fail.Exception` or a `FixtureLookupError` (if that's possible -- but even if not, probably a good idea to move).

testing/test_reports.py
```diff
@@ -396,6 +396,18 @@ def check_longrepr(longrepr):
...
 
+    def test_report_prevent_ConftestImportFailure_hiding_exception(self, testdir):
+        sub_dir = testdir.tmpdir.join("ns").ensure_dir()
+        sub_dir.join("conftest").new(ext=".py").write("import unknown")
+
+        result = testdir.runpytest_subprocess(".")
+        result.stdout.fnmatch_lines(
+            ["E   ModuleNotFoundError: No module named 'unknown'"]
+        )
+        result.stdout.no_fnmatch_line(
```
bluetech, line 407: To make this more robust, I'd match just `*ConftestImportFailure*`.

src/_pytest/nodes.py
```diff
@@ -331,13 +332,19 @@ def _prunetraceback(self, excinfo):
         pass
 
     def _repr_failure_py(
-        self, excinfo: ExceptionInfo[Union[Failed, FixtureLookupError]], style=None
+        self,
+        excinfo: ExceptionInfo[
```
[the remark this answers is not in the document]
DahlitzFlorian: done

src/_pytest/nodes.py
```diff
@@ -331,13 +332,19 @@ def _prunetraceback(self, excinfo):
...
+        ],
+        style=None,
     ) -> Union[str, ReprExceptionInfo, ExceptionChainRepr, FixtureLookupErrorRepr]:
         if isinstance(excinfo.value, fail.Exception):
             if not excinfo.value.pytrace:
                 return str(excinfo.value)
         if isinstance(excinfo.value, FixtureLookupError):
             return excinfo.value.formatrepr()
+        if isinstance(excinfo.value, ConftestImportFailure):
+            excinfo = ExceptionInfo(excinfo.value.excinfo)  # type: ignore
```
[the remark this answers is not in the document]
DahlitzFlorian: Solved by using `BaseException` for `excinfo`

src/_pytest/nodes.py
DahlitzFlorian: done

testing/test_reports.py
DahlitzFlorian: I needed to include `ERROR  - *ConftestImportFailure*` as otherwise tests including `ConftestImportFailure` in their name would fail.

DahlitzFlorian:
Thanks for the feedback @bluetech! I applied the suggested changes.

bluetech:
The code LGTM, just two things:

- The tests fail on Python 3.5 -- seems like the exception type is different there.

- According to the "patch" coverage report, the `if isinstance(excinfo.value, ConftestImportFailure):` branch isn't covered, which means the tests don't exercise it. Is it wrong or are the tests checking something else?

Also, did you see my comment about code sharing above? (It's OK if you don't want to dig into that)

DahlitzFlorian:
Thanks for the feedback!

- I don't know why it's failing, but I saw a bunch of PRs failing on Python 3.5 but got merged.
- The test covers exactly than bit of code, but I do not know why coverage flags it as not covered.
- Yes, I had a look at it, but wasn't able to understand it in a way to be able to extend it and share the code (sorry for that)

bluetech:
> I don't know why it's failing, but I saw a bunch of PRs failing on Python 3.5 but got merged.

The CI is gated on Python 3.5, i.e. the merge button is disabled if it is not passing.

Looking at the [CI logs](https://github.com/pytest-dev/pytest/pull/7257/checks?check_run_id=708615453), the issue is that the test asserts for

```
E   ModuleNotFoundError: No module named 'unknown'
```

but on Python 3.5 it is

```
E   ImportError: No module named 'unknown'
```

so one way to take care of it is to assert just for

```
E   *Error: No module named 'unknown'
```

> The test covers exactly than bit of code, but I do not know why coverage flags it as not covered.

Yes that's weird. I verified now that the `if` is indeed entered by the test.

DahlitzFlorian:
Fixed the py35 build.

DahlitzFlorian:
@asottile What do you think about the patch? Can it be merged?

PR #7258, opened 2020-05-25
Issue 6471 - Introduce new option to skip header and summary

### LINKED ISSUE #6471: Introduce new option to skip header and summary

piotrhm:
This pull request implements suggested approach for issue #6471.

Fix #6471.

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -117,15 +117,15 @@ def pytest_addoption(parser: Parser) -> None:
     )
     group._addoption(
         "--no-header",
-        action="count",
-        default=0,
+        action="store_true",
+        default=False,
         dest="no_header",
         help="disable header",
     )
     group._addoption(
         "--no-summary",
-        action="count",
-        default=0,
+        action="store_true",
+        default=False,
         dest="no_summary",
         help="disable summary",
     )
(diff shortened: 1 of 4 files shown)


src/_pytest/terminal.py
```diff
@@ -84,6 +84,20 @@ def pytest_addoption(parser):
         dest="verbose",
         help="increase verbosity.",
     )
+    group._addoption(
```
symonk, line 87: would using a store_true/store_false be more appropriate here? assuming we expect this only once on the command line (unlike -v) we behaves differently based on the count

symonk:
Hi @piotrhm thanks for working on this.  Could you add some test(s) to cover such scenarios as well if you get a chance? :)

src/_pytest/terminal.py
piotrhm: done

piotrhm:
@symonk thanks for comments! I added tests and changed ._addoption signature. I will be grateful if you take a look at current code.

nicoddemus:
Sorry, should have squashed this one... 😞

PR #7260, opened 2020-05-25
document autouse scope fixture priority 

### LINKED ISSUE #3404: Can't override an autouse fixture ordering
pytest 3.5.0 on Python 3.6.5 (macOS).  This result surprised me:

```python
import pytest

d = {'k': 'v_init'}

@pytest.fixture(autouse=True)
def fixauto(monkeypatch):
    monkeypatch.setitem(d, 'k', 'v_auto')

@pytest.fixture
def fix(monkeypatch):
    monkeypatch.setitem(d, 'k', 'v_fix')

def test_auto():
    assert d['k'] == 'v_auto'

def test_fix(fix):
    assert d['k'] == 'v_fix'

def test_fixauto_fix(fixauto, fix):
    assert d['k'] == 'v_fix'

def test_fix_fixauto(fix, fixauto):
    assert d['k'] == 'v_auto'
```

To be clear, first 3 tests are passing OK but the `test_fix_fixauto` fails.  I would have expected that explicitly injecting an autouse fixture could be used to override the usual order.  Feature or bug?

symonk:
Hi team, added some brief documentation explaining that autouse fixtures will be prioritized even when declared after non auto use fixtures in a test signature.  closes #3404

### DIFF
--- a/doc/en/example/fixtures/test_fixtures_autouse_order.py
+++ b/doc/en/example/fixtures/test_fixtures_autouse_order.py
@@ -16,9 +16,6 @@ def non_autouse_function_fixture():
 def test_fixture_autouse_ordering(
     non_autouse_function_fixture, autouse_function_fixture
 ):
-    """
-    You may (mistakenly) expect the test declaration order of fixtures to hold true here.
-    However autouse=True fixtures are always given priority on a per scope basis.
-    """
-    assert items == ["non_auto_use_fixture", "auto_use_fixture"]  # failure
-    assert items == ["auto_use_fixture", "non_auto_use_fixture"]  # pass
+    # You may (mistakenly) expect the test declaration order of fixtures to hold true here.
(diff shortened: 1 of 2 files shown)


symonk:
Whoops seems the tests in there get actually ran, will address (on mobile right now)

bluetech:
Hi @symonk, isn't that sufficiently covered a few lines above? Specifically by this line:

> a1: is a function-scoped autouse fixture: it will be instantiated before other fixtures within the same scope.

symonk:
Thanks @bluetech I do see `a1` now which outlines it,  I will defer to you on this one.  I think:

- An example in code can help articulate the example nicely

But yep, that a1 line basically covers it

bluetech:
It's also mentioned in the paragraph introducing the existing example:

> Autouse fixtures will be instantiated before explicitly used fixtures.

So I don't think another emphasis is needed.

symonk:
Yes sir, closing 👍

PR #7263, opened 2020-05-26
Fix typo in capture.rst documentation

xuiqzy:
Rename ``capfsysbinary`` to ``capsysbinary`` as the former does not exist as far as i can see.

### DIFF
--- a/doc/en/capture.rst
+++ b/doc/en/capture.rst
@@ -145,7 +145,7 @@ The return value from ``readouterr`` changed to a ``namedtuple`` with two attrib
 
 If the code under test writes non-textual data, you can capture this using
 the ``capsysbinary`` fixture which instead returns ``bytes`` from
-the ``readouterr`` method.  The ``capfsysbinary`` fixture is currently only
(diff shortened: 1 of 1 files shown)


doc/en/capture.rst
```diff
@@ -145,7 +145,7 @@ The return value from ``readouterr`` changed to a ``namedtuple`` with two attrib
 
 If the code under test writes non-textual data, you can capture this using
 the ``capsysbinary`` fixture which instead returns ``bytes`` from
-the ``readouterr`` method.  The ``capfsysbinary`` fixture is currently only
+the ``readouterr`` method.  The ``capsysbinary`` fixture is currently only
 available in python 3.
```
hugovk, line 149: Could zap this at the same time :)

```suggestion
available in Python 3.
```

symonk:
thanks @xuiqzy LGTM :) no access rights to merge/approve however, hopefully someone else will be along shortly

bluetech:
Since we only support Python 3, I think we should just remove this sentence entirely.

PR #7264, opened 2020-05-26
Improve our own wcwidth implementation and remove dependency on wcwidth package

bluetech:
`TerminalWriter`, imported recently from `py`, contains its own incomplete wcwidth (`char_with`/`get_line_width`) implementation. The `TerminalReporter` also needs this, but uses the external `wcwidth` package.

This commit brings the `TerminalWriter` implementation up-to-par with `wcwidth`, moves the implementation to a new file `_pytest._io.wcwidth` module which is used everywhere, and removes the dependency.

The differences compared to the `wcwidth` package are:

- Normalizes the string before counting.

- Uses Python's `unicodedata` instead of vendored Unicode tables. This means the data corresponds to the Python's version Unicode version instead of the `wcwidth`'s package version.

- Applies some optimizations.

### DIFF
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -14,7 +14,9 @@ import pluggy
 import py
 
 import _pytest.config
+import _pytest.terminal
 import pytest
+from _pytest._io.wcwidth import wcswidth
 from _pytest.config import ExitCode
 from _pytest.pytester import Testdir
 from _pytest.reports import BaseReport
@@ -2027,9 +2029,6 @@ def test_skip_reasons_folding():
 
 
 def test_line_with_reprcrash(monkeypatch):
-    import _pytest.terminal
-    from wcwidth import wcswidth
-
     mocked_verbose_word = "FAILED"
 
     mocked_pos = "some::nodeid"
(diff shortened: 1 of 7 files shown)


testing/test_terminal.py
```diff
@@ -2079,19 +2078,19 @@ def check(msg, width, expected):
     check("some\nmessage", 80, "FAILED some::nodeid - some")
 
     # Test unicode safety.
-    check("😄😄😄😄😄\n2nd line", 25, "FAILED some::nodeid - ...")
-    check("😄😄😄😄😄\n2nd line", 26, "FAILED some::nodeid - ...")
-    check("😄😄😄😄😄\n2nd line", 27, "FAILED some::nodeid - 😄...")
-    check("😄😄😄😄😄\n2nd line", 28, "FAILED some::nodeid - 😄...")
-    check("😄😄😄😄😄\n2nd line", 29, "FAILED some::nodeid - 😄😄...")
+    check("🧛🧛🧛🧛🧛\n2nd line", 25, "FAILED some::nodeid - ...")
```
bluetech, line 2081: This change is because the width of the 😄 character (U+1F604) happened to change from N (width 1) to W (width 2) in Unicode 9, but Python 3.5 uses Unicode 8, so the test started to fail there. I changed to a different character which was always wide.

bluetech: OK, Vampire doesn't work on Pypy (Unicode 9), it was only added in Unicode 10. So I changed to 🉐 now (Japanese “Bargain” Button) which should hopefully work in all versions..

nicoddemus:
> moves the implementation to a new file _pytest._io.wcwidth module which is used everywhere

By "moves" you mean we copied it from the `wcwidth` package? It is MIT Licensed, but if that's the case, I think we should add a docstring stating it was copied from their work.

testing/test_terminal.py
jquast: Yes, wcwidth wants to match modern terminal versions in behavior by default, to be most up-to-date. we are also trying to tackle multi-code point emoji and other difficulties, anyway best wishes from wcwidth package author.

PR #7265, opened 2020-05-26
Report the right package name for nested packages

s0undt3ch:

### DIFF
--- a/changelog/7265.bugfix.rst
+++ b/changelog/7265.bugfix.rst
@@ -1,0 +1,1 @@
+`_pytest.python.Package` instances now report the full package name instead of the top level package
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -4,6 +4,7 @@ import fnmatch
 import inspect
 import itertools
 import os
+import pathlib
 import sys
 import typing
 import warnings
@@ -571,7 +572,11 @@ class Package(Module):
         nodes.FSCollector.__init__(
             self, fspath, parent=parent, config=config, session=session, nodeid=nodeid
         )
-        self.name = os.path.basename(str(fspath.dirname))
+        self.name = str(
+            pathlib.Path(fspath.dirname).relative_to(
+                pathlib.Path(str(session.config.rootdir))
+            )
+        ).replace(os.sep, ".")
 
     def setup(self):
         # not using fixtures to call setup_module here because autouse fixtures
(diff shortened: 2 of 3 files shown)


nicoddemus:
Thanks @s0undt3ch! 

Not sure this is wanted though... the name of all other nodes are only the leaf names, not the full name.

s0undt3ch:
Hmm, sure, but it just reads wrong.
Do you suggest another attribute?

nicoddemus:
To elaborate a bit more, I believe we should be consistent:

```
<Package tests>
  <Module __init__.py>
    <Function test_init>
  <Module test_foo.py>
    <Function test_foo>
<Package tests.pkg>
  <Module test_foo.py>
    <Function test_foo>
```

Note that `Module` does not contain the full name, so `Package` should follow suit.

I believe we can introduce a new `@property` in `Node` that gives the `full_name` or `qual_name` in Python terms, which would iterate over all parent nodes constructing the proper full qualified name:

```python
assert tests_pkg.name == "pkg"
assert tests_pkg.qual_name == "tests.pkg"
```

And would even work for any node:

```python
assert test_foo_func.name == "test_foo"
assert test_foo_func.qual_name == "tests.pkg.test_foo.test_foo"
```

What do you think?

nicoddemus:
@s0undt3ch gentle ping. 😁

s0undt3ch:
Sorry, totally missed your comment.
Yes, I believe `qual_name` is indeed a better solution.
Is the above "pseudo-code" or you actually worked on it and this PR should just be closed?

nicoddemus:
> Is the above "pseudo-code" or you actually worked on it and this PR should just be closed?

Pseudo-code 😁 

Feel free to close this PR and re-open another one with my suggestion, or create an issue if you feel you won't have to tackle it. 👍

s0undt3ch:
I don't mind tackling it, and since it's not a high priority, ie, no one else needs it right away, I can try and squeeze it on my free time slots.
Obrigado pela sugestão da direção a tomar ;)

nicoddemus:
Sure thing, let's close this for now then, I've created https://github.com/pytest-dev/pytest/issues/7365 to track the new feature suggested here. 

> Obrigado pela sugestão da direção a tomar ;)

Claro, sem problemas, valeu! 😁

symonk:
@nicoddemus  gonna grab this if its ok

PR #7274, opened 2020-05-28
code: remove last usage of py.error

bluetech:
`str(self.path)` can't raise at all, so it can just be removed.

https://github.com/pytest-dev/pytest/blob/b3db440d4c41c91dd761dee98d7014149e1d7c08/src/_pytest/_code/code.py#L74-L90

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -268,10 +268,6 @@ class TracebackEntry:
         return tbh
 
     def __str__(self) -> str:
-        try:
-            fn = str(self.path)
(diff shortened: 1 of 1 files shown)


bluetech:
`py.error.Error` inherits from `EnvironmentError`, which is an alias for `OSError` in Python 3. The `p.check()` call is wrapped with `except OSError`. So the `except py.error.Error` could never have been reached anyway.

bluetech:
Thanks for reviewing @RonnyPfannschmidt!

PR #7280, opened 2020-05-29
Issue 7202 - Point development guide to contributing section

### LINKED ISSUE #7202: Contributing guide vs. development guide
We have two documents that serve a similar purpose:

- https://docs.pytest.org/en/latest/contributing.html (source: https://github.com/pytest-dev/pytest/blob/master/CONTRIBUTING.rst)
- https://docs.pytest.org/en/latest/development_guide.html (source: https://github.com/pytest-dev/pytest/blob/master/doc/en/development_guide.rst)

The development guide includes some of its own content, and also embeds https://github.com/pytest-dev/pytest/blob/master/RELEASING.rst.

The development guide's own content seems mostly outdated or redundant with the CONTRIBUTING.

My proposal to fix this is:

- Add a new releasing section to the docs, embedding RELEASING.rst. Or just remove it -- I'm not sure that anyone doing pytest releases will need to read the instructions on the rendered docs..?
- Change the development guide to only link to the contributing and releasing guides, mostly to avoid breaking the URL.

mcsitter:
- [X] Create a new changelog file
- [X] Tox docs environment passed successfully

Resolves #7202 

As proposed the development guide now points to the contributing section.

Instead of embedding `RELEASING.rst` a link to GitHub is now present. Since only a few people who contribute need to now the release procedure this solution seems appropriate. Adding a new releasing section to the documentation would provoke the same maintenance issues the development guide has now to occur on the releasing section.

I also allowed myself to add a note about the maintenance of the development guide.

Open for feedback!


.pre-commit-config.yaml
```diff
@@ -1,72 +1,72 @@
-repos:
```
bluetech, line 1: Please remove the changes to this file (from Unix line endings to Windows line endings). Probably something your editor or git did unintentionally.

doc/en/development_guide.rst
```diff
@@ -2,59 +2,9 @@
...
-
-* ``temporary: EP2017 sprint``: candidate issues or PRs tackled during the EuroPython 2017
-
-Issues created at those events should have other relevant labels added as well.
-
-Those labels should be removed after they are no longer relevant.
-
-
-.. include:: ../../RELEASING.rst
+This site will not be maintained and may be removed in the future.
```
bluetech, line 10: I think we can remove this sentence. Just to keep the link stable, I don't think we should remove it, and because it only contains a couple of links, there is not much to maintain :smile:

.pre-commit-config.yaml
mcsitter: Fixed!

Pre-commit threw an error and recommended to add the YAML to the commit. This was an unforseen consequence. Different clients do not always show when the line endings of a file changed or at least not as explicit as is shown here.

doc/en/development_guide.rst
mcsitter: Fixed!

webknjaz (commented):
^

webknjaz:
Why was the mention of the label management removed? @nicoddemus do you know? Is the automation still in use?

cc @RonnyPfannschmidt

nicoddemus:
TBH I completely forgot about that, and suspect we have never used it really, so that `labels.toml` file has not been used in awhile.

PR #7288, opened 2020-05-31
warnings: fix missing None in existing hook & add some docs

bluetech:
This covers some loose ends from #7255 and fixes the issue reported later in #4049.

### DIFF
--- a/doc/en/deprecations.rst
+++ b/doc/en/deprecations.rst
@@ -20,6 +20,17 @@ Below is a complete list of all pytest features which are considered deprecated.
 :ref:`standard warning filters <warnings>`.
 
 
+The ``pytest_warning_captured`` hook
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 6.0
+
+This hook has an `item` parameter which cannot be serialized by `pytest-xdist`.
+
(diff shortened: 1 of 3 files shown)


bluetech:
cc @gnikonorov. Wanted to submit this before we forget about this topic.

doc/en/deprecations.rst
```diff
@@ -20,6 +20,17 @@ Below is a complete list of all pytest features which are considered deprecated.
 :ref:`standard warning filters <warnings>`.
 
 
+The ``pytest_warning_captured`` hook
```
gnikonorov, line 23: I didn't realize we had a deprecations file. Thanks @bluetech

src/_pytest/warnings.py
```diff
@@ -112,7 +112,12 @@ def catch_warnings_for_item(config, ihook, when, item):
 
         for warning_message in log:
             ihook.pytest_warning_captured.call_historic(
```
gnikonorov, line 114: Do we usually update deprecated hooks? Asking for my own knowledge. I wasn't sure when I made the change for `pytest_warning_recorded`

bluetech: I think it makes sense to update deprecated hooks if it's a simple bugfix like this.

PR #7290, opened 2020-06-01
Issue 993 - improve handling exceptions in conftest

### LINKED ISSUE #993: ImportError raised by plugins make pytest return EXIT_TESTSFAILED
Currently there's some inconsistency on which exit code is returned by pytest when import errors happen during startup.

**example 1**

``` python
# contents of conftest.py
raise ImportError

# contents of test_foo.py
def test_foo():
    pass
```

Running `py.test` in this directory produces error code `4 - EXIT_USAGEERROR`.

**example 2**

``` python
# contents of conftest.py
pytest_plugins = 'plugin'

# contents of plugin.py
raise ImportError

# contents of test_foo.py
def test_foo():
    pass
```

Here `py.test` produces error code `1 - EXIT_TESTSFAILED`.

Shouldn't they all return the same error code? Furthermore, shouldn't the error code be `3 - EXIT_INTERNALERROR` instead?

CarycaKatarzyna:
Fixes #993. 
I've changed output of exceptions like in example 2 from #993. Now the output of example 2 is:
```terminal
ImportError while loading conftest '/home/xyz/conftest.py',
    while importing plugin 'plugin'.
plugin.py:1: in <module>
    raise ImportError
E   ImportError
```
and is more similar to output of example 1. The output of example 1 is all the time:
```terminal
ImportError while loading conftest '/home/xyz/conftest.py'.
plugin.py:1: in <module>
    raise ImportError
E   ImportError
```

I also changed exit code from ``ExitCode.USAGE_ERROR`` to ``ExitCode.INTERNAL_ERROR`` for all exceptions during conftest's startup. Please tell me what you think about that.


testing/test_pluginmanager.py
```diff
@@ -188,7 +188,7 @@ def test_traceback():
     with pytest.raises(ImportError) as excinfo:
         pytestpm.import_plugin("qwe")
 
-    assert str(excinfo.value).endswith(
+    assert str(excinfo.value).find(
```
asottile, line 191: find is not a helpful assertion:

```pycon
>>> assert 'watwatwat'.find('asdf')
>>>
```

src/_pytest/config/__init__.py
```diff
@@ -123,7 +145,7 @@ def main(args=None, plugins=None) -> Union[int, ExitCode]:
             formatted_tb = str(exc_repr)
             for line in formatted_tb.splitlines():
                 tw.line(line.rstrip(), red=True)
-            return ExitCode.USAGE_ERROR
+            return ExitCode.INTERNAL_ERROR
```
asottile, line 148: INTERNAL_ERROR I believe is reserved for when pytest itself crashes -- this is still a user-space error so I think the return code should stay the same here?

src/_pytest/config/__init__.py
```diff
@@ -651,6 +671,11 @@ def import_plugin(self, modname, consider_entry_points=False):
                 self.hook,
                 stacklevel=2,
             )
+        except Exception as e:
+            raise (PluginImportFailure(modname, sys.exc_info())).with_traceback(
```
asottile, line 675: there's an extra set of parens here that aren't needed -- also black kinda mangled this didn't it :S

src/_pytest/config/__init__.py
```diff
@@ -86,17 +86,30 @@ class ExitCode(enum.IntEnum):
...
 
     def __str__(self):
         return "{}: {} (from {})".format(
             self.excinfo[0].__name__, self.excinfo[1], self.path
         )
 
 
+class PluginImportFailure(ImportError):
+    def __init__(self, plugin, excinfo):
+        super().__init__(plugin, excinfo)
```
asottile, line 103: I think this `super()` call is maybe not right?  I think this puts the wrong data into `ImportError`'s members -- should this extend a different type?

src/_pytest/config/__init__.py
CarycaKatarzyna: In #993 was suggestion to change ``USAGE_ERROR`` into ``INTERNAL_ERROR``. But I personally agree with your reasoning.

src/_pytest/config/__init__.py
CarycaKatarzyna: Please see if you meant it (remove ``from e``)?

src/_pytest/config/__init__.py
CarycaKatarzyna: I've used ``ImportError`` to avoid problems with these kinds of tests: https://github.com/pytest-dev/pytest/blob/19ad5889353c7f5f2b65cc2acd346b7a9e95dfcd/testing/test_pluginmanager.py#L295
But I think removing that line with ``super()`` won't mess anything. WDYT?

testing/test_pluginmanager.py
CarycaKatarzyna: Yes, the result of test with ``find`` confused me.

bluetech:
This change brings up a "philosophical" question. In the scenario

- A top-level conftest imports a plugin through `pytest_plugins = ["foo"]`
- The plugin "foo" raises an exception during import time

is it considered a problem with the conftest (`ConftestImportError`) or a problem with the plugin?

Generally it seems pytest is not very tolerant of import-time errors in plugins, so when a plugin has an import-time error it just propagates, e.g. `pytest -p foo` just crashes if `foo` does not exist or raises any errors.

This PR changes the particular scenario (plugin thru top-level conftest) to be treated as a `ContestImportError` which results in a nicer formatted error, but hides the stack trace.

changelog/993.breaking.rst
```diff
@@ -0,0 +1 @@
+The output if an exception occurs during the plugin import in conftest is now similar to the output if the exception is directly in conftest. All exceptions during conftest's startup are now ExitCode.INTERNAL_ERROR. Previously it was ExitCode.USAGE_ERROR.
```
nicoddemus, line 1:
```suggestion
Exceptions raised while importing `conftest.py` files now cause pytest to exit with `ExitCode.INTERNAL_ERROR` instead of `ExitCode.USAGE_ERROR`.

Also, exceptions that happen during plugin import in `conftest.py` files are now similar to the output obtained from exceptions raised directly in `conftest.py` files. 
```

src/_pytest/config/__init__.py
nicoddemus: It is probably best to pass just a `msg` argument to the parent class, for example:

```python
class PluginImportFailure(ImportError):
    def __init__(self, plugin, excinfo):
        self.plugin = plugin
        self.excinfo = excinfo  # type: Tuple[Type[Exception], Exception, TracebackType]
        super().__init__(str(self))

```

src/_pytest/config/__init__.py
nicoddemus: I think @asottile meant to remove the parenthesis around `raise`: `raise (...)`

changelog/993.breaking.rst
CarycaKatarzyna: I have already restored ``ExitCode.USAGE_ERROR`` after @asottile comment. Should it ultimately be ``ExitCode.INTERNAL_ERROR``?

nicoddemus:
Closing this for now as it has been a long time since this has seen an update... @CarycaKatarzyna please feel free to reopen this if you want to retake this. 👍

PR #7299, opened 2020-06-02
[5.4.x] Merge pull request #7294 from nicoddemus/codecov-adjustments

nicoddemus:


bluetech:
Although it looks like codecov only looks at `master`...

nicoddemus:
> Although it looks like codecov only looks at master...

Hmm I think you are right, I remember reading something in that sense... 🤔 

Btw: when merging backports, make sure to go back to the original PR and mark it as backported. 😉

PR #7303, opened 2020-06-02
Show invalid ini keys sorted

nicoddemus:
Otherwise this relies on the dictionary order of `config.inicfg`, which
is insertion order in py36+ but "random" order in py35.

This lead to an unrelated failure in https://github.com/pytest-dev/pytest/pull/7247/checks?check_run_id=732077801.


gnikonorov:
Sorry about that. Thanks for patching @asottile

PR #7316, opened 2020-06-03
Sort install dependencies in setup.py

nicoddemus:
This is something that always annoyed me a little. 😁

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,18 +1,17 @@
 from setuptools import setup
 
-# TODO: if py gets upgrade to >=1.6,
-#       remove _width_of_current_line in terminal.py
 INSTALL_REQUIRES = [
-    "py>=1.5.0",
-    "packaging",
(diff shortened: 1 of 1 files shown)


asottile:
if we move these to `setup.cfg` and use [setup-cfg-fmt](https://github.com/asottile/setup-cfg-fmt) then they'll stay sorted via pre-commit without needing manual intervention

nicoddemus:
> if we move these to setup.cfg and use setup-cfg-fmt then they'll stay sorted via pre-commit without needing manual intervention

I'm fine with that. 😁 

I remember we didn't move to full `setup.cfg` before because we had some hacks involving environment variables to test pluggy from `master` (or something like this), but it seems we don't have that anymore.

Would you like to give this a go @asottile? That would supersede this PR then.

asottile:
yep! I'll put that together

PR #7321, opened 2020-06-05
Add reference to builtin markers to doc

xinbinhuang:
Small doc improvement to add API reference to Marker doc.

Issue ref: #7270

PR #7326, opened 2020-06-05
Remove start_doc_server.sh script

xinbinhuang:
In #7321 , I forgot to add a note about the utility script `start_doc_server` in the contrubuting.rst.

This add it back..  

cc: @nicoddemus

### DIFF
--- a/doc/en/start_doc_server.sh
+++ b/doc/en/start_doc_server.sh
@@ -1,5 +1,0 @@
-#!/usr/bin/env bash
-
-MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-cd "${MY_DIR}"/_build/html || exit
-python -m http.server 8000


xinbinhuang:
The failed test does not seem relevant.

CONTRIBUTING.rst
```diff
@@ -88,6 +88,8 @@ without using a local copy.  This can be convenient for small fixes.
 
     The built documentation should be available in ``doc/en/_build/html``,
     where 'en' refers to the documentation language.
+    You can launch a simple local server by running ``bash doc/en/start_doc_server.sh``
```
asottile, line 91: oh uh, I'm not sure we need a server at all -- usually you can just `firefox doc/en/index.html` or whatever -- we should proably delete `start_doc_server.sh` (especially because it is not portable to windows)

nicoddemus: Indeed, do you mind removing the local server and removing this docs @xinbinhuang ?

xinbinhuang: I generally launch a simple server because I believe it provides better navigation than the `file://`.   But probably not needed for sphinx built documentation. We can add it back if really need to.

I just removed the `start_doc_server.sh`. PTAL :)

xinbinhuang:
I like the cat! lol

PR #7330, opened 2020-06-06
Add new INI config key 'required_plugins' that defines a list of plugins that must be present for a run of pytest

gnikonorov:
closes https://github.com/pytest-dev/pytest/issues/7305

Changes:
- Introduce a new config key called `require_plugins` that is a space separated list of plugins required for pytest to run
- Make the `--strict-config` flag ( added [here](https://github.com/pytest-dev/pytest/pull/7286) treat these warnings as errors instead

### DIFF
--- a/changelog/7305.feature.rst
+++ b/changelog/7305.feature.rst
@@ -1,3 +1,1 @@
-New `require_plugins` configuration option allows the user to specify a list of plugins required for pytest to run. Warnings are raised if these plugins are not found when running pytest.
-
-The `--strict-config` flag can be used to treat these warnings as errors.
+New `required_plugins` configuration option allows the user to specify a list of plugins required for pytest to run. An error is raised if any required plugins are not found when running pytest.
--- a/changelog/7305.feature.rst
+++ b/changelog/7305.feature.rst
@@ -1,0 +1,3 @@
+A new INI key `require_plugins` has been added that allows the user to specify a list of plugins required for pytest to run.
+
+The `--strict-config` flag can be used to treat these warnings as errors.
--- a/changelog/7305.feature.rst
+++ b/changelog/7305.feature.rst
@@ -1,1 +1,1 @@
-New `required_plugins` configuration option allows the user to specify a list of plugins required for pytest to run. An error is raised if any required plugins are not found when running pytest.
+New ``required_plugins`` configuration option allows the user to specify a list of plugins required for pytest to run. An error is raised if any required plugins are not found when running pytest.
--- a/changelog/7305.feature.rst
+++ b/changelog/7305.feature.rst
@@ -1,3 +1,3 @@
-A new INI key `require_plugins` has been added that allows the user to specify a list of plugins required for pytest to run.
+New `require_plugins` configuration option allows the user to specify a list of plugins required for pytest to run.
 
 The `--strict-config` flag can be used to treat these warnings as errors.
--- a/changelog/7305.feature.rst
+++ b/changelog/7305.feature.rst
@@ -1,3 +1,3 @@
-New `require_plugins` configuration option allows the user to specify a list of plugins required for pytest to run.
+New `require_plugins` configuration option allows the user to specify a list of plugins required for pytest to run. Warnings are raised if these plugins are not found when running pytest.
 
 The `--strict-config` flag can be used to treat these warnings as errors.
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1604,3 +1604,13 @@ passed multiple times. The expected format is ``name=value``. For example::
 
         [pytest]
         xfail_strict = True
+
+
+.. confval:: required_plugins
+
+   A space seperated list of plugins that must be present for pytest to run
+
+   .. code-block:: ini
+
+       [pytest]
+       require_plugins = pluginA pluginB pluginC
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1564,7 +1564,7 @@ passed multiple times. The expected format is ``name=value``. For example::
 .. confval:: required_plugins
 
    A space separated list of plugins that must be present for pytest to run.
-   If any one of the plugins is not found, emit a error.
+   If any one of the plugins is not found, emit an error.
 
    .. code-block:: ini
 
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1561,16 +1561,15 @@ passed multiple times. The expected format is ``name=value``. For example::
    See :ref:`change naming conventions` for more detailed examples.
 
 
-.. confval:: require_plugins
+.. confval:: required_plugins
 
    A space separated list of plugins that must be present for pytest to run.
-   If any one of the plugins is not found, emit a warning.
-   If pytest is run with ``--strict-config`` exceptions are raised in place of warnings.
+   If any one of the plugins is not found, emit a error.
 
    .. code-block:: ini
 
        [pytest]
-       require_plugins = html xdist
+       required_plugins = pytest-html pytest-xdist
 
 
 .. confval:: testpaths
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1570,7 +1570,7 @@ passed multiple times. The expected format is ``name=value``. For example::
    .. code-block:: ini
 
        [pytest]
-       require_plugins = pluginA pluginB pluginC
+       require_plugins = pytest-xdist pytest-mock
 
 
 .. confval:: testpaths
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1561,6 +1561,18 @@ passed multiple times. The expected format is ``name=value``. For example::
    See :ref:`change naming conventions` for more detailed examples.
 
 
+.. confval:: require_plugins
+
+   A space separated list of plugins that must be present for pytest to run.
+   If any one of the plugins is not found, emit a warning.
+   If pytest is run with ``--strict-config`` exceptions are raised in place of warnings.
+
+   .. code-block:: ini
+
+       [pytest]
+       require_plugins = pluginA pluginB pluginC
+
+
 .. confval:: testpaths
 
 
@@ -1604,13 +1616,3 @@ passed multiple times. The expected format is ``name=value``. For example::
 
         [pytest]
         xfail_strict = True
-
-
-.. confval:: required_plugins
-
-   A space seperated list of plugins that must be present for pytest to run
-
-   .. code-block:: ini
-
-       [pytest]
-       require_plugins = pluginA pluginB pluginC
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1570,7 +1570,7 @@ passed multiple times. The expected format is ``name=value``. For example::
    .. code-block:: ini
 
        [pytest]
-       require_plugins = pytest-xdist pytest-mock
+       require_plugins = html xdist
 
 
 .. confval:: testpaths
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1087,7 +1087,7 @@ class Config:
 
     def _validate_keys(self) -> None:
         for key in sorted(self._get_unknown_ini_keys()):
-            self._emit_warning_or_fail("Unknown config ini key: {}\n".format(key))
+            self._warn_or_fail_if_strict("Unknown config ini key: {}\n".format(key))
 
     def _validate_plugins(self) -> None:
         required_plugins = sorted(self.getini("required_plugins"))
@@ -1108,7 +1108,7 @@ class Config:
                 pytrace=False,
             )
 
-    def _emit_warning_or_fail(self, message: str) -> None:
+    def _warn_or_fail_if_strict(self, message: str) -> None:
         if self.known_args_namespace.strict_config:
             fail(message, pytrace=False)
         sys.stderr.write("WARNING: {}".format(message))
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -953,7 +953,7 @@ class Config:
         self._parser.addini("addopts", "extra command line options", "args")
         self._parser.addini("minversion", "minimally required pytest version")
         self._parser.addini(
-            "require_plugins",
+            "required_plugins",
             "plugins that must be present for pytest to run",
             type="args",
             default=[],
@@ -1090,11 +1090,21 @@ class Config:
             self._emit_warning_or_fail("Unknown config ini key: {}\n".format(key))
 
     def _validate_plugins(self) -> None:
-        for plugin in self.getini("require_plugins"):
-            if not self.pluginmanager.hasplugin(plugin):
-                self._emit_warning_or_fail(
-                    "Missing required plugin: {}\n".format(plugin)
-                )
+        plugin_info = self.pluginmanager.list_plugin_distinfo()
+        plugin_dist_names = [
+            "{dist.project_name}".format(dist=dist) for _, dist in plugin_info
+        ]
+
+        required_plugin_list = []
+        for plugin in sorted(self.getini("required_plugins")):
+            if plugin not in plugin_dist_names:
+                required_plugin_list.append(plugin)
+
+        if required_plugin_list:
+            fail(
+                "Missing required plugins: {}".format(", ".join(required_plugin_list)),
+                pytrace=False,
+            )
 
     def _emit_warning_or_fail(self, message: str) -> None:
         if self.known_args_namespace.strict_config:
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1090,19 +1090,21 @@ class Config:
             self._emit_warning_or_fail("Unknown config ini key: {}\n".format(key))
 
     def _validate_plugins(self) -> None:
+        required_plugins = sorted(self.getini("required_plugins"))
+        if not required_plugins:
+            return
+
         plugin_info = self.pluginmanager.list_plugin_distinfo()
-        plugin_dist_names = [
-            "{dist.project_name}".format(dist=dist) for _, dist in plugin_info
-        ]
+        plugin_dist_names = [dist.project_name for _, dist in plugin_info]
 
-        required_plugin_list = []
-        for plugin in sorted(self.getini("required_plugins")):
+        missing_plugins = []
+        for plugin in required_plugins:
             if plugin not in plugin_dist_names:
-                required_plugin_list.append(plugin)
+                missing_plugins.append(plugin)
 
-        if required_plugin_list:
+        if missing_plugins:
             fail(
-                "Missing required plugins: {}".format(", ".join(required_plugin_list)),
+                "Missing required plugins: {}".format(", ".join(missing_plugins)),
                 pytrace=False,
             )
 
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1090,11 +1090,6 @@ class Config:
             self._emit_warning_or_fail("Unknown config ini key: {}\n".format(key))
 
     def _validate_plugins(self) -> None:
-        # so iterate over all required plugins and see if pluginmanager hasplugin
-        # NOTE: This also account for -p no:<plugin> ( e.g: -p no:celery )
-        # raise ValueError(self._parser._inidict['requiredplugins'])
-        # raise ValueError(self.getini("requiredplugins"))
-        # raise ValueError(self.pluginmanager.hasplugin('debugging'))
         for plugin in self.getini("require_plugins"):
             if not self.pluginmanager.hasplugin(plugin):
                 self._emit_warning_or_fail(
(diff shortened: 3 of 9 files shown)


gnikonorov:
Hey @RonnyPfannschmidt - I think we have a flaky test for py36. What's the process for dealing with it? Looking at the logs I don't think it's related to my change

RonnyPfannschmidt:
I retried the build

gnikonorov:
Thanks @RonnyPfannschmidt - looks like that fixed the build ( fingers crossed for no flaky tests in py37 )

doc/en/reference.rst
```diff
@@ -1561,6 +1561,18 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+.. confval:: require_plugins
+
+   A space separated list of plugins that must be present for pytest to run.
+   If any one of the plugins is not found, emit a warning.
+   If pytest is run with ``--strict-config`` exceptions are raised in place of warnings.
+
+   .. code-block:: ini
+
+       [pytest]
+       require_plugins = pluginA pluginB pluginC
```
nicoddemus, line 1573: Can we add here a real list? For example `require_plugins = pytest-xdist pytest-mock`?

I ask because it is not clear here if the user should add the package name or plugin name.

gnikonorov:
Thanks for the review, @nicoddemus! I updated the documentation like you suggested

gnikonorov:
Friendly ping for a merge @nicoddemus, @RonnyPfannschmidt. Or, should I merge myself now that I have the ability to and this has been approved?

RonnyPfannschmidt:
Sorry for the delay, I'm moving

src/_pytest/config/__init__.py
```diff
@@ -1035,7 +1041,8 @@ def _preparse(self, args: List[str], addopts: bool = True) -> None:
         self.known_args_namespace = ns = self._parser.parse_known_args(
             args, namespace=copy.copy(self.option)
         )
-        self._validatekeys()
+        self._validate_keys()
```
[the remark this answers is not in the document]
gnikonorov: you're right, we should do that @RonnyPfannschmidt. Just pushed a commit with that. Thanks for catching

gnikonorov:
No worries @RonnyPfannschmidt, hope your move goes smoothly. Looks like py36 is flaky on the same tests as last time

changelog/7305.feature.rst
```diff
@@ -0,0 +1,3 @@
+New `require_plugins` configuration option allows the user to specify a list of plugins required for pytest to run.
+
+The `--strict-config` flag can be used to treat these warnings as errors.
```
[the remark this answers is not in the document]
gnikonorov: 'those warnings'....were omitted. Adding in a description now

doc/en/reference.rst
```diff
@@ -1561,6 +1561,18 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+.. confval:: require_plugins
+
+   A space separated list of plugins that must be present for pytest to run.
+   If any one of the plugins is not found, emit a warning.
+   If pytest is run with ``--strict-config`` exceptions are raised in place of warnings.
+
+   .. code-block:: ini
+
+       [pytest]
+       require_plugins = pytest-xdist pytest-mock
```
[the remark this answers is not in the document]
gnikonorov: It doesn't, pushing a change now.
```
(.venv) gnikonorov:~/pytest/gleb_test$ cat pytest.ini
[pytest]
minversion = 3.0.0
require_plugins = html xdist

(.venv) gnikonorov:~/pytest/gleb_test$ python3 -m pytest test_xfail.py
platform linux -- Python 3.6.9, pytest-5.4.1.dev465+gd2bb67bfd.d20200607, py-1.8.1, pluggy-0.13.1
plugins: xdist-1.32.0, hypothesis-5.15.1, mock-3.1.1, html-2.1.1, metadata-1.9.0, forked-1.1.3
collected 1 item

test_xfail.py x                                                                                                                                                                                                                      [100%]

============================================================================================================ 1 xfailed in 0.05s ============================================================================================================
```

gnikonorov:
( should be ) good to merge now @nicoddemus

gnikonorov:
The failure seems to be the same unrelated flaky one from before. @RonnyPfannschmidt's re-run of the build 'fixed' it last time

gnikonorov:
Friendly bump for a test rerun and/or merge @nicoddemus

doc/en/reference.rst
```diff
@@ -1561,6 +1561,18 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+.. confval:: require_plugins
+
+   A space separated list of plugins that must be present for pytest to run.
+   If any one of the plugins is not found, emit a warning.
+   If pytest is run with ``--strict-config`` exceptions are raised in place of warnings.
+
+   .. code-block:: ini
+
+       [pytest]
+       require_plugins = html xdist
```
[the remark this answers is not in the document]
gnikonorov: @RonnyPfannschmidt @nicoddemus what if we just checked for both `<plugin>` and `pytest-<plugin>` ( e.g.: `html` and `pytest-html` ) as suggested in the beginning of the comment. To @nicoddemus' point, there is no standard way to define plugins, and once all the plugins are loaded we store them in the way I wrote up with this pr ( `pytest-html` becomes `html` ).

I think looking for both would be valid, since one is the internal representation, and the other is the user friendly name, and it would be simpler than trying to user the logic in `terminal.py` which we'd probably have to groom a little for this use case

src/_pytest/config/__init__.py
```diff
@@ -952,6 +952,12 @@ def _initini(self, args: Sequence[str]) -> None:
         self._parser.extra_info["inifile"] = self.inifile
         self._parser.addini("addopts", "extra command line options", "args")
         self._parser.addini("minversion", "minimally required pytest version")
+        self._parser.addini(
+            "require_plugins",
```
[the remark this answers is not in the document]
gnikonorov: I can go either way. I'll change it to `required_plugins`, unless someone has a strong objection

src/_pytest/config/__init__.py
```diff
@@ -1078,12 +1085,21 @@ def _checkversion(self):
...
-            if self.known_args_namespace.strict_config:
-                fail(message, pytrace=False)
-            sys.stderr.write("WARNING: {}".format(message))
+            self._emit_warning_or_fail("Unknown config ini key: {}\n".format(key))
+
+    def _validate_plugins(self) -> None:
+        for plugin in self.getini("require_plugins"):
+            if not self.pluginmanager.hasplugin(plugin):
+                self._emit_warning_or_fail(
+                    "Missing required plugin: {}\n".format(plugin)
```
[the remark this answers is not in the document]
gnikonorov: @nicoddemus - I feel like it's unlikely, but a suite could still potentially work without a required plugn. For example, you could have required [pytest-greendots](https://pypi.org/project/pytest-greendots/), but since it has no configuration options that are required the suite will work with and without the plugin

doc/en/reference.rst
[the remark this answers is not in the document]
RonnyPfannschmidt: We shouldn't guess, the giben names should be python distribution names

With module names also being accepted

There should be a way to add version constraints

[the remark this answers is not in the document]
gnikonorov: @RonnyPfannschmidt - so you mean we should prefer `pytest-html` over `html`

Agreed on the ability to add version constraints. Can we add this in a follow up PR to not bloat this one?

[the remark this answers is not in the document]
gnikonorov: @RonnyPfannschmidt - so you mean we should prefer `pytest-html` over `html`

Agreed on the ability to add version constraints. Can we add this in a follow up PR to not bloat this one?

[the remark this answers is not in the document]
gnikonorov: Also, @RonnyPfannschmidt @nicoddemus is it a requirement that plugins for pytest start with `pytest-`? It seems like yes from the documentation, but I am not sure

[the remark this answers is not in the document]
RonnyPfannschmidt: Sure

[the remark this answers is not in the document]
gnikonorov: @nicoddemus - thanks for sharing the snippet in `pytest/src/_pytest/terminal.py`, we  can definitely use that as a solution here

src/_pytest/config/__init__.py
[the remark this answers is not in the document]
nicoddemus: Indeed, but I think it is reasonable to stop executing if a plugin listed in `required_plugins` is not really installed.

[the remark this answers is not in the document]
gnikonorov: Now that I think about it again, I agree with you @nicoddemus. I mean, if it emits a warning but is 'required' I think that's pretty confusing. Let me change it to always stop executing. Any objections @RonnyPfannschmidt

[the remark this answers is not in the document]
nicoddemus: Great!

Make sure to display all missing plugins, instead of stopping on the first one missing.

gnikonorov:
Thanks for the review and thoughtful feedback. Updated the PR @nicoddemus and @RonnyPfannschmidt

gnikonorov:
Thank you @nicoddemus! Looking forward to getting this in and starting on the followup task

changelog/7305.feature.rst
```diff
@@ -0,0 +1 @@
+New `required_plugins` configuration option allows the user to specify a list of plugins required for pytest to run. An error is raised if any required plugins are not found when running pytest.
```
bluetech, line 1: I think RST requires double ticks though I'm never sure TBH...

```suggestion
New ``required_plugins`` configuration option allows the user to specify a list of plugins required for pytest to run. An error is raised if any required plugins are not found when running pytest.
```

doc/en/reference.rst
```diff
@@ -1561,6 +1561,17 @@ passed multiple times. The expected format is ``name=value``. For example::
    See :ref:`change naming conventions` for more detailed examples.
 
 
+.. confval:: required_plugins
+
+   A space separated list of plugins that must be present for pytest to run.
+   If any one of the plugins is not found, emit a error.
```
bluetech, line 1567:
```suggestion
   If any one of the plugins is not found, pytest emits an error and exits with code XXX (unless ``strict_config`` is ``false``).
```

The XXX should be filled here ^^^ (which is it?)

src/_pytest/config/__init__.py
```diff
@@ -1078,12 +1085,33 @@ def _checkversion(self):
...
-            message = "Unknown config ini key: {}\n".format(key)
-            if self.known_args_namespace.strict_config:
-                fail(message, pytrace=False)
-            sys.stderr.write("WARNING: {}".format(message))
+            self._emit_warning_or_fail("Unknown config ini key: {}\n".format(key))
+
+    def _validate_plugins(self) -> None:
+        required_plugins = sorted(self.getini("required_plugins"))
+        if not required_plugins:
+            return
```
bluetech, line 1095: I'd omit these two lines as they're not really necessary.

```suggestion
```

src/_pytest/config/__init__.py
```diff
@@ -1078,12 +1085,33 @@ def _checkversion(self):
...
+
+        plugin_info = self.pluginmanager.list_plugin_distinfo()
```
bluetech, line 1097: Any particular reason to only include setuptools plugins here?

src/_pytest/config/__init__.py
```diff
@@ -1078,12 +1085,33 @@ def _checkversion(self):
...
+        plugin_dist_names = [dist.project_name for _, dist in plugin_info]
+
+        missing_plugins = []
+        for plugin in required_plugins:
+            if plugin not in plugin_dist_names:
+                missing_plugins.append(plugin)
+
+        if missing_plugins:
+            fail(
```
bluetech, line 1106: Is it possible the intention was to use `self._emit_warning_or_fail` here?

src/_pytest/config/__init__.py
gnikonorov: This was intentional. The point is to not fetch plugin info if it’s not necessary. The function is useless if there are no required plugins, so we exit

src/_pytest/config/__init__.py
gnikonorov: No, we always want to fail ( see earlier in the PR comments for justification )

src/_pytest/config/__init__.py
gnikonorov: This isn’t just setup tools it is all plugins, but with the proper distribution info. Correct me if I’m wrong @nicoddemus

doc/en/reference.rst
gnikonorov: strict_config has influence on this as we always fail ( see earlier in PR for reasoning )

src/_pytest/config/__init__.py
bluetech: So I was thrown off by the fact that `_emit_warning_or_fail` is extracted to a separate function, and that the help text for `strict_config` was changed. So I'd roll back these two changes -- maybe they were relevant for a previous PR iteration.

doc/en/reference.rst
bluetech: So since `strict_config` is not consulted in this case, will edit the suggestion above to remove mention of it.

src/_pytest/config/__init__.py
bluetech: I don't think this optimization is worth it, the code below doesn't do anything expensive. I think it is better to avoid special cases when they are not necessary. But it's OK if you want to keep it.

doc/en/reference.rst
gnikonorov: Sure I can update this to show the exit code but I think that’s a bit redundant and becomes a maintenance headache since we seem to centralize our exit code information here: https://docs.pytest.org/en/stable/usage.html. I think we should keep exit code information there. If we don’t things will eventually diverge and confuse users

changelog/7305.feature.rst
gnikonorov: It does I’ll make the change. Thank you for catching

src/_pytest/config/__init__.py
gnikonorov: Resolving as above. Please reopen if I’m wrong

src/_pytest/config/__init__.py
gnikonorov: The call to get dist info is expensive ( as per @nicoddemus’ earlier comment )

src/_pytest/config/__init__.py
gnikonorov: Why roll back the extraction? I see no harm in leaving it in. Functionality is neither added or removed with it

doc/en/reference.rst
gnikonorov: Thoughts @nicoddemus @RonnyPfannschmidt

gnikonorov: It actually looks like it exits with exit code 1, which sounds valid to me for the use case

nicoddemus: I think it is OK to leave out the exact error code here, but I don't have strong feelings either.

src/_pytest/config/__init__.py
nicoddemus: Yeah I was under that impression, but [I was wrong](https://github.com/pytest-dev/pluggy/blob/f17598a7b422a8a7566a258d2f79fae5fde6b78a/src/pluggy/manager.py#L286-L289), sorry for not doing the proper research before.

Regardless I think the change is harmless and particularly I would prefer to bail out early. 😁

doc/en/reference.rst
gnikonorov: If we’re all ok with it, I’d like to leave all exit codes in the usage doc I posted above. Otherwise the error code reporting may diverge over time and people will be confused. I also wouldn’t know to change this location easily as maintainer who has never seen this PR

src/_pytest/config/__init__.py
gnikonorov: +1 on bailing early. It’s unnecessary otherwise. We’re not writing high frequency trading software but at the same time we should do basic optimizations

doc/en/reference.rst
bluetech: Yea makes sense to leave the exit code out. So just a typo "a error" -> "an error"

src/_pytest/config/__init__.py
bluetech: The main thing is that the condition makes it look like a Falsey `required_plugins` has a special meaning. If it is removed, the reader doesn't need to figure out that it doesn't.

But just in general, if I see e.g.  `sum` implemented like this:

```py
def sum(xs):
    if not xs:
        return 0
    s = 0
    for x in xs:
        s += x
    return s
```

that seems redundant...

src/_pytest/config/__init__.py
bluetech: `list_plugin_distinfo` only lists setuptools plugins, at least this is what I understand from the pluggy code. That might be what we want here, not sure myself, but unresolving so can figure it out.

src/_pytest/config/__init__.py
bluetech: I think because it examines `strict_config` the name is a bad confusing, would call it `_emit_warning_or_fail_if_strict`. But that's a handful, and since it's only used once, I'd remove the indirection and bring it back inline.

src/_pytest/config/__init__.py
nicoddemus: Good point @bluetech.

Should we consider somehow plugins specified via `-p` and `PYTEST_PLUGINS`?

src/_pytest/main.py
```diff
@@ -78,7 +78,7 @@ def pytest_addoption(parser: Parser) -> None:
     group._addoption(
         "--strict-config",
         action="store_true",
-        help="invalid ini keys for the `pytest` section of the configuration file raise errors.",
+        help="any warnings encountered while parsing the `pytest` section of the configuration file raise errors.",
```
bluetech, line 81: I think this should be reverted now?

bluetech:
Sorry for being extra nitpicky @gnikonorov :)

src/_pytest/config/__init__.py
nicoddemus: Perhaps using `warn` as a verb directly, as in `_warn_or_fail_if_strict`, is better?

src/_pytest/main.py
gnikonorov: I don’t see why. This statement is valid. The one warning in the config will be raised as an error

src/_pytest/config/__init__.py
gnikonorov: Sure, I’ll change it to the name suggested by @nicoddemus

src/_pytest/main.py
gnikonorov: And more generally I think strict_config should apply to the entire config file parsing, for consistencies sake

src/_pytest/config/__init__.py
gnikonorov: Do we want to add support for -p and PYTEST_PLUGINS in a follow up PR or does that feel like we’re submitting a half complete feature @nicoddemus @bluetech

gnikonorov:
No need to be sorry at all! I appreciate the thoroughness, thank you!

src/_pytest/config/__init__.py
gnikonorov: I mean, technically speaking a falsely `required_plugins` means to the function and the reader of the function "nothing important to do here now, just carry on"

In this case we also don't have to assemble plugin_dist_names `plugin_dist_names = [dist.project_name for _, dist in plugin_info]` which we still would if we don't bail early

src/_pytest/config/__init__.py
bluetech: I haven't checked what each means, but just from perusing the `PlugingManager` code, `list_name_plugin()` seems good?

I guess it depends on why the code uses `dist.project_name` instead of the name the plugin was registered with? (Sorry if was discussed before)

gnikonorov: This was discussed here: https://github.com/pytest-dev/pytest/pull/7330#discussion_r437135961

gnikonorov: From reading the help, I think we should defer `-p` if we add it at all. It sounds like a user is intentionally fiddling around with the run
```
-p name               early-load given plugin module name or entry point
                            (multi-allowed).
```

The same goes for the environment variable `PYTEST_PLUGINS`. I'm not opposed to adding them in, but to me it really sounds like they're a different use case than this and would warrant their own PR

gnikonorov:
Friendly ping to @RonnyPfannschmidt for a review

src/_pytest/config/__init__.py
nicoddemus: Agree, not sure this makes sense when used with `-p` or `PYTEST_PLUGINS`.

gnikonorov:
@nicoddemus and/or @bluetech if there are no more questions can we actually merge this into master? It’s blocking a few issues I want to start working on

gnikonorov:
Thanks for the  thoughtful reviews @nicoddemus and @bluetech
I’ll wait for @RonnyPfannschmidt

nicoddemus:
Thanks @gnikonorov!

PR #7351, opened 2020-06-11
Fix exception causes in config/__init__.py

cool-RR:
I recently went over [Matplotlib](https://github.com/matplotlib/matplotlib/pull/16706), [Pandas](https://github.com/pandas-dev/pandas/pull/32322) and [NumPy](https://github.com/numpy/numpy/pull/15731), fixing a small mistake in the way that Python 3's exception chaining is used. If you're interested, I can do it here too. I've done it on just one file right now. 

The mistake is this: In some parts of the code, an exception is being caught and replaced with a more user-friendly error. In these cases the syntax `raise new_error from old_error` needs to be used.

Python 3's exception chaining means it shows not only the traceback of the current exception, but that of the original exception (and possibly more.) This is regardless of `raise from`. The usage of `raise from` tells Python to put a more accurate message between the tracebacks. Instead of this: 

You'll get this: 

    The above exception was the direct cause of the following exception:

The first is inaccurate, because it signifies a bug in the exception-handling code itself, which is a separate situation than wrapping an exception.

Let me know what you think!

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -641,7 +641,7 @@ class PytestPluginManager(PluginManager):
         except ImportError as e:
             raise ImportError(
                 'Error importing plugin "{}": {}'.format(modname, str(e.args[0]))
-            ).with_traceback(e.__traceback__)
+            ).with_traceback(e.__traceback__) from e
 
         except Skipped as e:
             from _pytest.warnings import _issue_warning_captured
@@ -1197,12 +1197,12 @@ class Config:
         for ini_config in self._override_ini:
             try:
                 key, user_ini_value = ini_config.split("=", 1)
-            except ValueError:
+            except ValueError as e:
                 raise UsageError(
                     "-o/--override-ini expects option=value style (got: {!r}).".format(
                         ini_config
                     )
-                )
+                ) from e
             else:
                 if key == name:
                     value = user_ini_value


nicoddemus:
I'm 👍 on the idea, thanks for the offer!

Are there still more places in the codebase where this adjustment is needed, or this PR covers them all?

cool-RR:
Awesome!

No, there are a few more:

```
src/_pytest/_code/source.py: 218
src/_pytest/config/__init__.py: 642
src/_pytest/config/__init__.py: 1134
src/_pytest/config/__init__.py: 1201
src/_pytest/config/__init__.py: 1233
src/_pytest/config/argparsing.py: 270
src/_pytest/config/findpaths.py: 29
src/_pytest/debugging.py: 30
src/_pytest/debugging.py: 129
src/_pytest/fixtures.py: 942
src/_pytest/logging.py: 492
src/_pytest/monkeypatch.py: 64
src/_pytest/monkeypatch.py: 73
src/_pytest/python.py: 562
src/_pytest/python.py: 564
src/_pytest/python.py: 583
src/_pytest/python.py: 592
src/_pytest/python.py: 847
src/_pytest/python.py: 1085
src/_pytest/runner.py: 160
src/_pytest/runner.py: 428
src/_pytest/warnings.py: 52
testing/test_config.py: 1647
testing/test_runner.py: 538
```

Some of these will be false-positives, I'll go over all of them and give you a separate PR after we're done with this PR.

nicoddemus:
Thanks!

Wouldn't be better to just change them in this PR instead though?

cool-RR:
I could do that, but before I spend time going over these cases, I want to be sure there wouldn't be any roadblock. Getting this PR merged would be a good indication on whether there'll be a roadblock.

nicoddemus:
Well the tests passing is a good indicator that things will probably go smooth. 😉 

But let's wait for other maintainers to chime in before you spend more time on this. 👍

src/_pytest/config/__init__.py
```diff
@@ -1130,8 +1130,8 @@ def getini(self, name: str):
     def _getini(self, name: str) -> Any:
         try:
             description, type, default = self._parser._inidict[name]
-        except KeyError:
-            raise ValueError("unknown configuration value: {!r}".format(name))
+        except KeyError as e:
+            raise ValueError("unknown configuration value: {!r}".format(name)) from e
```
bluetech, line 1134: Here I think `from None` would be better.

```suggestion
        except KeyError:
            raise ValueError("unknown configuration value: {!r}".format(name)) from None
```

src/_pytest/config/__init__.py
```diff
@@ -1223,14 +1223,14 @@ def getoption(self, name: str, default=notset, skip: bool = False):
             if val is None and skip:
                 raise AttributeError(name)
             return val
-        except AttributeError:
+        except AttributeError as e:
```
bluetech, line 1226:
```suggestion
        except AttributeError:
```

src/_pytest/config/__init__.py
```diff
@@ -1223,14 +1223,14 @@ def getoption(self, name: str, default=notset, skip: bool = False):
...
             if default is not notset:
                 return default
             if skip:
                 import pytest
 
                 pytest.skip("no {!r} option found".format(name))
-            raise ValueError("no option named {!r}".format(name))
+            raise ValueError("no option named {!r}".format(name)) from e
```
bluetech, line 1233:
```suggestion
            raise ValueError("no option named {!r}".format(name)) from None
```

src/_pytest/config/__init__.py
```diff
@@ -641,7 +641,7 @@ def import_plugin(self, modname: str, consider_entry_points: bool = False) -> No
         except ImportError as e:
             raise ImportError(
                 'Error importing plugin "{}": {}'.format(modname, str(e.args[0]))
-            ).with_traceback(e.__traceback__)
+            ).with_traceback(e.__traceback__) from e
```
bluetech, line 644: The `with_traceback` makes me think maybe the intention here is `from None`.

cool-RR:
My position is that I prefer `from old_exception` everywhere. I'm okay with skipping a few cases that we don't have agreement on, but I won't add `from None`, because I don't want to be the reason that someone debugging a problem doesn't have access to potentially helpful information. I'm okay with anyone else adding `from None` on a separate PR.

If you agree, I'll just remove my changes for all the cases where you've done `from None` and leave just the ones we agree about.

bluetech:
@cool-RR Of course I agree :) I'd be interested to understand your position though.

For example in this case

```py
try:
    description, type, default = self._parser._inidict[name]
except KeyError as e:
    raise ValueError("unknown configuration value: {!r}".format(name)) from e
```

The `KeyError` I think would be plain noise, exposing an internal implementation detail. It could have been written just the same as

```py
if name in self._parser._inidict:
    description, type, default = self._parser._inidict[name]
else:
    raise ValueError("unknown configuration value: {!r}".format(name))
```

cool-RR:
@bluetech 

I'll happily explain my position :)

I personally hate `from None`, and I think that when people are debugging they have a very different mindset than when they are writing code. When you're debugging, and you get a very limited amount of information that may have been copy-pasted in an email by a non-technical person, you're very desperate for the information.

In this case, I would have wanted to see the `self._parser._inidict[name]` traceback. I would maybe get a hint on what happened. If it's a dead end for exploration, I at least want to know its content before I rule it out as a dead end.

Another issue is that some error-reporting systems show not only the traceback, but all the local variables on each of these levels. These supposedly irrelevant levels could have variables that make it clearer what this exception is about.

bluetech:
Thanks for explaining your position! So in some cases, like debugging crashes or unexpected exceptions, I fully agree. But in the cases in the PR, the intention is to report some error to the user, and here I subscribe to a different approach, which is to make the errors as concise and to-the-point as possible.

But, I shall not force you to go against your principles :smile: Since a direct cause is already an improvement over an indirect cause (which to me in Python 3 means a programmer error), it's fine by me.

cool-RR:
Cool :)

I changed the cases you mentioned, except the one on line 644. I think that the `.with_traceback(e.__traceback__)`, which can actually be removed, means that the traceback of the original exception is relevant for understanding the problem.

bluetech:
Thanks @cool-RR!

cool-RR:
I just made a blog post about this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

PR #7356, opened 2020-06-12
Finish deprecation of "slave"

Zac-HD:
See #7344, https://github.com/pytest-dev/pytest-xdist/issues/234, https://github.com/pytest-dev/pytest-xdist/pull/541.

CC @RonnyPfannschmidt


Zac-HD:
I'll just add the changelog entry, then merge :grin:

PR #7358, opened 2020-06-12
More type annotations, fix some typing bugs

bluetech:
Please see the commits, but here's a summary:

- Remove unused type: ignores
- Add type annotations to `MonkeyPatch`
- Add type annotations to `pytest_keyboard_interrupt` and `pytest_internalerror` hooks
- Fix type errors when running with typed `py.path.local`
- Fix type errors when running with typed `packaging`

### DIFF
--- a/src/_pytest/config/findpaths.py
+++ b/src/_pytest/config/findpaths.py
@@ -63,7 +63,7 @@ def load_config_dict_from_file(
     elif filepath.ext == ".toml":
         import toml
 
-        config = toml.load(filepath)
+        config = toml.load(str(filepath))
 
         result = config.get("tool", {}).get("pytest", {}).get("ini_options", None)
         if result is not None:
@@ -161,16 +161,18 @@ def determine_setup(
     args: List[str],
     rootdir_cmd_arg: Optional[str] = None,
     config: Optional["Config"] = None,
-) -> Tuple[py.path.local, Optional[str], Dict[str, Union[str, List[str]]]]:
+) -> Tuple[py.path.local, Optional[py.path.local], Dict[str, Union[str, List[str]]]]:
     rootdir = None
     dirs = get_dirs_from_args(args)
     if inifile:
-        inicfg = load_config_dict_from_file(py.path.local(inifile)) or {}
+        inipath_ = py.path.local(inifile)
+        inipath = inipath_  # type: Optional[py.path.local]
+        inicfg = load_config_dict_from_file(inipath_) or {}
         if rootdir_cmd_arg is None:
             rootdir = get_common_ancestor(dirs)
     else:
         ancestor = get_common_ancestor(dirs)
-        rootdir, inifile, inicfg = locate_config([ancestor])
+        rootdir, inipath, inicfg = locate_config([ancestor])
         if rootdir is None and rootdir_cmd_arg is None:
             for possible_rootdir in ancestor.parts(reverse=True):
                 if possible_rootdir.join("setup.py").exists():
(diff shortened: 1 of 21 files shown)


bluetech:
@symonk The py.path.local fixes should take care of the typing issues you're having in #7350.

nicoddemus:
Cool!

> @symonk The py.path.local fixes should take care of the typing issues you're having in #7350.

I suppose the idea is for @symonk to rebase his branch once this gets merged, correct?

src/_pytest/config/findpaths.py
```diff
@@ -161,24 +161,26 @@ def determine_setup(
     args: List[str],
     rootdir_cmd_arg: Optional[str] = None,
     config: Optional["Config"] = None,
-) -> Tuple[py.path.local, Optional[str], Dict[str, Union[str, List[str]]]]:
+) -> Tuple[py.path.local, Optional[py.path.local], Dict[str, Union[str, List[str]]]]:
     rootdir = None
     dirs = get_dirs_from_args(args)
     if inifile:
-        inicfg = load_config_dict_from_file(py.path.local(inifile)) or {}
+        inipath_ = py.path.local(inifile)
```
nicoddemus, line 168: I'm curious, why this `inipath_/inipath` dance here? 🤔

src/_pytest/terminal.py
```diff
@@ -315,6 +318,9 @@ def __init__(self, config: Config, file: Optional[TextIO] = None) -> None:
         self._show_progress_info = self._determine_show_progress_info()
         self._collect_report_last_write = None  # type: Optional[float]
         self._already_displayed_warnings = None  # type: Optional[int]
+        self._keyboardinterrupt_memo = (
```
nicoddemus: 🙏

symonk:
Thanks @nicoddemus / @bluetech Im a bit busy for a day or two and then il revisit my other PR; this is likely merged by then and il bring it in.  I need to revisit as to some information Ran has mentioned.

nicoddemus:
> Thanks @nicoddemus / @bluetech Im a bit busy for a day or two and then il revisit my other PR; this is likely merged by then and il bring it in. I need to revisit as to some information Ran has mentioned.

Sure thing, take your time. 👍

src/_pytest/config/findpaths.py
bluetech: If I had used `inipath = ... # type: Optional[py.path.local]` here, then I wouldn't be able to pass it to `load_config_dict_from_file` which requires `py.path.local`.

In Python 3.6 I would have declared `inipath: Optional[py.path.local]` at the top of the function but for now we need to use the comments.

nicoddemus: Not sure I follow, don't `inipath` and `inipath_` point to the same object? I mean, `inipath_` is a `py.path.local` object, how can `inipath = inipath_` be typed as `Optional[py.path.local]`?

bluetech: Stripping to only the relevant bits:

```py
if inifile:
    inipath = py.path.local(inifile)  # A
    inicfg = load_config_dict_from_file(inipath)  # B
else:
   inipath = something that returns Optional[py.path.local]  # C
```

As written, A will set the type of `inipath` as `py.path.local`, and then C is invalid.

The usual solution to this is to change A to

```py
    inipath = py.path.local(inifile) # type: Optional[py.path.local]
```

but this causes a new problem in line B, that the argument is required to be `py.path.local` but is now `Optional[py.path.local]`.

So I worked around it with the `inipath_`. Let me know if you think of a better way :)

nicoddemus: Thanks for taking the time to explain, appreciate it!

PR #7372, opened 2020-06-15
Require py>=1.8.2 so we can rely on correct hash() of py.path.local n Windows

### LINKED ISSUE #7357: Audit sets/dicts of `py.path.local`s
`py.path.local` has a footgun flaw: `__eq__` on Windows is case-insensitive, but `__hash__` returns the hash of the `strpath` directly, meaning `a == b` does not imply `hash(a) == hash(b)`. So things like this are possible:

```py
import sys, py
assert sys.platform == "win32"

a = py.path.local("/some/path")
b = py.path.local("/some/PATH")
assert a == b  # Passes
assert a in {b}  # Fails
assert a in {b: 'b'}  # Fails
```

This probably causes some subtle bugs in pytest, which uses several dicts & sets of `py.path.local`s. Since `py` is frozen, it's not possible to fix there.

I haven't checked it yet, but wanted to make a note for later.

bluetech:
See https://github.com/pytest-dev/py/blob/1.8.2/CHANGELOG#L4.
Fixes #7357.

### DIFF
--- a/changelog/7357.trivial.rst
+++ b/changelog/7357.trivial.rst
@@ -1,0 +1,1 @@
+py>=1.8.2 is now required.
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,7 @@ install_requires =
     more-itertools>=4.0.0
     packaging
     pluggy>=0.12,<1.0
-    py>=1.5.0
+    py>=1.8.2
     toml
     atomicwrites>=1.0;sys_platform=="win32"
     colorama;sys_platform=="win32"


nicoddemus:
Thanks!

While at it, would you mind taking care of this comment?

https://github.com/pytest-dev/pytest/blob/b322004047774f4b0992bfc8b9b1123069a6d0f1/setup.py#L3-L4

(We lost it by accident it seems when moving settings to `setup.cfg`)

nicoddemus:
Also, I think it is worth adding a `trivial` CHANGELOG entry for this. :+

bluetech:
@nicoddemus I added a changelog entry. The setup.cfg comment is actually no longer relevant ever since we imported `TerminalWriter` into pytest.

nicoddemus:
Ahh right, thanks! 👍

PR #7401, opened 2020-06-22
config: improve typing

bluetech:
This improves the typing around `_pytest.config`, and also makes pytest completely clean with regards to the py typings in https://github.com/pytest-dev/py/pull/232.

### DIFF
--- a/src/_pytest/_code/__init__.py
+++ b/src/_pytest/_code/__init__.py
@@ -6,6 +6,7 @@ from .code import Frame
 from .code import getfslineno
 from .code import getrawcode
 from .code import Traceback
+from .code import TracebackEntry
 from .source import compile_ as compile
 from .source import Source
 
@@ -17,6 +18,7 @@ __all__ = [
     "getfslineno",
     "getrawcode",
     "Traceback",
+    "TracebackEntry",
     "compile",
     "Source",
 ]
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -213,7 +213,7 @@ class TracebackEntry:
         return source.getstatement(self.lineno)
 
     @property
-    def path(self):
+    def path(self) -> Union[py.path.local, str]:
         """ path to the source code """
         return self.frame.code.path
 
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -15,10 +15,13 @@ from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import IO
+from typing import Iterable
+from typing import Iterator
 from typing import List
 from typing import Optional
 from typing import Sequence
 from typing import Set
+from typing import TextIO
 from typing import Tuple
 from typing import Union
 
@@ -42,6 +45,7 @@ from _pytest.compat import TYPE_CHECKING
 from _pytest.outcomes import fail
 from _pytest.outcomes import Skipped
 from _pytest.pathlib import import_path
+from _pytest.pathlib import ImportMode
 from _pytest.pathlib import Path
 from _pytest.store import Store
 from _pytest.warning_types import PytestConfigWarning
@@ -50,6 +54,7 @@ if TYPE_CHECKING:
     from typing import Type
 
     from _pytest._code.code import _TracebackStyle
+    from _pytest.terminal import TerminalReporter
     from .argparsing import Argument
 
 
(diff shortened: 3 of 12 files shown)


src/_pytest/_code/__init__.py
```diff
@@ -17,6 +18,7 @@
     "getfslineno",
     "getrawcode",
     "Traceback",
+    "TracebackEntry",
```
bluetech, line 21: `TracebackEntry` is exposed through `Traceback` so it should be visible too (though this is an internal package anyway).

src/_pytest/_code/code.py
```diff
@@ -213,7 +213,7 @@ def statement(self) -> "Source":
         return source.getstatement(self.lineno)
 
     @property
-    def path(self):
+    def path(self) -> Union[py.path.local, str]:
```
bluetech, line 216: `Code.path` has this annoying type due to some error path, we should fix it sometime...

src/_pytest/config/__init__.py
```diff
@@ -313,13 +333,11 @@ def __init__(self) -> None:
...
+        self._conftest_plugins = set()  # type: Set[types.ModuleType]
 
-        # state related to local conftest plugins
-        # Maps a py.path.local to a list of module objects.
-        self._dirpath2confmods = {}  # type: Dict[Any, List[object]]
-        # Maps a py.path.local to a module object.
-        self._conftestpath2mod = {}  # type: Dict[Any, object]
+        # State related to local conftest plugins.
+        self._dirpath2confmods = {}  # type: Dict[py.path.local, List[types.ModuleType]]
+        self._conftestpath2mod = {}  # type: Dict[Path, types.ModuleType]
```
bluetech, line 340: Actually stores `Path` and not what the previous comment said.

I removed the comments since they are redundant with the types.

src/_pytest/config/__init__.py
```diff
@@ -636,7 +671,6 @@ def import_plugin(self, modname: str, consider_entry_points: bool = False) -> No
         assert isinstance(modname, str), (
             "module name as text required, got %r" % modname
         )
-        modname = str(modname)
```
bluetech, line 639: Not needed per assert above.

src/_pytest/config/__init__.py
```diff
@@ -793,9 +831,9 @@ class InvocationParams:
             Plugins accessing ``InvocationParams`` must be aware of that.
         """
 
-        args = attr.ib(converter=tuple)
+        args = attr.ib(type=Tuple[str, ...], converter=_args_converter)
```
bluetech, line 834: Explicit converter instead of `tuple` needed just for the type annotation to work

src/_pytest/config/__init__.py
```diff
@@ -1331,21 +1374,23 @@ def _warn_about_missing_assertion(mode):
...
     if config.option.color == "yes":
         tw.hasmarkup = True
     if config.option.color == "no":
         tw.hasmarkup = False
     return tw
 
 
-def _strtobool(val):
-    """Convert a string representation of truth to true (1) or false (0).
+def _strtobool(val: str) -> bool:
```
bluetech, line 1392: Changed `_strtobool` to return bool instead of 0/1, seems more sensible.

src/_pytest/logging.py
```diff
@@ -141,9 +141,14 @@ def _get_auto_indent(auto_indent_option: Union[int, str, bool, None]) -> int:
 
         if auto_indent_option is None:
             return 0
-        elif type(auto_indent_option) is int:
```
bluetech, line 144: mypy doesn't do type narrowing on `type(...) is ..`, so I changed the code to use `isinstance` instead. Then things need to move a bit because `isinstance(True, int)` is true.

testing/test_config.py
```diff
@@ -1778,5 +1778,7 @@ def test_conftest_import_error_repr(tmpdir):
     ):
         try:
             raise RuntimeError("some error")
-        except Exception as e:
-            raise ConftestImportFailure(path, sys.exc_info()) from e
+        except Exception as exc:
+            assert exc.__traceback__ is not None
+            exc_info = (type(exc), exc, exc.__traceback__)
```
bluetech, line 1783: This is needed because `sys.exc_info()` has types with `BaseException` while we want just `Exception`.

src/_pytest/config/__init__.py
```diff
@@ -668,7 +702,7 @@ def import_plugin(self, modname: str, consider_entry_points: bool = False) -> No
             self.register(mod, modname)
 
 
-def _get_plugin_specs_as_list(specs):
+def _get_plugin_specs_as_list(specs) -> List[str]:
```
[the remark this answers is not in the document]
bluetech: I added another commit which does this now, but it also refactors the function a bit. I was thrown off by the `ModuleType` check but found the reference issue.

PR #7405, opened 2020-06-23
testing: fix flaky tests on pypy3 due to resource warnings in stderr

bluetech:
This happens a lot recently, for example https://github.com/pytest-dev/pytest/runs/800088944. I tried a bit to find the underlying cause but failed.

Use the same workaround as used previously in test_runner.py.

### DIFF
--- a/testing/test_stepwise.py
+++ b/testing/test_stepwise.py
@@ -75,6 +75,16 @@ def broken_testdir(testdir):
     return testdir
 
 
+def _strip_resource_warnings(lines):
+    # Assert no output on stderr, except for unreliable ResourceWarnings.
+    # (https://github.com/pytest-dev/pytest/issues/5088)
+    return [
+        x
+        for x in lines
+        if not x.startswith(("Exception ignored in:", "ResourceWarning"))
+    ]
+
+
 def test_run_without_stepwise(stepwise_testdir):
     result = stepwise_testdir.runpytest("-v", "--strict-markers", "--fail")
 
@@ -88,7 +98,7 @@ def test_fail_and_continue_with_stepwise(stepwise_testdir):
     result = stepwise_testdir.runpytest(
         "-v", "--strict-markers", "--stepwise", "--fail"
     )
-    assert not result.stderr.str()
+    assert _strip_resource_warnings(result.stderr.lines) == []
 
     stdout = result.stdout.str()
     # Make sure we stop after first failing test.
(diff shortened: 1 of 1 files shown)


gnikonorov:
Can we make an issue to track this separately @bluetech ? I'm +1 on temporarily removing these to not have the pr pipeline break but I feel like we should track down what's causing this and try to fix it at some point, since it's a legitimate error

bluetech:
> I'm +1 on temporarily removing these to not have the pr pipeline break but I feel like we should track down what's causing this and try to fix it at some point, since it's a legitimate error

We can reopen #5088 if you think it's worth it, however I suspect it will languish as the issue is rather elusive and hard to pin down. It doesn't come from the failing test itself, but some place before, and it also doesn't reproduce locally, and only occasionally on CI. It tried to insert `gc.collect` calls in various places to narrow it down, tried ` -X track-resources` and some other stuff.

gnikonorov:
I’ll defer reopening to someone with more background. I’m not opposed to leaving it closed. Can we just post the information in your above comment to the issue so future readers know why it’s commented out?

testing/test_stepwise.py
```diff
@@ -75,6 +75,16 @@ def broken_testdir(testdir):
     return testdir
 
 
+def _strip_resource_warnings(lines):
+    # Assert no output on stderr, except for unreliable ResourceWarnings.
```
gnikonorov, line 79: Can we change this comment? The way it’s written right now is misleading since we don’t assert in the function but after we return.

Alternatively, maybe we could move the assert into the function?

bluetech, line 79:
```suggestion
    # Strip unreliable ResourceWarnings, so no-output assertions on stderr can work.
```

bluetech: Thanks, I updated the comment.

> Alternatively, maybe we could move the assert into the function?

I usually prefer the assert to be in the function itself, that way the stacktrace is nicer.

nicoddemus:
> ’ll defer reopening to someone with more background. I’m not opposed to leaving it closed. Can we just post the information in your above comment to the issue so future readers know why it’s commented out?

I agree to leave it closed for the reasons @bluetech brought up; having a comment in the code and in the issue we can at least track back to the reasons that lead to it.

nicoddemus:
After this gets merged, we should add a note on #5088 explaining that this PR did not fix the underlying cause, but did workaround it because we could not pin the underlying issue. 👍

bluetech:
@gnikonorov Does that sound OK to you?

gnikonorov:
That sounds great to me @bluetech, approved

nicoddemus:
Thanks folks!

PR #7412, opened 2020-06-24
Improvement/recursive diff output

### LINKED ISSUE #7348: Improve recursive diff output for comparison asserts on dataclasses/attrs
pytest 6 will have a new feature (PR #6835) for displaying nicer diff output on large dataclasses. Often dataclasses have other dataclasses or other types nested inside them, and the assertion fails because one or more deeply nested fields differ. Before the diff was just on the top-level `repr` of the objects, which was very hard to decipher for larger dataclasses. After the PR, the output looks like this (see https://github.com/pytest-dev/pytest/pull/6835#issuecomment-637545552):

```
    def test_cmp():
>       assert obj == obj.replace(assentior=obj.assentior.replace(culpa=7, aliquip=Aliquip.Aliquip3))
E       AssertionError: assert Dignissim(assentior=Assentior(aliquip=<Aliquip.Aliquip1: 1>, culpa=6, fugiat=Fugiat(eiusmod='aaaaaaaaaaaaaaaa', tempor='bbbbbb', incididunt='CCCCCCCCCCCCCCCCCCC', labore='dddddddddd', dolore='eeeeeeeeee', magna='fffffffffff', aliqua='gggggggggggggg', veniam=None, nostrud=None, exercitation=None, ullamco=None, laboris=None, commodo=None, consequat=None, aute=None), invenire=Invenire(irure=53, reprehenderit=153, voluptate=242, velit=100, esse=5035, cillum=53, eepcillum=422), deserunt=True, lorem=Lorem(ipsum=b'', dolor=['xx'], sit=['yy'], amet=['zz'], consectetur=b'sdfsd', adipiscing=b'adfsdf'), mollit=4294967294, laborums=[Laborum(type=23, urangulal=b'abc', ipsumal=None, occaecat=b'sdf', cupidatat=13, proident=None)], tantas='sdfsdlxcv49249sdfs90sdf==', nominati=-1, fabulas=False, tritani=None), new_cupidatat=13, laoreet=1, rationibus=False) == Dignissim(assentior=Assentior(aliquip=<Aliquip.Aliquip3: 3>, culpa=7, fugiat=Fugiat(eiusmod='aaaaaaaaaaaaaaaa', tempor='bbbbbb', incididunt='CCCCCCCCCCCCCCCCCCC', labore='dddddddddd', dolore='eeeeeeeeee', magna='fffffffffff', aliqua='gggggggggggggg', veniam=None, nostrud=None, exercitation=None, ullamco=None, laboris=None, commodo=None, consequat=None, aute=None), invenire=Invenire(irure=53, reprehenderit=153, voluptate=242, velit=100, esse=5035, cillum=53, eepcillum=422), deserunt=True, lorem=Lorem(ipsum=b'', dolor=['xx'], sit=['yy'], amet=['zz'], consectetur=b'sdfsd', adipiscing=b'adfsdf'), mollit=4294967294, laborums=[Laborum(type=23, urangulal=b'abc', ipsumal=None, occaecat=b'sdf', cupidatat=13, proident=None)], tantas='sdfsdlxcv49249sdfs90sdf==', nominati=-1, fabulas=False, tritani=None), new_cupidatat=13, laoreet=1, rationibus=False)
E         Matching attributes:
E         ['new_cupidatat', 'laoreet', 'rationibus']
E         Differing attributes:
E         assentior: Assentior(aliquip=<Aliquip.Aliquip1: 1>, culpa=6, fugiat=Fugiat(eiusmod='aaaaaaaaaaaaaaaa', tempor='bbbbbb', incididunt='CCCCCCCCCCCCCCCCCCC', labore='dddddddddd', dolore='eeeeeeeeee', magna='fffffffffff', aliqua='gggggggggggggg', veniam=None, nostrud=None, exercitation=None, ullamco=None, laboris=None, commodo=None, consequat=None, aute=None), invenire=Invenire(irure=53, reprehenderit=153, voluptate=242, velit=100, esse=5035, cillum=53, eepcillum=422), deserunt=True, lorem=Lorem(ipsum=b'', dolor=['xx'], sit=['yy'], amet=['zz'], consectetur=b'sdfsd', adipiscing=b'adfsdf'), mollit=4294967294, laborums=[Laborum(type=23, urangulal=b'abc', ipsumal=None, occaecat=b'sdf', cupidatat=13, proident=None)], tantas='sdfsdlxcv49249sdfs90sdf==', nominati=-1, fabulas=False, tritani=None) != Assentior(aliquip=<Aliquip.Aliquip3: 3>, culpa=7, fugiat=Fugiat(eiusmod='aaaaaaaaaaaaaaaa', tempor='bbbbbb', incididunt='CCCCCCCCCCCCCCCCCCC', labore='dddddddddd', dolore='eeeeeeeeee', magna='fffffffffff', aliqua='gggggggggggggg', veniam=None, nostrud=None, exercitation=None, ullamco=None, laboris=None, commodo=None, consequat=None, aute=None), invenire=Invenire(irure=53, reprehenderit=153, voluptate=242, velit=100, esse=5035, cillum=53, eepcillum=422), deserunt=True, lorem=Lorem(ipsum=b'', dolor=['xx'], sit=['yy'], amet=['zz'], consectetur=b'sdfsd', adipiscing=b'adfsdf'), mollit=4294967294, laborums=[Laborum(type=23, urangulal=b'abc', ipsumal=None, occaecat=b'sdf', cupidatat=13, proident=None)], tantas='sdfsdlxcv49249sdfs90sdf==', nominati=-1, fabulas=False, tritani=None)
E         
E         Drill down into differing attribute assentior
E         Matching attributes:
E         ['fugiat',
E          'invenire',
E          'deserunt',
E          'lorem',
E          'mollit',
E          'laborums',
E          'tantas',
E          'nominati',
E          'fabulas',
E          'tritani']
E         Differing attributes:
E         aliquip: <Aliquip.Aliquip1: 1> != <Aliquip.Aliquip3: 3>
E         
E         Drill down into differing attribute aliquip
E         +<Aliquip.Aliquip1: 1>
E         -<Aliquip.Aliquip3: 3>
E         culpa: 6 != 7
E         
E         Drill down into differing attribute culpa
E         +6
E         -7

x.py:169: AssertionError
```

This is a great improvement, but the drill down structure is a bit hard to see. In this case particularly, there are two differing nested attributes, `aliquip` and `culpa`, but the `culpa` one starts just after the output of the `aliquip` without any spacing, which is confusing. Also, it is not visible at which nesting level the differing attribute is.

Vlad-Radz:
closes #7348


nicoddemus:
Hi @Vlad-Radz!

Thanks for the contribution, but the commit history is messy, it contains some other changes which are already on master. Can you try to clean it up? Let us know if you need help with that. 👍

Vlad-Radz:
Hello @nicoddemus! Did it - now it has only 2 commits, both of mine.

Vlad-Radz:
Hello @nicoddemus! Did it - now it has only 3 commits, all of mine. All tests are through.

src/_pytest/assertion/util.py
```diff
@@ -169,23 +190,56 @@ def assertrepr_compare(config, op: str, left: Any, right: Any) -> Optional[List[
...
+    :param verbose
+    :type verbose: int
+    .. note:: indendation used for better visual perception of the comparison report.
+    :return: explanation of an AssertionError
+    :rtype: List[str]
+    """
+    explanation = []
+
+    my_name = cast(types.FrameType, inspect.currentframe()).f_code.co_name
+    num_calls = _get_number_of_calls(func_name=my_name)  # type: int
```
nicoddemus, line 210: I see why you are using `_get_number_of_calls` here, but couldn't we instead receive a `indentation` parameter as argument, and forward that down the stack like it was done with the other methods? Seems less "magic" than fiddling with frame objects.

src/_pytest/assertion/util.py
```diff
@@ -169,23 +190,56 @@ def assertrepr_compare(config, op: str, left: Any, right: Any) -> Optional[List[
 
 
 def _compare_eq_any(left: Any, right: Any, verbose: int = 0) -> List[str]:
-    explanation = []  # type: List[str]
+    """
+    Compare 2 objects.
+    The generic comparison function, which calls other functions based on the input's type.
+
+    :param left
```
nicoddemus, line 197: Probably the empty parameter docs were added by your IDE automatically, but could you please remove the empty ones, or even better, add a description?

Also I think in general `:type` is redundant when we are already annotating the function signature (as is the case here), so they should be removed.

src/_pytest/assertion/util.py
nicoddemus: Also it is probably easier to treat indentation as a single number that increases by one for each indentation level, and then each function that receives the indentation level just calls an utility function to build the indent string:

```python
def _compare_eq_verbose(left: Any, right: Any, indent_level: int) -> List[str]:
    indent_level += 1
    indent = _get_indent(indent_level)

...

def _get_indent(indent_level):
    return " " * (indent_level * 4)
```

Vlad-Radz: Thank you @nicoddemus for reviewing the changes! 
I am not sure that I understand your proposal right: you propose to pass the `indentation` (or `indent_level`) as argument to `_compare_eq_any()`, right? But how would this function (or the calling function - `assertrepr_compare`) know, how many times was it called (and it needs that to make the right indentation, dependent on the level of nestedness of data)? We could make a wrapper around the function, which would store the state (number of executions), but that would be reinvention of the call stack. Or do I get your idea completely wrong?

Regarding having `_get_indent()` - you propose to call this function directly from every function, that needs indentation, right? Wouldn't it create additional overhead? An option would be to call this function just from `_compare_eq_any()`, instead of creating the indent string there, and pass the string as parametere, as I do now. But when you think, that overhead is fine, and from design perspective it's better to call the `_get_indent()` directly from every needed function, I'll do that.

src/_pytest/assertion/util.py
Vlad-Radz: Sure, I agree with both.

src/_pytest/assertion/util.py
Vlad-Radz: I think, I got some clarification from @bluetech

Vlad-Radz:
@bluetech Thank you for the review. I left 2 comments there, to the points where you are defineitely right, but I had some issues.

src/_pytest/assertion/util.py
nicoddemus: Thanks! I'm assuming you got my suggestion then, but let me know otherwise and I will try to explain it.

src/_pytest/assertion/util.py
```diff
@@ -96,6 +96,13 @@ def _format_lines(lines: Sequence[str]) -> List[str]:
     return result
 
 
+def _get_indent(indent_level):
```
bluetech, line 99:
```suggestion
def _get_indent(indent_level: int) -> str:
```

src/_pytest/assertion/util.py
```diff
@@ -168,22 +176,47 @@ def assertrepr_compare(config, op: str, left: Any, right: Any) -> Optional[List[
     return [summary] + explanation
 
 
-def _compare_eq_any(left: Any, right: Any, verbose: int = 0) -> List[str]:
-    explanation = []  # type: List[str]
+def _compare_eq_any(
+    left: Any, right: Any, indent_level: int, verbose: int = 0
+) -> List[str]:
+    """
+    Generic comparison function, dispatches to other comparison functions based on the inputs' type.
```
bluetech, line 183:
```suggestion
    Generic comparison function, dispatches to other comparison functions based on the inputs' type.

```

src/_pytest/assertion/util.py
```diff
@@ -168,22 +176,47 @@ def assertrepr_compare(config, op: str, left: Any, right: Any) -> Optional[List[
...
+    :param left: 1st object passed for comparison.
+    :param right: 2nd object passed for comparison.
+    :param indent_level: indentation level, for better visual perception of the output.
+    :param verbose: decides whether to print full error report or only a part of it.
+    """
+    if not isinstance(indent_level, int):
```
bluetech, line 189: I suggest removing this, usually we don't type-check private functions, we assume the caller passes the correct types (or that mypy would catch the mistake).

src/_pytest/assertion/util.py
```diff
@@ -168,22 +176,47 @@ def assertrepr_compare(config, op: str, left: Any, right: Any) -> Optional[List[
...
+        raise TypeError(
+            "Wrong type for argument `indent_level`! Expected: int, got: {}, value: {}".format(
+                type(indent_level), indent_level
+            )
+        )
+    elif indent_level < -1:
```
bluetech, line 195: I'd remove this too, or at most, change it into `assert indent_level >= -1`.

src/_pytest/assertion/util.py
```diff
@@ -168,22 +176,47 @@ def assertrepr_compare(config, op: str, left: Any, right: Any) -> Optional[List[
...
+        raise ValueError(
+            "Argument `indent_level` can't be less than -1! Got following value: {}".format(
+                indent_level
+            )
+        )
+    else:
+        indent_level += 1
```
bluetech, line 202: Instead of doing this, it will be better to start out with `0` in `assertrepr_compare`, and add when when drilling down, i.e. pass `indent_level + 1` in the arguments where it's needed.

testing/test_assertion.py
```diff
@@ -910,7 +909,9 @@ class SimpleDataObject:
 
         lines = callequal(left, right)
         assert lines is not None
-        assert "field_d: 'a' != 'b'" in lines
+        assert (
+            "    field_d: 'a' != 'b'" in lines
+        )  # indentation in output because of nested object structure
```
bluetech, line 914:
```suggestion
        # Indentation in output because of nested object structure.
        assert "    field_d: 'a' != 'b'" in lines
```

Vlad-Radz:
Hello @bluetech! I fixed the most of the things, please take a look at the comments. Checks for windows-py35 failed because of the timeout - I am not sure it's code issue. Can I somehow restart the tests?

src/_pytest/assertion/util.py
```diff
@@ -436,16 +478,20 @@ def _compare_eq_cls(
     if same and verbose < 2:
         explanation.append("Omitting %s identical items, use -vv to show" % len(same))
     elif same:
-        explanation += ["Matching attributes:"]
-        explanation += pprint.pformat(same).splitlines()
+        explanation += ["{}Matching attributes:".format(indent)]
+        mystr = indent + str(pprint.pformat(same).splitlines())
```
bluetech, line 482: Here, instead of calling `str` on the list, we should instead indent every line on its own, and append them to `explanation`, like before. Otherwise the attributes are shown with some repr format (this is why you had to change a bunch of unrelated parts ot the tests).

Vlad-Radz:
Hello @bluetech, I fixed the issues and really liked your proposal. All tests are through. 

Thank you for your patience - I appreciate that you always try to explain the things.

testing/example_scripts/dataclasses/test_compare_recursive_dataclasses.py
```diff
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
```
bluetech, line 1: I think the names here are bit too long/unwieldy for the tests, and also it doesn't test the situation perfectly.

I played with it a bit and I think this file would be a good replacement:

```py


@dataclass
class S:
    a: int
    b: str


@dataclass
class C:
    c: S
    d: S


@dataclass
class C2:
    e: C
    f: S


@dataclass
class C3:
    g: S
    h: C2
    i: str
    j: str


def test_recursive_dataclasses():
    left = C3(
        S(10, "ten"), C2(C(S(1, "one"), S(2, "two")), S(2, "three")), "equal", "left",
    )
    right = C3(
        S(20, "xxx"), C2(C(S(1, "one"), S(2, "yyy")), S(3, "three")), "equal", "right",
    )

    assert left == right
```

And IMO this would be the desirable output (but let me know if you think differently):

```
E       AssertionError: assert C3(g=S(a=10, b='ten'), h=C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='two')), f=S(a=2, b='three')), i='equal', j='left') == C3(g=S(a=20, b='xxx'), h=C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='yyy')), f=S(a=3, b='three')), i='equal', j='right')
E         
E         Matching attributes:
E         ['i']
E         Differing attributes:
E         ['g', 'h', 'j']
E         
E         Drill down into differing attribute g:
E           g: S(a=10, b='ten') != S(a=20, b='xxx')
E           
E           Differing attributes:
E           ['a', 'b']
E           
E           Drill down into differing attribute a:
E             a: 10 != 20
E             +10
E             -20
E           
E           Drill down into differing attribute b:
E             b: 'ten' != 'xxx'
E             - xxx
E             + ten
E         
E         Drill down into differing attribute h:
E           h: C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='two')), f=S(a=2, b='three')) != C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='yyy')), f=S(a=3, b='three'))
E           
E           Differing attributes:
E           ['e', 'f']
E           
E           Drill down into differing attribute e:
E             e: C(c=S(a=1, b='one'), d=S(a=2, b='two')) != C(c=S(a=1, b='one'), d=S(a=2, b='yyy'))
E             
E             Matching attributes:
E             ['c']
E             Differing attributes:
E             ['d']
E             
E             Drill down into differing attribute d:
E               d: S(a=2, b='two') != S(a=2, b='yyy')
E               
E               Matching attributes:
E               ['a']
E               Differing attributes:
E               ['b']
E               
E               Drill down into differing attribute b:
E                 b: 'two' != 'yyy'
E                 - yyy
E                 + two
E           
E           Drill down into differing attribute f:
E             f: S(a=2, b='three') != S(a=3, b='three')
E             
E             Matching attributes:
E             ['b']
E             Differing attributes:
E             ['a']
E             
E             Drill down into differing attribute a:
E               a: 2 != 3
E               +2
E               -3
E         
E         Drill down into differing attribute j:
E           j: 'left' != 'right'
E           - right
E           + left
```
```

File: src/_pytest/assertion/util.py [hunk 11]
```diff
@@ -436,17 +451,25 @@ def _compare_eq_cls(
...
-        explanation += pprint.pformat(same).splitlines()
+        explanation += ["{}Matching attributes:".format(indent)]
+        lines = pprint.pformat(same).splitlines()
+        for line in lines:
+            explanation += [indent + str(line)]
     if diff:
-        explanation += ["Differing attributes:"]
+        explanation += ["{}Differing attributes:".format(indent)]
         for field in diff:
+            indent_level += 1
```
bluetech, line 461: This one is not exactly right; it causes differing attributes at the same tree level to be indented differently.

Generally since the indentation is only done one level now, I think we can do away with the `indent_level` variable and just add the indentation directly.

src/_pytest/assertion/util.py
```diff
@@ -96,6 +96,13 @@ def _format_lines(lines: Sequence[str]) -> List[str]:
     return result
 
 
+def _get_indent(indent_level: int) -> str:
+    """
+    Returns the indentation string based on required indentation level.
+    """
+    return " " * (indent_level * 4)
```
bluetech, line 103: It seems that `_format_lines` already uses 2-space indents we should probably be consistent and use that instead of 4-space indents.

src/_pytest/assertion/util.py
```diff
@@ -169,14 +176,22 @@ def assertrepr_compare(config, op: str, left: Any, right: Any) -> Optional[List[
...
-    explanation = []  # type: List[str]
+    """
+    Generic comparison function, dispatches to other comparison functions based on the inputs' type.
+    :param left: 1st object passed for comparison.
+    :param right: 2nd object passed for comparison.
+    :param verbose: decides whether to print full error report or only a part of it.
+    """
+
+    explanation = []
+
```
bluetech, line 187: This initialization is not needed:

```suggestion

    explanation = []

```

src/_pytest/assertion/util.py
Vlad-Radz: I got your point - sorry for not doing enough tests, I was sure, that it is necessary to increase `indent_level` by one

testing/example_scripts/dataclasses/test_compare_recursive_dataclasses.py
Vlad-Radz: @bluetech does this output look OK to you? Difference to your proposed output is that:
- the lines where differing attributes are being compared (like: `g: S(a=10, b='ten') != S(a=20, b='xxx')`) 
- and the lines where differing fields of "simple" data structures are being outputted (like `+10`)

are on the same level as the expression `Drill down into differing attribute <attr_name>`. I could put the first type of lines on an 1x more nested level, but with the second type it would be more complicated, I guess: we either need to check the output of `_compare_eq_any()`, or pass the `indentation_level` again to all comparison functions. Or am I wrong and miss something?

Current output (not pushed yet):

```
E       AssertionError: assert C3(g=S(a=10, b='ten'), h=C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='two')), f=S(a=2, b='three')), i='equal', j='left') == C3(g=S(a=20, b='xxx'), h=C2(e=C(c=S(a=1, b='one'), d=S(a=2, b=
'yyy')), f=S(a=3, b='three')), i='equal', j='right')
E         
E           Matching attributes:
E           ['i']
E           Differing attributes:
E           ['g', 'h', 'j']
E         
E           Drill down into differing attribute g:
E           g: S(a=10, b='ten') != S(a=20, b='xxx')
E           
E             Differing attributes:
E             ['a', 'b']
E           
E             Drill down into differing attribute a:
E             a: 10 != 20
E             +10
E             -20
E           
E             Drill down into differing attribute b:
E             b: 'ten' != 'xxx'
E             - xxx
E             + ten
E         
E           Drill down into differing attribute h:
E           h: C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='two')), f=S(a=2, b='three')) != C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='yyy')), f=S(a=3, b='three'))
E           
E             Differing attributes:
E             ['e', 'f']
E           
E             Drill down into differing attribute e:
E             e: C(c=S(a=1, b='one'), d=S(a=2, b='two')) != C(c=S(a=1, b='one'), d=S(a=2, b='yyy'))
E             
E               Matching attributes:
E               ['c']
E               Differing attributes:
E               ['d']
E             
E               Drill down into differing attribute d:
E               d: S(a=2, b='two') != S(a=2, b='yyy')
E               
E                 Matching attributes:
E                 ['a']
E                 Differing attributes:
E                 ['b']
E               
E                 Drill down into differing attribute b:
E                 b: 'two' != 'yyy'
E                 - yyy
E                 + two
E           
E             Drill down into differing attribute f:
E             f: S(a=2, b='three') != S(a=3, b='three')
E             
E               Matching attributes:
E               ['b']
E               Differing attributes:
E               ['a']
E             
E               Drill down into differing attribute a:
E               a: 2 != 3
E               +2
E               -3
E         
E           Drill down into differing attribute j:
E           j: 'left' != 'right'
E           - right
E           + left
```

@nicoddemus  for your info.

bluetech: That's already pretty good!

I do prefer my version because for visual consistency I think it's better if each "Drill down" actually drills down instead of sometimes yes and sometimes no. The idea is that the user gets the idea just from the visual tree structure even before reading the details.

> I could put the first type of lines on an 1x more nested level

Right, this one line you can just indent.

> but with the second type it would be more complicated, I guess: we either need to check the output of _compare_eq_any(), or pass the indentation_level again to all comparison functions. Or am I wrong and miss something?

If we indent all of the lines of the recursive `_compare_eq_any` (in `_compare_eq_cls`, not by passing indent level) then it should work out well unless I'm missing something. Is that what you mean by "check the output of `_compare_eq_any()`"?

nicoddemus: > I do prefer my version because for visual consistency I think it's better if each "Drill down" actually drills down instead of sometimes yes and sometimes no. The idea is that the user gets the idea just from the visual tree structure even before reading the details.

I agree, because then the `Drill down into` acts like a header.

Vlad-Radz: @bluetech Not exactly what I mean. 

I already indent all the lines of the recursive `_compare_eq_any` (in `_compare_eq_cls`). But since in our example the last values in the hierarchy are differing integer / string values (not dataclasses / attrs), we indent them with same level, as the `Drill down into`. If we indent everything from `_compare_eq_any` with double level, then it will lead to a new problem: analysis of the difference (e.g. `Differing attributes:`) will be on diff. level of nestedness, than e.g. `g: S(a=10, b='ten') != S(a=20, b='xxx')`. 

So, basically, we can't have same indentation level for both types of lines, nor different indentation level. That's why I said, that I see only one possibility to fix this: check the output of `_compare_eq_any()`. By that I mean the following: if the line starts with `"+"` or `"-"`, double the indentation for these lines.

Vlad-Radz: I will give an example. This is what happens when I indent both types of lines with double level:

```
E       AssertionError: assert C3(g=S(a=10, b='ten'), h=C2(e=C(c=S(a=1, b='one'), d=S(a=2, b='two')), f=S(a=2, b='three')), i='equal', j='left') == C3(g=S(a=20, b='xxx'), h=C2(e=C(c=S(a=1, b='one'), d=S(a=2, b=
'yyy')), f=S(a=3, b='three')), i='equal', j='right')
E         
E           Matching attributes:
E           ['i']
E           Differing attributes:
E           ['g', 'h', 'j']
E         
E           Drill down into differing attribute g:
E             g: S(a=10, b='ten') != S(a=20, b='xxx')
E             
E               Differing attributes:
E               ['a', 'b']
```

bluetech: In general I don't see a reason for this to happen. Can you push your latest version, then we can figure it out? (The tests don't need to pass in the meanwhile).

Vlad-Radz: @bluetech did it

src/_pytest/assertion/util.py
```diff
@@ -350,9 +361,7 @@ def _compare_eq_sequence(
     return explanation
 
 
-def _compare_eq_set(
-    left: AbstractSet[Any], right: AbstractSet[Any], verbose: int = 0
-) -> List[str]:
+def _compare_eq_set(left: AbstractSet[Any], right: AbstractSet[Any], verbose: int = 0) -> List[str]:
```
bluetech, line 364: Here too, should change to the way it was.

src/_pytest/assertion/util.py
```diff
@@ -368,15 +377,17 @@ def _compare_eq_set(
 
 
 def _compare_eq_dict(
-    left: Mapping[Any, Any], right: Mapping[Any, Any], verbose: int = 0
+    left: Mapping[Any, Any], right: Mapping[Any, Any], verbose: int = 0,
```
bluetech, line 380: Here too, should change to the way it was.

src/_pytest/assertion/util.py
```diff
@@ -368,15 +377,17 @@ def _compare_eq_set(
...
 ) -> List[str]:
     explanation = []  # type: List[str]
     set_left = set(left)
     set_right = set(right)
     common = set_left.intersection(set_right)
     same = {k: left[k] for k in common if left[k] == right[k]}
     if same and verbose < 2:
-        explanation += ["Omitting %s identical items, use -vv to show" % len(same)]
+        explanation += [
```
bluetech, line 388: Changing `%` to `format` is an unrelated change as well, so would change it back to the way it was just to keep the diff focused.

src/_pytest/assertion/util.py
```diff
@@ -432,21 +444,34 @@ def _compare_eq_cls(
         else:
             diff.append(field)
 
-    explanation = []
+    explanation = [""]
```
bluetech, line 447: It will be slightly safer to change this to

```suggestion
    explanation = []
    if same or diff:
        explanation += [""]
```

Just in case sometime equal objects are passed in, we don't want the empty line.

testing/example_scripts/dataclasses/test_compare_recursive_dataclasses.py
bluetech: Thanks! I left comments on the changes I think we should make.

testing/test_assertion.py
```diff
@@ -772,56 +772,70 @@ def test_mojibake(self) -> None:
 
 class TestAssert_reprcompare_dataclass:
     @pytest.mark.skipif(sys.version_info < (3, 7), reason="Dataclasses in Python3.7+")
-    def test_dataclasses(self, testdir):
+    def test_dataclasses(self, testdir, consecutive=True):
```
bluetech, line 775: Note that `consecutive=True` should be passed as a keyword argument to `fnmatch_lines`; the above has no effect.

Vlad-Radz: get it, thank you, tried to check if I did it right, but didn't understand it.

Vlad-Radz:
@bluetech Thank you for patience and support - your ideas made my initial idea much simplier

PR #7414, opened 2020-06-24
Rephrase entry-point to avoid conveying setuptools must be used.

Carreau:
Entrypoints are not setuptools specific. The phrasing make it looks like
one must use setuptools.

Link to other project allowing developers to create entrypoints.

### DIFF
--- a/doc/en/writing_plugins.rst
+++ b/doc/en/writing_plugins.rst
@@ -17,7 +17,7 @@ reporting by calling :ref:`well specified hooks <hook-reference>` of the followi
 * builtin plugins: loaded from pytest's internal ``_pytest`` directory.
 
 * :ref:`external plugins <extplugins>`: modules discovered through
-  `setuptools entry points`_
+  `entry points`_
 
 * `conftest.py plugins`_: modules auto-discovered in test directories
 
@@ -35,7 +35,7 @@ Plugin discovery order at tool startup
 
 * by loading all builtin plugins
 
-* by loading all plugins registered through `setuptools entry points`_.
+* by loading all plugins registered through `entry points`_.
 
 * by pre-scanning the command line for the ``-p name`` option
   and loading the specified plugin before actual command line parsing.
@@ -129,6 +129,7 @@ Also consider :ref:`contributing your plugin to pytest-dev<submitplugin>`
 once it has some happy users other than yourself.
 
 
+.. _`entry points`:
 .. _`setuptools entry points`:
 .. _`pip-installable plugins`:
 
@@ -138,10 +139,10 @@ Making your plugin installable by others
 If you want to make your plugin externally available, you
 may define a so-called entry point for your distribution so
 that ``pytest`` finds your plugin module.  Entry points are
-a feature that is provided by `setuptools`_. pytest looks up
-the ``pytest11`` entrypoint to discover its
+a feature that is provided by `setuptools`_ and many other packaging tools.
+pytest looks up the ``pytest11`` entrypoint to discover its
 plugins and you can thus make your plugin available by defining
-it in your setuptools-invocation:
+it in for example your setuptools-invocation:
 
 .. sourcecode:: python
 
@@ -156,6 +157,10 @@ it in your setuptools-invocation:
         # custom PyPI classifier for pytest plugins
         classifiers=["Framework :: Pytest"],
     )
+
+Or via tools dependants mechanism when using other packaging framework like `poetry
+<https://python-poetry.org/docs/pyproject/#plugins>`_ or `flit
+<https://flit.readthedocs.io/en/stable/entrypoints.html#specifying-entry-points>`_.
 
 If a package is installed this way, ``pytest`` will load
 ``myproject.pluginmodule`` as a plugin which can define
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks ! 

Also not sure it fits in Pytest docs, but https://pypi.org/project/entrypoints/ allow you to find and list entry points w/o the help of setuptools.

nicoddemus:
> Also not sure it fits in Pytest docs, but pypi.org/project/entrypoints allow you to find and list entry points w/o the help of setuptools.

Thanks, but I don't think it does, `entrypoints` seems targeted to framework authors than general users.

graingert:
https://pypi.org/project/importlib-metadata/ provides the standard way to get entrypoint information

doc/en/writing_plugins.rst
```diff
@@ -157,6 +158,10 @@ it in your setuptools-invocation:
         classifiers=["Framework :: Pytest"],
     )
 
+Or via tools dependants mechanism when using other packaging framework like `poetry
```
graingert, line 161:
```suggestion
Or via tool specific mechanism when using other packaging framework like `poetry
```

graingert:
the python.org docs seem very setuptoolsy:  Read the setuptools docs for more information on entrypoints, their definition, and usage.

doc/en/writing_plugins.rst
```diff
@@ -157,6 +158,10 @@ it in your setuptools-invocation:
         classifiers=["Framework :: Pytest"],
     )
 
+Or via tool specific mechanism when using other packaging framework like `poetry
+<https://python-poetry.org/docs/pyproject/#plugins>`_ or `flit
+<https://flit.readthedocs.io/en/stable/entrypoints.html#specifying-entry-points>`_.
```
asottile, line 163: I think this is not correct, these tools have their own settings for generating setuptools entry_points

Carreau: Well if you can register, and list, entry points without setuptools on your system, are they setuptools entry_points ? 

I guess it's a question of semantic, yes entry_point of this type have been initially created by setuptools sure they are "setuptools entry_points" the problem is that if you keep referring to them as "setuptools entry_points" user will assume you _need_ setuptools to create those.

asottile: afaik all of the example tools added in this PR use setuptools under the hood to create them so...

graingert: I didn't think flit used setuptools to make wheels

Carreau: Yes that is correct, I send the PR as I created https://github.com/Carreau/pytest_covid/blob/master/pyproject.toml using flit, that does works well, and does not use setuptools.

asottile: nah, it definitely uses setuptools, here's the code for it -- ^F entry_points: https://github.com/takluyver/flit/blob/master/flit/sdist.py

graingert: That's the sdist shim for build systems without PEP517/518 here's how it works when creating whl:

https://github.com/takluyver/flit/blob/master/flit_core/flit_core/wheel.py#L193

Carreau: Yes. You _can_ generate a `setup.py` for compatibility but don't have to. So definitively no you do not have to use setuptools to generate entry_points.

asottile: ah I see flit has changed, either way the idea of entry points was "standardized" by setuptools so I believe it is correct to call them "setuptools entrypoints"

graingert: Seems they're defined here now: https://packaging.python.org/specifications/entry-points/

graingert: > The entry point file format was originally developed to allow packages built with setuptools to provide integration point metadata that would be read at runtime with pkg_resources. It is now defined as a PyPA interoperability specification in order to allow build tools other than setuptools to publish pkg_resources compatible entry point metadata, and runtime libraries other than pkg_resources to portably read published entry point metadata (potentially with different caching and conflict resolution strategies).

graingert: Is there a list of other PyPA interoperability specifications? Or is it really just defacto pip implementation defined?

nicoddemus: IMHO opinion we should dwell too much into the semantics here: I think the most important message is that users are able to create plugins using a myriad of tools: setup.py/setuptools, flint, etc.

Carreau:
Seem we reach a dead end, closing. Let me know if you wish to reopen.

nicoddemus:
Thanks @Carreau for bringing this up.

I still have the same opinion though:

> IMHO opinion we should dwell too much into the semantics here: I think the most important message is that users are able to create plugins using a myriad of tools: setup.py/setuptools, flit, etc.

Sadly IMO the discussion drifted too much into specific semantics.

Carreau:
Thanks, I appreciate, sometime it's just better to close and retry latter. 

I'm personally trying to nudge the ecosystem toward declarative (non setup tools) base install so every bit help.

PR #7427, opened 2020-06-28
Display empty or None help parameter for parser option as empty string

### LINKED ISSUE #7394: Don't throw index error when running `pytest --help` when adding ini option via `parser.addini`  in `conftest.py` with no help
###  a detailed description of the bug or suggestion

When running `pytest --help` with a `conftest.py` file present that defines an ini option with an empty string as help text, an index error is thrown. We should not throw an error here, but what exactly we should do I think is open for discussion.

When passing `None`, a different error is throw. We should fix both the issue that happens with `None` and the empty string as part of this

### output of `pip list` from the virtual environment you are using

[pip_list_pytest_help_issue.txt](https://github.com/pytest-dev/pytest/files/4809858/pip_list_pytest_help_issue.txt)

### pytest and operating system versions
Pytest is built from master locally:
```
(.venv) gnikonorov:~/pytest/gleb_test$ pytest --version
pytest 5.4.1.dev546+g4cc4ebf3c
```

### minimal example if possible
Contents of `conftest.py`:
```
(.venv) gnikonorov:~/pytest/gleb_test$ cat conftest.py
def pytest_addoption(parser):
    parser.addini("my_ini", "", default=True, type="bool")
```

```
(.venv) gnikonorov:~/pytest/gleb_test$ pytest --help
usage: pytest [options] [file_or_dir] [file_or_dir] [...]
.....
                        Automatically find and add a Django project to the Python path.
  FAIL_INVALID_TEMPLATE_VARS (bool):
                        Fail for invalid variables in templates.
  my_ini (bool):        Traceback (most recent call last):
  File "/home/gnikonorov/pytest/.venv/bin/pytest", line 11, in <module>
    load_entry_point('pytest', 'console_scripts', 'pytest')()
  File "/home/gnikonorov/pytest/src/_pytest/config/__init__.py", line 165, in console_main
    code = main()
  File "/home/gnikonorov/pytest/src/_pytest/config/__init__.py", line 143, in main
    config=config
  File "/home/gnikonorov/pytest/.venv/lib/python3.6/site-packages/pluggy/hooks.py", line 286, in __call__
  File "/home/gnikonorov/pytest/.venv/lib/python3.6/site-packages/pluggy/manager.py", line 93, in _hookexec
  File "/home/gnikonorov/pytest/.venv/lib/python3.6/site-packages/pluggy/manager.py", line 87, in <lambda>
  File "/home/gnikonorov/pytest/.venv/lib/python3.6/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/gnikonorov/pytest/.venv/lib/python3.6/site-packages/pluggy/callers.py", line 80, in get_result
  File "/home/gnikonorov/pytest/.venv/lib/python3.6/site-packages/pluggy/callers.py", line 187, in _multicall
  File "/home/gnikonorov/pytest/src/_pytest/helpconfig.py", line 148, in pytest_cmdline_main
    showhelp(config)
  File "/home/gnikonorov/pytest/src/_pytest/helpconfig.py", line 194, in showhelp
    tw.line(wrapped[0])
IndexError: list index out of range
```

hp310780:
In response to Bug #7394 , this PR updates pytest --help to parse an empty string or None as an empty string for the `help` parameter.

closes #7394


src/_pytest/helpconfig.py
```diff
@@ -191,9 +194,13 @@ def showhelp(config: Config) -> None:
             tw.write(" " * (indent_len - spec_len - 2))
             wrapped = textwrap.wrap(help, columns - indent_len, break_on_hyphens=False)
 
-            tw.line(wrapped[0])
-            for line in wrapped[1:]:
-                tw.line(indent + line)
+            # If a help string wasn't specified, just write an empty line.
```
gnikonorov, line 197: This would definitely solve it, but do we want to allow this? From an end user perspective I feel like an empty line would confuse me, and I would think that the feature is unimplemented, or that pytest is broken in general and not outputting information.

testing/test_helpconfig.py
```diff
@@ -38,6 +38,34 @@ def test_help(testdir):
     )
 
 
+def test_empty_help_param(testdir):
```
gnikonorov, line 41: You could join the two test cases you've written into one by [parametrizing them](https://docs.pytest.org/en/stable/parametrize.html)

src/_pytest/helpconfig.py
hp310780: Do you think it'd be better to throw an error on an empty help or perhaps display a message in its place indicating it was empty? Open to all suggestions as I agree it may come across as a bug/something has broken.

gnikonorov: I’m personally leaning towards throwing an error clearly indicating to the user that the help text cannot be empty. I feel like we should never assume for the end user. Also, I’m unsure as to what an appropriate default help message would even contain

hp310780: I was thinking something like `[No description provided]` so the end user knows the state of the option. But, happy to change to throwing an error (a `TypeError` like a missing function param?) on help being `None` or `""`, let me know what you would like.

gnikonorov:
> I was thinking something like `[No description provided]` so the end user knows the state of the option. But, happy to change to throwing an error (a `TypeError` like a missing function param?) on help being `None` or `""`, let me know what you would like.

Let’s get some more input. @RonnyPfannschmidt @nicoddemus do we think that in the event of no help provided we should throw an error, or supply a default help

gnikonorov:
@hp310780 - I personally vote in favor of an exception. Are you ok with going with that plan while we wait for a review

hp310780:
> @hp310780 - I personally vote in favor of an exception. Are you ok with going with that plan while we wait for a review

Yep! Changed. Let me know of any feedback. Thanks.

src/_pytest/helpconfig.py
```diff
@@ -170,6 +170,8 @@ def showhelp(config: Config) -> None:
         help, type, default = config._parser._inidict[name]
         if type is None:
             type = "string"
+        if help in (None, ""):
+            raise TypeError("help argument must be provided for {}".format(name))
```
gnikonorov, line 174: just to avoid confusion, can you have the exception explicitly state why it was raised? E.g.: `"help argument cannot be empty for {}".format(name)`.  Otherwise the user may feel like the exception is an error ( if they provided an empty string explicitly, for example )

gnikonorov:
bump @RonnyPfannschmidt @nicoddemus @bluetech - could someone else take a look at this PR?

src/_pytest/helpconfig.py
```diff
@@ -170,6 +170,8 @@ def showhelp(config: Config) -> None:
         help, type, default = config._parser._inidict[name]
         if type is None:
             type = "string"
+        if help in (None, ""):
```
bluetech, line 173: I think `""` should be accepted, i.e. just make it not crash in this case. It's not very nice but I don't see a particular reason to prevent it if someone really wants to pass that. Further, the same issue happens with `"  "` (two spaces) etc..

gnikonorov: What would the benefit of a help command with no text be though @bluetech? I don't see a definite reason to ban it, but at the same time I don't see a reason to accept it.

Regarding arbitrary empty spaces, we could just strip the space off of help, if it's possible to retrieve such a value in the first place. ( @hp310780 can you confirm please )

bluetech: > but at the same time I don't see a reason to accept it.

The reason is just to avoid extra code. There are many unreasonable things a user can pass, like a 1000 word explanation, or writing it all in upper case, but we can't check them all. So I think it's better to just accept what the user specifies, as long as it has the right type.

gnikonorov: I agree @bluetech

hp310780: Okay, so this would accept any string (including the empty string) and only raises an error on `None`?

gnikonorov: I think that’s what @bluetech had in mind. The idea is we should never guess on behalf of the user, since we can never really exhaustively guess everything the user may do

testing/test_helpconfig.py
```diff
@@ -38,6 +38,35 @@ def test_help(testdir):
...
+    """
+    testdir.makeconftest(
+        """
+        def pytest_addoption(parser):
+            parser.addini("test_ini", "", default=True, type="bool")
+    """
+    )
+    result = testdir.runpytest("--help")
+    assert result.ret == 0
+    result.stdout.fnmatch_lines(["*test_ini*(bool):*"])
```
bluetech, line 67: This doesn't really test that the help is *displayed* correctly. I suggest:

- Include a few more surrounding lines
- Pass `conescutive=True` to `fnmatch_lines`
- Avoid using the leading `*`, in order to test the indentation

nicoddemus:
LGTM, the requests/suggestions by @bluetech are on point! 👍

testing/test_helpconfig.py
hp310780: Thanks for the feedback! I've updated the PR, let me know what you think.

src/_pytest/helpconfig.py
```diff
@@ -191,9 +193,12 @@ def showhelp(config: Config) -> None:
             tw.write(" " * (indent_len - spec_len - 2))
             wrapped = textwrap.wrap(help, columns - indent_len, break_on_hyphens=False)
 
-            tw.line(wrapped[0])
-            for line in wrapped[1:]:
-                tw.line(indent + line)
+            if not wrapped:
+                tw.write(" " * (indent_len - spec_len - 2))
```
bluetech, line 197: Can you explain why this line is necessary? From what I can see it will be better to omit it (I think it just adds trailing whitespace).

hp310780: To be honest, I added it for the whole help to display correctly, but testing without shows this is, as you say, trailing whitespace. I've removed it and updated the PR!

nicoddemus:
Backport: https://github.com/pytest-dev/pytest/pull/7577

PR #7428, opened 2020-06-29
Don't use reserved keywords as var names in config/__init__.py and helpconfig.py

gnikonorov:
Stop using reserved keywords for local variables noted in ` src/_pytest/config/__init__.py ` and `src/_pytest/helpconfig.py`. Came here after having it block another change I was making.

No changelog entry, as this change is trivial

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1239,7 +1239,7 @@ class Config:
 
     def _getini(self, name: str):
         try:
-            description, type, default = self._parser._inidict[name]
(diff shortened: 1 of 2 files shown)


gnikonorov:
I'm +1 on adding a plugin to the pipeline to ensure this @asottile 
I can open a separate issue to track adding a plugin / doing a cleanup of reserved keyword overriding

gnikonorov:
bump for a merge @nicoddemus @RonnyPfannschmidt

gnikonorov:
Closing as this will be handled by https://github.com/pytest-dev/pytest/issues/7433

PR #7434, opened 2020-07-01
logging: use unique handlers for caplog and reports

bluetech:
Replaces #7159.

Setting log_level via the CLI or .ini will control the log level of the report that is dumped upon failure of a test.

If caplog modified the log level during the execution of that test, it should not impact the level that is displayed upon failure in the "captured log report" section.

Changes I made:
- rebased
- reused handler
- changed store keys also to "caplog_handler_*"
- added changelog

all bugs are mine :)


bluetech:
Thanks @ruaridhw!

PR #7436, opened 2020-07-01
pytester: slightly clean up LsofFdLeakChecker

bluetech:
Just a little leftover from trying to hunt down `ResourceWarnings` a few weeks ago.

No semantic changes intended.

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -94,21 +94,16 @@ def pytest_configure(config: Config) -> None:
 
 
 class LsofFdLeakChecker:
-    def get_open_files(self):
-        out = self._exec_lsof()
-        open_files = self._parse_lsof_output(out)
-        return open_files
+    def get_open_files(self) -> List[Tuple[str, str]]:
+        out = subprocess.run(
+            ("lsof", "-Ffn0", "-p", str(os.getpid())),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
+            check=True,
+            universal_newlines=True,
+        ).stdout
 
-    def _exec_lsof(self):
-        pid = os.getpid()
-        # py3: use subprocess.DEVNULL directly.
-        with open(os.devnull, "wb") as devnull:
-            return subprocess.check_output(
-                ("lsof", "-Ffn0", "-p", str(pid)), stderr=devnull
-            ).decode()
-
-    def _parse_lsof_output(self, out):
-        def isopen(line):
+        def isopen(line: str) -> bool:
             return line.startswith("f") and (
                 "deleted" not in line
                 and "mem" not in line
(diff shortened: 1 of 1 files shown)


src/_pytest/pytester.py
```diff
@@ -130,9 +125,9 @@ def isopen(line):
 
         return open_files
 
-    def matching_platform(self):
+    def matching_platform(self) -> bool:
         try:
-            subprocess.check_output(("lsof", "-v"))
```
gnikonorov, line 135: just out of curiosity, why do you prefer `run` to `check_output` here?

bluetech: There's no good reason, only that the [subprocess docs](https://docs.python.org/3/library/subprocess.html) start with describing only the `run` function and say

> The run() function was added in Python 3.5; if you need to retain compatibility with older versions, see the Older high-level API section.

And that "Older high-level API" includes `check_output`. So I just use `run` for everything.

gnikonorov: thanks for explaining @bluetech

src/_pytest/pytester.py
```diff
@@ -94,21 +94,16 @@ def pytest_configure(config: Config) -> None:
...
-        pid = os.getpid()
-        # py3: use subprocess.DEVNULL directly.
-        with open(os.devnull, "wb") as devnull:
-            return subprocess.check_output(
-                ("lsof", "-Ffn0", "-p", str(pid)), stderr=devnull
-            ).decode()
-
-    def _parse_lsof_output(self, out):
-        def isopen(line):
+    def get_open_files(self) -> List[Tuple[str, str]]:
```
gnikonorov, line 97: why do we not handle any errors? Just curious since the original method doesn't seem to have done it either. But, it looks like it is possible for `subprocess.run` to throw at least `CalledProcessError ` and `TimeoutExpired`.

bluetech: Good point! I think it would be good to handle errors from the `lsof` call here.

If the `lsof` command doesn't exist, the `LsofFdLeakChecker` plugin isn't registered at all, so the issue can't be that `lsof` isn't found at all, but it can still return an error itself.

I don't think `TimeoutExpired` can happen because we don't pass `timeout` to `subprocess.run`, but `CalledProcessError` can happen.

Then there's a question what to do if an error occurs. I guess in this case the most reasonable thing is to just skip the check for the particular test, and issue a warning.

bluetech:
Thanks for reviewing @gnikonorov!

src/_pytest/pytester.py
```diff
@@ -94,21 +94,16 @@ def pytest_configure(config: Config) -> None:
```
roderickmarj, line 94: 2bcad38fbd1fb4022e2dc261817b9918cfbda43e

roderickmarj (approved):

roderickmarj (approved):

src/_pytest/pytester.py
roderickmarj: 2bcad38

roderickmarj (approved):

src/_pytest/pytester.py
roderickmarj: 😁

PR #7437, opened 2020-07-01
testing: fix flaky tests due to "coroutine never awaited" warnings

bluetech:
They sometime leak into other test's warnings and cause them to fail.

### DIFF
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -1160,6 +1160,9 @@ def test_usage_error_code(testdir):
 
 @pytest.mark.filterwarnings("default")
 def test_warn_on_async_function(testdir):
+    # In the below we .close() the coroutine only to avoid
+    # "RuntimeWarning: coroutine 'test_2' was never awaited"
+    # which messes with other tests.
     testdir.makepyfile(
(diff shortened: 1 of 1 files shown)


testing/acceptance_test.py
```diff
@@ -1160,14 +1160,19 @@ def test_usage_error_code(testdir):
...
     testdir.makepyfile(
         test_async="""
         async def test_1():
             pass
         async def test_2():
             pass
         def test_3():
-            return test_2()
+            coro = test_2()
+            coro.close()
```
gnikonorov, line 1174: wouldn't it be better to await the coroutine or call `asyncio.create_task()` as documented [here](https://docs.python.org/3/library/asyncio-dev.html#detect-never-awaited-coroutines)

bluetech: `await` requires that `test_3` itself become `async` which would not be appropriate for this test.

`asyncio.create_task` requires running under an asyncio event loop which isn't the case for pytest.

gnikonorov: thanks for clarifying @bluetech

Issue #7439, opened 2020-07-02
mypy shows 're.Match' undefined


gnikonorov:
As the title states, `mypy` says that `re.Match` is undefined as seen below:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/_pytest/junitxml.py:73: error: Name 're.Match' is not defined  [name-defined]
Found 1 error in 1 file (checked 1 source file)
```

My git commit log is even with master in this case, with an added test commit on my part:
```
commit ca43a0a5a9418c576e9df315cb4b80ba738ac42d (HEAD -> DELETE_ME_TEST_BRANCH)
Author: Gleb Nikonorov <gleb.i.nikonorov@gmail.com>
Date:   Thu Jul 2 00:36:03 2020 -0400

    this is a test

commit 74f20f9fba3d8990cfbb0cfa9b68062ad97d4f72 (upstream/master, origin/master, origin/HEAD, origin/DELETE_ME_TEST_BRANCH, master)
Merge: e6e300e72 304f2916f
Author: Ran Benita <ran@unusedvar.com>
Date:   Wed Jul 1 20:54:55 2020 +0300

    Merge pull request #7434 from bluetech/distinct_log_handlers-rebased

    logging: use unique handlers for caplog and reports
```

My last commit looked like this, and is not related to the `mypy` error as far as I can tell ( I've also been noticing this error for quite some time now in my workflows ):
```
gnikonorov:pytest [DELETE_ME_TEST_BRANCH] $ git diff ca43a0a5a9418c576e9df315cb4b80ba738ac42d
diff --git a/src/_pytest/junitxml.py b/src/_pytest/junitxml.py
index 4df7535de..bf1c643ac 100644
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -72,6 +72,7 @@ _py_ext_re = re.compile(r"\.py$")
 def bin_xml_escape(arg: str) -> py.xml.raw:
     def repl(matchobj: "re.Match[str]") -> str:
         i = ord(matchobj.group())
+        i = 0xFA
         if i <= 0xFF:
             return "#x%02X" % i
         else:
gnikonorov:pytest [DELETE_ME_TEST_BRANCH] $
```

to reproduce, run `tox -e linting`

bluetech:
That's weird! Can you try with this patch?

```diff
diff --git a/src/_pytest/junitxml.py b/src/_pytest/junitxml.py
[...]

Issue #7433, opened 2020-06-29
Investigate how we can warn of/prevent redefinition of built-ins


gnikonorov:
Building off of https://github.com/pytest-dev/pytest/pull/7428 we should investigate if there's an elegant way we can detect and prevent redefinition of built-ins. pylint supports this under [W0622](http://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=W0622#variables-checker-messages), but we do not use `pylint` ( rationale as to why can be found [here](https://github.com/pytest-dev/pytest/issues/7086) )

The high level steps of what we'd want to accomplish would then be:
- determine if we want to prevent redefinition of built ins, or stop short of blocking with some sort of warning, log message, etc. ( or if we even want to block this at all )
- determine which tool we could use to accomplish this
- perform a one time sweep to remove any overridden built ins
- hook it into our CI pipeline

asottile:
[`flake8-builtins`](https://pypi.org/project/flake8-builtins/) is probably the easiest way to do this

gnikonorov:
@asottile agreed
Do we have any concerns that it's owned and managed by only one person?

asottile:
we're already using a lot of one-person software so I highly doubt it

gnikonorov:
I'm +1 for flake8-builtins then

I suggest we run it on ci/cd and also as part of the linting target

asottile:
no need to do any special setup for it, just add it to the list of flake8 plugins in .pre-commit-config.yaml

gnikonorov:
Thanks @asottile - let me get a PR up for this tonight ( assuming no surprises )

bluetech:
@gnikonorov Can you describe what problem did this cause? I followed the link to #7428 but couldn't quite figure it out.

gnikonorov:
@bluetech - in my case I just split a function, and it redefined `type`. When I ran mypy, mypy complained about how I was using `type` in my new function.

There is no real problem going on with mypy or the code. It is dangerous to override builtins though, and for developer santity/peace of mind we shouldn't do it IMO. I remember it really confused me at first that we were overriding type ( to use this example ) when I first started looking at the file. If it were called `ini_type` I would have not been confused, since the overriding of `type` looked intentional when in reality it wasn't

bluetech:
Ah thanks for explaining. So I think it's a tradeoff. Using the name of a builtin can cause confusion as you describe. But some arguments against it:

[...]

PR #7442, opened 2020-07-02
fix mypy issue by using typing Match instead of re.Match

gnikonorov:
Closes https://github.com/pytest-dev/pytest/issues/7439

Use the `Match` in `Typing` instead of `re.Match` for junitxml.py to fix mypy error

### DIFF
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -16,6 +16,7 @@ import sys
 from datetime import datetime
 from typing import Dict
 from typing import List
+from typing import Match
 from typing import Optional
 from typing import Tuple
 from typing import Union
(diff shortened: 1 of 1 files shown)


src/_pytest/junitxml.py
```diff
@@ -70,7 +71,7 @@ class Junit(py.xml.Namespace):
 
 
 def bin_xml_escape(arg: str) -> py.xml.raw:
-    def repl(matchobj: "re.Match[str]") -> str:
+    def repl(matchobj: "Match[str]") -> str:
```
bluetech, line 74:
```suggestion
    def repl(matchobj: Match[str]) -> str:
```

Unless CI says it's necessary on Python 3.5 or such.

gnikonorov: let me try, I forgot you don't need to quote things imported from Typing. Thanks @bluetech

gnikonorov:
@bluetech @RonnyPfannschmidt can I ask for one of you to merge this upon review? It's blocking my ability to commit work for https://github.com/pytest-dev/pytest/issues/7433

bluetech:
Thanks @gnikonorov !

PR #7448, opened 2020-07-04
Publish our types

### LINKED ISSUE #3342: Support for static typing

bluetech:
Fixes #3342. All of the relevant TODOs outlined in https://github.com/pytest-dev/pytest/issues/3342#issuecomment-648301701 are covered now so I think we're good to go.

### DIFF
--- a/changelog/3342.feature.rst
+++ b/changelog/3342.feature.rst
@@ -1,0 +1,10 @@
+pytest now includes inline type annotations and exposes them to user programs.
+Most of the user-facing API is covered, as well as internal code.
+
+If you are running a type checker such as mypy on your tests, you may start
+noticing type errors indicating incorrect usage. If you run into an error that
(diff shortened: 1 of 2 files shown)


The-Compiler:
I was about to say "awesome work, I gave it a try with my codebase (qutebrowser) and didn't run into any issues". Then I discovered that mypy was only checking my `conftest.py` because of missing `__init__.py` files, and all the rest of my `test/` code wasn't ever  type-checked (despite `--check-untyped-defs`. Looks like I have some work to do now! :sweat_smile: 

Either way: Awesome work! :+1:

roderickmarj (approved):

roderickmarj (approved):

roderickmarj (approved):

PR #7449, opened 2020-07-04
Add flake8-builtins to linting target

gnikonorov:
Add [flake8-builtins](https://pypi.org/project/flake8-builtins/) to the linting target to prevent use of python builtins as user defined variables/methods/parameters and apply this plugin to the source, where the change was safe and non-breaking.

I went over the [API reference](https://docs.pytest.org/en/stable/reference.html) and made sure I didn't change anything in it. I was also intentionally cautious and opted to noqa things that weren't obviously safe changes. If we want to change these things, I think it should be done as a follow up ( and maybe even in it's own project )

Closes https://github.com/pytest-dev/pytest/issues/7433


src/_pytest/_code/code.py
```diff
@@ -217,8 +217,8 @@ def path(self) -> Union[py.path.local, str]:
         """ path to the source code """
         return self.frame.code.path
 
-    @property
-    def locals(self) -> Dict[str, Any]:
+    @property  # noqa : A003
```
nicoddemus, line 220: This doesn't seem correct right? We are not overriding `property`.

src/_pytest/_code/code.py
```diff
@@ -489,8 +489,8 @@ def fill_unfilled(self, exc_info: Tuple["Type[_E]", _E, TracebackType]) -> None:
         assert self._excinfo is None, "ExceptionInfo was already filled"
         self._excinfo = exc_info
 
-    @property
-    def type(self) -> "Type[_E]":
+    @property  # noqa: A003
```
nicoddemus, line 492: Ditto here for `property`

src/_pytest/_code/code.py
```diff
@@ -802,7 +802,7 @@ def _makepath(self, path):
     def repr_traceback(self, excinfo: ExceptionInfo) -> "ReprTraceback":
         traceback = excinfo.traceback
         if self.tbfilter:
-            traceback = traceback.filter()
+            traceback = traceback.apply_filter()
```
nicoddemus, line 805: We probably shouldn't change this function name; I'm not sure it is public, but doesn't seem worth the risk.

src/_pytest/mark/expression.py
```diff
@@ -51,7 +51,7 @@ class TokenType(enum.Enum):
 
 @attr.s(frozen=True, slots=True)
 class Token:
-    type = attr.ib(type=TokenType)
+    tokenType = attr.ib(type=TokenType)
```
nicoddemus, line 54: This should be `token_type` I think to keep to our naming convention.

src/_pytest/mark/expression.py
```diff
@@ -198,13 +199,13 @@ class Expression:
     def __init__(self, code: types.CodeType) -> None:
         self.code = code
 
-    @classmethod
-    def compile(self, input: str) -> "Expression":
+    @classmethod  # noqa: A003
```
nicoddemus, line 202: Same comment here regarding `property`.

src/_pytest/monkeypatch.py
```diff
@@ -147,19 +147,19 @@ def test_partial(monkeypatch):
         finally:
             m.undo()
 
-    @overload
-    def setattr(
+    @overload  # noqa: A003
```
nicoddemus, line 150: Ditto here about the decorator... not sure why it complains about decorator usage.

src/_pytest/pytester.py
```diff
@@ -1587,6 +1588,6 @@ def _fail(self, msg: str) -> None:
         self._log_output = []
         pytest.fail(log_text)
 
-    def str(self) -> str:
+    def string(self) -> str:
```
nicoddemus, line 1591: This is a breaking change, we should keep the original name

gnikonorov:
@nicoddemus - sadly we have to override decorators and also the function definitions with this plugin, which honestly I'm not too opposed to since it looks like we did it with other plugins ( e.g.: search for `F811` ):

Here's what happens when I remove some of the decorator comments:
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

src/_pytest/monkeypatch.py:150:5: A003 class attribute "setattr" is shadowing a python builtin
src/_pytest/monkeypatch.py:156:5: A003 class attribute "setattr" is shadowing a python builtin
src/_pytest/mark/expression.py:202:5: A003 class attribute "compile" is shadowing a python builtin
src/_pytest/_code/code.py:220:5: A003 class attribute "locals" is shadowing a python builtin
src/_pytest/_code/code.py:492:5: A003 class attribute "type" is shadowing a python builtin
```

src/_pytest/_code/code.py
gnikonorov: resolving as per my comment below your review. Please reopen if you disagree

src/_pytest/_code/code.py
gnikonorov: resolving as per my comment below your review. Please reopen if you disagree

src/_pytest/mark/expression.py
gnikonorov: resolving as per my comment below your review. Please reopen if you disagree

src/_pytest/monkeypatch.py
gnikonorov: resolving as per my comment below your review. Please reopen if you disagree

nicoddemus:
Hmmm I would be a bit annoyed to have to remember to do this all the time when it is perfectly legal/normal todo.

Given that having method names is not error prone (as opposed to locals and parameters), only a matter of (arguably) bad style/frowned upon (but not by me), how about we disable `A003` entirely?

gnikonorov:
I'd be -1 on disabling it, just because it could be a little confusing and it is uncommon.

I'm open to whatever group consensus is. In the argument of not disabling it, `tox -e linting` would tell you exactly what to do to fix the issue ( but I can see the argument for disabling it too )

src/_pytest/_code/code.py
gnikonorov: sure, I was going to make a follow up issue to track all the overrides we have, so we could always just do it later

src/_pytest/pytester.py
gnikonorov: thanks, I didn't realize. Just changed it back

src/_pytest/mark/expression.py
gnikonorov: thank you for catching, changing as suggested

nicoddemus:
> I'd be -1 on disabling it, just because it could be a little confusing 

Hmm can you elaborate why do you think it would be confusing? 

> Also uncommon?

Can you also elaborate on that? I mean it is very common to add exceptions to linters, even more so for flake8 rules, at least in my experience.

gnikonorov:
> > I'd be -1 on disabling it, just because it could be a little confusing
> 
> Hmm can you elaborate why do you think it would be confusing?

I didn't mean disabling rules would be confusing. I mean encountering an overridden built in could be confusing to a reader

> > Also uncommon?
> 
> Can you also elaborate on that? I mean it is very common to add exceptions to linters, even more so for flake8 rules, at least in my experience.

Same as above, I meant that encountering the overridden built in is uncommon, unless it's intentional ( at least, in my experience?  I don't make this claim with any authority ).

nicoddemus:
> I didn't mean disabling rules would be confusing. I mean encountering an overridden built in could be confusing to a reader 

Oh I see. Well many builtins are reused as method names, like `list`, `format`, etc. This happens through the Python ecosystem, and IMHO opinion is not a big deal because a method cannot cause the same kind of bugs as overriding a builtin with a local variable could. 

I guess it might be a bit confusing if an editor uses simplified regex-based syntax highlighting though.

nicoddemus:
Closing as per discussion in https://github.com/pytest-dev/pytest/issues/7433.

Thanks again @gnikonorov!

Issue #7443, opened 2020-07-03
No colors by default on GitHub Actions


The-Compiler:
When running pytest on GitHub Actions (like [we do ourselves](https://github.com/pytest-dev/pytest/runs/833494741?check_suite_focus=true#step:7:16), random example), the output is not colored.

pytest *does* color output on Travis CI by default, so it could be argued that this is somehow GitHub Action's fault. However, other tools (e.g. `eslint`, and I think `mypy`) seem to colorize their output on GitHub actions just fine, without needing to force them! From what I've seen, those tools seem to be influenced by the `TERM` environment variable, while pytest's (or pylib's?) autodetection seems to work differently?

As a workaround, this helps (as part of the step definition):

```yaml
      - name: "Run pytest"
        env:
          PYTEST_ADDOPTS: "--color=yes"
        run: pytest
```

(If you run pytest via tox, make sure to add `PYTEST_ADDOPTS` to `passenv`!)

hugovk:
Yep, this happens with pre-commit too, and many, many other tools. The problem is GitHub Actions does not provide a tty: https://github.com/actions/runner/issues/241. This issue mentions pytest, and GitHub say it's low in their backlog and they have no plans to fix it (https://github.com/actions/runner/issues/241#issuecomment-580480162). 

If you run pytest via tox, a similar workaround is to put it in config (eg. `setup.cfg`) so you don't need `passenv`:

* https://github.com/python-pillow/Pillow/pull/4716/files

hugovk:
Please see PR https://github.com/pytest-dev/pytest/pull/7462.

eine:
IMHO, envvars to force colouring are VERY inconsistent. This is true even in the Python ecosystem. For example, see this issue, tartley/colorama#230 (/cc @jhol @tartley), tox-dev/tox#1468 (/cc @gaborbernat)... I believe it would be desirable if the community could talk and agree on a couple of envvars.

I like @jhol's `FORCE_COLOR` and `NO_COLOR`.

The-Compiler:
@eine True, but this seems unrelated to this issue?

eine:
[...]

PR #7462, opened 2020-07-08
Enable colours on GitHub Actions by default

hugovk:
Workaround to fix https://github.com/pytest-dev/pytest/issues/7443.

GitHub Actions doesn't provide a tty, so many tools that normally output in colour instead output in monochrome: https://github.com/actions/runner/issues/241.
 
GitHub say it's low in their backlog and they have no plans to fix it (https://github.com/actions/runner/issues/241#issuecomment-580480162), so instead of requiring every pytest user to explicitly force colours for pytest, improve the pytest colour detection using the `GITHUB_ACTIONS` env var:

> **`GITHUB_ACTIONS`** Always set to **`true`** when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.

https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

---

Test using this branch:

* https://github.com/hugovk/test/runs/848991439?check_suite_focus=true#step:5:11

image

### DIFF
--- a/src/_pytest/_io/terminalwriter.py
+++ b/src/_pytest/_io/terminalwriter.py
@@ -27,9 +27,11 @@ def should_do_markup(file: TextIO) -> bool:
         return True
     if os.environ.get("PY_COLORS") == "0":
         return False
+    if os.environ.get("GITHUB_ACTIONS") == "true":
+        return True
     return (
         hasattr(file, "isatty") and file.isatty() and os.environ.get("TERM") != "dumb"
-    ) or os.environ.get("GITHUB_ACTIONS") == "true"
+    )
 
 
 class TerminalWriter:
--- a/src/_pytest/_io/terminalwriter.py
+++ b/src/_pytest/_io/terminalwriter.py
@@ -32,7 +32,7 @@ def should_do_markup(file: TextIO) -> bool:
         and file.isatty()
         and os.environ.get("TERM") != "dumb"
         and not (sys.platform.startswith("java") and os._name == "nt")
-    )
+    ) or os.environ.get("GITHUB_ACTIONS") == "true"
 
 
 class TerminalWriter:
--- a/src/_pytest/_io/terminalwriter.py
+++ b/src/_pytest/_io/terminalwriter.py
@@ -28,10 +28,7 @@ def should_do_markup(file: TextIO) -> bool:
     if os.environ.get("PY_COLORS") == "0":
         return False
     return (
-        hasattr(file, "isatty")
-        and file.isatty()
-        and os.environ.get("TERM") != "dumb"
-        and not (sys.platform.startswith("java") and os._name == "nt")
+        hasattr(file, "isatty") and file.isatty() and os.environ.get("TERM") != "dumb"
     ) or os.environ.get("GITHUB_ACTIONS") == "true"
 
 
--- a/changelog/7443.improvement.rst
+++ b/changelog/7443.improvement.rst
@@ -1,0 +1,1 @@
+Detect if running on GitHub Actions and enable colors by default.
--- a/testing/io/test_terminalwriter.py
+++ b/testing/io/test_terminalwriter.py
@@ -177,6 +177,18 @@ def test_should_do_markup_PY_COLORS_eq_0(monkeypatch: MonkeyPatch) -> None:
     assert s == "hello\n"
 
 
+def test_should_do_markup_GITHUB_ACTIONS_eq_true(monkeypatch: MonkeyPatch) -> None:
+    monkeypatch.setitem(os.environ, "GITHUB_ACTIONS", "true")
+    file = io.StringIO()
+    tw = terminalwriter.TerminalWriter(file)
+    assert tw.hasmarkup
+    tw.line("hello", bold=True)
+    s = file.getvalue()
+    assert len(s) > len("hello\n")
+    assert "\x1b[1m" in s
+    assert "\x1b[0m" in s
+
+
 class TestTerminalWriterLineWidth:
     def test_init(self) -> None:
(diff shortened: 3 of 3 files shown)


src/_pytest/_io/terminalwriter.py
```diff
@@ -32,7 +32,7 @@ def should_do_markup(file: TextIO) -> bool:
         and file.isatty()
         and os.environ.get("TERM") != "dumb"
         and not (sys.platform.startswith("java") and os._name == "nt")
-    )
+    ) or os.environ.get("GITHUB_ACTIONS") == "true"
```
bluetech, line 35: I think it will be a little clearer as a separate `if` before the `return` (can leave as is if you prefer)

The-Compiler: Sounds like a good idea, since we then handle all the special cases before the "ordinary" one.

src/_pytest/_io/terminalwriter.py
```diff
@@ -32,7 +32,7 @@ def should_do_markup(file: TextIO) -> bool:
         and file.isatty()
         and os.environ.get("TERM") != "dumb"
         and not (sys.platform.startswith("java") and os._name == "nt")
```
The-Compiler, line 34: Seems fine to me to remove this if we don't support Jython anymore - looks like it was originally added in https://github.com/pytest-dev/py/commit/536252cb2e661558bf9d9092a6036a42d31c46ad some 10 years ago :laughing:

hugovk: Removed!

src/_pytest/_io/terminalwriter.py
hugovk: Moved!

nicoddemus:
> the environment variable doesn't tell whether or not you're piped or other possible things where color is inappropriate 

But if you are piped you don't get a `atty`, right? We should then adapt the check to take that in account instead of forcing it everytime.

Your point about being specific is correct, but if we can get this correct in all cases I still see it as a win as every user will get this for free without additional configuration.

asottile:
> > the environment variable doesn't tell whether or not you're piped or other possible things where color is inappropriate
> 
> But if you are piped you don't get a `atty`, right? We should then adapt the check to take that in account instead of forcing it everytime.
> 
> Your point about being specific is correct, but if we can get this correct in all cases I still see it as a win as every user will get this for free without additional configuration.

right, but if you're running in github actions, you can't discern whether that output is going to the log or is being piped into something else since both cases would be (`GITHUB_ACTIONS` present, no tty)

The-Compiler:
This is a valid concern I'm afraid... and I can't think of any way to find out whether we're writing to a pipe/file or to the standard output on GitHub Actions. On the other hand, I wonder how common the case of someone piping pytest output somewhere on GitHub Actions would be.

Alternatively, maybe this should go into some `pytest-github-actions` plugin I guess? I've been toying with the idea of writing such a plugin to do [output grouping](https://github.com/actions/toolkit/blob/master/docs/commands.md#group-and-ungroup-log-lines) similar to [pytest-travis-fold](https://github.com/abusalimov/pytest-travis-fold), but as usual, haven't found the time yet :laughing:

The-Compiler:
I think this is only a problem with GHA, as others somehow provide a fake TTY. FWIW I ended up using this at the top of the workflow file:

```yaml
env:
  PY_COLORS: "1"
```

that way, colors will be enabled for both pytest and tox, and it won't influence other places where commandline arguments are passed or handled.

nicoddemus:
Let's close this for now, thanks @hugovk for giving it a go anyway! 👍

hugovk:
Sure, no problem!

umarcor:
> Alternatively, maybe this should go into some `pytest-github-actions` plugin I guess? I've been toying with the idea of writing such a plugin to do [output grouping](https://github.com/actions/toolkit/blob/master/docs/commands.md?rgh-link-date=2020-07-09T09%3A01%3A14Z#group-and-ungroup-log-lines) similar to [pytest-travis-fold](https://github.com/abusalimov/pytest-travis-fold), but as usual, haven't found the time yet 😆

@The-Compiler, we are already doing that in https://github.com/ghdl/ghdl-cosim/blob/master/test.py#L25-L30. See output: https://github.com/ghdl/ghdl-cosim/runs/854521453?check_suite_focus=true#step:4:13. Unfortunately, there is an annoyance which I could not solve: when pytest is used to execute makefiles, everything is correct; but when executing other python scripts, the output/log is empty. Compare https://github.com/ghdl/ghdl-cosim/runs/854521453?check_suite_focus=true#step:4:174 and any other of the "Log" collapsible groups. That's the reason that has prevented me from "announcing" the solution; which is really simple and effective otherwise.

The-Compiler:
@umarcor Interesting! I have no idea why that is. FWIW what I'd be interested in is a solution collapsing the output of pytest's own test failure output.

umarcor:
> FWIW what I'd be interested in is a solution collapsing the output of pytest's own test failure output.

All you need to do is print `##[group]` before and `##[endgroup]`. However, be aware that nested groups are NOT supported.

The-Compiler:
> All you need to do is print `##[group]` before and `##[endgroup]`. However, be aware that nested groups are NOT supported.

Note you're using an old/alternative/depecated? syntax there, that [should probably be](https://github.com/actions/toolkit/blob/master/docs/commands.md#group-and-ungroup-log-lines) `::group::title` and `::endgroup::` instead.

umarcor:
Yes, you are correct. It's an old, probably deprecated, syntax. Sorry about that. I'll fix it on my side.

roderickmarj (approved):

PR #7474, opened 2020-07-10
New doc role: globalvar for special variables

nicoddemus:
This introduces a new role, `:globalvar:`, so we can mark/reference variables like `pytest_plugins`, `pytestmark`, etc. This besides being useful also makes the documentation look more consistent.

Also uses proper references in the CHANGELOG for #7466 as commented in https://github.com/pytest-dev/pytest/pull/7466#discussion_r452798085.

### DIFF
--- a/doc/en/example/markers.rst
+++ b/doc/en/example/markers.rst
@@ -280,27 +280,7 @@ its test methods:
 This is equivalent to directly applying the decorator to the
 two test functions.
 
-Due to legacy reasons, it is possible to set the ``pytestmark`` attribute on a TestClass like this:
-
-.. code-block:: python
-
-    import pytest
-
-
-    class TestClass:
-        pytestmark = pytest.mark.webtest
-
-or if you need to use multiple markers you can use a list:
-
-.. code-block:: python
-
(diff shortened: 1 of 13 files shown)


doc/en/example/markers.rst
```diff
@@ -280,37 +280,26 @@ its test methods:
 This is equivalent to directly applying the decorator to the
 two test functions.
 
-Due to legacy reasons, it is possible to set the ``pytestmark`` attribute on a TestClass like this:
```
nicoddemus, line 283: Moved this legacy note to the bottom, the text reads better this way.

doc/en/reference.rst
```diff
@@ -971,12 +968,6 @@ test functions and methods. Can be either a single mark or a list of marks.
 
     pytestmark = [pytest.mark.integration, pytest.mark.slow]
 
-PYTEST_DONT_REWRITE (module docstring)
```
nicoddemus, line 974: This is not really a global variable, and is already mentioned in detail in the `assertion` chapter, so decided to remove this from here as it is pretty niche.

PR #7476, opened 2020-07-10
mark/structure: fix pylint complaining that builtin marks are not callable

### LINKED ISSUE #7473: pytest 6: pylint shows not-callable error for builtin pytest marks
With the current versions of pylint (2.5.3) and the underlying astroid (2.4.2), running pylint against a simple pytest 6.0.0rc1 test:

```python3
import pytest

@pytest.mark.parametrize('a', [1, 2])
def test_a(a):
    pass
```

results in:

```
************* Module test_pylint
test_pylint.py:3:1: E1102: pytest.mark.parametrize is not callable (not-callable)
```

which isn't the case with pytest 5.4.3. I was able to bisect this to b3fb5a2d47743a09c551555da22da27ce9e73f41 (#7379, "Type annotate pytest.mark.* builtin marks" by @bluetech).

Note that to my knowledge, pylint doesn't know about type annotations and relies on its inference only. Thus, I suppose it either only sees `parametrize = None`, or it sees `_ParametrizeMarkDecorator.__call__` raising `NotImplementedError`. Not sure what, though. Maybe @PCManticore can explain what exactly is going on there?

Not sure how to best make this work in a way that pylint understands it (or, alternatively, fix it in pylint?) though.

bluetech:
This fixes #7473.

It seems that pylint doesn't understand TYPE_CHECKING and/or the type comments, and thinks `parametrize` and friends are `None`, and then complains that they are not callable.

The solution is a little uglier but seems to do the trick. On Python 3.6 we can use the nice syntax, which I also tested and will work.

### DIFF
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -4,6 +4,7 @@ import typing
 import warnings
 from typing import Any
 from typing import Callable
+from typing import cast
 from typing import Iterable
 from typing import List
 from typing import Mapping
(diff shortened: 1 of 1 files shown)


src/_pytest/mark/structures.py
```diff
@@ -467,12 +468,12 @@ def test_function():
 
     # See TYPE_CHECKING above.
     if TYPE_CHECKING:
-        skip = None  # type: _SkipMarkDecorator
-        skipif = None  # type: _SkipifMarkDecorator
-        xfail = None  # type: _XfailMarkDecorator
-        parametrize = None  # type: _ParametrizeMarkDecorator
-        usefixtures = None  # type: _UsefixturesMarkDecorator
-        filterwarnings = None  # type: _FilterwarningsMarkDecorator
+        skip = cast(_SkipMarkDecorator, None)
```
nicoddemus, line 471: Perhaps leave a TODO note to use proper type annotations once we drop Python 3.5?

bluetech: I actually had it in my unsaved editor buffer :facepalm: Pushed now.

PR #7496, opened 2020-07-14
 Fix typing of params ids callable form 

### LINKED ISSUE #7494: 6.0.0rc1 pytest.fixture mypy error: Value of type "object" is not indexable
- [ ] a detailed description of the bug or suggestion

```
# file: pytest-mypy.py
import pytest


@pytest.fixture(params=[(0, 0), (1, 1)], ids=lambda x: x[0])
def b(request):
    return request.param


def test_b(b):
    pass
```

```
$ pytest ~/pytest-mypy.py -vv
=========================================================================== test session starts ============================================================================
platform win32 -- Python 3.8.3, pytest-6.0.0rc1, py-1.9.0, pluggy-0.13.1 -- c:\users\simon\anaconda3\envs\pandas-dev\python.exe
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('C:\\Users\\simon\\pandas\\.hypothesis\\examples')
rootdir: C:\Users\simon
plugins: hypothesis-5.16.3, asyncio-0.12.0, cov-2.10.0, forked-1.1.2, xdist-1.32.0
collected 2 items

..\pytest-mypy.py::test_b[0] PASSED                                                                                                                                   [ 50%]
..\pytest-mypy.py::test_b[1] PASSED                                                                                                                                   [100%]

============================================================================ 2 passed in 0.17s =============================================================================
```

```
$ mypy ~/pytest-mypy.py
C:\Users\simon\pytest-mypy.py:4: error: Value of type "object" is not indexable
Found 1 error in 1 file (checked 1 source file)
```

the type annotation for `ids` in `fixture` is

```
    ids: Optional[
        Union[
            Iterable[Union[None, str, float, int, bool]],
            Callable[[object], Optional[object]],
        ]
    ] = None,
```

i believe that the Callable should accept Any.

bluetech:
Fixes #7494.

There is also a commit which adds a `testing/typing_check.py` file whose purpose is only to not have any mypy errors.

### DIFF
--- a/testing/typing_checks.py
+++ b/testing/typing_checks.py
@@ -1,0 +1,12 @@
+"""File for checking typing issues.
+
+This file is not executed, it is only checked by mypy to ensure that
+none of the code triggers any mypy errors.
+"""
+import pytest
+
+
+# Issue #7488.
+@pytest.mark.xfail(raises=RuntimeError)
+def check_mark_xfail_raises() -> None:
+    pass
--- a/testing/typing_checks.py
+++ b/testing/typing_checks.py
@@ -10,3 +10,15 @@ import pytest
 @pytest.mark.xfail(raises=RuntimeError)
 def check_mark_xfail_raises() -> None:
     pass
+
+
+# Issue #7494.
+@pytest.fixture(params=[(0, 0), (1, 1)], ids=lambda x: str(x[0]))
+def check_fixture_ids_callable() -> None:
+    pass
+
+
+# Issue #7494.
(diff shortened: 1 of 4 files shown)


testing/typing_checks.py
```diff
@@ -0,0 +1,24 @@
+"""File for checking typing issues.
```
nicoddemus, line 1: Good idea! 👍

PR #7510, opened 2020-07-17
Format docstrings in a consistent style 

### LINKED ISSUE #7471: Preferred docstring style
pytest uses various docstring styles, which also leaks to the documentation, and I'm starting to develop an OCD from this!

I propose we use this style:

```py
def func(arg: ArgType) -> Foo:
    """Do important stuff.

    More detailed info here, in separate paragraphs from the subject line.

    Use proper sentences -- start sentences with capital letters and end with periods.

    Can include annotated documentation:

    :param ArgType arg: an argument which determines stuff.
    :returns the result.
    :rtype Foo
    :raises ValueError
    """
```

After we agree on a style I'll make a sweep over the code (possibly gradual) to update to it.

bluetech:
This PR updates the docstrings (and some comments) in `src/` to use a more consistent style as discussed in #7471.

The second commit documents this in CONTRIBUTING.rst.

The diff is very large and probably pretty boring to review in full, but you can sample it to get the idea. I tried not to make any material changes to the text, but sometimes couldn't resist some very minor improvements.

Fixes #7471. (@aklajnert suggestion to use sphinx-autodoc-typehints will come up in a future PR if it goes well)


src/_pytest/config/argparsing.py
```diff
@@ -141,19 +140,17 @@ def parse_known_args(
         args: Sequence[Union[str, py.path.local]],
         namespace: Optional[argparse.Namespace] = None,
     ) -> argparse.Namespace:
-        """parses and returns a namespace object with known arguments at this
-        point.
-        """
+        """Parses and returns a namespace object with known arguments at this
+        point."""
```
blueyed, line 144: Could/should use imperative mood (also below).
pydocstyle would have caught this.

bluetech: Thanks. So I ran pydocstyle now, and fixes issues that came up: 3b6eedadf33b93b4710b52496d1291f522699ebf

It does seem useful, if we only include some error codes (unfortunately the imperative mood one has too many false positives). I will probably send a PR to add it, but since it also requires fixes to docstrings in testing/ I will leave it to a separate PR.

bluetech:
Thanks for reviewing @nicoddemus!

As mentioned in a comment above, a followup adding pydocstyle is coming.

PR #7516, opened 2020-07-19
doc: note about mutation of parametrized values

### LINKED ISSUE #7514: Mutation in parameterized value is seen across repeated calls of the test-case
(couldn't find any relevant issue but #237)
I'm running a parameterized test-case with a mutable parameter `param_dict` that gets modified in the test-case code
(the param has a single value, for simplicity).
The test-case runs twice due to some other parameter, `run_twice`.
The modification leaks to the next test-case run;  
is that on purpose?

```python
import pytest

@pytest.fixture(params=[0, 1])
def run_twice(request):
    pass


@pytest.mark.parametrize("param_dict", [[1]])  # this parameter leaks across runs
def test_mutate_params(param_dict, run_twice):
    assert param_dict == [1]
    # mutation seen from subsequent test-case run.
    del param_dict[0]
```
...which passes only the 1st time:
```
platform linux -- Python 3.8.3, pytest-5.3.5, py-1.9.0, pluggy-0.13.1
rootdir: ../pytest-mutate_params
plugins: cov-2.8.1, xonsh-0.9.18
collected 2 items                                                                                                                                             

test_mutate_param.py .F                                                                                                                                 [100%]

================================== FAILURES ==================================
______________________ test_mutate_params[1-param_dict0] _____________________

param_dict = [], fixt = None

    @pytest.mark.parametrize("param_dict", [[1]])
    def test_mutate_params(param_dict, run_twice):
>       assert param_dict == [1]
E       assert [] == [1]
E         Right contains one more item: 1
E         Use -v to get the full diff

test_mutate_param.py:12: AssertionError
========================= 1 failed, 1 passed in 0.03s =========================

``` 

## Environment
```
$ pip list | grep pytest
pytest                             5.3.5
pytest-cov                         2.8.1
```
Linux, but i guess it doesn't make a difference.

ankostis:
Fix #7514 by augmenting Note with behaviour when parametrized values are mutated
(changes are reflected in subsequent test-case calls).

### DIFF
--- a/doc/en/parametrize.rst
+++ b/doc/en/parametrize.rst
@@ -83,7 +83,7 @@ them in turn:
 
     Parameter values are passed as-is to tests (no copy whatsoever).
 
-    For example, if you pass a list or a dict as a parameter value, and 
+    For example, if you pass a list or a dict as a parameter value, and
     the test case code mutates it, the mutations will be reflected in subsequent
     test case calls.
 
--- a/doc/en/parametrize.rst
+++ b/doc/en/parametrize.rst
@@ -81,7 +81,8 @@ them in turn:
 
 .. note::
 
-    Parameter values are passed as-is to tests (no copy whatsoever), so if you mutate them,
+    Parameter values are passed as-is to tests (no copy whatsoever).
+    For example, if you pass a list or a dict as a parameter value, and mutate it,
     the mutations will be reflected in subsequent test case calls.
 
 .. note::
--- a/doc/en/parametrize.rst
+++ b/doc/en/parametrize.rst
@@ -81,23 +81,25 @@ them in turn:
 
 .. note::
 
-    1. Parameter values are passed as-is to tests (no copy whatsoever), so if you mutate them,
-       the mutations will be reflected in subsequent test case calls.
+    Parameter values are passed as-is to tests (no copy whatsoever), so if you mutate them,
+    the mutations will be reflected in subsequent test case calls.
 
-    2. pytest by default escapes any non-ascii characters used in unicode strings
-       for the parametrization because it has several downsides.
-       If however you would like to use unicode strings in parametrization
-       and see them in the terminal as is (non-escaped), use this option
-       in your ``pytest.ini``:
+.. note::
 
-       .. code-block:: ini
+    pytest by default escapes any non-ascii characters used in unicode strings
+    for the parametrization because it has several downsides.
+    If however you would like to use unicode strings in parametrization
+    and see them in the terminal as is (non-escaped), use this option
+    in your ``pytest.ini``:
 
-           [pytest]
-           disable_test_id_escaping_and_forfeit_all_rights_to_community_support = True
+    .. code-block:: ini
 
-       Keep in mind however that this might cause unwanted side effects and
-       even bugs depending on the OS used and plugins currently installed,
-       so use it at your own risk.
+        [pytest]
+        disable_test_id_escaping_and_forfeit_all_rights_to_community_support = True
+
+    Keep in mind however that this might cause unwanted side effects and
+    even bugs depending on the OS used and plugins currently installed,
+    so use it at your own risk.
 
 
 As designed in this example, only one pair of input/output values fails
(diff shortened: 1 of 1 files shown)


ankostis:
Oups, ..my apologies, i forgot i had dev access in pytest, and github-UI opened my new PR in the original repo.

Sgould i delete it (if am i allowed to) and do the proper thing from my fork?

doc/en/parametrize.rst
```diff
@@ -81,17 +81,27 @@ them in turn:
 
 .. note::
 
-    pytest by default escapes any non-ascii characters used in unicode strings
-    for the parametrization because it has several downsides.
-    If however you would like to use unicode strings in parametrization and see them in the terminal as is (non-escaped), use this option in your ``pytest.ini``:
+    1. Parameter values are passed as-is to tests, and if you mutate them,
```
bluetech, line 84: Instead of using one note with a list inside, I think it will be better to use two notes.

doc/en/parametrize.rst
```diff
@@ -81,17 +81,27 @@ them in turn:
...
+       the mutations will be reflected in subsequent test case calls.
 
-    .. code-block:: ini
+       Specifically, pytest doesn't do any copying of the values, since that would
```
bluetech, line 87: I would omit this paragraph. I don't think the user really *needs* to know the rationale, they just need to be aware of it, and I'm inclined to keep explanations brief when possible.

WDYT?

The-Compiler:
> Oups, ..my apologies, i forgot i had dev access in pytest, and github-UI opened my new PR in the original repo.

I'd say don't worry about it. We can always delete the branch after the PR is merged.

doc/en/parametrize.rst
ankostis: Should they be both in the same place, one after the other, or maybe move them into 2 different places to catch the attention of the reader?
Any suggestions?

doc/en/parametrize.rst
ankostis: Done with new commit.

doc/en/parametrize.rst
bluetech: I think one after the other is fine, unless you think of a better place.

doc/en/parametrize.rst
```diff
@@ -81,17 +81,23 @@ them in turn:
 
 .. note::
 
-    pytest by default escapes any non-ascii characters used in unicode strings
-    for the parametrization because it has several downsides.
-    If however you would like to use unicode strings in parametrization and see them in the terminal as is (non-escaped), use this option in your ``pytest.ini``:
+    1. Parameter values are passed as-is to tests (no copy whatsoever), so if you mutate them,
+       the mutations will be reflected in subsequent test case calls.
```
bluetech, line 85: Another suggestion upon reading again, to clarify a bit:

```suggestion
    1. Parameter values are passed as-is to tests (no copy whatsoever), so if for example you use a list or a dict as a paramater value and mutate it,
       the mutations will be reflected in subsequent test case calls.
```

ankostis: Much better,thanks!

doc/en/parametrize.rst
ankostis: Done.

doc/en/parametrize.rst
ankostis: Just committed the example as a separate paragraph.

ankostis:
I'm seeing some failures in Windows...could it be that my last commit message contains the non-ASCII char `§`?

The-Compiler:
@ankostis That's not related to your change, most likely a change in the GitHub Actions environment. I opened #7516.

doc/en/parametrize.rst
```diff
@@ -79,19 +79,30 @@ them in turn:
     FAILED test_expectation.py::test_eval[6*9-42] - AssertionError: assert 54...
     ======================= 1 failed, 2 passed in 0.12s ========================
 
+.. note::
+
+    Parameter values are passed as-is to tests (no copy whatsoever).
+
+    For example, if you pass a list or a dict as a parameter value, and
```
bluetech, line 86: Linting complains about a trailing space here, applying suggestion to remove it.

```suggestion
    For example, if you pass a list or a dict as a parameter value, and
```

PR #7520, opened 2020-07-20
testing: use a tighter check if `bash` is available

### LINKED ISSUE #7518: Tests on Windows started failing since WSL was made available on GitHub Actions
See e.g. https://github.com/pytest-dev/pytest/pull/7516/checks?check_run_id=889498913#step:7:172 from #7516. 

That smells like a change in the GitHub Actions environments, as the error message reads (with a couple of NUL-bytes in between):

```
running: bash C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\popen-gw0\test_argcomplete0\test_argcomplete --fu
     in: C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\popen-gw0\test_argcomplete0
Windows Subsystem for Linux has no installed distributions.


Distributions can be installed by visiting the Microsoft Store:


https://aka.ms/wslstore
```

Probably due to this: https://github.com/actions/virtual-environments/pull/1081. 
Not sure what to do here. Should we just skip that test on Windows? Or perhaps if on Windows and on GitHub Actions?

bluetech:
Fixes #7518.

This fixes CI on Windows since GitHub Actions started installing WSL on their images which apparently installs some wrapper `bash` which does not run actual bash.

### DIFF
--- a/testing/test_parseopt.py
+++ b/testing/test_parseopt.py
@@ -1,7 +1,7 @@
 import argparse
 import os
 import shlex
-import shutil
+import subprocess
 import sys
 
 import py
@@ -288,8 +288,19 @@ class TestParser:
 
 
 def test_argcomplete(testdir, monkeypatch) -> None:
-    if not shutil.which("bash"):
-        pytest.skip("bash not available")
+    try:
+        bash_version = subprocess.run(
+            ["bash", "--version"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+        ).stdout
+    except OSError:
+        pytest.skip("bash is not available")
+    if "GNU bash" not in bash_version:
+        # See #7518.
+        pytest.skip("not a real bash")
+
     script = str(testdir.tmpdir.join("test_argcomplete"))
 
     with open(str(script), "w") as fp:


testing/test_parseopt.py
```diff
@@ -288,8 +288,16 @@ def test_multiple_metavar_help(self, parser: parseopt.Parser) -> None:
 
 
 def test_argcomplete(testdir, monkeypatch) -> None:
-    if not shutil.which("bash"):
-        pytest.skip("bash not available")
+    try:
+        bash_version = subprocess.run(
+            ["bash", "--version"], capture_output=True, universal_newlines=True
+        ).stdout
+    except Exception:
```
The-Compiler, line 295: `Exception` seems quite broad here, maybe just `FileNotFoundError` or `OSError`?

testing/test_parseopt.py
```diff
@@ -288,8 +288,16 @@ def test_multiple_metavar_help(self, parser: parseopt.Parser) -> None:
 
 
 def test_argcomplete(testdir, monkeypatch) -> None:
-    if not shutil.which("bash"):
-        pytest.skip("bash not available")
+    try:
+        bash_version = subprocess.run(
+            ["bash", "--version"], capture_output=True, universal_newlines=True
```
The-Compiler, line 293: `capture_output` was added in Python 3.7, so I suspect this will fail with Python 3.5 and 3.6 (but hidden by the overly broad catching below).

testing/test_parseopt.py
bluetech: Right, I think `OSError` should be good for this.

testing/test_parseopt.py
bluetech: Missed this, will fix.

The-Compiler:
The diff looks good, but I'm a bit confused about the [CI output now](https://github.com/pytest-dev/pytest/pull/7520/checks?check_run_id=890089864#step:6:46):

```
testing\test_parseopt.py ............................                    [ 15%]
```

that looks like those tests are actually passing now, rather than being skipped? :thinking:

bluetech:
The skip in `testing/test_parseopt` appears at a later line: https://github.com/pytest-dev/pytest/pull/7520/checks?check_run_id=890089914#step:7:115

testing/test_parseopt.py
```diff
@@ -288,8 +288,19 @@ def test_multiple_metavar_help(self, parser: parseopt.Parser) -> None:
 
 
 def test_argcomplete(testdir, monkeypatch) -> None:
-    if not shutil.which("bash"):
-        pytest.skip("bash not available")
+    try:
+        bash_version = subprocess.run(
+            ["bash", "--version"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
```
graingert, line 295: `stderr=subprocess.DEVNULL` ?

graingert:
No worries, I only got the email notification when the merge happened.

PR #7529, opened 2020-07-22
Make --strict imply both strictness options

The-Compiler:
RFC for #7503. It looks like there was a consensus that this is the right thing to do, but not on whether this can still go to pytest 6 in such a late stage or should wait for pytest 7.

IMHO it's fine to have in pytest 6 since we already documented this possibility in the changelog for pytest 4.5.0 (*"The existing --strict option has the same behavior currently, but can be augmented in the future for additional checks."*) and discouraged it in the documentation (*The use of ``--strict-markers`` is highly preferred. ``--strict`` was kept for backward compatibility only and may be confusing for others as it only applies to markers and not to other options.*)

Let me know what you think.

### DIFF
--- a/changelog/7503.breaking.rst
+++ b/changelog/7503.breaking.rst
@@ -1,0 +1,1 @@
+The ``--strict`` flag now implies both ``--strict-markers`` (added as recommended alternative for ``--strict`` in pytest 4.5.0) and ``--strict-config`` (new in this release).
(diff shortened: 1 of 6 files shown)


The-Compiler:
Looks like we're not reaching a consensus here in #7503, and it'd be better to deprecate `--strict` for 6.0 and then take care to get it right in 7.0. Thus I'm closing this.

PR #7550, opened 2020-07-28
Prepare release 6.0.0

pytestbot:
Created automatically from https://github.com/pytest-dev/pytest/issues/7547#issuecomment-664991147.

Once all builds pass and it has been **approved** by one or more maintainers, the build
can be released by pushing a tag `6.0.0` to this repository.


nicoddemus:
@asottile had the strong opinion that they should be kept separate (as is now), right @asottile?

The-Compiler:
Works for me, though then perhaps the 6.0.0 changelog should link to the RC changelog to make sure people don't miss that part?

Also I didn't realize turning `PytestDeprecationWarning` into errors happened after the rc, i.e. is mostly untested/unnoticed by people testing the rc. Well, I suppose we'll see how that goes!

nicoddemus:
> Also I didn't realize turning PytestDeprecationWarning into errors happened after the rc, i.e. is mostly untested/unnoticed by people testing the rc.

That was mostly by oversight from our part; the PR got stalled and I didn't realize it until the rc1 was out. I agree it can lead to surprises for people who tested rc1, but I think we can see how it goes as you suggest (we're already in a much better position than if we released 6.0.0 directly).

doc/en/announce/release-6.0.0.rst
```diff
@@ -0,0 +1,40 @@
+pytest-6.0.0
+=======================================
+
+The pytest team is proud to announce the 6.0.0 release!
+
+pytest is a mature Python testing tool with more than a 2000 tests
```
hugovk, line 6: Nit: no need for an indefinite article here:

```suggestion
pytest is a mature Python testing tool with more than 2000 tests
```

(See https://github.com/pytest-dev/pytest/pull/7552 to tidy the other files.)

nicoddemus: Indeed, thanks!

nicoddemus:
> @asottile had the strong opinion that they should be kept separate (as is now)

Any suggestions how we "link" with the `6.0.0rc1` changelog below? How about:

image

asottile:
> 
> 
> > @asottile had the strong opinion that they should be kept separate (as is now)
> 
> Any suggestions how we "link" with the `6.0.0rc1` changelog below? How about:
> 
> image

sounds good to me!

nicoddemus:
Done. 👍

bluetech:
I haven't checked the release action yet, but according to [RELEASING.md](https://github.com/pytest-dev/pytest/blob/master/RELEASING.rst#releasing) the release should be made to the `6.0.x` branch, but this PR is made to `master`. Do we want to change `RELEASING.md` (the argument for releasing from the branch is that once created it is no longer affected by concurrent merges; argument against is that the tag doesn't show up in `master`).

nicoddemus:
> I haven't checked the release action yet, but according to RELEASING.md the release should be made to the 6.0.x branch, but this PR is made to master. Do we want to change RELEASING.md (the argument for releasing from the branch is that once created it is no longer affected by concurrent merges; argument against is that the tag doesn't show up in master)

I was planning on pushing the `6.0.x` branch to point to the same commit as the tag (once pushed), but good call, better to stick to the existing process. I've created `6.0.x` at the base of this branch and changed the merge target. Thanks!

nicoddemus:
Woot!

PR #7553, opened 2020-07-28
Add support to display field names in namedtuple diffs.

### LINKED ISSUE #7527: Show namedtuple field name in assertion error message
```python
from typing import NamedTuple

class Cart(NamedTuple):
    num_items: int
    total_cost: float

def test_cart():
    assert Cart(1, 10) == Cart(1, 20)
```

pytest failure message:
```
    def test_cart():
>       assert Cart(1, 10) == Cart(1, 20)
E       assert Cart(num_item...total_cost=10) == Cart(num_item...total_cost=20)
E         At index 1 diff: 10 != 20
E         Use -v to get the full diff
```

It would nice if the field name was shown in the assertion failure message, rather than an index, eg:
```
    def test_cart():
>       assert Cart(1, 10) == Cart(1, 20)
E       assert Cart(num_item...total_cost=10) == Cart(num_item...total_cost=20)
E         At index 1 (total_cost) diff: 10 != 20
E         Use -v to get the full diff
```

Python 3.6.10
Pytest 5.4.3

tirkarthi:

If this change fixes an issue, please:

- [ ] Fixes #7527 

### DIFF
--- a/changelog/7527.improvement.rst
+++ b/changelog/7527.improvement.rst
@@ -1,0 +1,1 @@
+When a comparison between `namedtuple` instances of the same type fails, pytest now shows the differing field names (possibly nested) instead of their indexes.
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -110,6 +110,10 @@ def isset(x: Any) -> bool:
     return isinstance(x, (set, frozenset))
 
 
+def isnamedtuple(obj: Any) -> bool:
+    return isinstance(obj, tuple) and getattr(obj, "_fields", None) is not None
+
+
 def isdatacls(obj: Any) -> bool:
     return getattr(obj, "__dataclass_fields__", None) is not None
 
@@ -171,14 +175,20 @@ def _compare_eq_any(left: Any, right: Any, verbose: int = 0) -> List[str]:
     if istext(left) and istext(right):
         explanation = _diff_text(left, right, verbose)
     else:
-        if issequence(left) and issequence(right):
+        if type(left) == type(right) and (
+            isdatacls(left) or isattrs(left) or isnamedtuple(left)
+        ):
+            # Note: unlike dataclasses/attrs, namedtuples compare only the
+            # field values, not the type or field names. But this branch
+            # intentionally only handles the same-type case, which was often
+            # used in older code bases before dataclasses/attrs were available.
+            explanation = _compare_eq_cls(left, right, verbose)
+        elif issequence(left) and issequence(right):
             explanation = _compare_eq_sequence(left, right, verbose)
         elif isset(left) and isset(right):
             explanation = _compare_eq_set(left, right, verbose)
         elif isdict(left) and isdict(right):
             explanation = _compare_eq_dict(left, right, verbose)
-        elif type(left) == type(right) and (isdatacls(left) or isattrs(left)):
-            explanation = _compare_eq_cls(left, right, verbose)
         elif verbose > 0:
             explanation = _compare_eq_verbose(left, right)
         if isiterable(left) and isiterable(right):
@@ -408,6 +418,10 @@ def _compare_eq_cls(left: Any, right: Any, verbose: int) -> List[str]:
     elif isattrs(left):
         all_fields = left.__attrs_attrs__
         fields_to_check = [field.name for field in all_fields if getattr(field, "eq")]
+    elif isnamedtuple(left):
+        fields_to_check = left._fields
+    else:
+        assert False
 
     indent = "  "
     same = []
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -9,7 +9,6 @@ from typing import List
 from typing import Mapping
 from typing import Optional
 from typing import Sequence
-from typing import Tuple
 
 import _pytest._code
 from _pytest import outcomes
@@ -179,8 +178,7 @@ def _compare_eq_any(left: Any, right: Any, verbose: int = 0) -> List[str]:
         elif isdict(left) and isdict(right):
             explanation = _compare_eq_dict(left, right, verbose)
         elif type(left) == type(right) and (isdatacls(left) or isattrs(left)):
-            type_fn = (isdatacls, isattrs)
-            explanation = _compare_eq_cls(left, right, verbose, type_fn)
+            explanation = _compare_eq_cls(left, right, verbose)
         elif verbose > 0:
             explanation = _compare_eq_verbose(left, right)
         if isiterable(left) and isiterable(right):
@@ -403,13 +401,7 @@ def _compare_eq_dict(
     return explanation
 
 
-def _compare_eq_cls(
-    left: Any,
-    right: Any,
-    verbose: int,
-    type_fns: Tuple[Callable[[Any], bool], Callable[[Any], bool]],
-) -> List[str]:
-    isdatacls, isattrs = type_fns
+def _compare_eq_cls(left: Any, right: Any, verbose: int) -> List[str]:
     if isdatacls(left):
         all_fields = left.__dataclass_fields__
         fields_to_check = [field for field, info in all_fields.items() if info.compare]
--- a/testing/test_assertion.py
+++ b/testing/test_assertion.py
@@ -1,3 +1,4 @@
+import collections
 import sys
 import textwrap
 from typing import Any
@@ -985,6 +986,44 @@ class TestAssert_reprcompare_attrsclass:
 
         lines = callequal(left, right)
         assert lines is None
+
+
+class TestAssert_reprcompare_namedtuple:
+    def test_namedtuple(self) -> None:
+        NT = collections.namedtuple("NT", ["a", "b"])
+
+        left = NT(1, "b")
+        right = NT(1, "c")
+
+        lines = callequal(left, right)
+        assert lines == [
+            "NT(a=1, b='b') == NT(a=1, b='c')",
+            "",
+            "Omitting 1 identical items, use -vv to show",
+            "Differing attributes:",
+            "['b']",
+            "",
+            "Drill down into differing attribute b:",
+            "  b: 'b' != 'c'",
+            "  - c",
+            "  + b",
+            "Use -v to get the full diff",
+        ]
+
+    def test_comparing_two_different_namedtuple(self) -> None:
+        NT1 = collections.namedtuple("NT1", ["a", "b"])
+        NT2 = collections.namedtuple("NT2", ["a", "b"])
+
+        left = NT1(1, "b")
+        right = NT2(2, "b")
+
+        lines = callequal(left, right)
+        # Because the types are different, uses the generic sequence matcher.
+        assert lines == [
+            "NT1(a=1, b='b') == NT2(a=2, b='b')",
+            "At index 0 diff: 1 != 2",
+            "Use -v to get the full diff",
+        ]
 
 
 class TestFormatExplanation:
--- a/AUTHORS
+++ b/AUTHORS
@@ -155,6 +155,7 @@ Justyna Janczyszyn
 Kale Kundert
 Kamran Ahmad
 Karl O. Pinc
+Karthikeyan Singaravelan
 Katarzyna Jachim
 Katarzyna Król
 Katerina Koukiou


The-Compiler:
> I think we should fall back to the remaining tuple elements

What do you mean? If `_fields` somehow doesn't match up with the tuple length? I'd rather just have an additional `len(obj._fields) == len(obj)` check or so then (handling `TypeError` if `len` isn't available).

src/_pytest/assertion/util.py
```diff
@@ -423,6 +431,8 @@ def _compare_eq_cls(
         fields_to_check = [
             field.name for field in all_fields if getattr(field, ATTRS_EQ_FIELD)
         ]
+    elif isnamedtuple(left):
+        fields_to_check = left._fields
```
graingert, line 435: moved from: https://github.com/pytest-dev/pytest/issues/7527#issuecomment-665050656

FWIW, I like the "._fields" approach but I think it should fallback to regular tuple comparison if the attributes are not exhausted.

Also I think we should just change the *display* of fields using _fields, and continue looking up items in the tuple by index/iteration

graingert, line 435: @The-Compiler 

> > I think we should fall back to the remaining tuple elements

> What do you mean? If `_fields` somehow doesn't match up with the tuple length? I'd rather just have an additional `len(obj._fields) == len(obj)` check or so then (handling `TypeError` if `len` isn't available).


something like this?

```
fields_and_items = [
    (field if getattr(obj, field) is item else i), item
    for i, (field, item) in enumerate(zip_longest(obj._fields, obj))
]
```

tirkarthi: > FWIW, I like the "._fields" approach but I think it should fallback to regular tuple comparison if the attributes are not exhausted.

Is this about checking the length of `_fields` and the object itself in the other comment?

> Also I think we should just change the display of fields using _fields, and continue looking up items in the tuple by index/iteration

Sorry, I don't get this. Is this about `_fields` being potentially used for some other purpose in custom subclass of tuple and to rely on iterating the namedtuple by index?

testing/test_assertion.py
```diff
@@ -981,6 +982,30 @@ class SimpleDataObjectTwo:
         assert lines is None
 
 
+class TestAssert_reprcompare_namedtuple:
+    def test_namedtuple(self) -> None:
+        SimpleDataObject = namedtuple("SimpleDataObject", ["field_a", "field_b"])
```
asottile, line 987: since `collections` is already imported this could be `collections.namedtuple`

tirkarthi: `collections.abc` is imported. I don't see collections being imported. Are you suggesting to just do `import collections` and use `collections.namedtuple`?

graingert: `import collections.abc` is equivalent to
```
importlib.import_module("collections.abc")
import collections
```

tirkarthi: TIL, I mostly prefer it to be explicit. Even on `import collections` the pre-commit removes it. So I have made the changes as suggested.

asottile: the hook removes it because `import a` and `import a.b.c` are duplicate imports

tekumara:
Thanks so much for working on this @tirkarthi! I'd love to see this feature land.

What's the best way to test this? I pulled your fork and tried it out but I couldn't get the field names to show.

tirkarthi:
@tekumara You're welcome. I can see the changes with the example from the report #7527  . I tried cloning and did `python setup.py install`. Can you check the relevant installed files to see if the patch is there? The patch is also small that it extends on the work already done for dataclass and attrs.

testing/test_assertion.py
```diff
@@ -981,6 +981,36 @@ class SimpleDataObjectTwo:
...
+        left = SimpleDataObject(1, "b")
+        right = SimpleDataObject(1, "c")
+
+        lines = callequal(left, right)
+        assert lines is not None
+        assert lines[2].startswith("Omitting 1 identical item")
+        assert "Matching attributes" not in lines
+        for line in lines[2:]:
+            assert "field_a" not in line
+
```
bluetech, line 999: Please also add a test which tests that a comparison between a `namedtuple` and a plain tuple works as intended.

testing/test_assertion.py
```diff
@@ -981,6 +981,36 @@ class SimpleDataObjectTwo:
         assert lines is None
 
 
+class TestAssert_reprcompare_namedtuple:
+    def test_namedtuple(self) -> None:
```
graingert, line 985: Can you also add some tests with objects that have `_fields`, are tuples, but are not NamedTuples?

Also some tests were `_fields` doesn't match `__getattr__` and ` len(type(o)_fields) != len(o)`

src/_pytest/assertion/util.py
```diff
@@ -412,9 +418,11 @@ def _compare_eq_cls(
     left: Any,
     right: Any,
     verbose: int,
-    type_fns: Tuple[Callable[[Any], bool], Callable[[Any], bool]],
+    type_fns: Tuple[
```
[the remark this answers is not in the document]
The-Compiler: FWIW it was added in a663f60b054fa791c1cb9729b68dbddedfdff26f as part of #3776 - maybe @alysivji remembers what the intention behind it was?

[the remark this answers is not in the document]
alysivji: Believe I didn't inline the arguments as the code would have been formatted multiple lines.

tirkarthi:
@graingert Good points that illustrate it can result in false positives. I have come to below definition of isnameduple as below but this will still cause problems in below test case. I feel adding too much cases in isnamedtuple will cause also slowdown the diff output since isnamedtuple. There could be namedtuple like objects as below that might cause surprising outputs. I guess the PR is moving more towards the line where its not feasible to display the output user wants with no standard way to detect namedtuples.

```python
def isnamedtuple(obj: Any) -> bool:
    fields = getattr(obj, "_fields", None)
    return (
        isinstance(obj, tuple) # Check for tuple
        and fields is not None # Check for _fields
        and issequence(fields) # Check that its iterable
        and len(obj) == len(fields) # Check for len of both
        and all(hasattr(obj, field) for field in fields) # Check all attrs are present
    )

```

 2 elements from constructor. 2 attributes as `_fields` with argument to constructor and `_fields` being different

```
def test_comparing_custom_object_fields_set_different_arity() -> None:
    class SimpleDataObjectOne(tuple):
        _fields = ["field_a", "field_b"]
    
    left = SimpleDataObjectOne((1, 2))
    left.field_a = "a"
    left.field_b = "d"
    right = SimpleDataObjectOne((2, 1))
    left.field_a = "a"
    left.field_b = "b"

    assert left == right
```

Current pytest output 

```
pytest -vvv /tmp/test_namedtuple1.py -k test_comparing_custom_object_fields_set_different_arity
=============================== test session starts ===============================
platform linux -- Python 3.8.0, pytest-5.4.1, py-1.8.1, pluggy-0.13.1 -- /root/py38-venv/bin/python
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/root/pytest/.hypothesis/examples')
rootdir: /tmp
plugins: django-3.9.0, forked-1.1.3, env-0.6.2, hypothesis-5.20.2, xdist-1.31.0, aiohttp-0.3.0, mock-3.1.1, timeout-1.3.4, cov-2.8.1, pythonpath-0.7.3
collected 5 items / 4 deselected / 1 selected                                     

../../tmp/test_namedtuple1.py::test_comparing_custom_object_fields_set_different_arity FAILED [100%]

==================================== FAILURES =====================================
_____________ test_comparing_custom_object_fields_set_different_arity _____________

    def test_comparing_custom_object_fields_set_different_arity() -> None:
        class SimpleDataObjectOne(tuple):
            _fields = ["field_a", "field_b"]
    
        left = SimpleDataObjectOne((1, 2))
        left.field_a = "a"
        left.field_b = "d"
        right = SimpleDataObjectOne((2, 1))
        left.field_a = "a"
        left.field_b = "b"
    
>       assert left == right
E       assert (1, 2) == (2, 1)
E         At index 0 diff: 1 != 2
E         Full diff:
E         - (2, 1)
E         + (1, 2)

/tmp/test_namedtuple1.py:68: AssertionError
============================= short test summary info =============================
FAILED ../../tmp/test_namedtuple1.py::test_comparing_custom_object_fields_set_different_arity
========================= 1 failed, 4 deselected in 0.14s =========================
```

With patch though length and attrs are present the util function might get confused with _fields being same and the arguments to constructor to be different.

```
pytest -vvv /tmp/test_namedtuple1.py -k test_comparing_custom_object_fields_set_different_arity
=============================== test session starts ===============================
platform linux -- Python 3.9.0b5, pytest-6.0.0rc2.dev72+gbecc9f941.d20200819, py-1.9.0, pluggy-0.12.0 -- /root/py39-venv/bin/python3.9
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/root/pytest/.hypothesis/examples')
rootdir: /tmp
plugins: hypothesis-4.43.1, cov-2.8.1
collected 5 items / 4 deselected / 1 selected                                     

../../tmp/test_namedtuple1.py::test_comparing_custom_object_fields_set_different_arity FAILED [100%]

==================================== FAILURES =====================================
_____________ test_comparing_custom_object_fields_set_different_arity _____________

    def test_comparing_custom_object_fields_set_different_arity() -> None:
        class SimpleDataObjectOne(tuple):
            _fields = ["field_a", "field_b"]
    
        left = SimpleDataObjectOne((1, 2))
        left.field_a = "a"
        left.field_b = "d"
        right = SimpleDataObjectOne((2, 1))
        left.field_a = "a"
        left.field_b = "b"
    
>       assert left == right
E       AssertionError: assert (1, 2) == (2, 1)
E         (pytest_assertion plugin: representation of details failed: /root/pytest/src/_pytest/assertion/util.py:448: AttributeError: 'SimpleDataObjectOne' object has no attribute 'field_a'.

/tmp/test_namedtuple1.py:68: AssertionError
============================= short test summary info =============================
FAILED ../../tmp/test_namedtuple1.py::test_comparing_custom_object_fields_set_different_arity
========================= 1 failed, 4 deselected in 0.29s =========================
```

The-Compiler:
IMHO we shouldn't care (or [bikeshed](https://shed.bike/)) too much about cases which are very unlikely to happen. At least the Debian Code Search has [no results at all](https://codesearch.debian.net/search?q=%22_fields+%3D%22&literal=1) for `_fields =`.

We should make sure pytest doesn't explode entirely when comparing them, but the "representation of details failed" message from above seems quite reasonable to me.

tirkarthi:
I have seen projects using `_fields_`. There is one project schematics that seems to use `_fields` from initial grep like https://github.com/schematics/schematics/blob/90dee53fd1d5c29f2c947bec6f5ffe5f74305ab1/tests/test_datastructures.py#L95 . Searching for `_fields` in around 2000 top PyPI projects do return 236 matches. Some of them are tuple of tuples where they are filtered. I haven't explored all instances but full log at https://gist.github.com/tirkarthi/e71470c370a1910da58adfc1da8d5dad . Let me know if you need any other pattern searches.

```
rg -t py ' _fields =' | wc
    236     978   11409
```

bluetech:
IMO the current check `return isinstance(obj, tuple) and getattr(obj, "_fields", None) is not None` is sufficient, but if we want to be more confident we can also check for existence of `_field_defaults`.

tirkarthi:
@bluetech Sorry, I lost track of the discussion and might not be able to continue. Feel free to use the patch as needed. Thanks :)

bluetech (commented):
I pushed an updated version.

Note that this is strictly restricted to the case where the namedtuple *types* match, which avoids most of the pitfalls, while still catching 95% of the cases, in particular the "proto-dataclass" case.

bluetech (approved):

changelog/7527.improvement.rst
```diff
@@ -0,0 +1 @@
+Field name is now shown in the diff message for namedtuple instead of index.
```
nicoddemus, line 1:
```suggestion
When a comparison between `namedtuple` instances fail, pytest now shows the differing field names instead of their indexes.
```

nicoddemus (approved):
Jumping a bit late into the party, but thanks everyone!

Left just a small suggestion to the CHANGELOG message, other than that LGTM. 👍

changelog/7527.improvement.rst
bluetech: Thanks, changed to this with a couple of adjustments.

tirkarthi:
Thanks @bluetech and everybody for the review :)

PR #7572, opened 2020-07-29
Improve output for missing config keys

The-Compiler:
Before:

```
  File "./.tox/py38/bin/pytest", line 33, in <module>
    sys.exit(load_entry_point('pytest', 'console_scripts', 'pytest')())
  File "/home/florian/proj/pytest/src/_pytest/config/__init__.py", line 180, in console_main
    code = main()
  File "/home/florian/proj/pytest/src/_pytest/config/__init__.py", line 136, in main
  File "/home/florian/proj/pytest/src/_pytest/config/__init__.py", line 313, in _prepareconfig
    config = pluginmanager.hook.pytest_cmdline_parse(
  File "/home/florian/proj/qutebrowser/git/.tox/py38/lib/python3.8/site-packages/pluggy/hooks.py", line 286, in __call__
  File "/home/florian/proj/qutebrowser/git/.tox/py38/lib/python3.8/site-packages/pluggy/manager.py", line 93, in _hookexec
  File "/home/florian/proj/qutebrowser/git/.tox/py38/lib/python3.8/site-packages/pluggy/manager.py", line 84, in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
  File "/home/florian/proj/qutebrowser/git/.tox/py38/lib/python3.8/site-packages/pluggy/callers.py", line 203, in _multicall
    gen.send(outcome)
  File "/home/florian/proj/pytest/src/_pytest/helpconfig.py", line 99, in pytest_cmdline_parse
    config = outcome.get_result()  # type: Config
  File "/home/florian/proj/qutebrowser/git/.tox/py38/lib/python3.8/site-packages/pluggy/callers.py", line 80, in get_result
  File "/home/florian/proj/qutebrowser/git/.tox/py38/lib/python3.8/site-packages/pluggy/callers.py", line 187, in _multicall
  File "/home/florian/proj/pytest/src/_pytest/config/__init__.py", line 932, in pytest_cmdline_parse
    self.parse(args)
  File "/home/florian/proj/pytest/src/_pytest/config/__init__.py", line 1204, in parse
  File "/home/florian/proj/pytest/src/_pytest/config/__init__.py", line 1102, in _preparse
    self._validate_plugins()
  File "/home/florian/proj/pytest/src/_pytest/config/__init__.py", line 1177, in _validate_plugins
    fail(
  File "/home/florian/proj/pytest/src/_pytest/outcomes.py", line 156, in fail
    raise Failed(msg=msg, pytrace=pytrace)
Failed: Missing required plugins: pytest-bdd
```

after:

```
ERROR: Missing required plugins: pytest-bdd
```

(in red)

### DIFF
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -252,9 +252,10 @@ class TestParseIni:
             """
             def pytest_addoption(parser):
                 parser.addini("conftest_ini_key", "")
-        """
+            """
         )
-        testdir.tmpdir.join("pytest.ini").write(textwrap.dedent(ini_file_text))
+        testdir.makepyfile("def test(): pass")
+        testdir.makeini(ini_file_text)
 
         config = testdir.parseconfig()
         assert sorted(config._get_unknown_ini_keys()) == sorted(invalid_keys)
@@ -262,11 +263,13 @@ class TestParseIni:
         result = testdir.runpytest()
         result.stdout.fnmatch_lines(warning_output)
 
+        result = testdir.runpytest("--strict-config")
         if exception_text:
-            with pytest.raises(pytest.UsageError, match=exception_text):
-                testdir.runpytest("--strict-config")
+            result.stderr.fnmatch_lines("ERROR: " + exception_text)
+            assert result.ret == pytest.ExitCode.USAGE_ERROR
         else:
-            testdir.runpytest("--strict-config")
+            result.stderr.no_fnmatch_line(exception_text)
+            assert result.ret == pytest.ExitCode.OK
 
     @pytest.mark.filterwarnings("default")
     def test_silence_unknown_key_warning(self, testdir: Testdir) -> None:
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -258,8 +258,10 @@ class TestParseIni:
         result.stdout.fnmatch_lines(warning_output)
 
         if exception_text:
-            result = testdir.runpytest("--strict-config")
-            result.stdout.fnmatch_lines("INTERNALERROR>*" + exception_text)
+            with pytest.raises(pytest.UsageError, match=exception_text):
+                testdir.runpytest("--strict-config")
+        else:
+            testdir.runpytest("--strict-config")
 
     @pytest.mark.filterwarnings("default")
     def test_silence_unknown_key_warning(self, testdir: Testdir) -> None:
--- a/testing/test_config.py
+++ b/testing/test_config.py
@@ -181,12 +181,12 @@ class TestParseIni:
     @pytest.mark.parametrize(
         "ini_file_text, invalid_keys, warning_output, exception_text",
         [
-            (
+            pytest.param(
                 """
-          [pytest]
-          unknown_ini = value1
-          another_unknown_ini = value2
-          """,
+                [pytest]
+                unknown_ini = value1
+                another_unknown_ini = value2
+                """,
                 ["unknown_ini", "another_unknown_ini"],
                 [
                     "=*= warnings summary =*=",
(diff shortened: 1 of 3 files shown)


nicoddemus:
Nice!

nicoddemus:
This is a minor improvement, should we backport it?

The-Compiler:
Just noticed that this applies to `--strict-config` as well, so I pushed an updated commit which does the same there. It does not affect the warning for `pytest_plugins` in other `configtest.py` files, since that one actually happens after collection has started, so `pytrace=False` actually works there.

Tests fail locally for some reason I don't understand (I think I somehow have an older version of the code installed), so I'm pushing this and seeing what CI says :sweat_smile: 

@nicoddemus You mean backporting for 6.0.1? I suppose that'd make sense.

nicoddemus:
> You mean backporting for 6.0.1? I suppose that'd make sense.

In general we only backport bug-fixes, but there were a few cases in the past of backporting minor improvements like this one (for example https://github.com/pytest-dev/pytest/blob/master/doc/en/changelog.rst#pytest-531-2019-11-25).

bluetech:
:+1: on backporting given it was added in 6.0.0, might as well get it working nicely.

The-Compiler:
Okay, looks like the tests fail here as well, I'm stumped. Why would a `PytestConfigWarning` (rather than the error) be emitted just because I'm now raising `UsageError` instead of using `pytest.fail` (which also raises internally)?

gnikonorov:
It's also interesting to note that the tests behave as expected when you cd to the directory tox ran out of and run with these changes ( meaning, no warning is raised ):

tox run output:
```
___________________ TestParseIni.test_invalid_ini_keys[\n          [pytest]\n          unknown_ini = value1\n          minversion = 5.0.0\n          -invalid_keys1-warning_output1-Unknown config ini key: unknown_ini] ___________________

self = <test_config.TestParseIni object at 0x7f5ff3378310>, testdir = <Testdir local('/tmp/pytest-of-gnikonorov/pytest-1234/test_invalid_ini_keys1')>
```

Local pytest run from virtual env:
```
(.venv) gnikonorov:test_invalid_ini_keys1 $ pwd
/tmp/pytest-of-gnikonorov/pytest-1234/test_invalid_ini_keys1
(.venv) gnikonorov:test_invalid_ini_keys1 $ python3 -m pytest --strict-config
ERROR: Unknown config ini key: unknown_ini


(.venv) gnikonorov:test_invalid_ini_keys1 $
```

Maybe it's a `pytest.raises` bug? I have no proof of this, just a hunch

ssbarnea:
@nicoddemus @The-Compiler Can you please resolve conficts on this? This would clearly be a very good addition.

ssbarnea (approved):
I confirm a much better experience with it, but needs rebase.

The-Compiler:
Rebased, but the mysterious test failure remains... No idea what's going on there!

The-Compiler:
Rebased and renamed now that #7723 is in. I'd still appreciate some help with the test failure as I still don't have an idea what's going on there.

nicoddemus:
> Rebased and renamed now that #7723 is in. I'd still appreciate some help with the test failure as I still don't have an idea what's going on there.

Sorry, just got some time today to look into this.

The test currently does this to ensure `--strict-config` raises an error when there's an unknown config option:

```python
        if exception_text:
            with pytest.raises(pytest.UsageError, match=exception_text):
                testdir.runpytest("--strict-config")
        else:
            testdir.runpytest("--strict-config")
```

What happens though is that pytest always captures `pytest.UsageError` and displays a nice message, instead of letting the exception bubble up from `testdir.runpytest`. 

So we need to change the test because the behavior is different:

```python
        result = testdir.runpytest("--strict-config")
        if exception_text:
            result.stderr.fnmatch_lines("ERROR: " + exception_text)
            assert result.ret == pytest.ExitCode.USAGE_ERROR
        else:
            result.stderr.no_fnmatch_line(exception_text)
            assert result.ret == pytest.ExitCode.OK

```

Which is of course much nicer. 👍

testing/test_config.py
```diff
@@ -181,61 +181,66 @@ def test_confcutdir(self, testdir):
...
+                another_unknown_ini = value2
+                """,
                 ["unknown_ini", "another_unknown_ini"],
                 [
                     "=*= warnings summary =*=",
                     "*PytestConfigWarning:*Unknown config option: another_unknown_ini",
                     "*PytestConfigWarning:*Unknown config option: unknown_ini",
                 ],
                 "Unknown config option: another_unknown_ini",
+                id="2-unknowns",
```
nicoddemus, line 197: Also made use of `pytest.param(..., id=)` here so the have a recognizable node id.

nicoddemus:
@The-Compiler I think it is safer to *not* backport this, given that this also changes the pytest exit code, other than just the output.

The-Compiler:
Ah, now that I look at your explanation again that actually makes a lot of sense :laughing: Thanks for the follow-up!

PR #7587, opened 2020-07-30
Stop using more-itertools

bluetech:
We barely use it; the couple places that do are not really worth the extra dependency, I think the code is clearer without it.

Also simplifies one (regular) itertools usage.

Also improves a check and an error message in `pytest.raises`.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -1,6 +1,5 @@
 import functools
 import inspect
-import itertools
 import sys
 import warnings
 from collections import defaultdict
@@ -1489,10 +1488,10 @@ class FixtureManager:
         else:
             argnames = ()
 
-        usefixtures = itertools.chain.from_iterable(
-            mark.args for mark in node.iter_markers(name="usefixtures")
+        usefixtures = tuple(
+            arg for mark in node.iter_markers(name="usefixtures") for arg in mark.args
         )
-        initialnames = tuple(usefixtures) + argnames
+        initialnames = usefixtures + argnames
         fm = node.session._fixturemanager
         initialnames, names_closure, arg2fixturedefs = fm.getfixtureclosure(
             initialnames, node, ignore_args=self._get_direct_parametrize_args(node)
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -1,11 +1,9 @@
-import inspect
 import math
 import pprint
 from collections.abc import Iterable
 from collections.abc import Mapping
 from collections.abc import Sized
 from decimal import Decimal
-from itertools import filterfalse
 from numbers import Number
 from types import TracebackType
 from typing import Any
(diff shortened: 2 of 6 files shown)


src/_pytest/fixtures.py
```diff
@@ -1489,8 +1488,8 @@ def getfixtureinfo(
         else:
             argnames = ()
 
-        usefixtures = itertools.chain.from_iterable(
```
graingert, line 1492:
```suggestion
        return FuncFixtureInfo(
            argnames,
            *node.session._fixturemanager.getfixtureclosure(
                tuple(
                    arg
                    for mark in node.iter_markers(name="usefixtures")
                    for arg in mark.args
                )
                + argnames,
                node,
                ignore_args=self._get_direct_parametrize_args(node),
            )
        )
```

src/_pytest/fixtures.py
```diff
@@ -1489,8 +1488,8 @@ def getfixtureinfo(
...
-            mark.args for mark in node.iter_markers(name="usefixtures")
+        usefixtures = tuple(
+            arg for mark in node.iter_markers(name="usefixtures") for arg in mark.args
         )
         initialnames = tuple(usefixtures) + argnames
         fm = node.session._fixturemanager
```
graingert, line 1495:
```suggestion
        return FuncFixtureInfo(
            argnames,
            *node.session._fixturemanager.getfixtureclosure(
                tuple(
                    arg
                    for mark in node.iter_markers(name="usefixtures")
                    for arg in mark.args
                )
                + argnames,
                node,
                ignore_args=self._get_direct_parametrize_args(node),
            )
        )
```

nicoddemus: :-1: from me, the original code is much more readable in my opinion.

graingert: if that's too much of a mega-expression, I still think it's worth doing:


```python
initialnames = (
    tuple(arg for mark in node.iter_markers(name="usefixtures") for arg in mark.args)
    + argnames
)
```

src/_pytest/python_api.py
```diff
@@ -680,11 +676,16 @@ def raises(  # noqa: F811
         documentation for :ref:`the try statement <python:try>`.
     """
     __tracebackhide__ = True
-    for exc in filterfalse(
-        inspect.isclass, always_iterable(expected_exception, BASE_TYPE)
```
graingert, line 684: unused BASE_TYPE https://github.com/pytest-dev/pytest/pull/7587/files#diff-fc0b2b5b62bce9258d927b9a577d6956R29

bluetech: Good catch, thanks.

src/_pytest/fixtures.py
bluetech: I think it's nice to build it in two steps, so I'll keep it as is.

I did forget to remove the `tuple(..)` from the existing line though, will remove it.

PR #7603, opened 2020-08-01
Enforce some pydocstyle lints with flake8-docstrings & docstring fixes in testing/

bluetech:
As discussed in #7471, we now have a preferred docstring style. While I was initially reluctant to add lints for the style we settled on, based on past experience, I tried pydocstyle again and it was quite useful, and I changed my mind. So this PR adds flake8-docstrings to our pre-commit setup.

The previous PR #7510 fixed docstring issues in `src/`. Since the lint works on the entire code, this PR also has a commit to fix issues in `testing/`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -154,7 +154,17 @@ commands = python scripts/publish-gh-release-notes.py {posargs}
 
 [flake8]
 max-line-length = 120
-extend-ignore = E203
+extend-ignore =
+    ; whitespace before ':'
+    E203
+    ; Missing Docstrings
+    D100,D101,D102,D103,D104,D105,D106,D107
+    ; Whitespace Issues
+    D202,D203,D204,D205,D209,D213
+    ; Quotes Issues
+    D302
+    ; Docstring Content Issues
+    D400,D401,D401,D402,D405,D406,D407,D408,D409,D410,D411,D412,D413,D414,D415,D416,D417
 
 [isort]
 ; This config mimics what reorder-python-imports does.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,9 @@ repos:
     hooks:
     -   id: flake8
         language_version: python3
-        additional_dependencies: [flake8-typing-imports==1.9.0]
+        additional_dependencies:
+          - flake8-typing-imports==1.9.0
+          - flake8-docstrings==1.5.0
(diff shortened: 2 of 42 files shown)


tox.ini
```diff
@@ -154,6 +154,7 @@ commands = python scripts/publish-gh-release-notes.py {posargs}
 
 [flake8]
 max-line-length = 120
+select = E,F,W,C,TYP,D200,D201,D206,D207,D208,D210,D211,D212,D214,D215,D300,D301,D403,D404
```
bluetech, line 157: pydocstyle has an [extensive list](http://www.pydocstyle.org/en/stable/error_codes.html) of error codes. Most of them we don't want, and some have too many false-positives. So I opted to only select the error codes we want instead of excluding those we don't want.

I'm not sure if this is the best way to achieve this -- suggestions welcome.

bluetech:
Linting CI failed with

```
    ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
        mypy==0.780 from https://files.pythonhosted.org/packages/7c/1c/ce92c1506d9fc4ca7964a5167267cfd9f5d4bdf77bc5e62a8d8c1630a1be/mypy-0.780-cp38-cp38-manylinux1_x86_64.whl#sha256=a5e5bb12b7982b179af513dddb06fca12285f0316d74f3964078acbfcf4c68f2 (from pre-commit-dummy-package==0.0.0):
            Expected sha256 a5e5bb12b7982b179af513dddb06fca12285f0316d74f3964078acbfcf4c68f2
                 Got        ae8cd205abcbcd264fe457ea3a4f7c0d851a88bc6dbb0a7bc17138d41bc1f67c
```

Not sure what's going on there.

nicoddemus:
> So this PR adds flake8-docstrings to our pre-commit setup.

Does it fix the code for you automatically (like black), or does it just complain about it?

bluetech:
> Does it fix the code for you automatically (like black), or does it just complain about it?

It just complains.

nicoddemus:
Gotcha!

tox.ini
asottile: in general, extend-ignore is preferred over `select`

bluetech: OK, updated to use extend-ignore.

PR #7604, opened 2020-08-01
typing: set disallow_any_generics

### LINKED ISSUE #7599: Type hints for pytest.raises causes pylance to report partial type error
The code is in `_pytest/python_api.py`
```python
@overload
def raises(
    expected_exception: Union["Type[_E]", Tuple["Type[_E]", ...]],
    *,
    match: "Optional[Union[str, Pattern]]" = ...
) -> "RaisesContext[_E]":
    ...  # pragma: no cover


@overload  # noqa: F811
def raises(  # noqa: F811
    expected_exception: Union["Type[_E]", Tuple["Type[_E]", ...]],
    func: Callable,
    *args: Any,
    **kwargs: Any
) -> _pytest._code.ExceptionInfo[_E]:
    ...  # pragma: no cover
```

Details:
IDE: vscode
extensions: 
* python: v2020.7.96456
* pylance: v2020.7.4 (with type checking set to strict)

Error/Warning Message: Type of "raises" is partially unknown.

Looks like pylance is not happy with the fact that we didn't specify the generic parameters for Pattern and Callable.

### Match
```python
def raises(
    expected_exception: Union["Type[_E]", Tuple["Type[_E]", ...]],
    *,
    match: "Optional[Union[str, Pattern]]" = ... # <-------- this line
```
for this, I suggest that we change Pattern into Pattern[str] since the code actually always converts the value into str

In `ExceptionInfo.match`:
```python
assert re.search(regexp, str(self.value)), msg.format(regexp, str(self.value))
```

### Callable
```python
def raises( 
    expected_exception: Union["Type[_E]", Tuple["Type[_E]", ...]],
    func: Callable, # <--- this line
```

Here is where it is being used:
```python
try:
    func(*args[1:], **kwargs)
```

I think something like `Callabe[...,Any]` should work

bluetech:
Fixes #7599 (presumably; I didn't actually test the described setup)

This prevents referring to a generic type without filling in its generic type parameters.

The `FixtureDef` typing might need some more refining in the future.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -96,6 +96,7 @@ formats = sdist.tgz,bdist_wheel
 [mypy]
 mypy_path = src
 check_untyped_defs = True
+disallow_any_generics = True
 ignore_missing_imports = True
 no_implicit_optional = True
 show_error_codes = True
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -613,7 +613,7 @@ class ExceptionInfo(Generic[_E]):
         )
         return fmt.repr_excinfo(self)
 
-    def match(self, regexp: "Union[str, Pattern]") -> "Literal[True]":
+    def match(self, regexp: "Union[str, Pattern[str]]") -> "Literal[True]":
         """Check whether the regular expression `regexp` matches the string
         representation of the exception using :func:`python:re.search`.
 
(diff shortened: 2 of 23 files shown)


bluetech:
I removed the needs-backport label from this, it's a bit to heavy to backport and not really a regression.

PR #7607, opened 2020-08-01
Demonstrate that plain unittest does not support async tests

nicoddemus:
Related to pytest-dev/pytest-asyncio#180

### DIFF
--- a/testing/test_unittest.py
+++ b/testing/test_unittest.py
@@ -1243,3 +1243,19 @@ def test_asynctest_support(testdir):
     testdir.copy_example("unittest/test_unittest_asynctest.py")
     reprec = testdir.inline_run()
     reprec.assertoutcome(failed=1, passed=2)
+
+
+def test_plain_unittest_does_not_support_async(testdir):
+    """
+    Async functions in plain unittest.TestCase subclasses are not supported without plugins.
+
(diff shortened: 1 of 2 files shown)


testing/test_unittest.py
```diff
@@ -1243,3 +1243,19 @@ def test_asynctest_support(testdir):
     testdir.copy_example("unittest/test_unittest_asynctest.py")
     reprec = testdir.inline_run()
     reprec.assertoutcome(failed=1, passed=2)
+
+
+def test_plain_unittest_does_not_support_async(testdir):
+    """
+    Async functions in plain unittest.TestCase subclasses are not supported without plugins.
```
bluetech, line 1250:
```suggestion
    """Async functions in plain unittest.TestCase subclasses are not supported without plugins.
```

:wink:

nicoddemus: Oops thanks!

bluetech:
@nicoddemus looks like the pypy3 failure here was legitimate

nicoddemus:
Oops yeah my bad, missed that bit.

nicoddemus:
Thanks for the heads up: https://github.com/pytest-dev/pytest/issues/7624

PR #7610, opened 2020-08-02
Better document -k partial matching

### LINKED ISSUE #7546: Better document -k partial matching and what 'keyword' means
_Originally posted by @asottile in https://github.com/pytest-dev/pytest/issues/7539#issuecomment-664050187_

I think at least this should be documented and improved to avoid underscored names (this would solve this particular issue, but I think would be a reasonable baseline expectation)

The relevant docs are here: https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name

ikedaosushi:
Closes #7546. 

To be honest, I'm not sure if this description is the best, I would be happy if you can modify it freely.

### DIFF
--- a/doc/en/example/markers.rst
+++ b/doc/en/example/markers.rst
@@ -202,7 +202,7 @@ You can use ``and``, ``or``, ``not`` and parentheses.
 
 .. note::
     ``-k`` option will match not only function names but also `function attributes <https://www.python.org/dev/peps/pep-0232/>`_ .
-    For example, every function has attributes such as ``__call__`` by default, 
+    For example, every function has attributes such as ``__call__`` by default,
     and some libraries like `hypothesis <https://github.com/HypothesisWorks/hypothesis>`_ attach additional attributes to functions.
     So if you specify such a keyword, let's say ``-k "__call__"``, then ``-k`` will match all functions in the scope, because all functions have ``__call__`` attributes.
     Unfortunetelly you can't controll this behavior in current version, so please keep it in your mind.
--- a/doc/en/example/markers.rst
+++ b/doc/en/example/markers.rst
@@ -200,6 +200,13 @@ Or to select "http" and "quick" tests:
 
 You can use ``and``, ``or``, ``not`` and parentheses.
 
+.. note::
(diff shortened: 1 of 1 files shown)


doc/en/example/markers.rst
```diff
@@ -200,6 +200,13 @@ Or to select "http" and "quick" tests:
 
 You can use ``and``, ``or``, ``not`` and parentheses.
 
+.. note::
```
bluetech, line 203: I think this should be a regular paragraph, not a note.

doc/en/example/markers.rst
```diff
@@ -200,6 +200,13 @@ Or to select "http" and "quick" tests:
...
+    ``-k`` option will match not only function names but also `function attributes <https://www.python.org/dev/peps/pep-0232/>`_ .
+    For example, every function has attributes such as ``__call__`` by default,
+    and some libraries like `hypothesis <https://github.com/HypothesisWorks/hypothesis>`_ attach additional attributes to functions.
+    So if you specify such a keyword, let's say ``-k "__call__"``, then ``-k`` will match all functions in the scope, because all functions have ``__call__`` attributes.
+    Unfortunetelly you can't controll this behavior in current version, so please keep it in your mind.
```
bluetech, line 208: I think we should definitely lay out what is included in `-k`. Below is a suggestion with these differences:

- Added some more things which are matched against.
- Removed the reference to the ancient PEP, I don't think it adds too much.
- Removed the `__call__` example. Normally `__call__` is not found in a function's `__dict__` so is not matched. Maybe Hypothesis does something more customized, but I think it will be confusing for other users.
- Removed the "unfortunately" sentence. At least in pytest the behavior is more of a feature than a bug. I think just explaining the behavior is enough.

```suggestion
    In addition to the test's name, ``-k`` also matches the names of the test's parents (usually, the name of the file and class it's in), attributes set on the test function, markers applied to it or its parents and any :attr:`extra keywords <_pytest.nodes.Node.extra_keyword_matches>` explicitly added to it or its parents.
```

ikedaosushi:
@bluetech Thank you for reviewing, and sorry for my misunderstanding. All you pointed out makes sense, I applied your comments.

PR #7613, opened 2020-08-03
typing: set warn_unreachable

bluetech:
This makes mypy raise an error whenever it detects code which is statically unreachable, e.g.

```py
x: int
if isinstance(x, str):
    ... # Statement is unreachable  [unreachable]
```

This is really neat and finds quite a few logic and typing bugs.

Sometimes the code is intentionally unreachable in terms of types, e.g. raising TypeError when a function is given an argument with a wrong type. In these cases a `type: ignore[unreachable]` is needed, but I think it's a nice code hint.

### DIFF
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -176,8 +176,9 @@ def getfuncargnames(
         p.name
         for p in parameters.values()
         if (
-            p.kind is Parameter.POSITIONAL_OR_KEYWORD
-            or p.kind is Parameter.KEYWORD_ONLY
+            # TODO: Remove type ignore after https://github.com/python/typeshed/pull/4383
+            p.kind is Parameter.POSITIONAL_OR_KEYWORD  # type: ignore[unreachable]
+            or p.kind is Parameter.KEYWORD_ONLY  # type: ignore[unreachable]
         )
         and p.default is Parameter.empty
     )
--- a/testing/test_pytester.py
+++ b/testing/test_pytester.py
@@ -23,7 +23,9 @@ def test_make_hook_recorder(testdir) -> None:
     recorder = testdir.make_hook_recorder(item.config.pluginmanager)
     assert not recorder.getfailures()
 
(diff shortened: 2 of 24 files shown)


src/_pytest/compat.py
```diff
@@ -176,8 +176,9 @@ def getfuncargnames(
         p.name
         for p in parameters.values()
         if (
-            p.kind is Parameter.POSITIONAL_OR_KEYWORD
-            or p.kind is Parameter.KEYWORD_ONLY
+            # TODO: Remove type ignore after https://github.com/python/typeshed/pull/4383.
```
Zac-HD, line 179: The `.` at end of line breaks the link, FYI.

And perhaps we could add `warn_unused_ignores = True` to the config so that this will be caught by CI when we upgrade?

bluetech: > FYI the . at end of line breaks the link for me.

Removed.

>  And perhaps we could add warn_unused_ignores = True to the config so that this will be caught by CI when we upgrade?

We can't use `warn_unused_ignores` because there are some warnings that only happen on some Python versions, or some platforms, etc. which will then trigger the warning on others.

What I've done up to now is that when I upgrade the mypy version we use, I set it temporarily and fix what it brings up.

testing/test_pytester.py
```diff
@@ -23,7 +23,9 @@ def test_make_hook_recorder(testdir) -> None:
     recorder = testdir.make_hook_recorder(item.config.pluginmanager)
     assert not recorder.getfailures()
 
-    pytest.xfail("internal reportrecorder tests need refactoring")
+    # (The silly condition is to fool mypy that the code below this is reachable)
+    if 1 + 1 == 2:
```
nicoddemus, line 27: I'm curious, `# type: ignore[unreachable]` doesn't work here?

bluetech: It works, but then the code is considered unreachable, which I suspect might make the type checking less strict.

PR #7619, opened 2020-08-04
Some py.path -> pathlib conversions

bluetech:
The first commit adds analogues of `py.path.local`'s `bestrelpath` and `common` functions to `_pytest.pathlib`.

The second commit converts `_pytest.config.findpaths` to use pathlib.

The third commit coverts some random places to pathlib.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -41,6 +41,7 @@ from _pytest.compat import ATTRS_EQ_FIELD
 from _pytest.compat import get_real_func
 from _pytest.compat import overload
 from _pytest.compat import TYPE_CHECKING
+from _pytest.pathlib import Path
 
 if TYPE_CHECKING:
     from typing import Type
@@ -1190,12 +1191,12 @@ def getfslineno(obj: object) -> Tuple[Union[str, py.path.local], int]:
 # note: if we need to add more paths than what we have now we should probably use a list
 # for better maintenance.
 
-_PLUGGY_DIR = py.path.local(pluggy.__file__.rstrip("oc"))
+_PLUGGY_DIR = Path(pluggy.__file__.rstrip("oc"))
 # pluggy is either a package or a single module depending on the version
-if _PLUGGY_DIR.basename == "__init__.py":
-    _PLUGGY_DIR = _PLUGGY_DIR.dirpath()
-_PYTEST_DIR = py.path.local(_pytest.__file__).dirpath()
-_PY_DIR = py.path.local(py.__file__).dirpath()
+if _PLUGGY_DIR.name == "__init__.py":
+    _PLUGGY_DIR = _PLUGGY_DIR.parent
+_PYTEST_DIR = Path(_pytest.__file__).parent
+_PY_DIR = Path(py.__file__).parent
 
 
 def filter_traceback(entry: TracebackEntry) -> bool:
@@ -1213,9 +1214,17 @@ def filter_traceback(entry: TracebackEntry) -> bool:
     is_generated = "<" in raw_filename and ">" in raw_filename
     if is_generated:
         return False
+
     # entry.path might point to a non-existing file, in which case it will
     # also return a str object. See #1133.
-    p = py.path.local(entry.path)
-    return (
-        not p.relto(_PLUGGY_DIR) and not p.relto(_PYTEST_DIR) and not p.relto(_PY_DIR)
-    )
+    p = Path(entry.path)
+
+    parents = p.parents
+    if _PLUGGY_DIR in parents:
+        return False
+    if _PYTEST_DIR in parents:
+        return False
+    if _PY_DIR in parents:
+        return False
+
+    return True
--- a/src/_pytest/config/findpaths.py
+++ b/src/_pytest/config/findpaths.py
@@ -1,23 +1,28 @@
+import itertools
 import os
+import sys
 from typing import Dict
 from typing import Iterable
 from typing import List
 from typing import Optional
+from typing import Sequence
 from typing import Tuple
 from typing import Union
 
 import iniconfig
-import py
 
 from .exceptions import UsageError
 from _pytest.compat import TYPE_CHECKING
 from _pytest.outcomes import fail
+from _pytest.pathlib import absolutepath
+from _pytest.pathlib import commonpath
+from _pytest.pathlib import Path
 
 if TYPE_CHECKING:
     from . import Config
 
 
-def _parse_ini_config(path: py.path.local) -> iniconfig.IniConfig:
+def _parse_ini_config(path: Path) -> iniconfig.IniConfig:
     """Parse the given generic '.ini' file using legacy IniConfig parser, returning
     the parsed object.
 
@@ -30,7 +35,7 @@ def _parse_ini_config(path: py.path.local) -> iniconfig.IniConfig:
 
 
 def load_config_dict_from_file(
-    filepath: py.path.local,
+    filepath: Path,
 ) -> Optional[Dict[str, Union[str, List[str]]]]:
     """Load pytest configuration from the given file path, if supported.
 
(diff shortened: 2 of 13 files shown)


src/_pytest/config/findpaths.py
```diff
@@ -93,57 +96,57 @@ def locate_config(
     ]
     args = [x for x in args if not str(x).startswith("-")]
     if not args:
-        args = [py.path.local()]
+        args = [Path.cwd()]
     for arg in args:
-        arg = py.path.local(arg)
-        for base in arg.parts(reverse=True):
+        argpath = Path(arg).resolve()
```
nicoddemus, line 101: I think we should avoid `resolve()` in general because of what we learned in https://github.com/pytest-dev/pytest/pull/6523. I think this will provide the same results:

```suggestion
        argpath = Path(arg).absolute()
```

src/_pytest/config/findpaths.py
```diff
@@ -93,57 +96,57 @@ def locate_config(
...
+    def get_dir_from_path(path: Path) -> Path:
+        if path.is_dir():
             return path
-        return py.path.local(path.dirname)
+        return path.parent
 
     # These look like paths but may not exist
     possible_paths = (
-        py.path.local(get_file_part_from_node_id(arg))
+        Path(get_file_part_from_node_id(arg)).resolve()
```
nicoddemus, line 149: Same here about `resolve()`:

```suggestion
        Path(get_file_part_from_node_id(arg)).absolute()
```

src/_pytest/config/findpaths.py
```diff
@@ -156,41 +159,43 @@ def get_dir_from_path(path: py.path.local) -> py.path.local:
...
     rootdir_cmd_arg: Optional[str] = None,
     config: Optional["Config"] = None,
-) -> Tuple[py.path.local, Optional[py.path.local], Dict[str, Union[str, List[str]]]]:
+) -> Tuple[Path, Optional[Path], Dict[str, Union[str, List[str]]]]:
     rootdir = None
     dirs = get_dirs_from_args(args)
     if inifile:
-        inipath_ = py.path.local(inifile)
-        inipath = inipath_  # type: Optional[py.path.local]
+        inipath_ = Path(inifile).resolve()
```
nicoddemus, line 169:
```suggestion
        inipath_ = Path(inifile).absolute()
```

src/_pytest/config/findpaths.py
```diff
@@ -156,41 +159,43 @@ def get_dir_from_path(path: py.path.local) -> py.path.local:
...
     else:
         ancestor = get_common_ancestor(dirs)
         rootdir, inipath, inicfg = locate_config([ancestor])
         if rootdir is None and rootdir_cmd_arg is None:
-            for possible_rootdir in ancestor.parts(reverse=True):
-                if possible_rootdir.join("setup.py").exists():
+            for possible_rootdir in itertools.chain(
+                (ancestor,), reversed(ancestor.parents)
+            ):
+                if (possible_rootdir / "setup.py").exists():
```
nicoddemus, line 181: I know this is similar to the original code, but I think we should be more specific in general:

```suggestion
                if (possible_rootdir / "setup.py").is_file():
```

src/_pytest/config/findpaths.py
```diff
@@ -156,41 +159,43 @@ def get_dir_from_path(path: py.path.local) -> py.path.local:
...
-                        cwd = py.path.local()
+                        cwd = Path.cwd()
                     rootdir = get_common_ancestor([cwd, ancestor])
                     is_fs_root = os.path.splitdrive(str(rootdir))[1] == "/"
                     if is_fs_root:
                         rootdir = ancestor
     if rootdir_cmd_arg:
-        rootdir = py.path.local(os.path.expandvars(rootdir_cmd_arg))
-        if not rootdir.isdir():
+        rootdir = Path(os.path.expandvars(rootdir_cmd_arg)).resolve()
```
nicoddemus, line 197:
```suggestion
        rootdir = Path(os.path.expandvars(rootdir_cmd_arg)).absolute()
```

src/_pytest/pathlib.py
```diff
@@ -569,3 +569,35 @@ def visit(
...
+def commonpath(path1: Path, path2: Path) -> Optional[Path]:
+    """Return the common part shared with the other path, or None if there is
+    no common part."""
+    try:
+        return Path(os.path.commonpath((path1, path2)))
+    except ValueError:
+        return None
+
+
+def bestrelpath(directory: Path, dest: Path) -> str:
```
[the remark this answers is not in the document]
bluetech: I'd like to keep the same interface as `py.path.local` has for now, to ease the porting.

src/_pytest/config/findpaths.py
bluetech: `Path.absolute` is not public, and also doesn't resolve `.` and `..` which is necessary here. It also has XXX and FIXME in its implementation, seems like better avoided.

`py.path.local(...)` does an `os.path.abspath` internally, which does normalize.

If `resolve()` is no good and we want to keep the behavior I guess we'll need to do `Path(abspath(...))` instead.

src/_pytest/config/findpaths.py
bluetech: Applied, thanks.

src/_pytest/config/findpaths.py
nicoddemus: > If resolve() is no good

Yeah I'm afraid that's the case.

I suggest we add a `def absolute(p: Union[Path, str]) -> Path` to `_pytest.pathlib` which does `Path(abspath(...))` internally and use that then.

Having a separate function will avoid having to remember how to do it each time and also a good place to document why it exists in the first place.

bluetech: Agreed, I added `absolutepath` (see commit message in regards to the name).

nicoddemus: Perfect! 👍

src/_pytest/pathlib.py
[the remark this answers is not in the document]
nicoddemus: Fair enough!

src/_pytest/_code/code.py
```diff
@@ -1213,9 +1214,17 @@ def filter_traceback(entry: TracebackEntry) -> bool:
...
+
     # entry.path might point to a non-existing file, in which case it will
     # also return a str object. See #1133.
-    p = py.path.local(entry.path)
-    return (
-        not p.relto(_PLUGGY_DIR) and not p.relto(_PYTEST_DIR) and not p.relto(_PY_DIR)
-    )
+    p = Path(entry.path)
+
+    parents = p.parents
```
nicoddemus, line 1222: Gotta love `Path.parents`. 😁

src/_pytest/config/findpaths.py
```diff
@@ -93,104 +97,114 @@ def locate_config(
...
-    def get_dir_from_path(path: py.path.local) -> py.path.local:
-        if path.isdir():
+    def get_dir_from_path(path: Path) -> Path:
+        if path.is_dir():
             return path
-        return py.path.local(path.dirname)
+        return path.parent
+
+    def safe_exists(path: Path) -> bool:
+        # On Python<=3.7, this can throw on paths that contain characters
```
nicoddemus, line 149: Perhaps we should use `sys.version_info` here as guard, so `safe_exists` can be dropped once only Python 3.8+ is supported?

```python
if sys.version_info[:2] <= (3, 7):
    def safe_exists(path: Path) -> bool:
        # On Python<=3.7, this can throw on paths that contain characters
        # unrepresentable at the OS level.
        try:
            return path.exists()
        except OSError:
            return False
else:
    def safe_exists(path: Path) -> bool:
        return path.exists()

```

testing/test_config.py
```diff
@@ -21,35 +21,42 @@
...
+            ),
+            encoding="utf-8",
         )
         _, _, cfg = locate_config([sub])
         assert cfg["name"] == "value"
-        config = testdir.parseconfigure(sub)
+        config = testdir.parseconfigure(str(sub))
         assert config.inicfg["name"] == "value"
 
-    def test_getcfg_empty_path(self):
```
nicoddemus, line 49: Not sure why this test was here, was it removed on purpose?

src/_pytest/config/findpaths.py
bluetech: Good idea, will change it.

testing/test_config.py
bluetech: I removed it because `locate_config` now only accepts a `Path` and an empty `Path` is not really a thing.

src/_pytest/config/findpaths.py
bluetech: BTW. The reason this can happen at all is pretty funky.

pytest determines the rootdir by looking at all non-option arguments and finding the common ancestor etc.

But because this code runs before plugins are loaded, any plugin argument not prefixed by a `-`, such as `test*bla` in `pytest --tx test*bla` (as opposed to if the `--tx=test*bla` form was used) gets interpreted as a possible path. If it is actually a valid path then you'd have one seriously confused user...

I looked into fixing it but seems not really possible.

nicoddemus: Yeah it is not simple due to how options are parsed. 😕

PR #7627, opened 2020-08-07
pytester: some improvements and encoding fix to subprocess luanching

### LINKED ISSUE #7623: testdir.runpytest_subprocess() fails with UnicodeDecodeError in Windows
This seems similar to #298.

Original issue code: https://github.com/altendky/qtrio/tree/72863a1e769d9f942e2062904c9b6f3fb077d3b8
CI (GitHub Actions Windows): https://github.com/altendky/qtrio/pull/2/checks?check_run_id=947483753
```python-traceback

>       result = testdir.runpytest_subprocess(timeout=4 * timeout)

C:\hostedtoolcache\windows\Python\3.8.5\x64\lib\site-packages\qtrio\_tests\test_pytest.py:44: 

self = <encodings.utf_8.IncrementalDecoder object at 0x0000022A4FF3E520>
input = b'============================= test session starts =============================\r\nplatform win32 -- Python 3.8.5, p... qtrio._exceptions.RunnerTim...\r\n============================== 1 failed in 0.74s ==============================\r\n'
final = True

    def decode(self, input, final=False):
        # decode input (taking the buffer into account)
        data = self.buffer + input
>       (result, consumed) = self._buffer_decode(data, self.errors, final)
E       UnicodeDecodeError: 'utf-8' codec can't decode byte 0x96 in position 9091: invalid start byte

c:\hostedtoolcache\windows\python\3.8.5\x64\lib\codecs.py:322: UnicodeDecodeError
```

```pycon
>>> b[9050:9150]
b"quest is\r\n           *also* asynchronous \x96 you don't find out until later whether the\r\n           op"
```

The Trio source being included in the context resulting in the `\x96`: https://github.com/python-trio/trio/blob/15facd95379a496b234cc762c4ae69bf4fcdd0bf/trio/_core/_traps.py#L120

```
2020-08-05T02:10:08.7472418Z + python -m pip list
2020-08-05T02:10:09.3155684Z Package             Version
2020-08-05T02:10:09.3156457Z ------------------- ---------
2020-08-05T02:10:09.3156870Z async-generator     1.10
2020-08-05T02:10:09.3157158Z atomicwrites        1.4.0
2020-08-05T02:10:09.3157384Z attrs               19.3.0
2020-08-05T02:10:09.3157830Z cffi                1.14.1
2020-08-05T02:10:09.3158411Z click               7.1.2
2020-08-05T02:10:09.3159025Z colorama            0.4.3
2020-08-05T02:10:09.3159984Z coverage            5.2.1
2020-08-05T02:10:09.3160282Z decorator           4.4.2
2020-08-05T02:10:09.3161003Z idna                2.10
2020-08-05T02:10:09.3161594Z iniconfig           1.0.1
2020-08-05T02:10:09.3162319Z more-itertools      8.4.0
2020-08-05T02:10:09.3162739Z outcome             1.0.1
2020-08-05T02:10:09.3163670Z packaging           20.4
2020-08-05T02:10:09.3164165Z pip                 20.2.1
2020-08-05T02:10:09.3164901Z pluggy              0.13.1
2020-08-05T02:10:09.3165352Z py                  1.9.0
2020-08-05T02:10:09.3166163Z pycparser           2.20
2020-08-05T02:10:09.3166517Z pyparsing           2.4.7
2020-08-05T02:10:09.3167314Z PySide2             5.15.0
2020-08-05T02:10:09.3167907Z pytest              6.0.1
2020-08-05T02:10:09.3168666Z pytest-cov          2.10.0
2020-08-05T02:10:09.3169237Z pytest-faulthandler 2.0.1
2020-08-05T02:10:09.3169938Z pytest-qt           3.3.0
2020-08-05T02:10:09.3170361Z QtPy                1.9.0
2020-08-05T02:10:09.3171850Z qtrio               0.1.0+dev
2020-08-05T02:10:09.3172256Z setuptools          49.2.1
2020-08-05T02:10:09.3172457Z shiboken2           5.15.0
2020-08-05T02:10:09.3172703Z six                 1.15.0
2020-08-05T02:10:09.3172974Z sniffio             1.1.0
2020-08-05T02:10:09.3173542Z sortedcontainers    2.2.2
2020-08-05T02:10:09.3174126Z toml                0.10.1
2020-08-05T02:10:09.3175258Z trio                0.16.0
2020-08-05T02:10:09.3175712Z wheel               0.34.2
```

So what should be different?  I think there are two pieces.  First, pytest should probably attempt to control the subprocess' output (and input?) encoding.  Second, pytest should not fail on decoding with only the context of the single byte value and location in the otherwise inaccessible output.

Hopefully I'll get to adding a test forcing the issue.

bluetech:
The first commit is a little cleanup.

The second and third commits add some features to `Testdir.run` that are needed for the next commit but are good to have anyway.

The last commit fixes #7623. I haven't actually tested it, because getting a non UTF-8 encoding (or a subset of it) on a modern Linux is almost impossible. But I think it should work. Pasting its commit message here:

To ensure that pytester subprocess tests are not affected by host system IO encoding (as do the inprocess tests, thanks to pytest's capturing), force its stdout/stderr to a known good encoding (UTF-8).

The alternative to this approach is, instead of trying to force the subprocess's encoding, try to adapt to it during decoding. Practically, it would mean trying to decode with `locale.getpreferredencoding()` or such. I opted against it in order to promote consistency across machines, and for consistency with inprocess tests, even though it makes the tests less realistic on the host machine.

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1218,7 +1218,10 @@ class Testdir:
 
         You probably want to use :py:meth:`run` instead.
         """
-        env = os.environ.copy()
+        if kw.get("env") is not None:
+            env = kw["env"].copy()
+        else:
+            env = os.environ.copy()
         env["PYTHONPATH"] = os.pathsep.join(
             filter(None, [os.getcwd(), env.get("PYTHONPATH", "")])
         )
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1245,7 +1245,13 @@ class Testdir:
         return popen
 
     def run(
-        self, *cmdargs, timeout: Optional[float] = None, stdin=CLOSE_STDIN
+        self,
+        *cmdargs,
+        timeout: Optional[float] = None,
+        stdin=CLOSE_STDIN,
+        encoding: str = "utf-8",
+        errors: str = "strict",
+        env=None
     ) -> RunResult:
         """Run a command with arguments.
 
@@ -1261,6 +1267,13 @@ class Testdir:
             the pipe, otherwise it is passed through to ``popen``.
             Defaults to ``CLOSE_STDIN``, which translates to using a pipe
             (``subprocess.PIPE``) that gets closed.
+        :param encoding:
+            The encoding used to decode the stdout and stderr of the process.
+        :param errors:
+            The error handler to use when decoding the stdout and stderr of the process.
+        :param env:
+            The enviornment variables for the process.
+            The default is to inherit the environment of the current process.
         """
         __tracebackhide__ = True
 
@@ -1282,6 +1295,7 @@ class Testdir:
                 stdout=stdout,
                 stderr=stderr,
                 close_fds=(sys.platform != "win32"),
+                env=env,
             )
             if isinstance(stdin, bytes):
                 popen.stdin.close()
@@ -1299,8 +1313,8 @@ class Testdir:
 
             duration = timing.perf_counter() - start
 
-        with open(stdout_path, encoding="utf8") as stdout_text, open(
-            stderr_path, encoding="utf8"
+        with open(stdout_path, encoding=encoding, errors=errors) as stdout_text, open(
+            stderr_path, encoding=encoding, errors=errors
         ) as stderr_text:
             out = stdout_text.read().splitlines()
             err = stderr_text.read().splitlines()
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1345,14 +1345,26 @@ class Testdir:
 
         :rtype: RunResult
         """
-        return self.run(sys.executable, script)
+        # Use a consistent encoding, isolated from the host system.
+        encoding = "utf-8"
+        env = {
+            **os.environ.copy(),
+            "PYTHONIOENCODING": encoding,
+        }
+        return self.run(sys.executable, script, encoding=encoding, env=env)
 
-    def runpython_c(self, command):
+    def runpython_c(self, command) -> RunResult:
         """Run python -c "command".
 
         :rtype: RunResult
         """
-        return self.run(sys.executable, "-c", command)
+        # Use a consistent encoding, isolated from the host system.
+        encoding = "utf-8"
+        env = {
+            **os.environ.copy(),
+            "PYTHONIOENCODING": encoding,
+        }
+        return self.run(sys.executable, "-c", command, encoding=encoding, env=env)
 
     def runpytest_subprocess(self, *args, timeout: Optional[float] = None) -> RunResult:
         """Run pytest as a subprocess with given arguments.
@@ -1378,7 +1390,13 @@ class Testdir:
         if plugins:
             args = ("-p", plugins[0]) + args
         args = self._getpytestargs() + args
-        return self.run(*args, timeout=timeout)
+        # Use a consistent encoding, isolated from the host system.
+        encoding = "utf-8"
+        env = {
+            **os.environ.copy(),
+            "PYTHONIOENCODING": encoding,
+        }
+        return self.run(*args, timeout=timeout, encoding=encoding, env=env)
 
     def spawn_pytest(
         self, string: str, expect_timeout: float = 10.0
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -1254,80 +1254,71 @@ class Testdir:
             The period in seconds after which to timeout and raise
             :py:class:`Testdir.TimeoutExpired`.
         :param stdin:
-            Optional standard input.  Bytes are being send, closing
+            Optional standard input.  If `bytes`, they are sent, closing
             the pipe, otherwise it is passed through to ``popen``.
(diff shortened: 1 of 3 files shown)


src/_pytest/pytester.py
```diff
@@ -1370,7 +1391,13 @@ def runpytest_subprocess(self, *args, timeout: Optional[float] = None) -> RunRes
         if plugins:
             args = ("-p", plugins[0]) + args
         args = self._getpytestargs() + args
-        return self.run(*args, timeout=timeout)
+        # Use a consistent encoding, isolated from the host system.
+        encoding = "utf-8"
+        env = {
+            **os.environ.copy(),
+            "PYTHONIOENCODING": encoding,
+        }
```
asottile, line 1399: this makes it hard/impossible to test under other encoding scenarios.

src/_pytest/pytester.py
```diff
@@ -1254,80 +1263,80 @@ def run(
...
                 )
+                raise self.TimeoutExpired(message)
 
-                popen.kill()
-                popen.wait()
-                raise self.TimeoutExpired(timeout_message)
+            duration = timing.perf_counter() - start
+
+        stdout_text = open(stdout_path, encoding=encoding, errors=errors)
+        stderr_text = open(stderr_path, encoding=encoding, errors=errors)
```
asottile, line 1318: there's a potential leak of the first file descriptor if this call were to fail (same above)

src/_pytest/pytester.py
```diff
@@ -1254,80 +1263,80 @@ def run(
...
-        self._dump_lines(err, sys.stderr)
+            for line in out:
+                print(line, file=sys.stdout)
+        except UnicodeEncodeError:
+            print("couldn't print to stdout because of encoding")
+        try:
+            for line in err:
+                print(line, file=sys.stderr)
+        except UnicodeEncodeError:
+            print("couldn't print to stderr because of encoding")
```
asottile, line 1332: these try/excepts would cause partial output to be written (and stopping at the problematic line) -- maybe consider writing everything at once instead so it's all-or-nothing

src/_pytest/pytester.py
```diff
@@ -1254,80 +1263,80 @@ def run(
...
+
             popen = self.popen(
                 cmdargs,
                 stdin=stdin,
-                stdout=f1,
-                stderr=f2,
+                stdout=stdout,
+                stderr=stderr,
                 close_fds=(sys.platform != "win32"),
             )
```
altendky, line 1300:
```suggestion
                env=env,
            )
```

I'll go make another branch and try this out.

altendky: Looks like that did the trick...  along with the hundreds of other changes you made.  `:]`

https://github.com/altendky/qtrio/pull/138/checks?check_run_id=964937422

bluetech: Whoops, fixed, thanks!

And thanks for verifying it fixes the problem.

src/_pytest/pytester.py
bluetech: Can add an `env` argument to `runpytest_subprocess`, but I prefer to punt on this as it doesn't seem like it was necessary up to now.

src/_pytest/pytester.py
bluetech: Fixed, thanks.

src/_pytest/pytester.py
bluetech: A partial output is somewhat nicer for debug-ability I think?

My own preference would be to remove this except entirely and see if anyone complains :)

src/_pytest/pytester.py
asottile: actually on second thought I don't think we should do this at all (but perhaps provide `encoding` as an option).  it potentially masks a class of common encoding errors by waving away the default behaviour

bluetech: It masks configuration issues on the host machine, not a bug in the tested program. I think such isolation is generally good. Another reason is to have consistency with inprocess tests (where the output is captured, and stdout/stderr are always utf-8).

But just to be clear, the current behavior is definitely wrong, it just *assumes* UTF-8. In the PR message I described an alternative approach (use `locale.getpreferredencoding()`) and why I rejected it, do you prefer that anyway? (or another approach)

asottile: hmmmm yeah that's tricky

I think what I'd do is grown an `encoding=...` parameter and default it to `None`, pass it along if non-None as PYTHONIOENCODING, and decode using either `getpreferredencoding()` (nothing passed) or `encoding` if passed?

I think that would be the most ~realistic setup (so the testsuite doesn't have some surprising behaviour that's different from reality)

bluetech: I'm fine with this plan, but would prefer the default to be `utf-8`, not `None`. In this case, I prefer the tests to be consistent/isolated over being realistic.

asottile: my fear is that someone will assume pytest is following the default behaviour, write a test against the pytest behaviour, and then be surprised when it's actually broken in reality

bluetech:
Going to close this for now, might get back to it at some point.

PR #7635, opened 2020-08-11
Remove faq.rst from docs

### LINKED ISSUE #1477: FAQ: improve comparison with nose and unittest
From #1476:

> Overall I would suggest still adding a bit more of an explanation to https://pytest.org/latest/faq.html#how-does-pytest-relate-to-nose-and-unittest, outlining some of the major features/differences

cc @philfreo

mcsitter:

In `adopt.rst` is a reference to `faq.html` for adoption of pytest when coming from nose or unittest. I think linking to `nose.html` and `unittest.html` accordingly, literally just changing the link to 2 links and not changing the text, should also be in this PR. Happy to add it right away if you agree!

closes #1477


nicoddemus:
> Happy to add it right away if you agree!

Yes please, otherwise we will have broken links. 😁 

I also noticed we have some references to the faq in the changelog.rst and release-2.3.0.rst files.

Not sure if we should remove `faq.rst` altogether or leave the file just with a note saying that it was removed... what do others think? @The-Compiler @bluetech @asottile @RonnyPfannschmidt ?

mcsitter:
I updated the links!

I of course also saw the references you mentioned. I added the changelog for the removal of the faq so it is not gone without notice.

Newbie question: Is it good practice to change links in old changelogs?

If so, one could simply point to the latest version in which the faq still existed?

nicoddemus:
> Newbie question: Is it good practice to change links in old changelogs?

It is not obvious actually, but I think keeping a broken link there doesn't really help. 😁 

> If so, one could simply point to the latest version in which the faq still existed?

Ahh good idea, I'm OK with going with that solution for the old changelogs. 👍

nicoddemus:
Thanks @mcsitter, appreciate it!

PR #7639, opened 2020-08-12
get_dirs_from_args handles paths with invalid syntax

### LINKED ISSUE #7638: get_dirs_from_args() breaks on Windows with options that include full paths
pytest-xdist rsync feature uses options in the command line that looks like files, but can themselves point to full paths:

```
C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_popen_rsync_subdir0\\popen\\chdir=C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pytest-of-appveyor\\pytest-0\\test_popen_rsync_subdir1\\dest
```

(From https://ci.appveyor.com/project/pytestbot/pytest-xdist/build/job/r7el1489cu4ipfmp).

Preparing a PR shortly. 👍

nicoddemus:
Fix #7638

### DIFF
--- a/changelog/7638.bugfix.rst
+++ b/changelog/7638.bugfix.rst
@@ -1,0 +1,1 @@
+Fix handling of command-line options that appear as paths but trigger an OS-level syntax error on Windows, such as the options used internally by ``pytest-xdist``.
--- a/src/_pytest/config/findpaths.py
+++ b/src/_pytest/config/findpaths.py
@@ -147,7 +147,7 @@ def get_dirs_from_args(args: Iterable[str]) -> List[Path]:
 
     def safe_exists(path: Path) -> bool:
         # This can throw on paths that contain characters unrepresentable at the OS level,
-        # or with invalid syntax on Windows
+        # or with invalid syntax on Windows (https://bugs.python.org/issue35306)
         try:
             return path.exists()
         except OSError:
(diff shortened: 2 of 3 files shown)


bluetech:
I don't have a windows machine to test, but by the removal of the `sys.version_info < (3, 8)` check I assume this (`OSError` on invalid syntax) can happen even on Python>=3.8? If so it might be worth reporting to Python upstream, it's just really annoying that these methods can raise and they already changed it do the right thing regarding "unrepresentable characters" ([ref](https://docs.python.org/3.8/library/pathlib.html#methods) second paragraph).

Regarding the `needs backport` label, I think this problem only happens in the new pathlib code which is not in 6.0.x. The AppVeyor run you linked seems to confirm it only fails on master. So IIUC it doesn't need a backport.

nicoddemus:
> I don't have a windows machine to test, but by the removal of the sys.version_info < (3, 8) check I assume this (OSError on invalid syntax) can happen even on Python>=3.8? If so it might be worth reporting to Python upstream, it's just really annoying that these methods can raise and they already changed it do the right thing regarding "unrepresentable characters" (ref second paragraph).

Good point, was about to reported it but then noticed that was already the case: https://bugs.python.org/issue35306

I've added that link as a comment. 👍 

>

PR #7643, opened 2020-08-14
Fix Module.name from full path without drive letter

### LINKED ISSUE #7628: pytest-6.0.1 does not find custom fixtures, test filename incorrect (just ..\..\)
```
platform win32 -- Python 3.7.0, pytest-6.0.1, py-1.9.0, pluggy-0.13.1 -- c:\prog
ram files (x86)\python37-32\python.exe
cachedir: .pytest_cache
metadata: {'Python': '3.7.0', 'Platform': 'Windows-2008ServerR2-6.1.7601-SP1', '
Packages': {'pytest': '6.0.1', 'py': '1.9.0', 'pluggy': '0.13.1'}, 'Plugins': {'
html': '2.1.1', 'metadata': '1.10.0'}}
rootdir: C:\
plugins: html-2.1.1, metadata-1.10.0
collected 2 items

..\..\::test_install[uninstall_S0_0] ERROR (0 secs)

..\..\::test_install[ver_1.6.1_S0_1] ERROR (0 secs)


=================================== ERRORS ====================================
________________ ERROR at setup of test_install[uninstall_S0_0] _________________
file C:\install\tests\install_test.py, line 77
  @pytest.mark.parametrize("rel_info", release_dirs, ids=test_ids)
  def test_install(install_fixture, rel_info):
file C:\install\tests\install_test.py, line 26
  @pytest.fixture(scope='session')
  def module_init_fixture(session_fixture):
E       fixture 'session_fixture' not found
>       available fixtures: cache, capfd, capfdbinary, caplog, capsys, capsysbin
ary, doctest_namespace, extra, install_fixture, metadata, module_init_fixture, m
onkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml
_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

```
Package            Version
------------------ -------
atomicwrites       1.4.0
attrs              19.3.0
bcrypt             3.1.7
cffi               1.14.1
colorama           0.4.3
cryptography       3.0
importlib-metadata 1.7.0
iniconfig          1.0.1
json-flatten       0.1
more-itertools     8.4.0
packaging          20.4
paramiko           2.7.1
pip                10.0.1
pluggy             0.13.1
py                 1.9.0
pycparser          2.20
PyNaCl             1.4.0
pyparsing          2.4.7
pytest             6.0.1
pytest-html        2.1.1
pytest-metadata    1.10.0
scp                0.13.2
selenium           3.141.0
setuptools         39.0.1
six                1.15.0
toml               0.10.1
urllib3            1.25.10
zipp               3.1.0
```

 Python 3.7.0, pytest-6.0.1
OS Name:                   Microsoft Windows Server 2008 R2 Enterprise
OS Version:                6.1.7601 Service Pack 1 Build 7601

The same test work with pytest-5.3.2

nicoddemus:
Fix #7628

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -549,7 +549,9 @@ class Session(nodes.FSCollector):
         self._initial_parts = []  # type: List[Tuple[py.path.local, List[str]]]
         self.items = items = []  # type: List[nodes.Item]
         for arg in args:
-            fspath, parts = self._parsearg(arg)
+            fspath, parts = resolve_collection_argument(
+                self.config.invocation_dir, arg, as_pypath=self.config.option.pyargs
+            )
             self._initial_parts.append((fspath, parts))
             initialpaths.append(fspath)
         self._initialpaths = frozenset(initialpaths)
@@ -673,37 +675,6 @@ class Session(nodes.FSCollector):
                 return
             yield from m
 
-    def _tryconvertpyarg(self, x: str) -> str:
-        """Convert a dotted module name to path."""
-        try:
-            spec = importlib.util.find_spec(x)
-        # AttributeError: looks like package module, but actually filename
-        # ImportError: module does not exist
-        # ValueError: not a module name
-        except (AttributeError, ImportError, ValueError):
-            return x
-        if spec is None or spec.origin is None or spec.origin == "namespace":
-            return x
-        elif spec.submodule_search_locations:
-            return os.path.dirname(spec.origin)
-        else:
-            return spec.origin
-
-    def _parsearg(self, arg: str) -> Tuple[py.path.local, List[str]]:
-        """Return (fspath, names) tuple after checking the file exists."""
-        strpath, *parts = str(arg).split("::")
-        if self.config.option.pyargs:
-            strpath = self._tryconvertpyarg(strpath)
-        fspath = Path(str(self.config.invocation_dir), strpath)
-        fspath = absolutepath(fspath)
-        if not fspath.exists():
-            if self.config.option.pyargs:
-                raise UsageError(
-                    "file or package not found: " + arg + " (missing __init__.py?)"
-                )
-            raise UsageError("file not found: " + arg)
-        return py.path.local(str(fspath)), parts
-
     def matchnodes(
         self, matching: Sequence[Union[nodes.Item, nodes.Collector]], names: List[str],
     ) -> Sequence[Union[nodes.Item, nodes.Collector]]:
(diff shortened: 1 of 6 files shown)


src/_pytest/nodes.py
```diff
@@ -515,6 +515,7 @@ def __init__(
         session: Optional["Session"] = None,
         nodeid: Optional[str] = None,
     ) -> None:
+        fspath = py.path.local(os.path.abspath(str(fspath)))
```
bluetech, line 518: `py.path.local` already runs `abspath` on its input:

https://github.com/pytest-dev/py/blob/2da2caea38812eaa3ce09dd5292e3635ce9b16c8/py/_path/local.py#L153-L163

So I wonder how this line has any effect. I am assuming that `abspath` is idempotent, but maybe it isn't on Windows? Or maybe the troublesome path comes from the `os.getcwd()` code path, and the result of that isn't absolute?

Basically I'm interested to know a bit more what is going on here:

1. Where does the input `fspath` come from, and what string was used to initialize it
2. How is it transformed by this line

nicoddemus: Good questions, I didn't realize that `py.path.local` would call `abspath` on its input.

Digging further, I found the place where the path is not correctly realized:

https://github.com/pytest-dev/pytest/blob/3f0abcc6a5e0ddf1c17d42125421c6816fcf3178/src/_pytest/main.py#L696-L697

```python
>>> import py.path
>>> py.path.local(r'd:\projects').join(r'\foo', abs=True)
local('\\foo')
```

So it is a problem with `py.path.local.join()`.

Decided to throw out that bit of code and use the `Path` API locally, which doesn't have that bug.

Thanks for the inquiry, that lead me to investigate it further and find the real cause of the issue. 👍 😅

nicoddemus:
Decided to refactor `_parsearg` into a free function and add more unittests. 👍

src/_pytest/main.py
```diff
@@ -693,15 +694,15 @@ def _parsearg(self, arg: str) -> Tuple[py.path.local, List[str]]:
         strpath, *parts = str(arg).split("::")
         if self.config.option.pyargs:
             strpath = self._tryconvertpyarg(strpath)
-        relpath = strpath.replace("/", os.sep)
-        fspath = self.config.invocation_dir.join(relpath, abs=True)
-        if not fspath.check():
+        fspath = Path(str(self.config.invocation_dir), strpath)
```
bluetech:
```suggestion
        fspath = self.config.invocation_params.dir / strpath
```

src/_pytest/main.py
```diff
@@ -770,3 +741,59 @@ def genitems(
...
+    if spec is None or spec.origin is None or spec.origin == "namespace":
+        return module_name
+    elif spec.submodule_search_locations:
+        return os.path.dirname(spec.origin)
+    else:
+        return spec.origin
+
+
+def resolve_path_argument(
+    invocation_dir: py.path.local, arg: str, *, as_pypath=False
```
bluetech, line 764:
```suggestion
    invocation_dir: py.path.local, arg: str, *, as_pypath: bool = False
```

src/_pytest/main.py
```diff
@@ -770,3 +741,59 @@ def genitems(
...
+) -> Tuple[py.path.local, List[str]]:
+    """Parse path arguments optionally containing selection parts and return (fspath, names).
```
bluetech, line 766: Great comment

src/_pytest/main.py
```diff
@@ -770,3 +741,59 @@ def genitems(
...
+        return module_name
+    if spec is None or spec.origin is None or spec.origin == "namespace":
+        return module_name
+    elif spec.submodule_search_locations:
+        return os.path.dirname(spec.origin)
+    else:
+        return spec.origin
+
+
+def resolve_path_argument(
```
bluetech, line 763: I think "path argument" is perhaps *slightly* misleading given that it can be a dotted python path and the selection part.

Maybe `resolve_selection_argument`? `resolve_collection_argument`?

nicoddemus: Good point, went with `resolve_collection_argument`. 👍

nicoddemus:
Hmmm I was hoping to backport this, but this would also need the pathlib improvements we have on master right now (notably `absolutepath`). Guess this fix can wait for 6.1 then.

PR #7647, opened 2020-08-14
Move rule characters into class attributes of TerminalReporter

doerwalter:
Move characters used for drawing horizontal rules into class attributes of `TerminalReporter`.

This allows those characters to be replaced by plugins (i.e. with Unicode box drawing characters) and documents which characters are used.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -298,6 +298,7 @@ Vlad Dragos
 Vlad Radziuk
 Vladyslav Rachek
 Volodymyr Piskun
+Walter Dörwald
 Wei Lin
 Wil Cooley
 William Lee
--- a/changelog/7608.feature.rst
+++ b/changelog/7608.feature.rst
@@ -1,0 +1,5 @@
+Move characters used for drawing horizontal rules into class attributes of
+TerminalReporter.
+
+This allows those characters to be replaced by plugins (i.e. with Unicode
+box drawing characters) and documents which characters are used.
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -680,7 +680,9 @@ class LogXML:
         logfile.close()
 
     def pytest_terminal_summary(self, terminalreporter: TerminalReporter) -> None:
-        terminalreporter.write_sep("-", "generated xml file: {}".format(self.logfile))
+        terminalreporter.write_hrule_single(
+            "generated xml file: {}".format(self.logfile)
+        )
 
     def add_global_property(self, name: str, value: object) -> None:
         __tracebackhide__ = True
--- a/src/_pytest/pastebin.py
+++ b/src/_pytest/pastebin.py
@@ -61,7 +61,7 @@ def pytest_unconfigure(config: Config) -> None:
         tr = config.pluginmanager.getplugin("terminalreporter")
         del tr._tw.__dict__["write"]
         # Write summary.
-        tr.write_sep("=", "Sending information to Paste Service")
+        tr.write_hrule_double("Sending information to Paste Service")
         pastebinurl = create_new_paste(sessionlog)
         tr.write_line("pastebin session-log: %s\n" % pastebinurl)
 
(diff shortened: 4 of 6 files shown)


bluetech:
Thanks for splitting this! A couple of questions to start with:

1. How are plugins intended to replace the characters? Is it subclassing `TerminalReporter`, replacing it entirely, or mutating the class attributes, or something else?
2. You add both class attributes `hrule_single` etc. and methods `write_hrule_single()` etc. Which of these is the intended customization point?

doerwalter:
> Thanks for splitting this! A couple of questions to start with:
> 
>     1. How are plugins intended to replace the characters? Is it subclassing `TerminalReporter`, replacing it entirely, or mutating the class attributes, or something else?

I can imagine all of those variants.

>     2. You add both class attributes `hrule_single` etc. and methods `write_hrule_single()` etc. Which of these is the intended customization point?

I think for simple customizations replacing the characters (either directly or by overrwriting the attributes in a subclasses) is the way to go. But I can imagine bars that are done by using a different background color or by drawing a frame around the title. For this overwriting the method would be necessary.

But then the question becomes whether `write_hrule_double()` etc. are the best names for the methods. What if someone wants to have different rules for different situations? Maybe the names should describe some kind of hierarchy like `write_title_level1()`, `write_title_level2()`? Or maybe each distinct call needs it's own `write_hrule_...()` implemenation?

nicoddemus:
> I think for simple customizations replacing the characters (either directly or by overrwriting the attributes in a subclasses) is the way to go. But I can imagine bars that are done by using a different background color or by drawing a frame around the title. For this overwriting the method would be necessary.

Indeed, that was my understanding as well.

> But then the question becomes whether write_hrule_double() etc. are the best names for the methods. What if someone wants to have different rules for different situations? Maybe the names should describe some kind of hierarchy like write_title_level1(), write_title_level2()? Or maybe each distinct call needs it's own write_hrule_...() implemenation?

Those are good points. To me the `level_1`, `level_2` makes sense, seems similar to different "headings" in markups.

doerwalter:
> [...]
> > But then the question becomes whether write_hrule_double() etc. are the best names for the methods. What if someone wants to have different rules for different situations? Maybe the names should describe some kind of hierarchy like write_title_level1(), write_title_level2()? Or maybe each distinct call needs it's own write_hrule_...() implemenation?
> 
> Those are good points. To me the `level_1`, `level_2` makes sense, seems similar to different "headings" in markups.

OK, then `write_hrule_double()` would probably be `write_hrule_level1()`. But I don't know whether `write_hrule_single()` or `write_hrule_lower()` would be level 2. And `write_hrule_error()` doesn't fit into this hierarchy anyway.

graingert:
`write_hrule(level: LevelEnum) -> int:`?

doerwalter:
> `write_hrule(level: LevelEnum) -> int:`?

Nice idea. But I'm not sure what the return value of type `int` is supposed to represent.

And if `pytest` decides to add new `LevelEnum` values, plugins might implement a fallback for values they don't know, which makes the whole thing somewhat backwards compatible.

I'm no longer sure if `write_hrule...` is the correct name for this. Maybe something like `write_header()`?

graingert:
To return the int from https://docs.python.org/3/library/io.html#io.TextIOBase.write

doerwalter:
Ah, OK. Currently none of the `write_...()` methods return this value. I don't know if there would be any use for it.

nicoddemus:
I like `def def write_header(level: int = 0) -> None:`.

@graingert do you see a use case for returning an int from that method?

graingert:
Ok `-> None:` is good too

doerwalter:
But this version wouldn't allow doing something fancier. Why can't the hook do it's own drawing? Because this would have to callback into some basic drawing functionality?

nicoddemus:
> Why can't the hook do it's own drawing? Because this would have to callback into some basic drawing functionality?

That's a possibility indeed; in fact we already have a hook that does provide the `terminalreporter` to hook implementations to do as they please: 

```python
def pytest_terminal_summary(
    terminalreporter: "TerminalReporter", exitstatus: "ExitCode", config: "Config",
) -> None:
```

bluetech:
As per the discussion above, this feature would need some more design work, so I'll close this PR. Thanks @doerwalter.

PR #7651, opened 2020-08-15
capture: fix disabled()/global_and_fixture_disabled() enabling capturing when it was disabled

bluetech:
The `CaptureManager.global_and_fixture_disabled()` context manager (and `CaptureFixture.disabled()` which calls it) did `suspend(); ...; resume()` but if the capturing was already suspended, the `resume()` would resume it when it shouldn't.

This caused caused some messages to be swallowed when `--log-cli` is used because it uses `global_and_fixture_disabled` when capturing is not necessarily resumed.

### DIFF
--- a/changelog/7148.bugfix.rst
+++ b/changelog/7148.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``--log-cli`` potentially causing unrelated ``print`` output to be swallowed.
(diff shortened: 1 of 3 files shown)


bluetech:
Thanks for the link @nicoddemus, looks a bit too involved for me to get working ATM, so I opened #7679 to track this.

PR #7660, opened 2020-08-17
Remove deprecated features

### LINKED ISSUE #5585: Remove deprecated features in 6.0
Refs: https://github.com/pytest-dev/pytest/pull/5529, https://github.com/pytest-dev/pytest/issues/5180

nicoddemus:
This removes all features which have been deprecated over 2 minor versions ago or more.

@RonnyPfannschmidt also removed `resultlog`, seems like there's has been enough time for people to try `pytest-reportlog` and provide feedback.

Close #5585


doc/en/deprecations.rst
```diff
@@ -59,38 +72,46 @@ display captured output when tests fail: ``no``, ``stdout``, ``stderr``, ``log``
 
 
 
-Node Construction changed to ``Node.from_parent``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Result log (``--result-log``)
```
nicoddemus, line 75: This diff is pretty confusing, suggest to read the full file instead.

src/_pytest/fixtures.py
```diff
@@ -1315,53 +1306,6 @@ def fixture(  # noqa: F811
         name the decorated function ``fixture_<fixturename>`` and then use
         ``@pytest.fixture(name='<fixturename>')``.
     """
-    # Positional arguments backward compatibility.
```
nicoddemus, line 1318: image

src/_pytest/nodes.py
```diff
@@ -97,8 +96,13 @@ def ischildnode(baseid: str, nodeid: str) -> bool:
...
     def __call__(self, *k, **kw):
-        warnings.warn(NODE_USE_FROM_PARENT.format(name=self.__name__), stacklevel=2)
-        return super().__call__(*k, **kw)
+        msg = (
+            "Direct construction of {name} has been deprecated, please use {name}.from_parent.\n"
+            "See "
+            "https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent"
+            " for more details."
+        ).format(name=self.__name__)
+        fail(msg, pytrace=False)
```
nicoddemus, line 105: @RonnyPfannschmidt I suggest that was the intention in the end, correct?

nicoddemus:
> The problem is that the from_parent API cannot be typed properly, and once we disallow direct construction there is no turning back ever.

Actually I recall @RonnyPfannschmidt mentioning this was supposed to be transitory, but don't remember the details now.

RonnyPfannschmidt:
Yup, until we get rid of the tricky mess in some of the ctors from_parent is a necessity mess

nicoddemus:
FTR that's what https://github.com/pytest-dev/pytest/projects/3 is about.

PR #7661, opened 2020-08-18
Update talks/trainings page

The-Compiler:
- Remove past webinar
- Add new open training
- Add some talks/webinars by Oliver Bestwalter and by me - I'm sure there have been more talks than those, but I don't have the time to go through years of conferences, so I just added the ones I remember
- Remove some stale link targets

Also, what do you think about moving the "Next Open Trainings" sidebar from the talks page to the main page? Right now it's quite hard to find the talks/tutorials page (and the training announcement) because it's also not part of the sidebar menu yet (I think it was with the old theme?).

IMHO it'd be nice to show it on the main page instead (since it's often only around for a couple of weeks/months anyways) to give those trainings a bit more visiblity - but since I'm probably the main person benefiting from that kind of thing that's not a decision I want to take alone. What do others think?

### DIFF
--- a/doc/en/index.rst
+++ b/doc/en/index.rst
@@ -1,5 +1,12 @@
 :orphan:
 
+.. sidebar:: Next Open Trainings
+
+   - `pytest: Test Driven Development (nicht nur) für Python <https://workshoptage.ch/workshops/2020/pytest-test-driven-development-nicht-nur-fuer-python/>`_ (German) at the `CH Open Workshoptage <https://workshoptage.ch/>`_, September 8 2020, HSLU Campus Rotkreuz (ZG), Switzerland.
+   - `Professional testing with Python <https://www.python-academy.com/courses/specialtopics/python_course_testing.html>`_, via Python Academy, February 1-3 2021, Leipzig (Germany) and remote.
+
+   Also see `previous talks and blogposts <talks.html>`_.
+
 .. _features:
 
 pytest: helps you write better programs
--- a/doc/en/talks.rst
+++ b/doc/en/talks.rst
@@ -4,10 +4,8 @@ Talks and Tutorials
 
 .. sidebar:: Next Open Trainings
 
-   - `Free 1h webinar: "pytest: Test Driven Development für Python" <https://mylearning.ch/kurse/online-kurse/tech-webinar/>`_ (German), online, August 18 2020.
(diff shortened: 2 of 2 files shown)


nicoddemus:
> IMHO it'd be nice to show it on the main page instead (since it's often only around for a couple of weeks/months anyways) to give those trainings a bit more visiblity - but since I'm probably the main person benefiting from that kind of thing that's not a decision I want to take alone. What do others think?

I agree, go for it. More visibility benefits everyone.

In fact we could consider using the pytest twitter account to announce those talks too, for more visibility? cc @hackebrot @pfctdayelise

The-Compiler:
I pushed a commit to move the block to the main page. Here's how it looks:

image

Not exactly optimal, but I haven't found a way to customize this without having to fork the Pallets Sphinx theme. Given that it'll get smaller in three weeks (when the first training is over) I'm okay with that, unless someone objects. Alternatively, I guess we could add some extra text to the first paragraph of the page if someone can think of something useful.

> In fact we could consider using the pytest twitter account to announce those talks too, for more visibility? cc @hackebrot @pfctdayelise

I do have access to the Twitter account as well, and I do sometimes use it for [shamless self promotion](https://twitter.com/pytestdotorg/status/1293202922185068550) already. Will do the same for the Python Academy training as well, as I think it's fair to assume that its followers might be interested in such a thing.

nicoddemus:
(Btw we can see the preview now thanks to read the docs: https://pytest--7661.org.readthedocs.build/en/7661/)

> Alternatively, I guess we could add some extra text to the first paragraph of the page if someone can think of something useful.

Hmm perhaps removing the `helps you write better programs` from the title would make it look better already?

The-Compiler:
> Hmm perhaps removing the `helps you write better programs` from the title would make it look better already?

I think that'd make it worse, because it increases the big space between "An example of a simple test:" and the source (which is the main thing I'm worried about).

nicoddemus:
> I think that'd make it worse, because it increases the big space between "An example of a simple test:" and the source (which is the main thing I'm worried about).

Oh I see, I was more worried about the title that got a bit munched in there, but I don't mind it too much either. 👍

The-Compiler:
Taking the freedom to merge this given that it's been approved and nobody objects - the sign-up for the workshop in September will close on 31st, so I'll open another PR then to remove the first line, which will hopefully improve how things look.

PR #7666, opened 2020-08-22
move tox.ini into pyproject.toml

graingert:


asottile:
agreed, this is worse -- since the consensus seems in that direction I'm going to close this

PR #7667, opened 2020-08-22
Make min duration configurable for slowest tests

ssbarnea:
Enables users to use custom values for the threshold used to determine slow tests.

### DIFF
--- a/changelog/7667.feature.rst
+++ b/changelog/7667.feature.rst
@@ -1,0 +1,1 @@
+Make min duration configurable for slowest tests.
--- a/changelog/7667.feature.rst
+++ b/changelog/7667.feature.rst
@@ -1,1 +1,1 @@
-Make min duration configurable for slowest tests.
+New `--durations-min` command-line flag controls the minimal duration for inclusion in the slowest list of tests shown by `--durations`. Previously this was hard-coded to `0.005s`.
--- a/changelog/7667.feature.rst
+++ b/changelog/7667.feature.rst
@@ -1,1 +1,1 @@
-New `--durations-min` command-line flag controls the minimal duration for inclusion in the slowest list of tests shown by `--durations`. Previously this was hard-coded to `0.005s`.
+New ``--durations-min`` command-line flag controls the minimal duration for inclusion in the slowest list of tests shown by ``--durations``. Previously this was hard-coded to ``0.005s``.
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -52,10 +52,19 @@ def pytest_addoption(parser: Parser) -> None:
         metavar="N",
         help="show N slowest setup/test durations (N=0 for all).",
     )
+    group.addoption(
+        "--durations-min",
+        action="store",
+        type=float,
+        default=0.005,
+        metavar="N",
+        help="Minimal duration in seconds for inclusion in slowest list. Default 0.005",
+    )
 
 
 def pytest_terminal_summary(terminalreporter: "TerminalReporter") -> None:
     durations = terminalreporter.config.option.durations
+    durations_min = terminalreporter.config.option.durations_min
     verbose = terminalreporter.config.getvalue("verbose")
     if durations is None:
         return
@@ -76,11 +85,11 @@ def pytest_terminal_summary(terminalreporter: "TerminalReporter") -> None:
         dlist = dlist[:durations]
 
     for i, rep in enumerate(dlist):
-        if verbose < 2 and rep.duration < 0.005:
+        if verbose < 2 and rep.duration < durations_min:
             tr.write_line("")
             tr.write_line(
-                "(%s durations < 0.005s hidden.  Use -vv to show these durations.)"
-                % (len(dlist) - i)
+                "(%s durations < %gs hidden.  Use -vv to show these durations.)"
+                % (len(dlist) - i, durations_min)
             )
             break
         tr.write_line("{:02.2f}s {:<8} {}".format(rep.duration, rep.when, rep.nodeid))
--- a/doc/en/usage.rst
+++ b/doc/en/usage.rst
@@ -426,14 +426,15 @@ Pytest supports the use of ``breakpoint()`` with the following behaviours:
 Profiling test execution duration
 -------------------------------------
 
+.. versionchanged:: 6.0
 
-To get a list of the slowest 10 test durations:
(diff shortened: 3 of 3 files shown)


ssbarnea:
> I prefer `--durations-threshold` over `--durations-min`, I think it is more understandable. WDYT?
> 
> Also, do you foresee this being used more at the project level, or the single `pytest` invocation level? If it is something that makes more sense to set project-wide, then an INI would make more sense than a command line flag.

I prefer min as is easier to type, not really confusing. Regarding config, I partially agree because the only reason I did not do it is because current implementation make is harder to do it. I would prefer spending time on replacing the config engine with something that maps CLI arguments to config file items and env vars naturally, without extra code, so when someone adds a new configurable, it will be exposed to all places automatically.

We can still configure options in config files using addopts, so not a deal breaker.

changelog/7667.feature.rst
```diff
@@ -0,0 +1 @@
+New `--durations-min` command-line flag controls the minimal duration for inclusion in the slowest list of tests shown by `--durations`. Previously this was hard-coded to `0.005s`.
```
webknjaz, line 1: It's RST, double backticks should be used

ssbarnea: Done.

src/_pytest/runner.py
```diff
@@ -76,11 +85,11 @@ def pytest_terminal_summary(terminalreporter: "TerminalReporter") -> None:
         dlist = dlist[:durations]
 
     for i, rep in enumerate(dlist):
-        if verbose < 2 and rep.duration < 0.005:
+        if verbose < 2 and rep.duration < durations_min:
             tr.write_line("")
             tr.write_line(
-                "(%s durations < 0.005s hidden.  Use -vv to show these durations.)"
-                % (len(dlist) - i)
+                "(%s durations < %gs hidden.  Use -vv to show these durations.)"
```
webknjaz, line 91: Is "%gs" valid?

webknjaz:
```suggestion
                "(%s durations < %ss hidden.  Use -vv to show these durations.)"
```
Typo?

src/_pytest/runner.py
```diff
@@ -76,11 +85,11 @@ def pytest_terminal_summary(terminalreporter: "TerminalReporter") -> None:
...
+                % (len(dlist) - i, durations_min)
```
webknjaz, line 92: FWIW it's probably easier to just use an f-string

src/_pytest/runner.py
```diff
@@ -52,10 +52,19 @@ def pytest_addoption(parser: Parser) -> None:
         metavar="N",
         help="show N slowest setup/test durations (N=0 for all).",
     )
+    group.addoption(
+        "--durations-min",
+        action="store",
+        type=float,
+        default=0.005,
+        metavar="N",
+        help="Minimal duration in seconds for inclusion in slowest list. Default 0.005",
```
webknjaz, line 61:
```suggestion
        help="Minimal duration in seconds for inclusion in slowest list.",
```
Default is probably already appended by argparse, right?

webknjaz:
@bluetech I think "threshold" may mean min/max depending on the PoV/context. So I'd avoid using it.

src/_pytest/runner.py
ssbarnea: `%gs` is valid and used there on purpose, in order to print variable precision, aka human numbers. "1.000" will render as "1" but "1.010" as "1.01".

src/_pytest/runner.py
ssbarnea: pytest did not switch to `require_python>=3.6` yet, but I would support dropping py35.

changelog/7667.feature.rst
nicoddemus: > It's RST, double backticks should be used

You are correct, but we changed our sphinx configuration to consider single ticks as monospaced font, so either way is OK for the docs/changelog entries.

nicoddemus:
> @bluetech I think "threshold" may mean min/max depending on the PoV/context. So I'd avoid using it.

Good point.

@bluetech would you like to take a second look or we can merge this?

src/_pytest/runner.py
ssbarnea: It is not, that is why I added it explicitly. I am not sure why but I know is not printed.

bluetech (approved):
Yes, LGTM.

Regarding the name, a few thoughts, "take it or leave it":

Maybe it's just me but I do find the name `--durations-min` and the explanation not entirely obvious, that is it takes me a moment to understand what it does. For me, the word "minimum" indicates a specific value from a set, not a threshold/limit, but maybe it's just me.

I think a more direct help text would be "Only include durations longer than this value (in seconds)". Which suggests other names for the flag, maybe `--durations-above=0.005`.

Or we can think about the other way and say `--durations-hide=0.005` with a help text "Hide durations below this value (in seconds)". This corresponds nicely to the "20 durations < 0.005s hidden" message.

webknjaz:
Good. I like `--durations-above`. How about `--durations-at-least`?

ssbarnea:
I cannot refrain to mention that two weeks after creating this PR is still not merged, main reason being our inability to agree on naming.

IMHO, it is often worse not to make a decision than making a bad one. In my particular case there is not real pressure to get this new feature merged but I can see how the same kind of issue may deter others from becoming active contributors.

Any core can edit this PR to change the name. I would not mind.

> ...me thinking that maybe someone wants to modify argparse to make it more flexible and allow `--durations-(min|threshold|above|at-least|hide|omg|wtf|...)` ;)

graingert:
If it's only the name, I think we should merge this, then change the name in a follow-up

graingert (approved):

ssbarnea:
Time to pray for someone to press the merge button :D

nicoddemus:
> In my particular case there is not real pressure to get this new feature merged but I can see how the same kind of issue may deter others from becoming active contributors.

I understand how one can feel like that, but we're really trying our best: we work on pytest on our own free time, and time is often short and things come up which sometimes take precedence. Also note that this PR doesn't need to be rushed as it only affects the next minor release anyway, which is something we try to do around monthly.

> Time to pray for someone to press the merge button :D

Indeed, let's get this in. I'm fine with the current name, but please if somebody else prefers to change it open a new PR and it will be welcome! 😁

PR #7685, opened 2020-08-25
config: start migrating Config.{rootdir,inifile} from py.path.local to pathlib

bluetech:
The first commit is a small doc improvement to make the Config info in the API reference more complete.

The second commit adds two new attributes to Config, `rootpath` and `inipath`, which are the `Path` equivalents of `rootdir` and `inifile`, and moves `rootdir` and `inifile` to properties. For now the new attributes are not documented, and the documentation still refers to the old attributes. I can change this if it seems like a good idea.

The third commit converts some code internally from the old attributes to the new. There is always a chance of some subtle change in behavior here; I tried to be careful but we'll have to live and see...

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -820,13 +820,13 @@ class Config:
     :param PytestPluginManager pluginmanager:
 
     :param InvocationParams invocation_params:
-        Object containing the parameters regarding the ``pytest.main``
+        Object containing parameters regarding the :func:`pytest.main`
         invocation.
     """
 
     @attr.s(frozen=True)
     class InvocationParams:
-        """Holds parameters passed during ``pytest.main()``
+        """Holds parameters passed during :func:`pytest.main`.
 
         The object attributes are read-only.
 
@@ -841,11 +841,20 @@ class Config:
         """
 
         args = attr.ib(type=Tuple[str, ...], converter=_args_converter)
-        """Tuple of command-line arguments as passed to ``pytest.main()``."""
+        """The command-line arguments as passed to :func:`pytest.main`.
+
+        :type: Tuple[str, ...]
+        """
         plugins = attr.ib(type=Optional[Sequence[Union[str, _PluggyPlugin]]])
-        """List of extra plugins, might be `None`."""
+        """Extra plugins, might be `None`.
+
+        :type: Optional[Sequence[Union[str, plugin]]]
+        """
         dir = attr.ib(type=Path)
-        """Directory from which ``pytest.main()`` was invoked."""
+        """The directory from which :func:`pytest.main` was invoked.
+
+        :type: pathlib.Path
+        """
 
     def __init__(
         self,
@@ -867,6 +876,10 @@ class Config:
         """
 
         self.invocation_params = invocation_params
+        """The parameters with which pytest was invoked.
+
+        :type: InvocationParams
+        """
 
         _a = FILE_OR_DIR
         self._parser = Parser(
@@ -876,7 +889,7 @@ class Config:
         self.pluginmanager = pluginmanager
         """The plugin manager handles plugin registration and hook invocation.
 
-        :type: PytestPluginManager.
+        :type: PytestPluginManager
         """
 
         self.trace = self.pluginmanager.trace.root.get("config")
@@ -901,7 +914,12 @@ class Config:
 
     @property
     def invocation_dir(self) -> py.path.local:
-        """Backward compatibility."""
+        """The directory from which pytest was invoked.
+
+        Prefer to use :attr:`invocation_params.dir <InvocationParams.dir>`.
+
+        :type: py.path.local
+        """
         return py.path.local(str(self.invocation_params.dir))
 
     def add_cleanup(self, func: Callable[[], None]) -> None:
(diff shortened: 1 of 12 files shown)


src/_pytest/config/__init__.py
```diff
@@ -1029,8 +1045,8 @@ def _initini(self, args: Sequence[str]) -> None:
             rootdir_cmd_arg=ns.rootdir or None,
             config=self,
         )
-        self.rootdir = py.path.local(str(rootpath))
-        self.inifile = py.path.local(str(inipath)) if inipath else None
+        self.rootpath = rootpath
```
nicoddemus, line 1049: While at it, we might decide to also make those read-only properties and hide them behind private attributes `_rootpath` and `_inipath`.

src/_pytest/config/__init__.py
```diff
@@ -898,9 +912,30 @@ def __init__(
...
+        Prefer to use :attr:`invocation_params.dir <InvocationParams.dir>`.
+
+        :type: py.path.local
+        """
         return py.path.local(str(self.invocation_params.dir))
 
+    @property
+    def rootdir(self) -> py.path.local:
+        """The path to the :ref:`rootdir <rootdir>`.
+
```
nicoddemus, line 926: If we do document `rootpath`, then we should mention here that users should use `rootpath` instead.

```suggestion

Prefer to use :attr:`rootpath` instead.
```

src/_pytest/config/__init__.py
```diff
@@ -898,9 +912,30 @@ def __init__(
...
+        :type: py.path.local
+        """
+        return py.path.local(str(self.rootpath))
+
+    @property
+    def inifile(self) -> Optional[py.path.local]:
+        """The path to the :ref:`configfile <configfiles>`.
+
```
nicoddemus, line 934: If we do document `inipath`, then we should mention here that users should use it instead.

```suggestion

Prefer to use :attr:`inipath` instead.
```

bluetech:
Thanks for the review @nicoddemus. I documented the new attributes and added a changelog.

src/_pytest/config/__init__.py
bluetech: I didn't do this one, it feels a bit too heavy handed/boilerplate-y to me ATM. I did think to add a `Final` type annotation but it doesn't work in this case because these attributes are not assigned in the `__init__`, only later.

bluetech:
BTW, I also though to change `inipath` to `configfile` which is the term the documentation now uses, but chose to keep the `ini` to be consistent with all the other places which use the `ini` naming.

src/_pytest/config/__init__.py
nicoddemus: > I didn't do this one, it feels a bit too heavy handed/boilerplate-y to me ATM

TBH I prefer we do this. We don't want users to go around and change those attributes at all, I would rather to get this *right* from the start.

nicoddemus:
> BTW, I also though to change inipath to configfile which is the term the documentation now uses, but chose to keep the ini to be consistent with all the other places which use the ini naming.

That's a good point. Even thought it might be a bit inconsistent, I think you got the right idea here: given we are introducing a new name, let's take the chance to also use the name we find more appropriate in modern pytest (one where we support other file formats other than ini, and where we are migrating to `Path`.

In that light we should probably name it `configpath`.

src/_pytest/config/__init__.py
bluetech: OK, changed to properties.

nicoddemus: Thanks a lot! 🙇

nicoddemus (approved):

src/_pytest/config/__init__.py
```diff
@@ -898,9 +912,55 @@ def __init__(
...
+        """The path to the :ref:`rootdir <rootdir>`.
+
+        Prefer to use :attr:`rootpath`, which is a :class:`pathlib.Path`.
+
+        :type: py.path.local
+        """
+        return py.path.local(str(self.rootpath))
+
+    @property
+    def inipath(self) -> Optional[Path]:
```
nicoddemus, line 945: Btw what do you think about calling this `configpath` as commented here: https://github.com/pytest-dev/pytest/pull/7685/#issuecomment-682480373

bluetech:
Thanks for the review @nicoddemus!

Also updated customizing.rst now to mention the new properties first, with a versionadded note mentioning the older attributes.

> In that light we should probably name it configpath.

I'd like to push back on this a bit more, but if you remain unconvinced let me know.

I think it would create an inconsistency with all of the existing `ini` names -- like the `--override-ini` flag, and APIs like `addini`, `inicfg`, `getini`, and the existing `inifile` name. I think that unless we want to add alternative names for these as well, the consistency and familiarity is more important. And I'm not sure adding alternative names is worth the trouble.

nicoddemus:
> I think it would create an inconsistency with all of the existing ini names -- like the --override-ini flag, and APIs like addini, inicfg, getini, and the existing inifile name.

Ahh you convinced me there. Indeed a bunch of existing names/options exist that use `ini`. Completely agree with you. 👍

nicoddemus:
Feel free to go ahead and merge this yourself, I'm not sure you want to tweak the history first. 👍

Issue #7696, opened 2020-08-27
Issues with PytestDeprecationWarning on direct construction of pytest.File/Item


andy-maier:
Hi. We have implemented a pytest collector quite some time ago that gets testcases from YAML files. I have noticed the following deprecation warnings recently:
```
tests/functiontest/conftest.py:221
  /Users/maiera/PycharmProjects/pywbem/pywbem/tests/functiontest/conftest.py:221:
  PytestDeprecationWarning: direct construction of YamlFile has been deprecated, please use YamlFile.from_parent
    return YamlFile(path, parent)

tests/functiontest/conftest.py:249
  /Users/maiera/PycharmProjects/pywbem/pywbem/tests/functiontest/conftest.py:249:
  PytestDeprecationWarning: direct construction of YamlItem has been deprecated, please use YamlItem.from_parent
    yield YamlItem(tc_name, self, testcase, filepath)
```

The source code for the collection is here: https://github.com/pywbem/pywbem/blob/master/tests/functiontest/conftest.py#L225

I have the following issues with that:

1. The `pytest.File` class is missing completely from the [pytest API Reference docs](https://docs.pytest.org/en/latest/reference.html).

2. The `pytest.Item` class does not list a `from_parent()` method in the [pytest API Reference docs for Item](https://docs.pytest.org/en/latest/reference.html#item).

3. For our `YamlFile`, I was able to map the arguments from the ctor to the `from_parent()` method, but for our `YamlItem` it turns out we are passing the YAML file name as an additional argument to the `YamlItem` ctor (for better error reporting), and I did not find a way to specify that via the `from_parent()` method. Is there some way for the `YamlItem` object to get at the file name of the YAML file? If not, could you possibly allow for additional kwargs that are passed on?

Versions:
```
Python 3.8.5 on MacOs
$ pytest --version
This is pytest version 5.4.3, imported from /Users/maiera/virtualenvs/pywbem38/lib/python3.8/site-packages/pytest/__init__.py
setuptools registered plugins:
  requests-mock-1.8.0 at /Users/maiera/virtualenvs/pywbem38/lib/python3.8/site-packages/requests_mock/contrib/_pytest_plugin.py
  yagot-0.5.0 at /Users/maiera/virtualenvs/pywbem38/lib/python3.8/site-packages/yagot_pytest/plugin.py
[...]

PR #7697, opened 2020-08-27
Add missing File reference to the docs

nicoddemus:
As related in #7696


bluetech (approved):
Current `File` docstring is "Base class for collecting tests from a file.". I suggest throwing a

```
:ref:`non-python tests`
```

in there for a usage explanation

nicoddemus:
Good idea @bluetech, done!

PR #7699, opened 2020-08-28
python: small optimization in PyCollector.collect()

bluetech:
Inline `_makeitem()` so that `self.ihook` (which is moderately expensive) can be called only once.

Note: the removed test "test_makeitem_non_underscore" comes from an old behavior of skipping names that start with `_` which has since been generalized, making the test no longer relevant.


RonnyPfannschmidt (approved):

nicoddemus:
Thanks @bluetech!

PR #7719, opened 2020-09-04
Use tox to execute release script

nicoddemus:
The release-on-comment script is always executed on *master*, so we should
execute the `release.py` script using tox to ensure we create the
right environment.

Also fixed errors in the error handling code.


nicoddemus:
Going ahead and merge this to put it into action.

PR #7721, opened 2020-09-05
Smoke tests for assorted plugins

ssbarnea:
This aims to create a tox environment that tests that installation of
popular plugins does not break pytest.

Related to https://github.com/pytest-dev/pytest-cov/issues/430

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -164,9 +164,10 @@ def async_warn_and_skip(nodeid: str) -> None:
     msg += (
         "You need to install a suitable plugin for your async framework, for example:\n"
     )
+    msg += "  - anyio\n"
     msg += "  - pytest-asyncio\n"
-    msg += "  - pytest-trio\n"
     msg += "  - pytest-tornasync\n"
+    msg += "  - pytest-trio\n"
     msg += "  - pytest-twisted"
     warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))
     skip(msg="async def function and no async plugin installed (see warnings)")
--- a/testing/plugins_integration/pytest_qt_integration.py
+++ b/testing/plugins_integration/pytest_qt_integration.py
@@ -1,5 +1,0 @@
-def test_hello_world(qtbot):
-    import PyQt5.QtWidgets
-
-    w = PyQt5.QtWidgets.QLabel()
-    qtbot.add_widget(w)
--- a/testing/plugins_integration/pytest_qt_integration.py
+++ b/testing/plugins_integration/pytest_qt_integration.py
@@ -1,0 +1,5 @@
+def test_hello_world(qtbot):
+    import PyQt5.QtWidgets
+
+    w = PyQt5.QtWidgets.QLabel()
+    qtbot.add_widget(w)
--- a/tox.ini
+++ b/tox.ini
@@ -122,6 +122,7 @@ deps =
     django
     pyqt5~=5.15
     pytest-asyncio
+    pytest-bdd
     pytest-cov
     pytest-django
     pytest-flakes
@@ -135,6 +136,7 @@ setenv =
     PYTHONPATH=.
 commands =
     pip check
+    pytest bdd_wallet.py
     pytest --cov=. simple_integration.py
     pytest --ds=django_settings simple_integration.py
     pytest --html=simple.html simple_integration.py
--- a/tox.ini
+++ b/tox.ini
@@ -124,6 +124,7 @@ deps =
     pytest-html
     pytest-sugar
     pytest-mock
+    pytest-asyncio
 setenv =
     PYTHONPATH=.
 commands =
@@ -132,6 +133,7 @@ commands =
     pytest --html=simple.html simple_integration.py
     pytest --ds=django_settings simple_integration.py
     pytest pytest_mock_integration.py
+    pytest pytest_asyncio_integration.py
 
 [testenv:py37-freeze]
 changedir = testing/freeze
--- a/tox.ini
+++ b/tox.ini
@@ -13,6 +13,7 @@ envlist =
     pypy3
     py37-{pexpect,xdist,unittestextras,numpy,pluggymaster}
     doctesting
+    plugins
     py37-freeze
     docs
     docs-checklinks
@@ -113,6 +114,21 @@ commands =
     rm -rf /tmp/doc-exec*
     rm -rf {envdir}/.pytest_cache
     make regen
+
+[testenv:plugins]
+changedir = testing/plugins_integration
+deps =
+    django
+    pytest-cov
+    pytest-django
+    pytest-html
+    pytest-sugar
+setenv =
+    PYTHONPATH=.
+commands =
+    pytest --cov=. simple_integration.py
+    pytest --html=simple.html simple_integration.py
+    pytest --ds=django_settings simple_integration.py
 
 [testenv:py37-freeze]
 changedir = testing/freeze
--- a/tox.ini
+++ b/tox.ini
@@ -126,6 +126,7 @@ deps =
     pytest-mock
     pytest-sugar
     pytest-trio
+    anyio[curio,trio]
 setenv =
     PYTHONPATH=.
 commands =
@@ -136,6 +137,7 @@ commands =
     pytest pytest_mock_integration.py
     pytest simple_integration.py --force-sugar
     pytest pytest_trio_integration.py
+    pytest pytest_anyio_integration.py
 
 [testenv:py37-freeze]
 changedir = testing/freeze
--- a/tox.ini
+++ b/tox.ini
@@ -130,6 +130,8 @@ deps =
     pytest-mock
     pytest-sugar
     pytest-trio
+    pytest-twisted
+    twisted
     pytest-xvfb
 setenv =
     PYTHONPATH=.
@@ -143,6 +145,7 @@ commands =
     pytest pytest_asyncio_integration.py
     pytest pytest_mock_integration.py
     pytest pytest_trio_integration.py
+    pytest pytest_twisted_integration.py
     pytest simple_integration.py --force-sugar --flakes
 
 [testenv:py37-freeze]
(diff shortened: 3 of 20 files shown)


nicoddemus:
@ssbarnea I've update the PR to include simple integration tests for `pytest-cov`, `pytest-django`, `pytest-html`, and `pytest-sugar`. I also fixed the issue you identified originally in `pytest-cov`.

We should backport the commit that adds the integration tests back to `6.0.x` after this gets merged.

Fix pytest-dev/pytest-cov#430.

nicoddemus:
I think this is a good starting point, but if others feel some plugin is missing, please speak up!

tox.ini
```diff
@@ -114,6 +115,28 @@ commands =
...
+changedir = testing/plugins_integration
+deps =
+    django
+    pytest-asyncio
+    pytest-cov
+    pytest-django
+    pytest-html
+    pytest-mock
+    pytest-sugar
+    pytest-trio
```
graingert, line 128: thoughts on a `pytest-anyio` test?

nicoddemus: Sounds good, feel free to push a new test to this branch. 👍

agronholm:
You need to ensure that the tests are actually run. Add a failing case with strict `xfail=True` to make sure. I learned this the hard way when I found out AnyIO's hypothesis support had been silently broken the whole time.

src/_pytest/python.py
```diff
@@ -164,9 +164,10 @@ def async_warn_and_skip(nodeid: str) -> None:
     msg += (
         "You need to install a suitable plugin for your async framework, for example:\n"
     )
+    msg += "  - anyio\n"
     msg += "  - pytest-asyncio\n"
-    msg += "  - pytest-trio\n"
     msg += "  - pytest-tornasync\n"
+    msg += "  - pytest-trio\n"
     msg += "  - pytest-twisted"
```
graingert, line 171: I ordered these alphabetically - maybe by github starts or pypi downloads would be better?

ssbarnea:
You may want to look at the PYTEST_REQPASS feature added by pytest-plus which assures that an exact number of tests passed. That was aimed for CI/CD usage for avoiding accidents where tests were skipped.

tox.ini
```diff
@@ -114,6 +115,30 @@ commands =
...
+    PYTHONPATH=.
+commands =
+    pytest pytest_asyncio_integration.py
+    pytest --cov=. simple_integration.py
+    pytest --ds=django_settings simple_integration.py
+    pytest --html=simple.html simple_integration.py
+    pytest pytest_mock_integration.py
+    pytest simple_integration.py --force-sugar
+    pytest pytest_trio_integration.py
+    pytest pytest_anyio_integration.py
```
graingert, line 140: I think we should test all the async frameworks simultaneously - they have all agreed to use marks to allow them to all interoperate in a single test run, so we should test that here

tox.ini
```diff
@@ -114,6 +115,30 @@ commands =
...
+deps =
+    django
+    pytest-asyncio
+    pytest-cov
+    pytest-django
+    pytest-html
+    pytest-mock
+    pytest-sugar
+    pytest-trio
+    anyio[curio,trio]
```
graingert, line 129: @altendky pytest-twisted ?

tox.ini
agronholm: At least as long as you don't use async fixtures. `pytest-asyncio` is known to appropriate those regardless of marks (see https://github.com/pytest-dev/pytest-asyncio/issues/124 for the issue).

src/_pytest/python.py
hugovk: Alphabetic is best for when looking at it later, for example, to check if a new one is already present or not.

Also no need to look up stars or downloads from elsewhere.

graingert:
@ssbarnea what's the scope of this you're happy to accept? I think as plugin maintainers become aware of this PR more will want included.

I don't want to see a situation where this scope creeps all the way to testing all of pypi before it's merged! So I think we should merge this sooner and rather than later so maintainers can add their own smoke tests as PRs directly to pytest, rather than this branch?

tox.ini
graingert: > At least as long as you don't use async fixtures. 

That's a very good point, we currently don't test any async fixtures here.

graingert:
> installation of popular plugins does not break pytest.

how popular is popular? Can we nail down a criteria for inclusion?

src/_pytest/python.py
altendky: Alphabetical seems good, even if it leaves 'me' at the end. `:]`

altendky:
> You may want to look at the PYTEST_REQPASS feature added by pytest-plus which assures that an exact number of tests passed. That was aimed for CI/CD usage for avoiding accidents where tests were skipped.

Is that better than just requiring coverage in your tests?

tox.ini
altendky: May as well alpha-sort here too?  Also `pytest-qt`.

altendky (commented):
The stated purpose is to test that pytest doesn't break.  Shouldn't the full pytest suite be run instead of just a trivial case for each plugin?

ssbarnea:
xdist and cov are ok dor start. good point about avoiding scope creep. we can always add more later.

tox.ini
```diff
@@ -114,6 +115,30 @@ commands =
...
+setenv =
+    PYTHONPATH=.
+commands =
```
graingert, line 132:
```suggestion
commands =
    [ $(pip list -o | wc -l) -eq 2 ]
```

I think we want something like this to make sure there's no version constraints holding us back from the latest releases

/cc @McSinyx

McSinyx: As I mentioned [when I suggested this on IRC](http://kafka.dcpython.org/day/pypa/2020-09-13#15.06.54.McSinyx[m]), it's quite of a hack.  I'm not sure how stable pip's output will be but I think at least adding `--quiet` would make it safer.

tox.ini
```diff
@@ -114,6 +115,30 @@ commands =
     rm -rf {envdir}/.pytest_cache
     make regen
 
+[testenv:plugins]
```
graingert, line 118: how about?

```suggestion
[testenv:plugins]
pip_pre=true
```
https://tox.readthedocs.io/en/latest/config.html#conf-pip_pre

nicoddemus:
> The stated purpose is to test that pytest doesn't break. Shouldn't the full pytest suite be run instead of just a trivial case for each plugin?

This makes sense, but I would rather start things slowly: 

1. Plugins might interact with pytest differently in a way that breaks the pytest suite, but are not wrong: for example `pytest-sugar` changes output, and several tests check output and would break, requiring us to special case tests, or choosing a subset.
2. We can always improve things later the more we get the feeling how this will work. Having it being a quick sanity check run is beneficial, and reliable.

> I don't want to see a situation where this scope creeps all the way to testing all of pypi before it's merged! So I think we should merge this sooner and rather than later so maintainers can add their own smoke tests as PRs directly to pytest, rather than this branch?

Agreed! This is already a large improvement over what we have today, so let's get this in as soon as we feel this is good enough.

tox.ini
ssbarnea: Just run `pip check` and it will fail if any conflicts are detected.

tox.ini
nicoddemus: To test pre-release of plugins you mean? I don't see many plugins doing that (never seen any TBH). You think it is worth it?

nicoddemus:
Add 3 more: `pytest-flakes`, `pytest-qt` and `pytest-bdd`. Also added `pip check` as suggested.

nicoddemus:
Hmm as I feared: pytest-qt, besides depending on a large library, also requires an X server. While it is possible, I don't want to introduce complicated setups at this stage.

tox.ini
altendky: What's the cost that could make it not worth it?  I don't pre-release in most cases but I feel like the world might be better if we all did and we call checked against those.  Maybe.

graingert: Yeah I think it's worth it, would encourage pre-releases even if none are made now

nicoddemus: Fair enough, thanks! Done.

nicoddemus:
Anything else? If not it would be nice we could get some approvals here and get this merged. 👍

hugovk (approved):
This is a great idea! Looks good. I agree on doing light testing of each plugin, this is a quick smoke test* to check nothing exploded. Should catch the major things, and plugins are still encouraged to test against pytest pre-releases.

(*[Etymology](https://stackoverflow.com/a/745279/724176):  'The term originates in hardware repair and has been applied to software. It's intended to be a quick test to see if the application "catches on fire" when run for the first time.')

hugovk:
Oh, one thing to keep in mind: it's possible a new plugin release will break current master, and a new PR will start to fail because of the plugin release, and not the PR changes.

In such cases, it's a good idea to check if a master build also fails in the same way.

It may be needed to temporarily skip smoke testing that plugin to prevent all PRs failing, until either pytest or the plugin has a fix (because it could be a problem on either side, or both!). 

But hopefully this would be rare. And in any case, it's good information to know there's a failure.

testing/plugins_integration/pytest_qt_integration.py
```diff
@@ -0,0 +1,5 @@
+def test_hello_world(qtbot):
```
altendky, line 1: Wasn't this supposed to get removed?

hugovk: Yep, missed in https://github.com/pytest-dev/pytest/pull/7721/commits/780752b25f976c539f79175b29c43d2ccbc9591e.

bluetech (approved):
As the person doing most of the breaking, this will be very helpful!

graingert (approved):

testing/plugins_integration/pytest_qt_integration.py
nicoddemus: 🤦

graingert: Makes me think we should be checking for 100% coverage for the smoke tests

nicoddemus:
Thanks everyone!

nicoddemus:
Of course everyone feel free to add more plugins that they consider relevant. 👍

PR #7733, opened 2020-09-08
Add example for __test__

JosephTLucas:
PR reference: https://twitter.com/pytestdotorg/status/1302444196842725376?s=20

This example page seemed to be the best location.  Anywhere else you recommend?

### DIFF
--- a/doc/en/example/pythoncollection.rst
+++ b/doc/en/example/pythoncollection.rst
@@ -314,7 +314,7 @@ interpreter:
     if sys.version_info[0] > 2:
         collect_ignore_glob = ["*_py2.py"]
 
-Since Pytest 2.6, users can prevent pytest from discovering classes that start 
+Since Pytest 2.6, users can prevent pytest from discovering classes that start
 with ``Test`` by setting a boolean ``__test__`` attribute to ``False``.
 
 .. code-block:: python
--- a/doc/en/example/pythoncollection.rst
+++ b/doc/en/example/pythoncollection.rst
@@ -314,16 +314,11 @@ interpreter:
     if sys.version_info[0] > 2:
         collect_ignore_glob = ["*_py2.py"]
 
-User modules with names matching ``test*.py`` may be discovered as tests. Since 
-Pytest 2.6, users can prevent this test discovery by setting a boolean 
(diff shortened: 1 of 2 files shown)


JosephTLucas:
Initially, I'd added my name to AUTHORS based on the PR prompt, but after reading the CONTRIBUTING guidelines in more detail, decided it wasn't warranted.  I also cut the documentation down to 80 chars per line.  For my development, is there something like `black` but for `.rst` files?

doc/en/example/pythoncollection.rst
```diff
@@ -313,3 +313,17 @@ interpreter:
...
         collect_ignore_glob = ["*_py2.py"]
+
+User modules with names matching ``test*.py`` may be discovered as tests. Since 
+Pytest 2.6, users can prevent this test discovery by setting a boolean 
+``__test__`` attribute.
+
+.. code-block:: python
+
+    # Will be discovered as a test
+    def test_class:
```
nicoddemus, line 324:
```suggestion
    class TestClass:
```

doc/en/example/pythoncollection.rst
```diff
@@ -313,3 +313,17 @@ interpreter:
     collect_ignore = ["setup.py"]
     if sys.version_info[0] > 2:
         collect_ignore_glob = ["*_py2.py"]
+
+User modules with names matching ``test*.py`` may be discovered as tests. Since 
+Pytest 2.6, users can prevent this test discovery by setting a boolean
```
nicoddemus, line 318:
```suggestion
Pytest 2.6, users can prevent pytest from discovering classes that start with ``Test`` by setting a boolean ``__test__`` attribute to ``False``.
```

Also I think we only need the example with `False` set; otherwise the example using `__test__ = True` implies that pytest will discover the class for tests even if it doesn't match the `python_classes` configuration option, which is not true.

nicoddemus (changes requested):
Thanks @JosephTLucas, please take a look at my comments. 👍

josephlucas-arc:
Thanks @nicoddemus !  Changes made.

nicoddemus:
Thanks @JosephTLucas!

PR #7734, opened 2020-09-08
Update pytest-dev requirements

asottile:
- be more vague about "packaging metadata" over explicitly mentioning `setup.py`
  (such that `pyproject.toml`-based distributions are allowed)
- drop extensions on `README.txt` / `LICENSE.txt` (it's more common to have `.md`
  / `.rst` / no extension)
- remove duplicate mention of license packaging metadata


nicoddemus (approved):
Thanks @asottile!

PR #7736, opened 2020-09-10
Allow ovewriting a parametrized fixture while reusing the parent fixture's value

### LINKED ISSUE #1953: Extending a fixture does not keep its params

When overriding a fixture, pytest allows use of the "super" fixture (http://docs.pytest.org/en/latest/fixture.html#override-a-fixture-on-a-folder-conftest-level). However, using such a fixture seems to not work the same as using other fixtures.

If I have a parametrized fixture `foo`, and another fixture `bar` that depends on `foo`, then `bar` will be parametrized on `foo`. And if there is a `test_bar` that uses `bar`, then `test_bar` will be run once for each of the params of `foo`.

Suppose there are two `bar` fixtures, and one extends the other. At test collection time, it seems that `test_bar` does not know it is supposed to be parametrized off the params of `foo`. But at test running time, pytest figures out that there is supposed to be parameterization, and fails itself.

I'm not sure if this is a missing feature, or a bug.

If this is intentional, is there any way to have the `bar` override automatically bring in the dependencies of the super fixture? Right now, the only workaround I can figure out is to do:

``` python
@pytest.fixture
def bar(bar, foo):
    ...
```

even if the new `bar` doesn't directly use `foo`.

Also, I don't know of any workaround if you extend a fixture that has params, e.g. if you want to extend `foo` based off the super fixture.
- [x] `pip list` of the virtual environment you are using

> $ .tox/py35/bin/pip list
> hypothesis (3.4.2)
> mock (2.0.0)
> nose (1.3.7)
> pbr (1.10.0)
> pip (8.1.2)
> py (1.4.31)
> pytest (3.0.3.dev0, /Users/jmoldow/repos/pytest)
> requests (2.11.1)
> setuptools (27.2.0)
> six (1.10.0)
> wheel (0.30.0a0)

pytest 3.0.2 and pytest 3.0.3.dev0 (https://github.com/pytest-dev/pytest/commit/7660a19d0a39adb9724b9acba05e64acf3fcc5ac).

OS X Yosemite Version 10.10.5

In `testing/python/fixture.py`, add the tests

``` python
    def test_extend_parametrized_fixture(self, testdir):
        testdir.makeconftest("""
            import pytest

            @pytest.fixture(params=[1])
            def foo(request):
                return request.param
        """)
        testfile = testdir.makepyfile("""
            import pytest

            @pytest.fixture
            def foo(foo):
                return foo * 2

            def test_spam(foo):
                assert foo == 2
        """)
        result = testdir.runpytest(testfile)
        result.stdout.fnmatch_lines(["*1 passed*"])

    def test_extend_fixture_with_parametrized_dependency(self, testdir):
        testdir.makeconftest("""
            import pytest

            @pytest.fixture(params=[1])
            def foo(request):
                return request.param

            @pytest.fixture
            def bar(foo):
                return foo
        """)
        testfile = testdir.makepyfile("""
            import pytest

            @pytest.fixture
            def bar(bar):
                return bar * 2

            def test_spam(bar):
                assert bar == 2
        """)
        result = testdir.runpytest(testfile)
        result.stdout.fnmatch_lines(["*1 passed*"])
```

The output from those tests is attached.
[captured_output.txt](https://github.com/pytest-dev/pytest/files/483332/captured_output.txt)

If you manually re-specify the `foo` dependency, then this test passes:

``` python
    def test_extend_fixture_with_parametrized_dependency(self, testdir):
        testdir.makeconftest("""
            import pytest

            @pytest.fixture(params=[1])
            def foo(request):
                return request.param

            @pytest.fixture
            def bar(foo):
                return foo
        """)
        testfile = testdir.makepyfile("""
            import pytest

            @pytest.fixture
            def bar(bar, foo):
                return bar * 2

            def test_spam(bar):
                assert bar == 2
        """)
        result = testdir.runpytest(testfile)
        result.stdout.fnmatch_lines(["*1 passed*"])
```

But I don't know of any way to fix the `test_extend_parametrized_fixture` test.

nicoddemus:
Fix #1953

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -47,6 +47,7 @@ from _pytest.config import _PluggyPlugin
 from _pytest.config import Config
 from _pytest.config.argparsing import Parser
 from _pytest.deprecated import FILLFUNCARGS
+from _pytest.mark import Mark
 from _pytest.mark import ParameterSet
 from _pytest.outcomes import fail
 from _pytest.outcomes import TEST_OUTCOME
@@ -1529,34 +1530,56 @@ class FixtureManager:
         return initialnames, fixturenames_closure, arg2fixturedefs
 
     def pytest_generate_tests(self, metafunc: "Metafunc") -> None:
-        for argname in metafunc.fixturenames:
-            faclist = metafunc._arg2fixturedefs.get(argname)
-            if faclist:
-                fixturedef = faclist[-1]
-                if fixturedef.params is not None:
-                    markers = list(metafunc.definition.iter_markers("parametrize"))
-                    for parametrize_mark in markers:
-                        if "argnames" in parametrize_mark.kwargs:
-                            argnames = parametrize_mark.kwargs["argnames"]
-                        else:
-                            argnames = parametrize_mark.args[0]
+        """Generate new tests based on parametrized fixtures used by the given metafunc"""
 
-                        if not isinstance(argnames, (tuple, list)):
-                            argnames = [
-                                x.strip() for x in argnames.split(",") if x.strip()
-                            ]
-                        if argname in argnames:
-                            break
-                    else:
-                        metafunc.parametrize(
-                            argname,
-                            fixturedef.params,
-                            indirect=True,
-                            scope=fixturedef.scope,
-                            ids=fixturedef.ids,
-                        )
+        def get_parametrize_mark_argnames(mark: Mark) -> Sequence[str]:
+            if "argnames" in mark.kwargs:
+                argnames = mark.kwargs[
+                    "argnames"
+                ]  # type: Union[str, Tuple[str, ...], List[str]]
             else:
-                continue  # Will raise FixtureLookupError at setup time.
+                argnames = mark.args[0]
+            if not isinstance(argnames, (tuple, list)):
+                argnames = [x.strip() for x in argnames.split(",") if x.strip()]
+            return argnames
+
+        for argname in metafunc.fixturenames:
+            # Get the FixtureDefs for the argname.
+            fixture_defs = metafunc._arg2fixturedefs.get(argname)
+            if not fixture_defs:
+                # Will raise FixtureLookupError at setup time if not parametrized somewhere
+                # else (e.g @pytest.mark.parametrize)
+                continue
+
+            # If the test itself parametrizes using this argname, give it
+            # precedence.
+            if any(
+                argname in get_parametrize_mark_argnames(mark)
+                for mark in metafunc.definition.iter_markers("parametrize")
+            ):
+                continue
+
+            # In the common case we only look at the fixture def with the
+            # closest scope (last in the list). But if the fixture overrides
+            # another fixture, while requesting the super fixture, keep going
+            # in case the super fixture is parametrized (#1953).
+            for fixturedef in reversed(fixture_defs):
+                # Fixture is parametrized, apply it and stop.
+                if fixturedef.params is not None:
+                    metafunc.parametrize(
+                        argname,
+                        fixturedef.params,
+                        indirect=True,
+                        scope=fixturedef.scope,
+                        ids=fixturedef.ids,
+                    )
+                    break
+
+                # Not requesting the overridden super fixture, stop.
+                if argname not in fixturedef.argnames:
+                    break
+
+                # Try next super fixture, if any.
 
     def pytest_collection_modifyitems(self, items: "List[nodes.Item]") -> None:
         # Separate parametrized setups.
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -1533,15 +1533,8 @@ class FixtureManager:
         """Generate new tests based on parametrized fixtures used by the given metafunc"""
 
         def get_parametrize_mark_argnames(mark: Mark) -> Sequence[str]:
-            if "argnames" in mark.kwargs:
-                argnames = mark.kwargs[
-                    "argnames"
-                ]  # type: Union[str, Tuple[str, ...], List[str]]
-            else:
-                argnames = mark.args[0]
-            if not isinstance(argnames, (tuple, list)):
-                argnames = [x.strip() for x in argnames.split(",") if x.strip()]
-            return argnames
+            args, _ = ParameterSet._parse_parametrize_args(*mark.args, **mark.kwargs)
+            return args
 
         for argname in metafunc.fixturenames:
             # Get the FixtureDefs for the argname.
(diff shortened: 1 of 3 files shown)


bluetech (changes requested):
The fix has a flaw in that if `metafunc.parametrize` happens, it still tries the super fixture, which leads to the parametrization possibly being attempted again. You can see it by changing the overriding fixture (in `test_foo.py`) to be parametrized itself, e.g. `@pytest.fixture(params=[10, 20])`. Then it crashes with `ValueError: duplicate 'foo'`.

The fix I think is just to stop when parametrizing. I would add the above case as an additional test either way.

BTW, when reviewing I refactored the function a bit to be more understandable (to me). You can apply it, if you think it's better of course. Based on master and contains the fix.

```diff
diff --git a/src/_pytest/fixtures.py b/src/_pytest/fixtures.py
index bf77d09f1..36cf15899 100644
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -47,6 +47,7 @@ from _pytest.config import _PluggyPlugin
 from _pytest.config import Config
 from _pytest.config.argparsing import Parser
 from _pytest.deprecated import FILLFUNCARGS
+from _pytest.mark import Mark
 from _pytest.mark import ParameterSet
 from _pytest.outcomes import fail
 from _pytest.outcomes import TEST_OUTCOME
@@ -1529,34 +1530,53 @@ class FixtureManager:
         return initialnames, fixturenames_closure, arg2fixturedefs
 
     def pytest_generate_tests(self, metafunc: "Metafunc") -> None:
+        def get_parametrize_mark_argnames(mark: Mark) -> Sequence[str]:
+            if "argnames" in mark.kwargs:
+                argnames = mark.kwargs[
+                    "argnames"
+                ]  # type: Union[str, Tuple[str, ...], List[str]]
+            else:
+                argnames = mark.args[0]
+            if not isinstance(argnames, (tuple, list)):
+                argnames = [x.strip() for x in argnames.split(",") if x.strip()]
+            return argnames
+
         for argname in metafunc.fixturenames:
-            faclist = metafunc._arg2fixturedefs.get(argname)
-            if faclist:
-                fixturedef = faclist[-1]
+            # Get the FixtureDefs for the argname.
+            fixture_defs = metafunc._arg2fixturedefs.get(argname)
+            if not fixture_defs:
+                # Will raise FixtureLookupError at setup time.
+                continue
+
+            # The test itself parametrizes using this argname, give it
+            # precedence.
+            if any(
+                argname in get_parametrize_mark_argnames(mark)
+                for mark in metafunc.definition.iter_markers("parametrize")
+            ):
+                break
+
+            # In the common case we only look at the fixture def with the
+            # closest scope (last in the list). But if the fixture overrides
+            # another fixture, while requesting the super fixture, keep going
+            # in case the super fixture is parametrized (#1953).
+            for fixturedef in reversed(fixture_defs):
+                # Fixture is parametrized, apply it and stop.
                 if fixturedef.params is not None:
-                    markers = list(metafunc.definition.iter_markers("parametrize"))
-                    for parametrize_mark in markers:
-                        if "argnames" in parametrize_mark.kwargs:
-                            argnames = parametrize_mark.kwargs["argnames"]
-                        else:
-                            argnames = parametrize_mark.args[0]
+                    metafunc.parametrize(
+                        argname,
+                        fixturedef.params,
+                        indirect=True,
+                        scope=fixturedef.scope,
+                        ids=fixturedef.ids,
+                    )
+                    break
 
-                        if not isinstance(argnames, (tuple, list)):
-                            argnames = [
-                                x.strip() for x in argnames.split(",") if x.strip()
-                            ]
-                        if argname in argnames:
-                            break
-                    else:
-                        metafunc.parametrize(
-                            argname,
-                            fixturedef.params,
-                            indirect=True,
-                            scope=fixturedef.scope,
-                            ids=fixturedef.ids,
-                        )
-            else:
-                continue  # Will raise FixtureLookupError at setup time.
+                # Not requesting the overridden super fixture, stop.
+                if argname not in fixturedef.argnames:
+                    break
+
+                # Try super fixture, if any.
 
     def pytest_collection_modifyitems(self, items: "List[nodes.Item]") -> None:
         # Separate parametrized setups.
```

bluetech:
I think the changes would be too risky for a backport, so I prefer we don't backport it.

nicoddemus:
> The fix has a flaw in that if metafunc.parametrize happens, it still tries the super fixture, which leads to the parametrization possibly being attempted again. You can see it by changing the overriding fixture (in test_foo.py) to be parametrized itself, e.g. @pytest.fixture(params=[10, 20]). Then it crashes with ValueError: duplicate 'foo'.

Ahh good catch, thanks! I will apply your improvements and add a new test (it really amazes me that, besides the huge number of test cases and scenarios we have, we still sometimes have gaps on them).

> BTW, when reviewing I refactored the function a bit to be more understandable (to me). You can apply it, if you think it's better of course. Based on master and contains the fix.

Much clearer, thanks, I will apply them. 👍 

> I think the changes would be too risky for a backport, so I prefer we don't backport it.

Fair enough. 👍

src/_pytest/fixtures.py
```diff
@@ -1529,34 +1530,56 @@ def sort_by_scope(arg_name: str) -> int:
...
+                # else (e.g @pytest.mark.parametrize)
+                continue
+
+            # The test itself parametrizes using this argname, give it
+            # precedence.
+            if any(
+                argname in get_parametrize_mark_argnames(mark)
+                for mark in metafunc.definition.iter_markers("parametrize")
+            ):
+                continue
```
nicoddemus, line 1560: Noticed this should be `continue` instead of `break`, otherwise we would skip evaluating the rest of the argnames for possible fixture parametrization (`test_override_parametrize_fixture_and_indirect` demonstrates this).

nicoddemus:
Applied the changes and tests as requested, thanks again for the great review @bluetech!

src/_pytest/fixtures.py
```diff
@@ -1529,34 +1530,49 @@ def sort_by_scope(arg_name: str) -> int:
         return initialnames, fixturenames_closure, arg2fixturedefs
 
     def pytest_generate_tests(self, metafunc: "Metafunc") -> None:
+        """Generate new tests based on parametrized fixtures used by the given metafunc"""
+
+        def get_parametrize_mark_argnames(mark: Mark) -> Sequence[str]:
```
bluetech, line 1535: We might as well inline it, unless the line becomes too ling.

bluetech (approved):
Great, LGTM!

src/_pytest/fixtures.py
nicoddemus: I would, but it is clunky because `ParameterSet._parse_parametrize_args` returns a tuple, and we are only interested in the first one:

```python
if any(
    argname in ParameterSet._parse_parametrize_args(*mark.args, **mark.kwargs)[0]
    for mark in metafunc.definition.iter_markers("parametrize")
):
```

PR #7739, opened 2020-09-10
Use multiple issue template types and mention Discussions

nicoddemus:


gnikonorov (approved):
This is awesome! Thanks @nicoddemus

asottile (approved):

The-Compiler:
Wouldn't it make more sense to use GitHub's issue type selectors to point to discussions? Here's an example:

- https://github.com/qutebrowser/qutebrowser/issues/new/choose
- https://github.com/qutebrowser/qutebrowser/tree/master/.github/ISSUE_TEMPLATE

I'm -0.5 on this, as it doesn't seem to be the right mechanism and will lead to some people most likely just leaving this text inside their issues instead of deleting it.

nicoddemus:
> Wouldn't it make more sense to use GitHub's issue type selectors to point to discussions?

Oh yeah I forgot a about that feature. I agree it is both much nicer as well as points people in the general direction!

Would you like to tackle that?

nicoddemus:
@The-Compiler done, thanks!

asottile (approved):

The-Compiler:
Looks great, thanks @nicoddemus and sorry for the delay!

PR #7747, opened 2020-09-12
Add full command-line flags to the reference docs

### LINKED ISSUE #7728: document strict-config and other command line flags
I was recently asked to explain why I was adding `--strict-config` to a pytest ini, "as it appears to be undocumented"

currently command line flags are only documented in the changelog and https://docs.pytest.org/en/stable/_modules/_pytest/main.html

I think we should automatically add the output of `pytest --help` to the docs, eg here:
https://docs.pytest.org/en/stable/usage.html#getting-help-on-version-option-names-environment-variables

nicoddemus:
Fix #7728

### DIFF
--- a/doc/en/reference.rst
+++ b/doc/en/reference.rst
@@ -1652,3 +1652,296 @@ passed multiple times. The expected format is ``name=value``. For example::
 
         [pytest]
         xfail_strict = True
+
+
+.. _`command-line-flags`:
+
+Command-line Flags
+------------------
+
+All the command-line flags can be obtained by running ``pytest --help``::
+
+    $ pytest --help
+    usage: pytest [options] [file_or_dir] [file_or_dir] [...]
+
+    positional arguments:
+      file_or_dir
+
+    general:
+      -k EXPRESSION         only run tests which match the given substring
+                            expression. An expression is a python evaluatable
+                            expression where all names are substring-matched
+                            against test names and their parent classes.
+                            Example: -k 'test_method or test_other' matches all
+                            test functions and classes whose name contains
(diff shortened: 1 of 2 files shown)


doc/en/reference.rst
```diff
@@ -1652,3 +1652,306 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+
+.. _`command-line-flags`:
+
+Command-line Flags
+------------------
+
+All the command-line flags can be obtained by running ``pytest --help``::
+
+    $ pytest --help
+    usage: pytest [options] [file_or_dir] [file_or_dir] [...]
```
nicoddemus, line 1665: This line an below were generated by `regendoc`.

doc/en/reference.rst
```diff
@@ -1652,3 +1652,306 @@ passed multiple times. The expected format is ``name=value``. For example::
...
+                            log file logging level.
+      --log-file-format=LOG_FILE_FORMAT
+                            log format as used by the logging module.
+      --log-file-date-format=LOG_FILE_DATE_FORMAT
+                            log date format as used by the logging module.
+      --log-auto-indent=LOG_AUTO_INDENT
+                            Auto-indent multiline messages passed to the logging
+                            module. Accepts true|on, false|off or an integer.
+
+    Hypothesis:
```
asottile, line 1858: this includes our plugins it looks like so we probably need to filter those out somehow?

nicoddemus: Ahh my bad. I ran regendoc manually. I will update this.

asottile (approved):

PR #7750, opened 2020-09-12
permit tox config in non-tox.ini files for pytest-dev member projects

graingert:

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -151,7 +151,8 @@ the following:
 - PyPI presence with packaging metadata that contains a ``pytest-``
   prefixed name, version number, authors, short and long description.
 
-- a ``tox.ini`` for running tests using `tox <https://tox.readthedocs.io>`_.
+- a  `tox configuration <https://tox.readthedocs.io/en/latest/config.html#configuration-discovery>`_
+  for running tests using `tox <https://tox.readthedocs.io>`_.
 
 - a ``README`` describing how to use the plugin and on which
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -151,7 +151,8 @@ the following:
 - PyPI presence with packaging metadata that contains a ``pytest-``
   prefixed name, version number, authors, short and long description.
 
-- a ``tox.ini`` for running tests using `tox <https://tox.readthedocs.io>`_.
+- a  `tox configuration <https://tox.readthedocs.io/en/latest/config.html#configuration-discovery>`_
+  for running tests using `tox <https://tox.readthedocs.io>`_.
```
graingert, line 155: we should probably allow tox,nox etc etc. Otherwise are people expected to create a shim tox file that runs their preferred test run?

asottile (approved):

CONTRIBUTING.rst
The-Compiler: I disagree, I think it's a good idea to require tox so that people don't need to find out what meta-testrunner they need to run for a given pytest project - after all, that's the whole point of tox.

nicoddemus: I disagree with @The-Compiler and @asottile here: I believe we should allow automated test runners similar to `tox` to be used, like `nox`. I wouldn't like to bar a project from joining the organization if they fulfill all requirements but prefer to use `nox`.

PR #7760, opened 2020-09-16
[Docs] remove semi-colon punctuation mark

finn0:
@gnikonorov now can you take a look into this PR? I've made suggested change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,21 +77,21 @@ Due to ``pytest``'s detailed assertion introspection, only plain ``assert`` stat
 Features
 --------
 
-- Detailed info on failing `assert statements <https://docs.pytest.org/en/stable/assert.html>`_ (no need to remember ``self.assert*`` names);
+- Detailed info on failing `assert statements <https://docs.pytest.org/en/stable/assert.html>`_ (no need to remember ``self.assert*`` names)
 
 - `Auto-discovery
   <https://docs.pytest.org/en/stable/goodpractices.html#python-test-discovery>`_
-  of test modules and functions;
+  of test modules and functions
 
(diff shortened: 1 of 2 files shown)


gnikonorov (changes requested):
Thank you for the PR @finn0!

You are right, semicolons are probably not the best punctuation to use here. Can you please remove the punctuation altogether.

The-Compiler (approved):
Looks good, thanks! :+1:

finn0:
Usually, we name `doc` folder as `docs`. Isn't it a good idea, we also rename `doc` folder to `docs`?

The-Compiler:
> Usually, we name `doc` folder as `docs`. Isn't it a good idea, we also rename `doc` folder to `docs`?

I've seen both. Various projects I'm involved in ([qutebrowser](https://github.com/qutebrowser/qutebrowser), [herbstluftwm](https://github.com/herbstluftwm/herbstluftwm/)) or use ([pylint](https://github.com/pycqa/pylint), [pygments](https://github.com/pygments/pygments), [coverage.py](https://github.com/nedbat/coveragepy), ...) use "doc" as well. Thus I'm -1 on that change, as it will introduce friction for no clear benefit :slightly_smiling_face:

gnikonorov:
> Usually, we name `doc` folder as `docs`. Isn't it a good idea, we also rename `doc` folder to `docs`?

My opinion is similar to @The-Compiler's above. I think that given convention `doc` is a valid folder name

gnikonorov (approved):
Thank you @finn0!

finn0:
@The-Compiler @gnikonorov Thanks for the reply, and example project which uses `doc` as folder name.

nicoddemus:
Thanks @finn0!

PR #7772, opened 2020-09-19
Add docs about reusing fixtures from other projects

### LINKED ISSUE #7754: Documentation on plugins should explain why importing them is bad
Intuitively, I re-used a fixture not by making it a plugin but by importing it from a library I created which is shared by several projects. This creates flake8 error F811 and caused me to read up why this should not be OK. This was not a quick journey and didn't lead to clear answers.

The problem I'm solving: I have a test library which contains a fixture for setting up database test data. It's shared by several projects: this means it can't be defined in a project's conftest.py (other than by copy-paste). The fixture is used to create project-specific fixtures.

This is the same case as nicely documented here:

https://mail.python.org/pipermail/code-quality/2017-May/000901.html

or here: 

https://stackoverflow.com/questions/43746413/how-do-i-get-flake8-to-work-with-f811-errors

Importing fixtures like this works. I haven't yet encountered any disadvantage. However after a bit of googling, reading and asking I found that this is not how one's supposed to do this but no explanation of why.

I'd like an update to the documentation here:

https://docs.pytest.org/en/latest/writing_plugins.html

It would be good if search terms such as "reuse pytest fixtures between projects or from a library" took me here.

It would be good if the documentation explained why re-using fixtures by importing from a pip installed library is a bad thing and if it really is a bad thing, perhaps the code should give some sort of warning if they're used that way.

It would be nice if the help text gave a direct answer to 'how do I write a pluging fixture' rather than having to trawl through the linked examples.

nicoddemus:
Fix #7754

### DIFF
--- a/doc/en/fixture.rst
+++ b/doc/en/fixture.rst
@@ -1136,8 +1136,8 @@ and teared down after every test that used it.
 
 .. _`usefixtures`:
 
-Using fixtures from classes, modules or projects
-----------------------------------------------------------------------
+Use fixtures implicitly with @pytest.mark.usefixtures
+-----------------------------------------------------
 
 .. regendoc:wipe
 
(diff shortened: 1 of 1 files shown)


doc/en/fixture.rst
```diff
@@ -1531,3 +1531,37 @@ Given the tests file structure is:
 In the example above, a parametrized fixture is overridden with a non-parametrized version, and
 a non-parametrized fixture is overridden with a parametrized version for certain test module.
 The same applies for the test folder level obviously.
+
+
+Using fixtures from other projects
+----------------------------------
+
+Usually projects that provide pytest support will use :ref:`entry points <setuptools entry points>`,
+so just by installing those projects into an environment will make those fixtures available for use.
```
bluetech, line 1540:
```suggestion
so just installing those projects into an environment will make those fixtures available for use.
```

doc/en/fixture.rst
```diff
@@ -1136,8 +1136,8 @@ and teared down after every test that used it.
 
 .. _`usefixtures`:
 
-Using fixtures from classes, modules or projects
-----------------------------------------------------------------------
+Use fixtures implicitly with @pytest.mark.usefixtures
```
bluetech, line 1139: I kinda prefer the old title here:

- The old one more directly states the main use case
- Not sure about the word "implicitly", I wouldn't say it's exactly implicit, and also has some overlap with autouse.

I think it'd be good to add the "with usefixtures" part though. However I'd omit the `@pytest.mark` part because this section also discusses the `usefixtures` ini option.

bluetech (approved):
LGTM with one comment

doc/en/fixture.rst
nicoddemus: Hmmm the old title read ambiguously to me, it could also meant "Using fixtures *defined* in other classes..."

But I agree with your points of course.

How about "Use fixtures in classes and modules with usefixtures"?

bluetech: :+1:

nicoddemus: Done, thanks

mgrazebrook:
I spent hours trying to work out the right way to use fixtures in a library, ended up with a wrong (but working) solution but at least knowing how I should have done it. Not only do you confirm what I learned, you also show me a better way. A real improvement, thanks.

PR #7780, opened 2020-09-21
Mark some public and to-be-public classes as `@final`

bluetech:
Some work toward #7469.

This ([ref](https://docs.python.org/3/library/typing.html#typing.final)) indicates at least for people using type checkers that these classes are not designed for inheritance and we make no stability guarantees when inheriting from them.

Currently this doesn't show up in the docs. Sphinx does actually support `@final`, however it only works when imported directly from `typing`, while we import from `_pytest.compat`.

In the future there might also be a `@sealed` decorator which would cover some more cases.

### DIFF
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -19,7 +19,6 @@ from typing import Union
 
 import attr
 
-from _pytest._io.saferepr import saferepr
 from _pytest.outcomes import fail
 from _pytest.outcomes import TEST_OUTCOME
 
@@ -297,6 +296,8 @@ def get_real_func(obj):
             break
         obj = new_obj
     else:
+        from _pytest._io.saferepr import saferepr
+
         raise ValueError(
             ("could not find real function of {start}\nstopped at {current}").format(
                 start=saferepr(start_obj), current=saferepr(obj)
@@ -354,6 +355,19 @@ def safe_isclass(obj: object) -> bool:
 if sys.version_info < (3, 5, 2):
 
     def overload(f):  # noqa: F811
+        return f
+
+
+if TYPE_CHECKING:
+    if sys.version_info >= (3, 8):
+        from typing import final as final
+    else:
+        from typing_extensions import final as final
+elif sys.version_info >= (3, 8):
+    from typing import final as final
+else:
+
+    def final(f):  # noqa: F811
         return f
 
 
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -32,6 +32,7 @@ from _pytest import timing
 from _pytest._code import ExceptionInfo
 from _pytest._code.code import ExceptionRepr
 from _pytest._io.wcwidth import wcswidth
+from _pytest.compat import final
 from _pytest.compat import order_preserving_dict
 from _pytest.compat import TYPE_CHECKING
 from _pytest.config import _PluggyPlugin
@@ -309,6 +310,7 @@ class WarningReport:
         return None
 
 
+@final
 class TerminalReporter:
     def __init__(self, config: Config, file: Optional[TextIO] = None) -> None:
         import _pytest.config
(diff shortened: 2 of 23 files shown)


src/_pytest/terminal.py
```diff
@@ -309,6 +310,7 @@ def get_location(self, config: Config) -> Optional[str]:
         return None
 
 
+@final
 class TerminalReporter:
```
nicoddemus, line 314: `TerminalReporter` is subclassed by both `pytest-xdist` and `pytest-sugar`. Not that this class is meant for subclassing, in fact we have broken `pytest-sugar` in the past a few times because of side effects of some refactorings, but I guess this makes it official. 😁

nicoddemus (approved):
I believe we should add an entry to the CHANGELOG mentioning this. 👍

src/_pytest/terminal.py
bluetech: I don't see it in `pytest-xdist`, but might be missing it.

As for pytest-sugar, it seems pretty clear to me that it is using private API. We can endeavor to keep it working but I don't think we should guarantee it. So I think the `@final` is good especially for these cases (making it official as you say).

bluetech:
Added a changelog

src/_pytest/terminal.py
nicoddemus: > I don't see it in pytest-xdist, but might be missing it.

Ahh sorry, I misremembered, it uses it by aggregation:

https://github.com/pytest-dev/pytest-xdist/blob/master/src/xdist/dsession.py#L331

But unfortunately our own `TerminalReporter` knows/handles xdist itself, hehehe.

> As for pytest-sugar, it seems pretty clear to me that it is using private API. We can endeavor to keep it working but I don't think we should guarantee it. So I think the @final is good especially for these cases (making it official as you say).

Completely agree. I just mentioned those for that effect: to mark it as officially non-supported.

nicoddemus (approved):
Great work!

domdfcoding:
Regarding marking the classes as "final" in the docs I had a look at how Sphinx does it for typing.final and came up with this: https://gist.github.com/domdfcoding/aec516077cf2d5c7ab1e2790ba074bce

It requires modifying `VariableCommentPicker` as the reference to it is buried deep in Sphinx, but it seems to work.

bluetech:
@domdfcoding Wow thanks for looking at this!

How safe (in terms of future compatibility) do you think it will be to add this to the pytest sphinx config?

BTW, we have the exact same issue with `_pytest.compat.overload`, but that will resolve itself sooner, once we dump Python 3.5.

domdfcoding:
@bluetech, sorry for the delay.

I don't think the changes would break the documentation builds, as they only affect the AST visitor which has a standard syntax. There's the potential that Sphinx adds new features to the modified functions, but the worst case is that the pytest docs miss out on those features.

I would be more confident in this approach working; the other way would require modifying functions in three different classes, which I think would be more likely to break with a future Sphinx release. 

How long until pytest dumps Python 3.5? It wouldn't be too hard to support `_pytest.compat.overload`, as it's handled by the same `VariableCommentPicker` class.

src/_pytest/compat.py
```diff
@@ -357,6 +358,19 @@ def overload(f):  # noqa: F811
         return f
 
 
+if TYPE_CHECKING:
+    if sys.version_info >= (3, 8):
+        from typing import final as final
+    else:
+        from typing_extensions import final as final
+elif sys.version_info >= (3, 8):
+    from typing import final as final
```
wimglenn, line 367: @bluetech  What is the reason for using `from typing import final as final` instead of just `from typing import final`?

The-Compiler: This is a common pattern if you import something deliberately to re-export it (rather than being an unused import for inside this module only) - see e.g. https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_reexport

PR #7782, opened 2020-09-22
Improved 'Declaring new hooks' section in docs.

### LINKED ISSUE #7706: Confusion in documentation: "Declaring New Hooks"
In [this section](https://docs.pytest.org/en/latest/writing_plugins.html#declaring-new-hooks) there are several details in the examples that could be improved for less confusion:

- The sample hook declaration file is `hook.py` but the import of it reads `import hooks`; the names do not match. Consider using a less generic name, such as `sample_hook.py`.
- Related to that is this line in the documentation: "hooks are called through the `hook` object, available in the config object." Is "the `hook` object" always named `hook` (yes it is, but that's the confusing point) or is it named `hook` because the example module is `hook.py`?
- Even tho it's not required (I guess,  this is not done in the xdist example), for clarity please decorate the `pytest_addhooks()` implementations with `hookimpl` (with the AT sign, of course)? Documentation somewhere that explains when/whether this decoration is optional would also be appreciated.
- "Hooks may be called both from fixtures or from other hooks." Please clarify that a hook may be invoked from a function or method that has access to the `Config` object; this is done in the xdist example, in fact.

kamahmad:
1. Renamed the  sample hook declaration file to "sample_hook.py" from "hooks.py".
2. Added the @hookimpl decorator to example functions.
3. Added name to AUTHORS.

closes #7706

### DIFF
--- a/doc/en/writing_plugins.rst
+++ b/doc/en/writing_plugins.rst
@@ -630,7 +630,7 @@ are expected. The names of the functions must start with `pytest_` otherwise pyt
 Here's an example. Let's assume this code is in the ``sample_hook.py`` module.
 
 .. code-block:: python
-    @hookimpl
+
     def pytest_my_hook(config):
         """
         Receives the pytest config and does things with it
@@ -641,7 +641,7 @@ class or module can then be passed to the ``pluginmanager`` using the ``pytest_a
 (which itself is a hook exposed by pytest).
 
 .. code-block:: python
-    @hookimpl
+
     def pytest_addhooks(pluginmanager):
         """ This example assumes the hooks are grouped in the 'sample_hook' module. """
         from my_app.tests import sample_hook
--- a/doc/en/writing_plugins.rst
+++ b/doc/en/writing_plugins.rst
@@ -614,6 +614,11 @@ among each other.
 Declaring new hooks
 ------------------------
 
+.. note::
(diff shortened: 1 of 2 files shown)


doc/en/writing_plugins.rst
```diff
@@ -643,8 +643,8 @@ class or module can then be passed to the ``pluginmanager`` using the ``pytest_a
 .. code-block:: python
 
     def pytest_addhooks(pluginmanager):
-        """ This example assumes the hooks are grouped in the 'hooks' module. """
-        from my_app.tests import hooks
+        """ This example assumes the hooks are grouped in the 'sample_hook' module. """
+        from my_app.tests import sample_hook
 
         pluginmanager.add_hookspecs(hooks)
```
nicoddemus, line 649:
```suggestion
        pluginmanager.add_hookspecs(sample_hook)
```

nicoddemus (approved):
Thanks @kamahmad, made some adjustments.

nicoddemus:
Thanks @kamahmad!

kamahmad:
Thanks @nicoddemus . Couldn't have done it without you help 👍 

Looking forward to working on more issues. :)

nicoddemus:
Awesome, thanks! 🙇

PR #7787, opened 2020-09-23
Fix `--debug` and setting of `PYTEST_CURRENT_TEST` in non-UTF8 locales

### LINKED ISSUE #7781: Pytest debug option "--debug" has problems to write non ASCII characters
- [X] a detailed description of the bug or problem you are having
- [X] output of `pip list` from the virtual environment you are using
- [X] pytest and operating system versions
- [X] minimal example if possible

# Introduction
In my tests, I have some tests to check nonascii names for files. The test consists in import and compiles theses tests and everything works fine, but at the moment that I try to take a look in the execution of the hooks using the ``pytest --debug`` option the execution fail because pytest debug can't write the ascii character present in the Node ID of these tests, returning a ``UnicodeEncodeError``.

# ``pip list`` command
aenum                         2.2.4
alabaster                     0.7.12
allure-pytest                 2.8.10
allure-python-commons         2.8.10
appy                          0.9.6
Arpeggio                      1.9.2
atomicwrites                  1.4.0
attrs                         20.2.0
Babel                         2.8.0
backcall                      0.2.0
backports.functools-lru-cache 1.6.1
bcrypt                        3.2.0
bitstruct                     8.11.0
cached-property               1.5.1
cantools                      32.16.0
certifi                       2020.6.20
cffi                          1.14.2
chardet                       3.0.4
Click                         7.0
colorama                      0.4.3
control                       0.8.3
cparser                       0.1.3
cryptography                  3.1
cycler                        0.10.0
Cython                        0.29.21
decorator                     4.4.2
diskcache                     5.0.3
docutils                      0.14
dss-python                    0.10.6
FMPy                          0.2.14
future                        0.18.2
idna                          2.10
imagesize                     1.2.0
iniconfig                     1.0.1
ipython                       7.18.1
ipython-genutils              0.2.0
jedi                          0.14.1
Jinja2                        2.11.2
kiwisolver                    1.2.0
lark-parser                   0.9.0
lxml                          4.5.2
MarkupSafe                    1.1.1
matplotlib                    3.1.2
model-converter               1.4.4
more-itertools                8.5.0
npTDMS                        0.23.0
numexpr                       2.7.1
numpy                         1.19.0
OpenDSSDirect.py              0.4.0
openpyxl                      1.7.0
packaging                     20.4
pandas                        1.0.5
paramiko                      2.7.1
parglare                      0.10.0
parso                         0.5.1
pathlib                       1.0.1
Paver                         1.3.4
PeakUtils                     1.3.3
pickleshare                   0.7.5
Pillow                        7.2.0
pip                           20.2.3
pluggy                        0.13.1
prompt-toolkit                3.0.7
psutil                        5.7.2
py                            1.9.0
pycparser                     2.20
pycryptodomex                 3.9.8
pyFFTW                        0.12.0
pygit2                        1.2.1
Pygments                      2.7.0
pyModbusTCP                   0.1.8
PyNaCl                        1.4.0
pyparsing                     2.4.0
PyQt5                         5.15.0
PyQt5-sip                     12.8.1
pyqtgraph                     0.11.0+gabfac52
PyQtWebEngine                 5.15.0
pyserial                      3.4
pytest                        5.4.1
pytest-assume                 2.2.0
python-can                    3.3.3
python-dateutil               2.8.0
pytz                          2019.1
pyusb                         1.0.2
pywin32                       228
pyzmq                         19.0.1
QScintilla                    2.11.4
requests                      2.24.0
scipy                         1.5.1
setuptools                    47.1.0
sip                           5.4.0
six                           1.12.0
snowballstemmer               2.0.0
Sphinx                        1.7.0
sphinxcontrib-programoutput   0.16
sphinxcontrib-serializinghtml 1.1.4
sphinxcontrib-websupport      1.2.4
tables                        3.6.1
tabulate                      0.8.3
textparser                    0.23.0
textX                         2.1.0
toml                          0.10.1
traitlets                     5.0.4
urllib3                       1.25.10
wcwidth                       0.2.5
wheel                         0.35.1
win-inet-pton                 1.1.0
windows-curses                2.1.0
wrapt                         1.12.1
yapf                          0.28.0

# Pytest and OS versions
pytest-6.0.2 - Windows 10 (Home Edition)

# A brief example
``` python
# In test_sample.py
def test4_чћшђ_čćšđ():
    assert True

# Called using ``pytest test_sample.py`` -> OK
# Called using ``pytest --debug test_sample.py`` -> UnicodeEncodeError
```

# Conclusion
Any doubt about my explanation about this issue just ping me, Thanks!

### LINKED ISSUE #7786: pytest fails setting PYTEST_CURRENT_TEST when test name cannot be encoded with current locale
```
$ LANG=C pytest t.py 
platform linux -- Python 3.6.12, pytest-6.0.2, py-1.9.0, pluggy-0.13.1
rootdir: /tmp/x/pytest, configfile: pyproject.toml
plugins: hypothesis-5.35.4
collected 1 item                                                               

t.py EE                                                                  [100%]

\x1b[31m\x1b[1m______________________ ERROR at setup of test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 _______________________\x1b[0m

self = environ({'LANG': 'C', 'SHELL': '/bin/bash', 'SESSION_MANAGER': 'local/babibox:@/tmp/.ICE-unix/1669,unix/babibox:/tmp/....DBUS_SESSION_BUS_ADDRESS': 'unix:path=/run/user/1000/bus', 'OLDPWD': '/tmp/x', '_': '/tmp/x/pytest/venv36/bin/pytest'})
key = b'PYTEST_CURRENT_TEST', value = 't.py::test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 (setup)'

    def __setitem__(self, key, value):
        key = self.encodekey(key)
>       value = self.encodevalue(value)

/usr/lib/python3.6/os.py:674: 

value = 't.py::test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 (setup)'

    def encode(value):
        if not isinstance(value, str):
            raise TypeError("str expected, not %s" % type(value).__name__)
>       return value.encode(encoding, 'surrogateescape')
E       UnicodeEncodeError: 'ascii' codec can't encode characters in position 12-15: ordinal not in range(128)

/usr/lib/python3.6/os.py:745: UnicodeEncodeError
\x1b[31m\x1b[1m_____________________ ERROR at teardown of test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 _____________________\x1b[0m

self = environ({'LANG': 'C', 'SHELL': '/bin/bash', 'SESSION_MANAGER': 'local/babibox:@/tmp/.ICE-unix/1669,unix/babibox:/tmp/....DBUS_SESSION_BUS_ADDRESS': 'unix:path=/run/user/1000/bus', 'OLDPWD': '/tmp/x', '_': '/tmp/x/pytest/venv36/bin/pytest'})
key = b'PYTEST_CURRENT_TEST', value = 't.py::test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 (teardown)'

    def __setitem__(self, key, value):
        key = self.encodekey(key)
>       value = self.encodevalue(value)

/usr/lib/python3.6/os.py:674: 

value = 't.py::test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 (teardown)'

    def encode(value):
        if not isinstance(value, str):
            raise TypeError("str expected, not %s" % type(value).__name__)
>       return value.encode(encoding, 'surrogateescape')
E       UnicodeEncodeError: 'ascii' codec can't encode characters in position 12-15: ordinal not in range(128)

/usr/lib/python3.6/os.py:745: UnicodeEncodeError
=========================== short test summary info ============================
ERROR t.py::test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 - UnicodeEncodeError: 'ascii' codec can't encode ...
ERROR t.py::test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111 - UnicodeEncodeError: 'ascii' codec can't encode ...
============================== 2 errors in 0.21s ===============================
```

Using the test from #7781

asottile:
Resolves #7781 
Resolves #7786

### DIFF
--- a/changelog/7781.bugfix.rst
+++ b/changelog/7781.bugfix.rst
@@ -1,0 +1,1 @@
+Fix writing non-encodable text to log file when using ``--debug``.
--- a/changelog/7786.bugfix.rst
+++ b/changelog/7786.bugfix.rst
@@ -1,0 +1,1 @@
+Fix setting of ``PYTEST_CURRENT_TEST`` when the current locale cannot encode the test name.
--- a/src/_pytest/helpconfig.py
+++ b/src/_pytest/helpconfig.py
@@ -100,7 +100,7 @@ def pytest_cmdline_parse():
     config = outcome.get_result()  # type: Config
     if config.option.debug:
         path = os.path.abspath("pytestdebug.log")
-        debugfile = open(path, "w")
+        debugfile = open(path, "w", encoding="UTF-8")
         debugfile.write(
             "versions pytest-%s, py-%s, "
             "python-%s\ncwd=%s\nargs=%s\n\n"
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -189,7 +189,13 @@ def _update_current_test_var(
         value = "{} ({})".format(item.nodeid, when)
         # don't allow null bytes on environment variables (see #2644, #2957)
         value = value.replace("\x00", "(null)")
-        os.environ[var_name] = value
+        # if we are running with a locale that cannot encode the test name
+        # and we're on a system where os.environb is supported, explicitly
+        # encode using UTF-8 (otherwise we get an EncodeError)
+        if os.supports_bytes_environ:
+            os.environb[var_name.encode()] = value.encode()
+        else:
+            os.environ[var_name] = value
     else:
         os.environ.pop(var_name)
 
--- a/testing/test_helpconfig.py
+++ b/testing/test_helpconfig.py
@@ -108,6 +108,21 @@ def test_debug(testdir):
     assert "pytest_sessionstart" in p.read()
 
 
+def test_debug_with_non_ascii_test_name(testdir, monkeypatch):
+    testdir.makepyfile(
+        """\
+        def test4_чћшђ_čćšđ():
+            pass
+        """
+    )
+    # force ASCII default io encoding (PYTHONIOENCODING didn't seem to work?)
+    monkeypatch.setenv("LANG", "C")
+    result = testdir.runpytest_subprocess("--debug")
+    assert result.ret == ExitCode.OK
+    log_contents = testdir.tmpdir.join("pytestdebug.log").read_binary()
+    assert b"pytest_sessionstart" in log_contents
+
+
 def test_PYTEST_DEBUG(testdir, monkeypatch):
     monkeypatch.setenv("PYTEST_DEBUG", "1")
     result = testdir.runpytest_subprocess()


The-Compiler (commented):
I'm not sure if I agree with the approach here.

For the debug file, I guess forcing it to be UTF-8 isn't too bad - even on Windows, I suppose most text editors will be able to detect and open UTF-8 (though I'm not sure if Windows Notepad does?)

But for the environment variable, this will just shift the error to the reader of the variable - surely the reader won't expect an UTF-8 encoded variable when the system encoding isn't UTF-8. Also I'm not sure how the situation is on Windows, are environment variables text/UTF-8 there? Otherwise this would still fail there, no?

The only other solution I can see is something like:

```
encoding = sys.getfilesystemencoding()
value = value.encode(encoding, errors='replace').decode(encoding)
```

asottile:
> 
> 
> I'm not sure if I agree with the approach here.
> 
> For the debug file, I guess forcing it to be UTF-8 isn't too bad - even on Windows, I suppose most text editors will be able to detect and open UTF-8 (though I'm not sure if Windows Notepad does?)
> 
> But for the environment variable, this will just shift the error to the reader of the variable - surely the reader won't expect an UTF-8 encoded variable when the system encoding isn't UTF-8. Also I'm not sure how the situation is on Windows, are environment variables text/UTF-8 there? Otherwise this would still fail there, no?
> 
> The only other solution I can see is something like:
> 
> ```
> encoding = sys.getfilesystemencoding()
> value = value.encode(encoding, errors='replace').decode(encoding)
> ```

- windows doesn't support bytes environ, so it is always text there
- I know of no modern system which doesn't use UTF-8 as a default locale

nicoddemus:
> windows doesn't support bytes environ, so it is always text there

True:

```python
Python 3.6.7 (default, Mar  4 2020, 17:08:00) [MSC v.1900 64 bit (AMD64)] on win32
>>> import os
>>> os.supports_bytes_environ
False
```

> I know of no modern system which doesn't use UTF-8 as a default locale

I think so too, but shouldn't we use the default locale then?

nicoddemus:
> For the debug file, I guess forcing it to be UTF-8 isn't too bad - even on Windows, I suppose most text editors will be able to detect and open UTF-8 (though I'm not sure if Windows Notepad does?)

I agree. The debug file is meant for advanced users, so it is fair to encode it using a known encoding and expect those advanced users to be able to open it. Even in situations where the user can't use a text editor that supports UTF-8, all it will happen is a few wrong characters *if* output contains non-ASCII output, which shouldn't be too bad.

asottile:
> I think so too, but shouldn't we use the default locale then?

if we use the default locale then there's information loss

nicoddemus:
> if we use the default locale then there's information loss

Hmm TBH I not sure I follow the underlying semantics here, but I will defer to your expertise that this is the correct approach. Other than that, the rest looks great. 👍

nicoddemus (approved):

bluetech:
Agree with the others that UTF-8 is good for `--debug`.

As for `PYTEST_CURRENT_TEST` I think UTF-8 is not really appropriate, because anyone *reading* the value would not do so correctly. You mentioned you don't want information loss but I think given how `PYTEST_CURRENT_TEST` is [described](https://docs.pytest.org/en/stable/example/simple.html#pytest-current-test-environment-variable) in the docs, specifically

> The contents of PYTEST_CURRENT_TEST is meant to be human readable and the actual format can be changed between releases (even bug fixes) so it shouldn’t be relied on for scripting or automation.

that information loss is preferable over bad encoding. So my suggestion would be to use `encode(sys.getfilesystemencoding(), "errors")`. For ascii it will be `b't.py::test4_????_????'`.

(`getfilesystemencoding` seems to be what python uses for envvars when `supports_bytes_environ` is true).

asottile:
I'm going to merge this as-is unless there's stronger opposition

python3.8+ in the same situation chooses UTF-8 so I think this choice is consistent and appropriate for cpython -- this is really only a bug in older python versions

the filesystem encoding also isn't appropriate for the environ, it's not a file -- there are however two equally bad choices: locale or filesystem

bluetech:
> python3.8+ in the same situation chooses UTF-8

Do you have a reference for that? Looking at the os.py code it seems the same to me.

asottile:
actually, it's 3.7+

```console
$ LANG= python3.6 -c 'import os; os.environ["x"] = "☃"'
Unable to decode the command from the command line:
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 30-32: surrogates not allowed
$ LANG= python3.7 -c 'import os; os.environ["x"] = "☃"'
$ LANG= python3.8 -c 'import os; os.environ["x"] = "☃"'
$
```

bluetech:
That's because these versions changed the interpretation of "C" to UTF-8. But doesn't work for other cases:

```
$ LANG=en_US.ISO-8859-1 python
Python 3.8.5 (default, Jul 27 2020, 08:42:51) 
[GCC 10.1.0] on linux
>>> import os; os.environ['x'] = 't.py::test4_\u0447\u045b\u0448\u0452_\u010d\u0107\u0161\u0111'
  File "/usr/lib/python3.8/os.py", line 680, in __setitem__
    value = self.encodevalue(value)
  File "/usr/lib/python3.8/os.py", line 751, in encode
    return value.encode(encoding, 'surrogateescape')
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 12-15: ordinal not in range(256)
```

asottile:
via #10935

PR #7806, opened 2020-09-27
Put smoke test deps in requirements.txt for Dependabot

hugovk:
Here's an idea.

This moves the smoke test deps from https://github.com/pytest-dev/pytest/pull/7721 into a `requirements.txt` so they can pinned.

Then, https://dependabot.com can be added to monitor this new `requirements.txt` and send a new PR when there's a new plugin release. Add `/testing/plugins_integration` as the Directory:

That means we get a trigger when new plugins are released, which increases confidence that current master works with each plugin.

For example, see https://github.com/hugovk/pytest/pull/6 and https://github.com/hugovk/pytest/pull/7 on a similar branch on my fork.

Upsides:
* Triggers a build on plugin releases
* Always smoke test latest plugin version (as before)

Downsides:
* More PRs.
  * If they're too frequent, the update schedule could be changed from live to daily, weekly or monthly. 
  * It's also possible to set them to merge automatically once status checks pass.

I've left anyio at the non-latest version, to allow a test bump from 2.0.0 to 2.0.2.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/testing/plugins_integration"
+  schedule:
+    interval: monthly
+    time: "03:00"
+  open-pull-requests-limit: 10
+  allow:
+  - dependency-type: direct
+  - dependency-type: indirect
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
 - package-ecosystem: pip
   directory: "/testing/plugins_integration"
   schedule:
-    interval: monthly
+    interval: weekly
     time: "03:00"
   open-pull-requests-limit: 10
   allow:
--- a/testing/plugins_integration/requirements.txt
+++ b/testing/plugins_integration/requirements.txt
@@ -1,0 +1,15 @@
+anyio[curio,trio]==2.0.0
+django==3.1.1
+pytest-asyncio==0.14.0
+pytest-bdd==4.0.1
+pytest-cov==2.10.1
+pytest-django==3.10.0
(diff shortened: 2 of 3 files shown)


Zac-HD (approved):
I'd go for daily + automerge, but let's try it out and see what happens :rocket:

bluetech:
Thanks @hugovk, this seems like a good way to keep the plugin smoketest fresh.

My preference is for a monthly update and without automerge (at least for start). I think daily would become annoying...

I believe dependabot is already configured for this repo, so we'd just need a suitable `dependabot.yml` file. Can you add one to this PR?

hugovk:
I've added a file for this exported config:



For info, the export said:

> Dependabot is moving natively into GitHub, and we need a quick update from you to finish the move! Click "Create pull request" to open a pull request on pytest with a new config file. When you merge the pull request, we'll swap outdependabot-preview for a new dependabot app, and you'll be all set.

> The pull request moves the UI configuration to a config file in.github/dependabot.yml. The config file uses the [new syntax](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates):

...

> Your account is using automerging. There is no automerging support in GitHub-native Dependabot, so these settings will not be added to the new config file. Several 3rd-party GitHub Actions and bots can replicate the automerge feature.

bluetech (approved):
Thanks, let's leave this up for a few days to let others a chance to chime in, then merge.

nicoddemus:
Great idea, thanks @hugovk!

> I'd go for daily + automerge
> My preference is for a monthly update

How about we compromise and have it once a week? Doesn't seem like this would be much overhead.

Regardless, :+1: from me!

nicoddemus (approved):

PR #7817, opened 2020-09-29
terminal: fix crash in header reporting when absolute testpaths is used

bluetech:
Regressed in 6.1.0 in 62e249a.

The `x` is an `str` but is expected to be a `pathlib.Path`. Not caught by mypy because `config.getini()` returns `Any`.

Fix by just removing the `bestrelpath` call:

- testpaths are always relative to the rootdir, it thus would be very unusual to specify an absolute path there.

- The code was wrong even before the regression: `py.path.local`'s `bestrelpath` function expects a `py.path.local`, not an `str`. But it had some weird `try ... except AttributeError` fallback which just returns the argument, i.e. it was a no-op. So there is no behavior change.

- It seems reasonable to me to just print the full path if that's what the ini specifies.

The two other commits are related improvements.

### DIFF
--- a/changelog/7814.bugfix.rst
+++ b/changelog/7814.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed crash in header reporting when :confval:`testpaths` is used and contains absolute paths (regression in 6.1.0).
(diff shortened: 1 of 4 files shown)


bluetech:
Note: I only intend to backport the "terminal: fix crash in header reporting when absolute testpaths is used" commit.

nicoddemus (approved):

PR #7811, opened 2020-09-28
[WIP] py36 plus

asottile:
still TODO:
- audit `TYPE_CHECKING`

PR #7826, opened 2020-10-01
doc: patch Sphinx to detect our `@final` for marking classes as `final`

bluetech:
Continuation to #7780, makes the `final class` also show up in the API Reference.

Thanks to @domdfcoding for code & assistance.

Compared to @domdfcoding's code, I made it less accurate (just accepts any `@final` as an indicator, disregarding where it's imported from), but simpler and hopefully more robust against changes in Sphinx. Since it's only scoped to pytest it shouldn't be a problem.

We also have a similar issue with `@overload`, but that should take care of itself after #7811 is merged (as confirmed by the RTD build of that PR).

### DIFF
--- a/changelog/7780.doc.rst
+++ b/changelog/7780.doc.rst
@@ -1,0 +1,1 @@
+Classes which should not be inherited from are now marked ``final class`` in the API reference.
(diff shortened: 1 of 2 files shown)


nicoddemus (approved):

bluetech:
Rebased to fix a tiny conflict

Issue #7808, opened 2020-09-28
dropping python3.5?


asottile:
since it is now end-of-lifed we should decide how much longer we'd like to support it

when dropping 3.5, we should maybe consider dropping 3.6.0 as well (3.6.0 has a broken implementation of `NamedTuple` for example) and only support 3.6.1+

The-Compiler:
We have some workarounds for Python 3.5.0 and 3.5.1 in which can be removed then (eg. #5752) together with Travis CI (which we only use to test Python 3.5.1).

Can you elaborate on the exact issues with 3.6.0? IIRC some people (@nicoddemus?) were against dropping 3.5.0 and 3.5.1 when I suggested it (can't find the thread anymore...) because it's easier to communicate "We require Python 3.5" than "We require Python 3.5.2+".

Zac-HD:
I'm in favor of dropping Python 3.5 as soon as practical, and updating `black` to `target_version=py36`.  
(for Hypothesis we also ran `pyupgrade --py36-plus`, but YMMV)

nicoddemus:
> IIRC some people (@nicoddemus?) were against dropping 3.5.0 and 3.5.1 when I suggested it (can't find the thread anymore...)

Don't recall either....

I guess we can then drop Python 3.5 in 6.2 then?

The-Compiler:
> I guess we can then drop Python 3.5 in 6.2 then?

I suppose following our own [backwards compatibility statement](https://github.com/pytest-dev/pytest/blob/master/doc/en/backwards-compatibility.rst) we'd first need to deprecate it and such? Then again maybe this is a bit of a special case since it's already EOL upstream.

nicoddemus:
In the past we just dropped versions in major releases without a warning in advance, but we can improve that for sure. How would we "warn" users in that aspect?

asottile:
It looks like in the past we dropped support for minor versions in minor releases (pytest 3.3 dropped python 2.6 and 3.3) -- the execption is when we dropped python2.7 but that seems exceptional on its own

nicoddemus:
True, minor releases (updated my comment).

asottile:
imo python-version-based warnings are unlikely to be helpful, most situations the user is unable to update their (eol) python version

my proposal:
- target the 6.2.0 pytest release
- allow `python_requires` to do its thing

this will allow us to use these typing imports we could not use before:
```
[...]

PR #7831, opened 2020-10-02
py36+: drop python3.5 in CI and setup.cfg

asottile:
this is the first part to #7808

I have ~13 other patches that can be previewed in #7811


changelog/7808.improvement.rst
```diff
@@ -0,0 +1 @@
+pytest now supports python3.6+ only.
```
nicoddemus, line 1: Should we consider this a breaking change perhaps?

asottile: I was on the fence, since we're going to release this in a minor version 🤔

nicoddemus: Yeah, but I think there's precedence: we removed a bunch of features in 6.1, and also [5.4](https://docs.pytest.org/en/stable/changelog.html#pytest-5-4-0-2020-03-12).

asottile: done!

nicoddemus (approved):
Thanks!

PR #7832, opened 2020-10-02
py36+: remove pathlib2 compatibility shim

asottile:
see #7808


nicoddemus (approved):

bluetech:
Perhaps we want to remove the `Path`, `PurePath` reexports from `_pytest.pathlib` now?

asottile:
> 
> 
> Perhaps we want to remove the `Path`, `PurePath` reexports from `_pytest.pathlib` now?

indeed, I did the simpler change in this patch to avoid conflicts in all the others

PR #7838, opened 2020-10-02
py36+: remove requires_ordered_markup

asottile:
See #7808

### DIFF
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -3,7 +3,6 @@ import sys
 from typing import List
 
 import pytest
-from _pytest.pytester import RunResult
 from _pytest.pytester import Testdir
 
 if sys.gettrace():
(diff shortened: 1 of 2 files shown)


nicoddemus (approved):
Awesome work @asottile, thanks a ton for this!

I suggest hold off merging a day or two to give other reviewers a chance to take a look.

Other than that: do you plan to run `pyupgrade --py36-plus` to change string.format to fstrings?

Particularly I'm -0 the changes to git blame are worth it (but of course it is refreshing to be able to use it in new code).

asottile:
> 
> 
> Awesome work @asottile, thanks a ton for this!
> 
> I suggest hold off merging a day or two to give other reviewers a chance to take a look.
> 
> Other than that: do you plan to run `pyupgrade --py36-plus` to change string.format to fstrings?
> 
> Particularly I'm -0 the changes to git blame are worth it (but of course it is refreshing to be able to use it in new code).

this is just the first round of changes, there's a whole set of other changes that conflicted / depended on these that I'd have to rebase

as these are the less controversial ones I'm just going to merge these now -- we can always revert later

I do want to run pyupgrade and black in py36+ mode (you can see what these look like in #7811 -- they touch way fewer lines than I expected) once the manual code changes are done

fstrings also offer some performance gains after all :) (there's other things as well that'll be upgraded beyond fstrings iirc) so I think we should do it anyway.  

additionally for blame, there's always [`--ignore-rev`](https://black.readthedocs.io/en/stable/installation_and_usage.html#migrating-your-code-style-without-ruining-git-blame) and/or `--ignore-revs-file` which fixes git blame for code reformats nicely!

I also still have to work out how to piece up the work for `# type:` comments since there's a significant amount that fail after automatic conversion.  that's going to be a whole project in its own :)

nicoddemus:
> as these are the less controversial ones I'm just going to merge these now -- we can always revert later

I agree they are not controversial, it was more to have a second set on eyes on the PRs, but if this will impact other immediate then by all means go for it.

> I do want to run pyupgrade and black in py36+ mode (you can see what these look like in #7811 -- they touch way fewer lines than I expected) once the manual code changes are done

Ahh OK. I would expect them to touch a whole lot, but if they are few then I agree it is fine. 👍 

> I also still have to work out how to piece up the work for # type: comments since there's a significant amount that fail after automatic conversion. that's going to be a whole project in its own :)

👍

PR #7839, opened 2020-10-03
py36+: remove _pytest.compat.fspath

asottile:
See #7808


bluetech (approved):
There are also a bunch of places where we do `something(str(path))` instead of just `something(path)` because the something only supports the fspath protocol since 3.6. But they are probably harder to track.

nicoddemus (approved):

PR #7840, opened 2020-10-03
py36+: from typing import Type: no longer need guard

asottile:
See #7808

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -21,6 +21,7 @@ from typing import Pattern
 from typing import Sequence
 from typing import Set
 from typing import Tuple
+from typing import Type
 from typing import TYPE_CHECKING
 from typing import TypeVar
 from typing import Union
@@ -44,7 +45,6 @@ from _pytest.compat import overload
 from _pytest.pathlib import Path
 
 if TYPE_CHECKING:
-    from typing import Type
     from typing_extensions import Literal
     from weakref import ReferenceType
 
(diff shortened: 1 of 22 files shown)


src/_pytest/_code/code.py
```diff
@@ -421,14 +421,14 @@ class ExceptionInfo(Generic[_E]):
 
     _assert_start_repr = "AssertionError('assert "
 
-    _excinfo = attr.ib(type=Optional[Tuple["Type[_E]", "_E", TracebackType]])
+    _excinfo = attr.ib(type=Optional[Tuple[Type["_E"], "_E", TracebackType]])
```
bluetech, line 424: Does `_E` needs to be quoted here?

bluetech (approved):

src/_pytest/_code/code.py
graingert:
```suggestion
    _excinfo = attr.ib(type=Optional[Tuple[Type[_E], _E, TracebackType]])
```

I don't think so

graingert: can't this be an actual type annotation now?

```suggestion
    _excinfo: Optional[Tuple[Type[_E], _E, TracebackType] = attr.ib()
```

or even with `auto_attribs=True`?

nicoddemus: Agreed

src/_pytest/warning_types.py
```diff
@@ -105,7 +102,7 @@ class UnformattedWarning(Generic[_W]):
     as opposed to a direct message.
     """
 
-    category = attr.ib(type="Type[_W]")
+    category = attr.ib(type=Type["_W"])
```
nicoddemus, line 105:
```suggestion
    category = attr.ib(type=Type[_W])
```

nicoddemus (approved):

src/_pytest/_code/code.py
asottile: there's a bug in mypy here fwiw

src/_pytest/warning_types.py
asottile: same here, bug in mypy

PR #7845, opened 2020-10-03
py36+: remove _pytest.compat.overload

asottile:
See #7808


bluetech (approved):
YES this one was so annoying

PR #7847, opened 2020-10-03
py36+: use new NamedTuple syntax

asottile:
See #7808

this one is slightly controversial, though I expect almost zero usage of 3.6.0

### DIFF
--- a/changelog/7808.breaking.rst
+++ b/changelog/7808.breaking.rst
@@ -1,1 +1,1 @@
-pytest now supports python3.6+ only.
+pytest now supports python3.6.1+ only.
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,7 @@ install_requires =
     atomicwrites>=1.0;sys_platform=="win32"
     colorama;sys_platform=="win32"
     importlib-metadata>=0.12;python_version<"3.8"
-python_requires = >=3.6
+python_requires = >=3.6.1
 package_dir =
     =src
 setup_requires =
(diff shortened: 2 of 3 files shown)


bluetech:
If it's just for this I think it's *probably* not worth the hassle of the "3.6 but not 3.6.0" subtlety? We did endure 3.5.{0,1} which were much worse than this...

asottile:
> 
> 
> If it's just for this I think it's _probably_ not worth the hassle of the "3.6 but not 3.6.0" subtlety? We did endure 3.5.{0,1} which were much worse than this...

there's also `NoReturn` in 3.6.2 but yeah this is pretty subtle.  for what it's worth pre-commit does 3.6.1+ and there hasn't been any complaints thus far

nicoddemus:
I see your point @asottile but I agree with @bluetech, I  would prefer to avoid having to special case, and this case doesn't really buy us much. 

However I think we can mark the code so it is clear that's only 3.6.0:

```
if sys.version_info[:3] == (3, 6, 0):
    # In 3.6.0 NamedTuples, ...
    class _ParameterSet(
        NamedTuple(
            "ParameterSet",
            [
                ("values", Sequence[Union[object, NotSetType]]),
                ("marks", Collection[Union["MarkDecorator", "Mark"]]),
                ("id", Optional[str]),
            ],
        )
    ):
    pass
else:    
    class _ParameterSet(NamedTuple):
        values: Sequence[Union[object, NotSetType]]
        marks: Collection[Union["MarkDecorator", "Mark"]]
        id: Optional[str]


class ParameterSet(_ParameterSet):
    ...
```

This helps finding the code when we drop 3.6.

PR #7848, opened 2020-10-03
py36+ tests are definition ordered

### LINKED ISSUE #5196: Use simple definition order in Python>3.6
> once we're py36+ I think we can stop sorting and let the definition order handle it (due to ordered mapping types by default)

https://github.com/pytest-dev/pytest/issues/5163#issuecomment-486873334_

asottile:
Resolves #5196
See #7808

### DIFF
--- a/changelog/5196.feature.rst
+++ b/changelog/5196.feature.rst
@@ -1,0 +1,1 @@
+Tests are now ordered by definition order in more cases.
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -442,11 +442,6 @@ class PyCollector(PyobjMixin, nodes.Collector):
                 else:
                     values.append(res)
 
-        def sort_key(item):
-            fspath, lineno, _ = item.reportinfo()
-            return (str(fspath), lineno)
-
-        values.sort(key=sort_key)
         return values
 
     def _genfunctions(self, name: str, funcobj) -> Iterator["Function"]:
--- a/testing/python/collect.py
+++ b/testing/python/collect.py
@@ -752,6 +752,30 @@ class TestSorting:
         assert len(colitems) == 2
         assert [item.name for item in colitems] == ["test_b", "test_a"]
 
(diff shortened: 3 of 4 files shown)


asottile:
an interesting side-effect of this is base classes appear *after* the subclass which is perhaps a little strange -- I can see if I can adjust that behaviour if it's undesirable (I would expect pytest to run base-class tests first, though it makes sense from an mro() perspective the way it is currently implemented) 🤔

bluetech:
So IIUC the previous behavior was to execute tests in the order in which they physically appear (the file path and the line number within the file). So tests from base classes would execute before or after depending on where the base class is in the file (nice IMO) or how its filename sorts (doesn't make sense IMO).

Certainly definition order makes more sense semantically, but I do indeed think that tests from base classes should run first, that's how I expect it anyway. I was going to say "just traverse the MRO in reverse" but since the code pokes at `__dict__`s directly that would be tricky.

asottile:
I don't think the mro thing would be that complicated -- I'll see what I can do (I agree the base classes should run first)

bluetech:
Looking at collection timings for pandas (182330 tests), this sort call features prominently. Particularly the slowdown comes from `sort` -> `sort_key` -> `PyObjMixin.reportinfo()` -> `getfslineno(obj)` -> `findsource(obj)` (only for objs without `__code__`, mostly test classes) -> `inspect.findsource(obj)`. And apparently [since Python 3.9](https://github.com/python/cpython/commit/696136b993e11b37c4f34d729a0375e5ad544ade), `inspect.findsource()` uses `ast` instead of regexes to find the line number of classes, which I suspect makes it even slower (didn't verify this).

Anyway, if we can get this PR to work it will be nice (maybe I'll try it myself).

asottile:
ah yeah I haven't had time to come back to this -- I think all that needs to happen is to consider the `mro` when ordering tests -- not sure when I'll have some time to look at it so feel free to carry the patch!

PR #7850, opened 2020-10-03
py36+: remove rexport of Path and PurePath

asottile:
See #7808


bluetech (approved):
Hopefully there aren't any plugins which do `from _pytest.pathlib import *`...

RonnyPfannschmidt (approved):

PR #7863, opened 2020-10-06
py36+: remove _pytest.compat.order_preserving_dict

asottile:
See #7808

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -22,7 +22,6 @@ from .reports import CollectReport
 from _pytest import nodes
 from _pytest._io import TerminalWriter
 from _pytest.compat import final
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):

bluetech:
I thought it was only since 3.7?

asottile:
> 
> 
> I thought it was only since 3.7?

implemented and provisional in cpython 3.6 (pypy since 3.5), and permanently specified in 3.7

bluetech (approved):
Huh so that should cover it.

nicoddemus:
> Huh so that should cover it.

Yep, seems like it, PyPy has that since 2015 so it should be covered there as well (in fact CPython borrowed the implementation IIRC).

nicoddemus (approved):

PR #7871, opened 2020-10-07
Add pythonic code style to validate boolean conditions

drsantos20:
**Small Code improvements**

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -71,6 +71,7 @@ Damian Skrzypczak
 Daniel Grana
 Daniel Hahler
 Daniel Nuri
+Daniel Santos
 Daniel Wandschneider
 Danielle Jenkins
 Daniil Galiev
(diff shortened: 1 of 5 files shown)


nicoddemus:
Hi @drsantos20,

Thanks for the PR, but I personally don't think replacing `x or y` by `any([x, y])` is either more Pythonic or an improvement by any means. This is a 👎 from me.

The-Compiler:
I don't think this is an improvement - IMHO, `if a or b:` is perfectly clear, and easier to read than `if any([a, b]):`. So basically what @nicoddemus said while I was typing this :slightly_smiling_face: 

Taking the freedom to close this - thanks for the contribution anyways!

gnikonorov:
@nicoddemus and @The-Compiler beat me to it. Also, this would slightly change behavior as any does not short circuit

PR #7874, opened 2020-10-08
Use collections.Counter instead of a dict - Closes #7868

### LINKED ISSUE #7868: Use `collections.Counter` instead of a `dict` in `terminal.py`
Consider the following code defining `counts`, and the similar [`counts_by_filename`](https://github.com/pytest-dev/pytest/blob/95917f8833bc71229c3ac4fb73cbf5ca5818201b/src/_pytest/terminal.py#L925-L928) below:

https://github.com/pytest-dev/pytest/blob/95917f8833bc71229c3ac4fb73cbf5ca5818201b/src/_pytest/terminal.py#L757-L760

This is a classic case for [`collections.Counter`](https://docs.python.org/3/library/collections.html#collections.Counter), and would make a great first pull request!

Example solution initialising the `Counter` from an iterable:

```python
counts = Counter(... for item in items)  # you fill in the ... part
```

tanvimehta:
Use `collections.Counter` instead of a `dict` in `terminal.py` Closes #7868

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -275,6 +275,7 @@ Sven-Hendrik Haase
 Sylvain Marié
 Tadek Teleżyński
 Takafumi Arakaki
+Tanvi Mehta
 Tarcisio Fischer
 Tareq Alayan
 Ted Xiao
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -920,7 +920,9 @@ class TerminalReporter:
                 if len(locations) < 10:
                     return "\n".join(map(str, locations))
 
-                counts_by_filename = Counter(str(loc).split("::", 1)[0] for loc in locations)
+                counts_by_filename = Counter(
+                    str(loc).split("::", 1)[0] for loc in locations
+                )
                 return "\n".join(
                     "{}: {} warning{}".format(k, v, "s" if v > 1 else "")
                     for k, v in counts_by_filename.items()
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -8,6 +8,7 @@ import inspect
 import platform
 import sys
 import warnings
+from collections import Counter
 from functools import partial
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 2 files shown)


tanvimehta:
@Zac-HD Would appreciate it if you could help review this. Thanks!

Zac-HD (approved):
This looks great to me - thanks @tanvimehta, and congrats on your first Pytest PR :tada: 

I'd also encourage you to add your name to the list of contributors in https://github.com/pytest-dev/pytest/blob/master/AUTHORS, and edit your PR description to [use the magic keywords](https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords) to automatically close the linked issue (e.g. `Closes #7868` instead of `closes issue #7868` - putting "issue" in the middle disables it).

Then we'll be ready to merge :star_struck:

Zac-HD:
Apparently keywords only work in the body, not the title - you should see a little line under the keyword (and tooltip) when it's working, including in the preview.

(FYI this is not at all a blocker, I'd merge it anyway, but the point here is to level you up as a OSS contributor :sparkling_heart:)

tanvimehta:
Thanks a lot @Zac-HD. I added the magic word to the body(confirmed that it works as expected) and added my name to the AUTHORS file. I assumed this change would be considered trivial so I wasn't sure about adding my name to the authors' list.

Zac-HD:
> I assumed this change would be considered trivial so I wasn't sure about adding my name to the authors' list.

Fair enough!  From one perspective it *is* trivial, in that it probably took me longer to write it up and review your PR than it would have to just write the patch myself.  But IMO that's the *wrong* perspective - we did it this way because mentoring and community matter for us (pytest and the wider Python OSS community), and you absolutely are a contributor.  So don't be shy about adding yourself to the list on other projects - the length of that list is a point of pride for us :grin: 

-------

One bit of advice for next time: I just noticed that you made the PR from your `master` branch - it's better to [use a PR-specific branch](https://guides.github.com/introduction/flow/), which lets you work on more than one PR or feature at a time.  Later on you can also do fancy tricks like rebasing on the latest upstream version to fix any conflicts.

And finally, :sparkles:CONGRATULATIONS:sparkles: on the PR :tada:

PR #7875, opened 2020-10-08
Improve summary stats when using '--collect-only'

### LINKED ISSUE #7701: Summary when using collection only is confusing
A this moment the summary line is the same regarless if you effectively executed any tests or if you just did a discovery with commands like `pytest --collect-only` or `pytest-collect-only -k foo`.

The expectation is to see in the summary how many tests were collected and how many tests were deselected.

The collected/deselected/selected is printed early but not part of the summary and if you have more than ~40-60 tests selected, you will need to scroll a lot to spot that information.

The proposal here is to include this information in the summary line, which is clearly easier to read.

# Current output
```
# pytest --collect-only
==== no tests ran in 1.66s ====
# pytest --collect-only -k filter
==== 544 deselected in 1.67s ====
```

# Expected output

```
# pytest --collect-only
==== 123 selected, no tests ran in 1.66s ====
# pytest --collect-only -k filter
==== 123 selected, 544 deselected in 1.67s ====
```

# Details

Some feedback regarding the formatting of the summary when run in collection mode would be appreciated.

Example of how visible is the summary line versus the "collected" line, the one that usually gets lost in the terminal output. (`pco` is my alias for `pytest --collect-only`)

caramelomartins:
Hey,

This PR intends to improve on the output of the summary stats line when using the `--collection-only`. I wrote it a bit "hackishly" because I was discovering the codebase so any feedback on improving the code, or the behavior, would be greatly appreciated. I also took the liberty of updating the existing tests to match this new behavior.

I tried to make the behavior similar to what was requested in #7701 and it results in something similar to:

```
==== 114 selected, 2891 deselected, no tests ran in 1.50s ====
```

Additionally, I noticed that in the examples found in the documentation, it only displays contexts where only "no tests ran in (...)s" happens...would it make sense to add more examples of different outputs?

Closes #7701.

Let me know how you feel about this.

Thanks!

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -1166,7 +1166,7 @@ class TerminalReporter:
 
     def build_summary_stats_line(self) -> Tuple[List[Tuple[str, Dict[str, bool]]], str]:
         main_color, known_types = self._get_main_color()
-        selected = self._numcollected
+        deselected = 0
         parts = []
         errors = 0
 
@@ -1181,7 +1181,7 @@ class TerminalReporter:
                 parts.append(("%d %s" % _make_plural(count, key), markup))
 
                 if key == "deselected":
-                    selected -= count
+                    deselected += count
 
                 if key == "error":
                     errors = count
@@ -1191,25 +1191,31 @@ class TerminalReporter:
 
         if self.config.getoption("collectonly"):
 
-            if self._numcollected == selected:
-                co_output = "%d %s found" % _make_plural(self._numcollected, "test")
-            else:
-                deselected = self._numcollected - selected
-                co_output = f"{selected}/{self._numcollected} tests found ({deselected} deselected)"
-
-            parts = [(co_output, {main_color: True})]
-
-            # Sticking with "0/0 tests found (0 deselected)" would be confusing.
+            # No tests
             if self._numcollected == 0:
                 parts = [("no tests found", {_color_for_type_default: True})]
 
-            # Sanity check for errors that might have ocurred. Otherwise, it
-            # will never let the user know that an error ocurred during
+            elif deselected == 0:
+                collected_output = "%d %s found" % _make_plural(
+                    self._numcollected, "test"
+                )
+                parts = [(collected_output, {main_color: True})]
+            else:
+                all_deselected = self._numcollected == deselected
+                if all_deselected:
+                    collected_output = f"no tests matched ({deselected} deselected)"
+                else:
+                    selected = self._numcollected - deselected
+                    collected_output = f"{selected}/{self._numcollected} tests matched ({deselected} deselected)"
+                parts = [(collected_output, {main_color: True})]
+
+            # Sanity check for errors that might have occurred. Otherwise, it
+            # will never let the user know that an error occurred during
             # collection.
             if errors:
                 color = _color_for_type.get(key, _color_for_type_default)
                 markup = {color: True, "bold": color == main_color}
-                parts = [("%d %s" % _make_plural(errors, "error"), markup)]
+                parts += [("%d %s" % _make_plural(errors, "error"), markup)]
 
         return parts, main_color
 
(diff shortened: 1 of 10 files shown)


caramelomartins:
It is okay @nicoddemus I just wasn't sure if was supposed to ask someone for a review or simply wait. 😂

src/_pytest/terminal.py
```diff
@@ -1178,9 +1179,22 @@ def build_summary_stats_line(self) -> Tuple[List[Tuple[str, Dict[str, bool]]], s
...
 
-        if not parts:
+                if key in ["deselected"]:
+                    selected -= count
+
+        if not parts and not self.config.getoption("collectonly"):
             parts = [("no tests ran", {_color_for_type_default: True})]
 
+        if self.config.getoption("collectonly"):
+            co_output = "%i/%i tests found (%i deselected)"
```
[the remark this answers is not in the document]
caramelomartins: Yeah, I agree this is an edge case that is missing. I thought about it this morning but your were too fast in reviewing the PR. 🙂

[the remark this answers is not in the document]
caramelomartins: @nicoddemus I have just updated the PR to reflect what we discussed here. Let me know how it looks now and if you'd like me to make more changes.

nicoddemus:
Hi @caramelomartins!

This looks good, but I noticed we didn't have a specific test for all the cases we wanted to handle, so I added it. While at it decided to refactor the function because it had grown too much functionality and it was hard to follow.

For my part this is now good to go, I will wait a while for other maintainers to take a look if they want to. 👍

src/_pytest/terminal.py
```diff
@@ -1165,24 +1165,90 @@ def _set_main_color(self) -> None:
...
+            return self._build_collect_only_summary_stats_line()
+        else:
+            return self._build_normal_only_summary_stats_line()
+
+    def _get_reports_to_display(self, key: str) -> List[Any]:
+        """Get test/collection reports for the given status key, such as `passed` or `error`."""
+        reports = self.stats.get(key, [])
+        return [x for x in reports if getattr(x, "count_towards_summary", True)]
+
+    def _build_normal_only_summary_stats_line(
```
bluetech, line 1197: The name "normal only" is a bit funny to me. I understand it is supposed to match the "collect only" but I would drop the "only" part here.

src/_pytest/terminal.py
```diff
@@ -1165,24 +1165,90 @@ def _set_main_color(self) -> None:
...
+        self,
+    ) -> Tuple[List[Tuple[str, Dict[str, bool]]], str]:
+        main_color, known_types = self._get_main_color()
         parts = []
+        errors = 0
```
bluetech, line 1202: I might be missing it but I don't see where `errors` is used.

bluetech (commented):
Nice improvement @caramelomartins. LGTM with a couple of comments inline, and also, I would change "X tests found" to "X tests collected" since that's the terminology we use.

src/_pytest/terminal.py
nicoddemus: Ahh ooops that was a mistake, it was supposed to be `_build_normal_summary_stats_line` or `_build_standard_summary_stats_line`, the "only" part got there by copy/paste. Fixed. 👍

src/_pytest/terminal.py
nicoddemus: Good catch, left over from the refactoring, removed. 👍

nicoddemus:
> I would change "X tests found" to "X tests collected" since that's the terminology we use.

Done. I did change to "matched" to convey when there was deselection, but it didn't work so well in the end. "collected" is perfect, don't know why it didn't occur to me, thanks!

All changes applied @bluetech. 👍 

(I didn't bother changing the docs because they will be updated next time we run regendoc during the release).

bluetech (approved):

nicoddemus:
Thanks again @caramelomartins!

caramelomartins:
@nicoddemus and @bluetech thanks for accepting this PR. Your suggestions and refactor made it much cleaner, thanks for the effort of making it better. :slightly_smiling_face:

PR #7877, opened 2020-10-09
better handle errors raised while applying filterwarnings (#7864)

### LINKED ISSUE #7864: warnings: better handle errors raised while applying filterwarnings
Refs: https://github.com/pytest-dev/pytest-django/issues/874

The `apply_warning_filters()` function, called during pytest initialization to parse `filterwarnings` ini among others, imports the module of the indicated warning type, which can raise. If the import fails, pytest crashes with an ugly traceback (see referenced issue). We should probably catch the exception and raise a nicer message (while still displaying the original traceback somehow).

charlesaracil-ulti:
closes #7864

Add a more explicit error when raising exception while trying to import filterwarnings.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -51,6 +51,7 @@ Cal Leeming
 Carl Friedrich Bolz
 Carlos Jenkins
 Ceridwen
+Charles Aracil
 Charles Cloud
 Charles Machalow
 Charnjit SiNGH (CCSJ)
--- a/changelog/7864.improvement.rst
+++ b/changelog/7864.improvement.rst
@@ -1,0 +1,1 @@
+Better handle errors raised while applying filterwarnings.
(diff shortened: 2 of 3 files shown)


bluetech (changes requested):
Thanks for working on this @charlesaracil-ulti.

Some comments:

1. It would be good if the error says which warning filter causes the problem.

2. The function `parse_warning_filter` can fail in other places than just the `warnings._getcategory` call. It would be good to handle all sort of invalid warning filters while we're at it.

3. Right now the exception would be raised "lazily", i.e. only is something actually triggers it -- which might be a config-time warning, or in the likely case that there isn't one, during the first test setup. We should maybe consider doing a "pre-check" so it's always triggered consistently and then can be assumed to be valid. Though it will slow down startup a tiny bit...

4. It would be nice (though not absolutely necessary) to format the exception a little more nicely, like a `ConftestImportFailure` is formatted for example. Might not be worth the effort though...

When I get some time I'll try to provide some pointers if needed.

charlesaracil-ulti:
Thanks @bluetech!
I'll be able to work on that next Friday, although I'm quite new to this project and I'll be happy to have some pointers anyway :)

charlesaracil-ulti:
Didn't have as many time today as I expected, still addressed your 1. to get one thing done, will get back to it asap

PR #7886, opened 2020-10-11
Add example for registering multiple custom mark

### LINKED ISSUE #7880: docs: Explain how to register multiple custom markers
In the [docs on registering markers](https://docs.pytest.org/en/latest/example/markers.html#registering-markers), it explains how to register a single custom marker but not how to register multiple custom markers.

It would be great if there were a sentence like, "To register multiple custom markers, put each one on its own line in your `pytest.ini` file," with or without an example.

gutsytechster:
This fixes #7880

### DIFF
--- a/doc/en/example/markers.rst
+++ b/doc/en/example/markers.rst
@@ -221,23 +221,18 @@ Registering markers for your test suite is simple:
     [pytest]
     markers =
         webtest: mark a test as a webtest.
-
-Multiple custom markers can be registered, by defining each one in its own line within the ``pytest.ini`` file. For e.g.
-
-.. code-block:: ini
-
-    # content of pytest.ini
-    [pytest]
-    markers =
-        webtest: mark a test as a webtest.
         slow: mark test as slow.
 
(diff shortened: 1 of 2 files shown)


gutsytechster:
Let me know if this requires an entry in the changelog. :)

bluetech (commented):
Thanks @gutsytechster!

To keep things a bit shorter, how about adding a second marker to the first example, instead of adding a second example? That, I think, would be sufficient for getting across how multiple markers are registered.

doc/en/example/markers.rst
```diff
@@ -222,6 +222,16 @@ Registering markers for your test suite is simple:
...
+
+.. code-block:: ini
+
+    # content of pytest.ini
+    [pytest]
+    markers =
+        webtest: mark a test as a webtest.
+        slow: mark test as slow.
+
 You can ask which markers exist for your test suite - the list includes our just defined ``webtest`` markers:
```
bluetech, line 235: This sentence (and the next code block) would need a little update as well.

bluetech (approved):
Thanks!

PR #7898, opened 2020-10-14
basic collect by line number implementation based on ast parsing

### LINKED ISSUE #3224: Allow test selection by file:lineno

mgaitan:
Closes #3224 . 

This is a very raw implementation that find the test to collect given a line number, by traversing the code via `ast` parsing. 

Doc and tests to be added.

### DIFF
--- a/src/_pytest/parser.py
+++ b/src/_pytest/parser.py
@@ -1,0 +1,107 @@
+"""
+Find a test class and/or test method in given source code line.
+
+Original code by Pavel Savchenko
+"""
+import ast
+
+class PytestParser(ast.NodeVisitor):
+    """
+    Given <source>, extract the top level class/function which contains given
+    line number.
+
+    Optional arguments:
+        * <ignore_bases> list of class base names to be skipped
+
+    """
+    nested_class = None
+
+    def __init__(self, source, ignore_bases=None):
+        self.source = source
+        self.ignore_bases = ignore_bases or []
+
+    def find(self, line):
+        self.nearest_class = None
+        self.nearest_func = None
+        self.nearest_ignored = None
+        self.lineno = line
+        tree = ast.parse(self.source)
+        self.visit(tree)
+        return [part.name for part in (self.nearest_class, self.nearest_func) if part]
+
+    def should_stop(self, node):
+        assert self.lineno, 'line number required'
+        return node.lineno > self.lineno
+
+    def ignore_class(self, node):
+        if not hasattr(node, 'bases'):
+            return False
+
+        def should_ignore(base):
+            if hasattr(base, 'id'):
+                return base.id in self.ignore_bases
+            elif hasattr(base, 'attr'):
+                return base.attr in self.ignore_bases
+            raise Exception('Unknown base node %s' % base)
+
+        return any(should_ignore(base) for base in node.bases)
+
+    def inside_class(self, node):
+        """
+        Test whether node is inside a class definition.
+        Test ignores classes whose base is in self.ignore_bases.
+        """
+        return (
+            # verify none of the bases are in ignored bases
+            not self.ignore_class(node) and
+            # check if currently already inside a class
+            self.nearest_class and
+            # compare col offset to check if we unindented
+            node.col_offset > self.nearest_class.col_offset and
+            # also make sure we're not currently in an ignored class
+            not self.nearest_ignored
+        )
+
+    def first_argument_is_self(self, node):
+        if not node.args.args:
+            return False
+        first_arg = node.args.args[0]
+        arg_name = getattr(first_arg, 'arg', getattr(first_arg, 'id', None))
+        return arg_name == 'self'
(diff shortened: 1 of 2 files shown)


nicoddemus:
I like the idea, thanks a lot @mgaitan! The parser implementation also is isolated and seems simple to maintain.

@bluetech @asottile @RonnyPfannschmidt @The-Compiler can we give @mgaitan the green light to continue?

src/_pytest/parser.py
```diff
@@ -0,0 +1,107 @@
+"""
+Find a test class and/or test method in given source code line.
+
+Original code by Pavel Savchenko
+"""
+import ast
+
+class PytestParser(ast.NodeVisitor):
```
nicoddemus, line 8: nitpick, but `PytestParser` doesn't tell much about what the class does, perhaps `TestLocator`?

I know you mentioned that we need to add docs already, just to highlight that we should have unittests for this class as well, in addition to 1 or more integration tests of the feature. 👍

asottile:
at first glance this seems like a lot of complexity to take on, and it is something that will get much more complicated with time (for instance, it's incorrect for decorators and blank lines between functions right now)

there's also some things it attempts to discovery based on static information that may be indeterminable (base class checking for instance isn't really achievable statically without solving a halting problem -- the normal case of importing from a conventional place sure, but that variable could come from anywhere) -- although it looks like the `ignore_bases` code is dead code?

ast parsing is also relatively expensive (and would happen twice in this case, delaying start by a bit) -- this is probably not a _huge_ deal

there's also a lot of tricky edge cases with classes that are difficult to handle:

```python
class C:
    if True:
        def f(self): ...  # this is a method

if True:
    def f(self): ...  # this is not
```

I think the biggest problem with this is when a test is defined outside of the module being run (for example, base class tests) -- this discovery method wouldn't work at all

I guess overall I'm -.5 on this

The-Compiler:
I'm also not sure about this. On one hand, it seems like a great path down a rabbit hole we'd maybe rather not go down. On the other hand, it doesn't need to be perfect and handle all imaginable corner cases to be useful.

nicoddemus:
> it doesn't need to be perfect and handle all imaginable corner cases to be useful.

Indeed, but I really didn't study the patch to the length @asottile did, and given he's -5 on it I guess we should drop this then.

Thanks a lot @mgaitan for the contribution, but we will be turning this down for now. 👍

The-Compiler:
@nicoddemus -.5 means -0.5 rather than -5 FWIW :wink:

nicoddemus:
Ahhh I've read `-5` sorry hehehe

Well let's see if others manifest themselves, we can always reopen if others are favorable to the idea.

mgaitan:
Let me try to defend a bit the approach, giving some answers to @asottile's points 

> at first glance this seems like a lot of complexity to take on, 

It's true that current implementation looks a bit more complicated than necessary and it needs to be polished (I've just [copied](https://github.com/asfaltboy/SublimeTestPlier/blob/develop/test_parser.py) it) , but a NodeVisitor's subclassing it's a standard, documented way to traverse the code for goals like this one. I mean, it's not an ad-hoc algorithm powered by tricky regex hard to maintain. 

> there's also some things it attempts to discovery based on static information that may be indeterminable

This is also true, but I think in this feature as a shortcut to call specific tests implemented as a function or method in an explicit way. 

For instance, if I pass a line that belong to a function that is not a test (or is one that won't be collected) that should be ok (at least in its first implementation), because the responsibility of this should be just the "mapping" and not to understand the full semantic of what the user is trying to execute. 

> ast parsing is also relatively expensive

as the main use case of this is run specific tests I guess the penalty could be dismissable. Moreover, there is a very simple improve to do, because we could ignore earlier all the nodes that ends before the given line number. 

> I think the biggest problem with this is when a test is defined outside of the module being run

I don't get this, sorry. How you could run, giving a module and a line number of that module, a test defined in other module from which the target class inherits from?  I mean, if I have  `test_base.TestBase.test_foo` defined in a module, and `test_specific.TestChild.test_bar` (where `TestChild` inherits from `TestBase`) I would be able to run  `test_specific:XX` to run `test_bar` but there is no way to run `test_foo` because there is no line number in that module that maps to that method.

bluetech:
I think a better implementation strategy is not to go the ast route to go from line -> class/function, but to go from node -> line range and filter on that. But I haven't looked at this in detail.

Aside from that, I'm somewhat interested on what is the use case for this feature. Maybe a "Run test at this line" editor action?

PR #7907, opened 2020-10-17
Update mypy 0.782 -> 0.790, iniconfig typing

bluetech:
Please see the two commits.

### DIFF
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -382,7 +382,7 @@ class NFPlugin:
             self.cached_nodeids.update(item.nodeid for item in items)
 
     def _get_increasing_order(self, items: Iterable[nodes.Item]) -> List[nodes.Item]:
-        return sorted(items, key=lambda item: item.fspath.mtime(), reverse=True)
+        return sorted(items, key=lambda item: item.fspath.mtime(), reverse=True)  # type: ignore[no-any-return]
 
     def pytest_sessionfinish(self) -> None:
         config = self.config
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -77,8 +77,7 @@ def pytest_terminal_summary(terminalreporter: "TerminalReporter") -> None:
                 dlist.append(rep)
     if not dlist:
         return
-    dlist.sort(key=lambda x: x.duration)  # type: ignore[no-any-return]
-    dlist.reverse()
+    dlist.sort(key=lambda x: x.duration, reverse=True)  # type: ignore[no-any-return]
     if not durations:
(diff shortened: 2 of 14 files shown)


src/_pytest/runner.py
```diff
@@ -77,7 +77,7 @@ def pytest_terminal_summary(terminalreporter: "TerminalReporter") -> None:
                 dlist.append(rep)
     if not dlist:
         return
-    dlist.sort(key=lambda x: x.duration)
+    dlist.sort(key=lambda x: x.duration)  # type: ignore[no-any-return]
     dlist.reverse()
```
Zac-HD, line 81:
```suggestion
    dlist.sort(key=lambda x: x.duration, reverse=True)  # type: ignore[no-any-return]
```

src/_pytest/cacheprovider.py
```diff
@@ -382,7 +382,7 @@ def pytest_collection_modifyitems(
             self.cached_nodeids.update(item.nodeid for item in items)
 
     def _get_increasing_order(self, items: Iterable[nodes.Item]) -> List[nodes.Item]:
-        return sorted(items, key=lambda item: item.fspath.mtime(), reverse=True)
+        return sorted(items, key=lambda item: item.fspath.mtime(), reverse=True)  # type: ignore[no-any-return]
```
Zac-HD, line 385: I presume this ignore is because Mypy doesn't know either that `item.fspath` is an `Optional[py.path.local]`, or that `py.path.local.mtime()` returns a number.  Probably because the `Optional` case isn't handled here, which we might want to fix?

Zac-HD (approved):
Two nitpicks below, but LGTM :+1:

src/_pytest/runner.py
bluetech: Good catch!

src/_pytest/cacheprovider.py
bluetech: Right now `item.fspath` is typed `Union[py.path, Any]` due to some `getattr` when it's assigned. Strictly it should be `Optional[py.path]`, because there are some nodes in which it can be `None`. But then there are a lot of places which assume it isn't (which can indeed be assumed for most nodes), and I am too lazy to sort this out just yet :)

Would be a good PR if someone wants to try it.

bluetech:
ubuntu-py37 job looks like it's going to timeout on the checkout step, unrelated, so ignoring that.

src/_pytest/cacheprovider.py
Zac-HD: Can you open an issue for that, so people know it's a PR they could make?

PR #7916, opened 2020-10-19
Changed the font of Security header to keep it inline with other headers

naman1303:
Increased the font of **Security** header in README.rst file to make all the headers look uniform while reading. 
I have tested and previewed the file before pushing it.


asottile:
I believe this is intentional, it's an extension of the tidelift section above it

it was added in one go here: https://github.com/pytest-dev/pytest/commit/dae455e8a32810aa3035ac3c5f2f802a06b851d3

PR #7921, opened 2020-10-21
Add plugin list

### LINKED ISSUE #5105: docs: plugins page no longer exists
It seems that https://docs.pytest.org/en/latest/plugins_index/index.html page no longer exists but is listed in the main index at https://docs.pytest.org/en/latest/

List of plugins exists for older releases but not for the latest / main documentation.

mcsitter:
## Features

As described in pytest-dev/pytest#5105 and mcsitter/pytest-plugin-list#1 this adds a pytest plugin list.

Previously a list of pytest plugins was retrieved from the package index and integration tests were performed. The table showed success and failure of the integration tests of the plugin for different python versions. This did not prove to be very useful so a simpler approach is desired.

A pytest plugin is defined as any package that matches the pattern pytest-*. Packages with the development status set to inactive as indicated by the trove classifier of the package are excluded.

## Columns

The plugin list contains the following columns:

- name of the plugin
- summary of the plugin functionality
- compatible python versions as indicated by the shield
- required pytest versions
- last release date

## Updates

A workflow runs `plugin_list.py` daily and automatically opens a PR for the updates. This PR should simply be merged automatically if the docs tests pass. If the PR is not merged, changes from future runs will be force pushed to the PR instead of opening a new one.

Closes #5105


scripts/plugin_list.py
```diff
@@ -0,0 +1,85 @@
+import datetime
```
nicoddemus, line 1: For consistency with the other scripts, let's use `-` as separator here.

Also nitpick of course, but how about naming this script `update-plugins-list.py`?

.github/workflows/plugin-list.yml
```diff
@@ -0,0 +1,32 @@
...
+          python-verion: 3.8
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install packaging requests tabulate[widechars]
+      - name: Update Plugin List
+        run: python scripts/plugin_list.py
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
```
nicoddemus, line 26: The PR opened by this will target `master`, correct?

`master` only generates the documentation for https://docs.pytest.org/en/latest, while `stable` points to the latest tagged release, and by default users are directed to the `stable` version when typing `docs.pytest.org`.

Would it be possible to target the latest released branch somehow? Otherwise the updated list of plugins will only be available on `latest`, and the default location reached by users will be outdated in general.

scripts/plugin_list.py
```diff
@@ -0,0 +1,85 @@
...
+            "last release": last_release,
+            "status": status,
+            "requires": requires,
+        }
+
+
+def main():
+    plugin_table = tabulate.tabulate(iter_plugins(), headers="keys", tablefmt="rst")
+    plugin_list = pathlib.Path("doc", "en", "plugin_list.rst")
+    plugin_list.parent.mkdir(parents=True, exist_ok=True)
```
nicoddemus, line 77: That should not be necessary because `doc/en` always exist in the repository, and probably it is best to avoid creating those if this script is executed somewhere else other than the root of the repository.

doc/en/_templates/links.html
```diff
@@ -2,7 +2,6 @@ <h3>Useful Links</h3>
 <ul>
   <li><a href="https://pypi.org/project/pytest/">pytest @ PyPI</a></li>
   <li><a href="https://github.com/pytest-dev/pytest/">pytest @ GitHub</a></li>
-  <li><a href="http://plugincompat.herokuapp.com/">3rd party plugins</a></li>
```
nicoddemus, line 5: While it makes sense to remove this from the external links list, I would still like to see a link to this page in the "Table of Contents" side bar:

image

scripts/plugin_list.py
```diff
@@ -0,0 +1,85 @@
...
+import packaging.version
+import requests
+import tabulate
+
+FILE_HEAD = r"""Plugins List
+============
+
+The plugins are listed automatically from PyPI.
+Only PyPI projects that match "pytest-\*" are considered plugins.
+Packages classified as inactive are also excluded.
```
nicoddemus, line 14: It would be interesting to have a count of the number of the plugins in the page, like we have currently with the plugincompat site:

image

(Just the plugin count is enough, no need to also show the latest version)

scripts/plugin_list.py
```diff
@@ -0,0 +1,85 @@
...
+                    break
+        releases = response.json()["releases"]
+        for release in sorted(releases, key=packaging.version.parse, reverse=True,):
+            if releases[release]:
+                release_date = datetime.date.fromisoformat(
+                    releases[release][-1]["upload_time_iso_8601"].split("T")[0]
+                )
+                last_release = release_date.strftime("%b %d, %Y")
+        name = f'`{info["name"]} <{info["project_url"]}>`_'
+        pyversions = f'.. image:: https://img.shields.io/pypi/pyversions/{info["name"]}'
```
nicoddemus, line 61: The images have varying sizes on the page:

image


Any idea on how to have them all their original sizes instead?

nicoddemus (changes requested):
Thanks @mcsitter!

This looks great for starters. Here's the direct link for the preview:

https://pytest--7921.org.readthedocs.build/en/7921/plugin_list.html

I've left a few comments, please take a look. 👍

scripts/plugin_list.py
```diff
@@ -0,0 +1,85 @@
...
+        else:
+            status = "NA"
+        requires = "NA"
+        if info["requires_dist"]:
+            for requirement in info["requires_dist"]:
+                if requirement == "pytest" or "pytest " in requirement:
+                    requires = requirement
+                    break
+        releases = response.json()["releases"]
+        for release in sorted(releases, key=packaging.version.parse, reverse=True,):
```
aklajnert, line 54: I think it shouldn't be reversed. The column says `last release`, where it actually shows the first release date. So my guess would be that it's the `reverse=True` fault.

mcsitter: Thanks for noticing the wrong displayed information. I fixed the issue! I forgot to add a break statement inside the if.

scripts/plugin_list.py
mcsitter: Done!

doc/en/_templates/links.html
mcsitter: I added it in between `Customize` and `Changelog`. Let me now if that is okay, else pls resolve.

scripts/plugin_list.py
mcsitter: I added the plugin count!

scripts/plugin_list.py
aklajnert: I still see the invalid data. Or does the generated documentation provided by @nicoddemus has to be somehow rebuilt manually?

aklajnert: OK, now I see it updated. All good, thanks.

.github/workflows/plugin-list.yml
mcsitter: As described [here](https://github.com/marketplace/actions/create-pull-request#action-inputs):

base | Sets the pull request base branch. | Defaults to the branch checked out in the workflow.
-- | -- | --

This actually makes a lot of sense to me. We can also pass a ref to the checkout according to its [documentation](https://github.com/marketplace/actions/checkout). It says the ref can be tag (like stable) but i don't know how it would open PR from a detached HEAD. This seems to be a problem to me? Since the [output of commands can be used in the workflow](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#set-an-output-parameter-set-output) one could maybe come up with a way to get the correct branch based on the tag and pass it to the `ref` option of the checkout. I am not sure how this would work in regards to keeping the master branch up-to-date. Maybe a matrix of environment variables has to be created?

scripts/plugin_list.py
mcsitter: Currently the shields are vector graphics be default. Maybe that can be fixed.

mcsitter: 1.1 One solution could be to scrape the dimensions of the shield from the respective link and set the dimensions explicitly using the width and height options of the image directive. This would add 2 lines to the file for each image. Using inline-notation so that the rst is more human readable would mean 3 more lines essentially making the file 4 times as large.
This would also double the requests to websites made by the script.
1.2 Instead of making the request one could manually parse the classifiers and based on the resulting list the dimensions could be defined as well. Parsing of the metadata is part of what makes shields nice though.

2. Shields.io also offers the badges in the PNG format, but the images are not as sharp of course. See for example  the [pytest-django pyversions shield (PNG)](https://raster.shields.io/pypi/pyversions/pytest-django).

Which option preferred? Any other ideas?

.github/workflows/update-plugin-list.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          python-verion: 3.8
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install packaging requests tabulate[widechars]
+      - name: Update Plugin List
+        run: python scripts/update-plugin-list.py
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
```
bluetech, line 25: Can we do it ourselves with https://github.com/actions/github-script? I prefer to avoid using external actions, especially in little-noticed automatic PRs. If it's difficult, then we should lock it to a specific commit so it doesn't auto-update.

scripts/update-plugin-list.py
```diff
@@ -0,0 +1,85 @@
...
+    "Development Status :: 4 - Beta",
+    "Development Status :: 5 - Production/Stable",
+    "Development Status :: 6 - Mature",
+    "Development Status :: 7 - Inactive",
+)
+
+
+def iter_plugins():
+    regex = r">([\d\w-]*)</a>"
+    response = requests.get("https://pypi.org/simple")
```
bluetech, line 28: Never knew such an endpoint exists, which just lists *all* of the PyPI packages. Seems quite non-scalable but if it works, it works :)

scripts/update-plugin-list.py
```diff
@@ -0,0 +1,85 @@
...
+    for match in re.finditer(regex, response.text):
+        name = match.groups()[0]
+        if not name.startswith("pytest-"):
+            continue
+        response = requests.get(f"https://pypi.org/pypi/{name}/json")
+        if not response.ok:
```
bluetech, line 34: If we fail to fetch, I think it is better to fail (or retry and fail) instead of silently skipping.

scripts/update-plugin-list.py
```diff
@@ -0,0 +1,85 @@
...
+        info = response.json()["info"]
+        if "Development Status :: 7 - Inactive" in info["classifiers"]:
+            continue
+        for classifier in DEVELOPMENT_STATUS_CLASSIFIERS:
+            if classifier in info["classifiers"]:
+                status = classifier[22:]
+                break
+        else:
+            status = "NA"
+        requires = "NA"
```
bluetech, line 45:
```suggestion
            status = "N/A"
        requires = "N/A"
```

scripts/update-plugin-list.py
```diff
@@ -0,0 +1,85 @@
...
+        if info["requires_dist"]:
+            for requirement in info["requires_dist"]:
+                if requirement == "pytest" or "pytest " in requirement:
+                    requires = requirement
+                    break
+        releases = response.json()["releases"]
+        for release in sorted(releases, key=packaging.version.parse, reverse=True,):
```
bluetech, line 52:
```suggestion
        for release in sorted(releases, key=packaging.version.parse, reverse=True):
```

bluetech (changes requested):
Really great work @mcsitter! This would be a good addition.

I have some concerns on the current version:

1. Is it possible to do this in a separate repository, and somehow include it in our docs as an external document?

   My main concern is the daily commits. For me as a developer it would be annoying to have our main repo filled with daily commits which update this file.

- The use of the shields as-is is excessive, and needs a different approach. Loading the page sends 790 HTTP requests to img.shields.io, which, if we add it prominently to our docs, would amount to a ddos attack on them. Already for me a lot of the requests fail with status 429 - too many requests.

- One requirement for being automatically picked up as a pytest plugin is having a `pytest11` entry point. That may be more accurate than matching `pytest-*`? I am not sure if it is include din the metadata however.

scripts/update-plugin-list.py
mcsitter: [The XML-RPC API will be deprecated in the future.](https://warehouse.pypa.io/api-reference/xml-rpc.html) Since i do not know any other way to get a full list from pypi, i used this one. Open for suggestions though!

Why would scalability be an issue down the line?

bluetech: It's perfectly fine, I was just very suprised that that is just one single page which lists all packages in PyPI like this.

> Why would scalability be an issue down the line?

My first thought was to look at npm which is the largest package registry around AFAIK, and indeed: https://blog.npmjs.org/post/157615772423/deprecating-the-all-registry-endpoint

Hopefully we won't see a similar post from the PyPI maintainers any time soon :)

.github/workflows/update-plugin-list.yml
mcsitter: Here is the [API reference](https://octokit.github.io/rest.js/v18#pulls-create) which is behind the linked action. The [`README`](https://github.com/actions/github-script) shows, that the action is basically a js script which calls the API. I have no idea how js works and what the environment of the script (provided variables etc.) is, but i am giving it a shot. Seems like i am not getting the `head` right for [this workflow](https://github.com/mcsitter/pytest-plugin-list/runs/1299984596?check_suite_focus=true).

While we are at it, [this](https://github.com/pytest-dev/pytest/pull/7921#discussion_r509431843) discussion is about the target branch of the pull request and could be solved in the same sweep if someone knows how to script this.

bluetech: Your workflow looks right to me. It fails because `head` is `invalid`. I think this is because the branch name `update-plugin-list` doesn't exist in your repo. You should run a previous step which creates it with a `git push --force` command. Also there needs to be some handling of the case where a PR for the branch is already open -- in this case can just reuse it I suppose.

.github/workflows/plugin-list.yml
nicoddemus: > I am not sure how this would work in regards to keeping the master branch up-to-date.

If we can make it work by just targeting the latest stable, IMHO that's good enough.

scripts/plugin_list.py
nicoddemus: Hmm as @bluetech mentions in https://github.com/pytest-dev/pytest/pull/7921#pullrequestreview-515705510, actually just referencing the outside images is problematic in this case. Perhaps the script can download the images when running and store them into the pytest's repo?

scripts/update-plugin-list.py
mcsitter: The first case in the list that will be an issue will be pytest-azurepipelines42. [This](https://pypi.org/project/pytest-azurepipelines42) would be the project page and [this](https://pypi.org/pypi/pytest-azurepipelines42/json) is the place for its JSON API. I do not know why they are empty, but here is a complete list of the current cases which pass the condition with the link to the API:
[pytest-azurepipelines42](https://pypi.org/pypi/pytest-azurepipelines42/json), [pytest-core](https://pypi.org/pypi/pytest-core/json), [pytest-locust](https://pypi.org/pypi/pytest-locust/json), [pytest-locustio](https://pypi.org/pypi/pytest-locustio/json), [pytest-regression](https://pypi.org/pypi/pytest-regression/json), [pytest-runtime-info](https://pypi.org/pypi/pytest-runtime-info/json), [pytest-testraill](https://pypi.org/pypi/pytest-testraill/json)

I would prefer not to have a blacklist. Maybe there will be working available packages in the future. This is not really failing to fetch, since there actually is nothing to fetch. I understand, that this kind of handling may cause problems, but i don't know how exactly, since i believe connectivity issues like timeouts will throw an error in the row before.

scripts/plugin_list.py
mcsitter: Do you mean having something like a `pyversions` folder in the docs with one shield for each of the almost 800 plugins or something closer to `1.2` of my previous comment? Storing the shields in the repo is probably a must either way.

If the same image is embedded multiple times in a single html document, will there be multiple requests for the same image?

mcsitter:
Thank you for your feedback, @bluetech! I hope we can make this a great feature 👍 

>     1. Is it possible to do this in a separate repository, and somehow include it in our docs as an external document?
>        My main concern is the daily commits. For me as a developer it would be annoying to have our main repo filled with daily commits which update this file.

I understand the concern for the "spam" of commits. I share that concern. Maybe a weekly update would solve that to a satisfactory degree? See the linked discussions in [the PR body](https://github.com/pytest-dev/pytest/pull/7921#issue-507493481) for the arguments made previously.

An external repository has a lot of overhead for what this list aims to achieve. Also the state of the plugin list for specific releases of pytest and thus the documentation would not be available, only the current version, which may not be that useful if say you are still on pytest 5 and wanted to see what plugins are available, but the popular ones already updated the requirement for the latest release. The idea is to have 1 script and 1 workflow to manage the whole thing. Of course this is a pull **request** and this should be discussed to find the preferred solution.

>     * The use of the shields as-is is excessive, and needs a different approach. Loading the page sends 790 HTTP requests to img.shields.io, which, if we add it prominently to our docs, would amount to a ddos attack on them. Already for me a lot of the requests fail with status 429 - too many requests.

I think we will tackle this according to the proposals in [this](https://github.com/pytest-dev/pytest/pull/7921#discussion_r509436872) discussion. I only encountered this issue once i reloaded the page a couple times within a short amount of time and did not think about it before. Of course we should try to minimize the load we put on their page.


If anyone knows how we could also reduce the requests to the pypi API ideas are welcome! Should not be that much of a problem though, since the requests are only done in the workflow. We could even add a sleep, but it would greatly increase the duration of the run.

>     * One requirement for being automatically picked up as a pytest plugin is having a `pytest11` entry point. That may be more accurate than matching `pytest-*`? I am not sure if it is include din the metadata however.

This information does not seem to be available in the metadata of the JSON API. If that metadata is available somewhere else it could be considered. The list will never be perfect though. For example the `Framework :: Pytest` classifier is basically useless without some way to filter for the package being a plugin instead of using pytest for testing. [`allure-pytest`](https://pypi.org/project/allure-pytest/) is not included in the list (has the classifier btw), just the legacy names `pytest-allure-adaptor` and `pytest-allure-adaptor2` are in the list, which probably won't get a new release.

scripts/plugin_list.py
nicoddemus: > One solution could be to scrape the dimensions of the shield from the respective link and set the dimensions explicitly using the width and height options of the image directive. This would add 2 lines to the file for each image. Using inline-notation so that the rst is more human readable would mean 3 more lines essentially making the file 4 times as large.

If it looks nice, I don't think the length of the rst file is an issue, users probably won't be reading it anyway, only looking at the rendered page.

> Do you mean having something like a pyversions folder in the docs with one shield for each of the almost 800 plugins or something closer to 1.2 of my previous comment?

I meant the former. Shouldn't be an issue as they are all very small SVG files.

> If the same image is embedded multiple times in a single html document, will there be multiple requests for the same image?

If they share the same filename, the browser makes only a single request.

asottile (commented):
this seems like... a lot (complexity, frequent commits/PRs, etc.)

could we just link to this instead? https://pypi.org/search/?c=Framework+%3A%3A+Pytest

nicoddemus:
> could we just link to this instead? pypi.org/search/?c=Framework+%3A%3A+Pytest

Unfortunately the "Framework :: Pytest" has been incorrectly used by many projects to indicate they _use_ pytest, not to identify themselves as plugins, which is unfortunate as completely defeated the purpose of the classifier.

But you have a point, what do we get from this that we don't get from a simple link to a search, say https://pypi.org/search/?q=pytest-&o=?

From the top of my head:

1. It returns too many false positives (over 2.5k+ projects, because the search engine searches for "pytest" anywhere in the name I suspect).
2. It only shows a handful of results per page. For example searching "pytest django" we get 10k hits, making it hard to find something useful if you are looking for django plugins. Having them all in the same page would help this.
3. I think this is just a first implementation, I can envision we improving the page a later to show a bunch of popular plugins first, for example.

Of course, we have to balance all the benefits with the downsides, but TBH a simple script and a cron job don't seem to me like it would bring too much of a burden (and we can always revisit this in the future if it does).

asottile:
this search seems way better: https://pypi.org/search/?c=Framework+%3A%3A+Pytest&q=pytest

scripts/plugin_list.py
mcsitter: I will have a go at implementing it so that the badges are part of the pytest docs and referenced locally. Quite busy atm, but should be done within a week.

asottile: 800 checked in and rapidly changing ~essentially binary files does not seem like a good idea to me.  it will quickly become a significant portion of the git history size

The-Compiler: An update script could deduplicate them before adding them to the repo - but I wouldn't find it too crazy to let git take care of that either.

Out of curiosity, I downloaded all 780 of them. They have a total size of 3.1 MB (1.1 to 1.2 KB each), but there are only 77 deduplicated variants.

If I commit them all without manual deduplication, I end up with a `.git` which is 240K - compared with 164K for an empty git repo (!) or 25M for pytest's current `.git`, that really doesn't seem like a lot. Also I don't see how they'd be rapidly changing - I'd expect a change once all couple of months.

asottile: they contain the package version (probably changes every few weeks) and the target python versions (probably changes twice a year) -- I think that's pretty often

aklajnert: Can't we just capture their text, and generate them on the spot in the docs build process with something like `anybadge`?  
Or we can just have a column for each python version supported by the pytest and just put `X` or `✓` there

The-Compiler: > they contain the package version 

I don't see a package version:

image

nicoddemus: > Or we can just have a column for each python version supported by the pytest and just put X or ✓ there

This can actually be a solution to another issue, which is the varying size of the columns because to the image dimensions.

nicoddemus:
Also realized we can later update the workflow to also run `regendoc`, so it would become actually a periodic "docs updater", which is something we have wanted for some time now. 👍

nicoddemus:
Any updates from the commenters on this? I'm still 👍 on the overall idea, barring some changes that are being discussed.

scripts/plugin_list.py
mcsitter: Storing the images locally was easy to implement, but i can't come up with a solution for the image size.
There are packages with tons of "compatible" python versions. Check out [pytest-pythonpath](https://img.shields.io/pypi/pyversions/pytest-pythonpath) and [pytest-dotenv](https://img.shields.io/pypi/pyversions/pytest-dotenv) and have a look the the rendered docs for the PR. The badges just get crazy long and are impossible to fit onto the page, not even talking about the column. Setting the dimensions of the images in the rst did not really work, but i fixed the scaling issue by setting the columnwidth to the biggest image width. This however means that half the table is the column for the badges, because some are so incredibly white. This means, that it is mostly whitespace, just to accomodate the outliers. Here are some ideas for how to handle this:
1. Do not show compatible python versions at all.
2. Limit the "relevant" versions only show relevant versions in the badge. This is easy, since the badges are already stored locally. I parsed the title of the shield, extracted the versions and then used the api to download unique version combinations into badges named something like `35_36_37_38_39.svg` into `doc/en/img/pyversions/`. We could just ignore anything below 3.5 (current minimal compatble python for pytest) for example. This would arguably still be to long.
3. Just put a link to the shield instead of embedding it.

I am in favor of option 1, since anyone interested can also go to the pypi page of the plugin, which is linked to already and the compatible pytest version can be used for a rough estimate, whether the plugin would be useful to the user.

The-Compiler: Thinking about it some more, :+1: from me as well for getting rid of the column entirely. Like you say, if a plugin supports up-to-date pytest versions, I guess the answer to "will this support Python version xyz I'm interested in" is rather likely to be "yup!"

mcsitter: I removed the column. Reimplementing some kind of pyversion compatibility can be done in a future PR if a manageable way can be found. Resolved!

gnikonorov:
Is this going to deprecate [plugincompat](https://github.com/pytest-dev/plugincompat) @nicoddemus ?

.github/workflows/plugin-list.yml
mcsitter: Maybe someone could help me out with this one. I do not know how to get the name of the branch with the latest release.  I don't think that checking out a tag would work. Also, i do not have the git repo structure set up to test this properly at [mcsitter/pytest-plugin-list](https://github.com/mcsitter/pytest-plugin-list). Just sorting by the tags on pytest currently spits out the dev version of version 6.2.0 as the latest tag . Should i just put `6.1.x` in there for the branch? That would mean that every minor version requires an update to the workflow.

mcsitter:
> Is this going to deprecate [plugincompat](https://github.com/pytest-dev/plugincompat) @nicoddemus ?

I believe so.

nicoddemus:
Yep it will most likely be retired.

gnikonorov:
> Yep it will most likely be retired, unless someone wants to maintain it.

Thanks for confirming! Just curious what benefits would maintaining it have over just using the product of this PR?

nicoddemus:
> Thanks for confirming! Just curious what benefits would maintaining it have over just using the product of this PR?

Not much... the current testing that the plugincompat does today is not very useful I think. Having a list of plugins in a single place seems simple and useful enough.

.github/workflows/update-plugin-list.yml
nicoddemus: I guess this point is still open right?

.github/workflows/plugin-list.yml
nicoddemus: (Sorry I missed the notification for this one)

I think it is safe to filter every tag which doesn't match `\d+\.\d\.x`, so it will leave out the dev tags or any other tag which doesn't match our version scheme for branches.

changelog/5105.doc.rst
```diff
@@ -0,0 +1 @@
+Add automatically generated `plugin_list.rst` and a workflow running on a cronjob for updates.
```
bluetech, line 1:
```suggestion
Add automatically generated `plugin_list.rst`. The list is updated on a periodic schedule.
```

scripts/update-plugin-list.py
bluetech: Really strange that some projects are listed but return 404.

Agree about the blacklist, we shouldn't do that.

What I suggest is something like this:

```py
response = requests.get(f"https://pypi.org/pypi/{name}/json")
if response.status_code == 404:
    # Some packages, like pytest-azurepipelines42, are included in https://pypi.org/simple but
    # return 404 on the JSON API. Skip.
    continue
response.raise_for_status()
```

.github/workflows/update-plugin-list.yml
bluetech: If the github-script or any other roll-our-own solution is too much work, we can keep using `peter-evans/create-pull-request@v3`, but we should at least pin it to a specific commit instead of a tag, so if it's somehow maliciously updated, we won't be affected.

.github/workflows/update-plugin-list.yml
```diff
@@ -0,0 +1,32 @@
+name: Update Plugin List
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
bluetech, line 5: For those of us who can't remember cron syntax even after using it a thousand times :)

```suggestion
    # Run daily at midnight.
    - cron: '0 0 * * *'
```

bluetech (changes requested):
I think is an improvement over the status quo so we should go ahead and merge this, just a couple of comments.

nicoddemus:
Howdy folks.

Sorry I lost track here, anything TODO here that is preventing this from getting merged?

bluetech:
My two small comments above still need to be fixed, I can handle them if @mcsitter prefers.

mcsitter:
Help on the workflow file is [very welcome](https://github.com/pytest-dev/pytest/pull/7921#discussion_r531698472)! Fixed the reponse code handling.

bluetech (approved):
OK, pushes a couple changes to pin the github action and fix a doc reference, other than that it's good! I'm assuming also from a comment above the @nicoddemus approves as well, so I'll go ahead and merge.

Thanks for this effort @mcsitter!

mcsitter:
Thanks for the feedback and helping out! I am happy to contribute.

nicoddemus:
Cool! Thanks everyone, specially to @mcsitter for the contribution and patience to get this in! 👍

PR #7930, opened 2020-10-23
pytester: minor doc fixes

bluetech:


nicoddemus (approved):
Thanks for catching those!

PR #7934, opened 2020-10-24
Replace term 'triple' with 'tuple' in hookspec docstrings

symonk:
Hey, very minor adjustment - was reading through hookspecs to build my understanding and noticed the `triple` term used in place of `tuple`, thought I should make that technically correct.

Thanks - have a wonderful weekend 👍


nicoddemus (approved):
Thanks!

PR #7935, opened 2020-10-24
Small refactor debugging.py: removed code duplicaton

proggga:
removed code duplicaton in _pytest/debugging.py

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -207,6 +207,7 @@ Mihai Capotă
 Mike Hoyle (hoylemd)
 Mike Lundy
 Miro Hrončok
+Mikhail Fesenko
 Nathaniel Compton
 Nathaniel Waisbrot
 Ned Batchelder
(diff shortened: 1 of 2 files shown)


proggga:
About code coverage, I removed big copypaste code, so present of coverage decreased, but actually nothing happend

image

bluetech:
Thanks @proggga.

I personally think that the previous code is a little easier to follow. It has some duplication but it's not too bad.

proggga:
> Thanks @proggga.
> 
> I personally think that the previous code is a little easier to follow. It has some duplication but it's not too bad.

If its not problem, pls add 'hacktoberfest-accepted' label

proggga:
And one more question, to you, asking like you a maintainer
why we can't merge this one? this looks little better (for me), if its the same, why we won't merge it ? idk

bluetech:
One reason is that I somewhat dislike format strings like `PDB continue (IO-capturing resumed%s)` because I often grep for certain string patterns I see as output in order to find where they are defined in the code, and this breaks it.

If you'd like to contribute to pytest, may I suggest issue #7942? You can pick any file you want, and any help here would be appreciated!

bluetech:
Per the comments I'll close this now, but thanks for trying to improve the code!

Issue #5147, opened 2019-04-19
Configuring discovery for paths with namespace packages


0cjs:
The structure of Python namespace packages cannot be determined just by looking at the filesystem. Consider the following directory tree:

    src/
        core.py
        aaa/
            test_core.py
        bbb/
            core.py
            test_core.py

If `src/` (but nothing underneath it) is in `sys.path`, we have the following modules:

    core
    aaa
    aaa.test_core
    bbb
    bbb.core
    bbb.test_core

where `aaa` and `bbb` are namespace modules. We might also have other parts of the namespace modules outside of this tree, e.g. by adding `othersrc/` to `sys.path` so that `import bbb.other` will work if `othersrc/bbb/other.py` exists.)

It's easy to configure pytest to have `src` in the path, but when discovering and loading `src/aaa/test_core.py`, [as described here][tpn], it adds `src/aaa` to `sys.path` and imports `src/aaa/test_core.py` using a different module structure than the programmer may have intended: `import test_core`. Further, the same thing happens again with `src/bbb/test_core.py` and now we get a conflict because there's already a `test_core` module.

It would be nice to be able to configure pytest so that we could tell it that `src/` may contain namespace packages and, if it does, they are all intended to be rooted at `src/`. That is, when discovering `src/{aaa,bbb}/test_core.py` do not add anything to `sys.path` but instead generate the fully qualified module name based on the file's position in the hierarchy relative to `src/`.

Another, more aggressive, option might be to see if the file is under any of the paths already in `sys.path` and determine the fully qualified module name by the hierarchy relative to the first path in `sys.path` under which the file falls.

The Stack Overflow question [How do I Pytest a project using PEP 420 namespace packages?][50174130] also discusses the issue.

The way I put these here makes this sound pretty simple (at least as far as description, if not implementation), but there may be other subtleties I'm missing here, about which I welcome discussion.

[tpn]: https://docs.pytest.org/en/latest/goodpractices.html#test-package-name
[50174130]: https://stackoverflow.com/q/50174130/107294

asottile:
I'd like to see pytest work better in this situation as well.  I think once we can leave python2 behind this becomes even easier (and more common at the same time) as `__init__.py` files become entirely optional.

I'm not quite sure the best way to get to that state without breaking everything though 🤔 

[...]

PR #7936, opened 2020-10-24
[BLOCKED] --import-mode=importlib: use sys.path relative path if present (wait for #7870)

davidhewitt:
This PR is inspired by https://github.com/pytest-dev/pytest/issues/5147#issuecomment-607075562 to use `sys.path` to discover package paths when attempting to import test modules in `importlib` mode.

This helps weird edge-case papercuts such as:
- relative imports in test files, which currently doesn't work with `importlib` mode (see https://github.com/pytest-dev/pytest/issues/7245#issuecomment-686398389)
- implicit namespace packages where the child package name clashes with a builtin module, which I have issues with on _all_ modes (not just importlib).

This patch only takes effect if the file being tested is in the user's sys.path. I think that usually this is only the case if the containing package was installed with `setup.py develop`.

I'm happy to write docs, changelog entry etc., but first wanted to collect feedback on this approach before doing so.

--- 

### DIFF
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -468,13 +468,16 @@ def import_path(
     if mode is ImportMode.importlib:
         module_name = path.stem
 
+        resolved_name = resolve_sys_path_module_name(path)
+        if resolved_name is not None:
+            return importlib.import_module(resolved_name)
+
         for meta_importer in sys.meta_path:
             spec = meta_importer.find_spec(module_name, [str(path.parent)])
             if spec is not None:
                 break
         else:
             spec = importlib.util.spec_from_file_location(module_name, str(path))
-
         if spec is None:
             raise ImportError(
                 "Can't find module {} at location {}".format(module_name, str(path))
@@ -486,10 +489,7 @@ def import_path(
     pkg_path = resolve_package_path(path)
     if pkg_path is not None:
         pkg_root = pkg_path.parent
-        names = list(path.with_suffix("").relative_to(pkg_root).parts)
-        if names[-1] == "__init__":
-            names.pop()
-        module_name = ".".join(names)
+        module_name = module_name_from_relative_path(pkg_root, path)
     else:
         pkg_root = path.parent
         module_name = path.stem
@@ -529,6 +529,34 @@ def import_path(
             raise ImportPathMismatchError(module_name, module_file, path)
 
     return mod
+
+
+def resolve_sys_path_module_name(path: Path) -> Optional[str]:
+    """Return the Python module name of path by looking in sys.path
+    for the first base path encompassing the target.
+
+    Returns None if it can not be determined.
+    """
+    path = path.absolute()
+    for entry in sys.path:
+        try:
+            return module_name_from_relative_path(Path(entry).absolute(), path)
+        except ValueError:
+            pass
+    return None
+
+
+def module_name_from_relative_path(base: Path, path: Path) -> str:
+    """Return the Python module name of path as it would be imported
+    from base.
+
+    :raises ValueError:
+        If path is not a subpath of base.
+    """
+    names = list(path.with_suffix("").relative_to(base).parts)
+    if names[-1] == "__init__":
+        names.pop()
+    return ".".join(names)
 
 
 def resolve_package_path(path: Path) -> Optional[Path]:
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -468,6 +468,9 @@ def import_path(
     if mode is ImportMode.importlib:
         module_name = path.stem
 
+        # If the module exists under a location in sys.path, just import it using that module name.
+        # This is as close to a "normal import" as possible - no need to modify sys.path and the
+        # module will have __package__ set correctly (so relative imports also work).
         resolved_name = resolve_sys_path_module_name(path)
         if resolved_name is not None:
             return importlib.import_module(resolved_name)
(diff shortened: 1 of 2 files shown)


src/_pytest/pathlib.py
```diff
@@ -468,13 +468,16 @@ def import_path(
     if mode is ImportMode.importlib:
         module_name = path.stem
 
+        resolved_name = resolve_sys_path_module_name(path)
```
bluetech, line 471: Could you add a comment above here, explaining with this code block is needed? IIUC, what it does is "If the path is importable as a name from the search path, then import it as such.". But the comment should say why it's done.

Also, AFAIK (not 100% sure) the previous importlib code doesn't insert the imported module into `sys.modules`, but `importlib.import_module` does. I think the idea was to *not* insert it, as mentioned in the `--import-mode=importlib` changelog entry:

> Traditionally pytest used `__import__` while changing sys.path to import test modules (which also changes sys.modules as a side-effect), which works but has a number of drawbacks, like requiring test modules that don't live in packages to have unique names (as they need to reside under a unique name in sys.modules).
>
> --import-mode=importlib uses more fine grained import mechanisms from importlib which don't require pytest to change sys.path or sys.modules at all, eliminating much of the drawbacks of the previous mode.

davidhewitt: Sure thing, very happy to add a comment. I'll push it as a separate commit.

> Also, AFAIK (not 100% sure) the previous importlib code doesn't insert the imported module into sys.modules, but importlib.import_module does.

This is true. I noticed that #7870 was actually proposing to change that though so that the imported module _is_ added to `sys.modules` (to fix the issues noted there).

An interesting thing to note about this approach is that the entry that ends up in `sys.modules` will be the same as if `import foo.bar.test_baz` was executed - e.g. `sys.modules["foo.bar.test_baz"]`.

With the approach in #7870 the concern is that test modules which have conflicting names will again have issues, as described in the changelog entry you quote. (The code in #7870 would add just `test_baz` to `sys.modules`, for example.)

This approach would add the fully qualified name `foo.bar.test_baz` to `sys.modules`, which may be interesting as an alternative approach to the fix in #7870.

bluetech: Thanks for the explanation.

So the approach does sounds appealing, *but* it does bring us again to have a dependence on `sys.path` and how it is configured (e.g. if it contains the test directory, at what depth, etc.). Basically it becomes pretty similar to `--import-mode=append`, no? Just without keeping the `sys.path` dirty.

davidhewitt: Yes, that's a very fair assessment of this approach.

Given that it's perhaps markedly different from the `importlib` mode, perhaps it makes more sense to makes this a new mode, maybe called `path`?

That way it'll have reliable behaviour - if the test file is not underneath a path entry, pytest will not be able to import it. So users who want to test using this mode will have to add the test files to their path (usually by `setup.py develop`).

The advantages of it are primarily for test files inside packages, so I don't think that's an unreasonable requirement.

nicoddemus: Thanks @davidhewitt for the patch! This is complicated topic so we are very grateful for the attempt.

This is an interesting approach, but doesn't handle one of the main uses of `--importmode=importlib`, which is using it to collect test files which are not in `sys.path` at all. Consider this common layout:

```
src/
   analysis/

tests/
   io/
      test_measurement.py
   model/
      test_measurment.py 
```


In this case `src` is added to `sys.path` due to `python setup.py develop` (or adding to `$PYTHONPATH` manually), however we don't have any parent of `tests/io/test_measurement.py` and `tests/model/test_measurement.py` in `sys.path`, which wouldn't trigger your patch and would still cause the issues with `pickle`/`dataclasses` mentioned in the original issue.

Your patch would work fine though when the tests are siblings of an entry in `sys.path`, which is common for "in-source" test layouts:

```
src/
   analysis/
     /io
       /tests
         test_measurement.py
     /model
       /tests    
         test_measurment.py 
```

But again this is not one of the use cases that `importlib` mode attempts to solve.

> Given that it's perhaps markedly different from the importlib mode, perhaps it makes more sense to make this a new mode, maybe called path?

That's sensible, but I'm not sure it is the best approach. Supposing we can fix #7870, what would be the advantage of `path` over `importlib`?

davidhewitt: As well as #7870 I have a separate issue with `importlib` mode, which is that relative imports no longer work in `importlib` mode. This patch originated as a attempt to figure out how to get `__package__` correctly set as suggested at https://github.com/pytest-dev/pytest/issues/7245#issuecomment-686245048.

I'm happy to close this PR if it doesn't look like the right solution.

nicoddemus: @davidhewitt again thanks for the patch!

Let's keep this open until we get #7870 working (even if only as a reference): we can then see if it also solves the relative import problem.

davidhewitt: 👍 

Note that the test case I've written here is _particularly_ awkward but is amazingly also a real-world use case I'm suffering with which motivated me to write the PR.

I'm attempting a relative import from a test file that is inside a pep 420 namespace package, where the first non-implicit package is called `csv`, and so collides with the builtin package of the same name.

Interestingly the `append` mode also cannot handle this case correctly, so I was wondering if this patch could also be relevant for the `append` mode.

nicoddemus:
#7870 has been merged, do we still need to revisit this?

Zac-HD:
Since this hasn't been updated in years, I'm going to assume that it's stale and close it - but if the issue persists, it would be _fantastic_ to rebase and open a new PR 🙏

Issue #7942, opened 2020-10-25
testing: migrate pytest's own tests from the testdir fixture to the pytester fixture


bluetech:
pytest has many tests for testing itself, under the `testing/` directory in this repo. Many of the tests invoke `pytest` and check its output and outcomes. pytest provides a fixture called `pytester` which aids in this, you can find its documentation [here](https://docs.pytest.org/en/latest/reference.html#pytester).

The `pytester` itself is new in pytest 6.2 (unreleased yet when this is written), but its functionality is also provided by an older plugin called [`testdir`](https://docs.pytest.org/en/latest/reference.html#testdir). The two fixtures have almost the same API, the only difference is that `testdir` takes and returns `py.path.local` objects, while `pytester` is based on `pathlib.Path`, which we are trying to transition to.

Almost all of pytest's test are currently written using `testdir`. This issue is about migrating to use `pytester` instead. The instructions for converting are:

- Pick a file under `testing/` which uses `testdir`.
- Go over each test which uses `tesdir` and change it to `pytester`. Beyond just replacing the name, some adjustments *might* be needed for using `pathlib` Paths instead of `py.path.local` paths.
- Bonus points: add a type annotation for the `pytester` argument.

Example:

Before:

```py
def test_rootdir_wrong_option_arg(testdir):
    result = testdir.runpytest("--rootdir=wrong_dir")
    result.stderr.fnmatch_lines(
        ["*Directory *wrong_dir* not found. Check your '--rootdir' option.*"]
    )
```

After:

```py
# At this import at the top of the file.
from _pytest.pytester import Pytester

def test_rootdir_wrong_option_arg(pytester: Pytester) -> None:
    result = pytester.runpytest("--rootdir=wrong_dir")
    result.stderr.fnmatch_lines(
        ["*Directory *wrong_dir* not found. Check your '--rootdir' option.*"]
    )
```

symonk:
Thanks @bluetech I will start with `testing/test_mark.py`

nicoddemus:
> Should this be branched off somewhere else than master for 6.2? or be destined for somewhere else? (I assume master is ok for both).

Yes `master` is fine. 👍

symonk:
`test_stepwise` next on the hit list, i've been refactoring the plugin so I want to follow up and port that over

cmecklenborg:
I'm working on `test_setuponly` now, then will go back and try to finish `test_assertrewrite`

pillemer:
[...]

PR #7944, opened 2020-10-25
Refactor test_warning_types.py & test_mark.py to use pytester 

symonk:
Good afternoon, This PR refactors `test_mark.py` & `test_warning_types.py` to use the newer `pytester` fixture in place of testdir.

Some type hinting was problematic, as was `pytester.tmpdir` / `basename`

relates to #7942

### DIFF
--- a/testing/test_mark.py
+++ b/testing/test_mark.py
@@ -39,13 +39,13 @@ class TestMark:
         assert pytest.mark.foo(SomeClass) is SomeClass
         assert pytest.mark.foo.with_args(SomeClass) is not SomeClass  # type: ignore[comparison-overlap]
 
-    def test_pytest_mark_name_starts_with_underscore(self):
+    def test_pytest_mark_name_starts_with_underscore(self) -> None:
         mark = MarkGenerator()
         with pytest.raises(AttributeError):
             mark._some_name
 
 
-def test_marked_class_run_twice(pytester: Pytester):
+def test_marked_class_run_twice(pytester: Pytester) -> None:
     """Test fails file is run twice that contains marked class.
     See issue#683.
     """
@@ -63,7 +63,7 @@ def test_marked_class_run_twice(pytester: Pytester):
     rec.assertoutcome(passed=6)
 
 
-def test_ini_markers(pytester: Pytester):
+def test_ini_markers(pytester: Pytester) -> None:
     pytester.makeini(
         """
         [pytest]
@@ -86,7 +86,7 @@ def test_ini_markers(pytester: Pytester):
     rec.assertoutcome(passed=1)
 
 
-def test_markers_option(pytester: Pytester):
+def test_markers_option(pytester: Pytester) -> None:
     pytester.makeini(
         """
         [pytest]
@@ -102,7 +102,7 @@ def test_markers_option(pytester: Pytester):
     )
 
 
-def test_ini_markers_whitespace(pytester: Pytester):
+def test_ini_markers_whitespace(pytester: Pytester) -> None:
     pytester.makeini(
         """
         [pytest]
@@ -123,7 +123,7 @@ def test_ini_markers_whitespace(pytester: Pytester):
     rec.assertoutcome(passed=1)
 
 
-def test_marker_without_description(pytester: Pytester):
+def test_marker_without_description(pytester: Pytester) -> None:
     pytester.makefile(
         ".cfg",
         setup="""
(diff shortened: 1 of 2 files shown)


testing/test_mark.py
```diff
@@ -37,16 +40,16 @@ class SomeClass:
...
 
     def test_pytest_mark_name_starts_with_underscore(self):
-        mark = Mark()
+        mark = MarkGenerator()
         with pytest.raises(AttributeError):
             mark._some_name
 
 
-def test_marked_class_run_twice(testdir):
+def test_marked_class_run_twice(pytester: Pytester):
```
bluetech, line 48: To complete the type annotation, please also add `-> None`. Otherwise mypy still considers it incomplete.

testing/test_mark.py
```diff
@@ -557,14 +570,14 @@ def test_has_inherited(self):
 
         """
         )
-        items, rec = testdir.inline_genitems(p)
+        items, rec = pytester.inline_genitems(p)
         has_own, has_inherited = items
-        assert has_own.get_closest_marker("c").kwargs == {"location": "function"}
-        assert has_inherited.get_closest_marker("c").kwargs == {"location": "class"}
+        assert has_own.get_closest_marker("c").kwargs == {"location": "function"}  # type: ignore
```
bluetech, line 575: When mypy emits an error, it also emits an error code, like `attr-defined` or such. We usually specify it in the `type: ignore` comment, like this: `type: ignore[attr-defined]`. This has two advantages: 1) it helps the reader understand what is the problem being ignored, and 2) if the error changes unexpectedly, it will start failing again, which is good.

testing/test_mark.py
```diff
@@ -913,9 +926,9 @@ def test_aliases(self) -> None:
 
 
 @pytest.mark.parametrize("mark", [None, "", "skip", "xfail"])
-def test_parameterset_for_parametrize_marks(testdir, mark):
+def test_parameterset_for_parametrize_marks(pytester: Pytester, mark):
```
bluetech, line 929:
```suggestion
def test_parameterset_for_parametrize_marks(pytester: Pytester, mark: Optional[str]) -> None:
```

bluetech (commented):
Great work, LGTM with a couple comments and after CI passes.

symonk:
@bluetech will tidy up after dinner, I threw an extra commit for warning_types.py too, it was a simple one - Is this stuff better managed as part of one very large PR? or should be atomic in smaller PRs going forward?

testing/test_mark.py
symonk: Thanks, I was kinda banging my head with this mypy stuff, get_cloest_marker returns an Optional[Mark] but we are calling .kwargs (None may not have it obviously) I suppressed but what is the correct approach to this scenario here?

testing/test_mark.py
```diff
@@ -176,21 +179,21 @@ def test_hello():
...
     )
-    reprec = testdir.inline_run()
+    reprec = pytester.inline_run()
     reprec.assertoutcome(passed=1)
 
 
 @pytest.mark.parametrize("option_name", ["--strict-markers", "--strict"])
-def test_strict_prohibits_unregistered_markers(testdir, option_name):
-    testdir.makepyfile(
+def test_strict_prohibits_unregistered_markers(pytester: Pytester, option_name) -> None:
```
bluetech, line 187:
```suggestion
def test_strict_prohibits_unregistered_markers(pytester: Pytester, option_name: str) -> None:
```

bluetech (commented):
> Is this stuff better managed as part of one very large PR? or should be atomic in smaller PRs going forward?

I'd say not one very large PR since that'd be hard to review, but a few files at once is not a problem.

Ping me when you're done with this one!

testing/test_mark.py
bluetech: The common way to appease mypy is to use `assert is not None`, in this case

```py
marker = assert has_own.get_closest_marker("c")
assert marker is not None
assert marker.kwargs = ....
```

symonk:
should be sorted, pending CI

bluetech (approved):

nicoddemus:
Note for the future: IMHO this PR probably should have been squashed given the commits by themselves don't seem too interesting.

bluetech:
Right, I sometimes forget...

nicoddemus:
> Right, I sometimes forget...

No worries, just a friendly reminder. 😁

PR #7948, opened 2020-10-26
testing: make conftest stuff check for pytester not testdir

bluetech:
testdir uses pytester so this applies to it as well, but now includes pytester as well.


nicoddemus (approved):
Great find, thanks @bluetech!

PR #7952, opened 2020-10-27
Migrate test_assertrewrite.py from testdir to pytester

cmecklenborg:
Partial fix for #7942

Converts test_assertrewrite.py from testdir to pytester

### DIFF
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -140,13 +140,12 @@ class TestAssertionRewrite:
         result = pytester.runpytest()
         result.stdout.fnmatch_lines(["*assert 1 == 2*"])
 
-    def test_honors_pep_235(self, pytester: Pytester, monkeypatch):
+    def test_honors_pep_235(self, pytester: Pytester, monkeypatch) -> None:
         # note: couldn't make it fail on macos with a single `sys.path` entry
         # note: these modules are named `test_*` to trigger rewriting
-        pytester.makepyfile("test_y", "x = 1")
+        pytester.makepyfile(test_y="x = 1")
         xdir = pytester.mkdir("x")
-        xdir.joinpath("test_Y").mkdir()
-        xdir.joinpath("test_Y").joinpath("__init__.py").touch()
+        pytester.mkpydir(str(xdir.joinpath("test_Y")))
         xdir.joinpath("test_Y").joinpath("__init__.py").write_text("x = 2")
         pytester.makepyfile(
             "import test_y\n"
@@ -155,7 +154,7 @@ class TestAssertionRewrite:
             "    assert test_y.x == 1\n"
             "    assert test_Y.x == 2\n"
         )
-        monkeypatch.syspath_prepend(xdir)
+        monkeypatch.syspath_prepend(str(xdir))
         pytester.runpytest().assert_outcomes(passed=1)
 
     def test_name(self, request) -> None:
@@ -1298,7 +1297,7 @@ class TestEarlyRewriteBailout:
         pytester.makepyfile(test_foo="def test_foo(): pass")
         pytester.makepyfile(bar="def bar(): pass")
         foobar_path = pytester.makepyfile(foobar="def foobar(): pass")
-        self.initial_paths.add(foobar_path)
+        self.initial_paths.add(str(foobar_path))
 
         # conftest files should always be rewritten
         assert hook.find_spec("conftest") is not None
@@ -1322,7 +1321,7 @@ class TestEarlyRewriteBailout:
         """If one of the python_files patterns contain subdirectories ("tests/**.py") we can't bailout early
         because we need to match with the full path, which can only be found by calling PathFinder.find_spec
         """
-        p = pytester.makepyfile(
+        pytester.makepyfile(
             **{
                 "tests/file.py": """\
                     def test_simple_failure():
@@ -1330,7 +1329,7 @@ class TestEarlyRewriteBailout:
                 """
             }
         )
-        pytester.syspathinsert(str(pytester.path))
+        pytester.syspathinsert("tests")
         hook.fnpats[:] = ["tests/**.py"]
         assert hook.find_spec("file") is not None
         assert self.find_spec_calls == ["file"]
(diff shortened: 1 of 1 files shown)


testing/test_assertrewrite.py
```diff
@@ -115,18 +115,18 @@ def test_dont_rewrite(self) -> None:
         assert isinstance(m.body[1], ast.Assert)
         assert m.body[1].msg is None
 
-    def test_dont_rewrite_plugin(self, testdir):
+    def test_dont_rewrite_plugin(self, pytester: Pytester):
```
bluetech, line 118: It will be great (though not absolutely necessary) if, while adding the `: Pytester` annotation, you also add `-> None` annotations to tests. Without it, mypy (the type checker) still considers the typing to be incomplete. In this case it will be

```suggestion
    def test_dont_rewrite_plugin(self, pytester: Pytester) -> None:
```

testing/test_assertrewrite.py
```diff
@@ -135,26 +135,26 @@ def test_rewrites_plugin_as_a_package(self, testdir):
...
-        testdir.makeconftest('pytest_plugins = ["plugin"]')
-        testdir.makepyfile("def test(special_asserter): special_asserter(1, 2)\n")
-        result = testdir.runpytest()
+        pytester.makeconftest('pytest_plugins = ["plugin"]')
+        pytester.makepyfile("def test(special_asserter): special_asserter(1, 2)\n")
+        result = pytester.runpytest()
         result.stdout.fnmatch_lines(["*assert 1 == 2*"])
 
-    def test_honors_pep_235(self, testdir, monkeypatch):
+    def test_honors_pep_235(self, pytester, monkeypatch):
```
bluetech, line 143:
```suggestion
    def test_honors_pep_235(self, pytester: Pytester, monkeypatch) -> None:
```

testing/test_assertrewrite.py
```diff
@@ -135,26 +135,26 @@ def test_rewrites_plugin_as_a_package(self, testdir):
...
         # note: couldn't make it fail on macos with a single `sys.path` entry
         # note: these modules are named `test_*` to trigger rewriting
-        testdir.tmpdir.join("test_y.py").write("x = 1")
-        xdir = testdir.tmpdir.join("x").ensure_dir()
+        pytester.tmpdir.join("test_y.py").write("x = 1")
```
bluetech, line 146: With `pytester`, `testdir.tmpdir` becomes `pytester.path`. It also becomes a `pathlib.Path`, so the code would need to be a little different. In this case:

```suggestion
        pytester.tmpdir.joinpath("test_y.py").write_text("x = 1", "utf-8")
```

testing/test_assertrewrite.py
```diff
@@ -962,26 +962,26 @@ def test():
...
+        pytester.makepyfile(
             """
             def test_meta_path():
                 import sys; sys.meta_path = []"""
         )
-        assert testdir.runpytest().ret == 0
+        assert pytester.runpytest().ret == 0
 
-    def test_write_pyc(self, testdir: Testdir, tmpdir, monkeypatch) -> None:
+    def test_write_pyc(self, pytester: Testdir, tmpdir, monkeypatch) -> None:
```
bluetech, line 980: Here I suggest also changing `tmpdir` to `tmp_path`, which is the `pathlib.Path` version of `tmpdir`, so it works better with `pytester`.

testing/test_assertrewrite.py
```diff
@@ -1062,15 +1062,15 @@ def test_read_pyc(self, tmp_path: Path) -> None:
 
         assert _read_pyc(source, pyc) is None  # no error
 
-    def test_reload_is_same_and_reloads(self, testdir: Testdir) -> None:
+    def test_reload_is_same_and_reloads(self, pytester: Testdir) -> None:
```
bluetech, line 1065:
```suggestion
    def test_reload_is_same_and_reloads(self, pytester: Pytester) -> None:
```

testing/test_assertrewrite.py
```diff
@@ -1627,11 +1627,11 @@ def test_foo():
...
             }
         )
-        result = testdir.runpytest()
+        result = pytester.runpytest()
         assert result.ret == 0
 
-        test_foo = Path(testdir.tmpdir) / "src/test_foo.py"
-        bar_init = Path(testdir.tmpdir) / "src/bar/__init__.py"
+        test_foo = Path(pytester.tmpdir) / "src/test_foo.py"
+        bar_init = Path(pytester.tmpdir) / "src/bar/__init__.py"
```
bluetech, line 1634:
```suggestion
        test_foo = pytester.path / "src/test_foo.py"
        bar_init = pytester.path / "src/bar/__init__.py"
```

bluetech (changes requested):
Thanks @cmecklenborg, that's a good start on a big file. I left some comments.

cmecklenborg:
I moved this to draft status. I addressed PR comments and fixed most of the failing tests, but I still have work to do on 3:
```============================================================================================ short test summary info ============================================================================================
FAILED testing/test_assertrewrite.py::TestAssertionRewrite::test_honors_pep_235 - AssertionError: assert {'errors': 1,...pped': 0, ...} == {'errors': 0,...pped': 0, ...}
FAILED testing/test_assertrewrite.py::TestEarlyRewriteBailout::test_basic - AssertionError: assert None is not None
FAILED testing/test_assertrewrite.py::TestEarlyRewriteBailout::test_pattern_contains_subdirectories - AssertionError: assert None is not None
```

I'll look at this more with fresh eyes tomorrow.

cmecklenborg:
@bluetech Do you have any thoughts on why `test_assertrewrite.py::TestEarlyRewriteBailout::test_basic` might be failing on the windows builds?

bluetech:
> Do you have any thoughts on why `test_assertrewrite.py::TestEarlyRewriteBailout::test_basic` might be failing on the windows builds?

The issue is this:

```
testing/test_assertrewrite.py:1300: error: Argument 1 to "add" of "set" has incompatible type "str"; expected "local"  [arg-type]
```

It did not show up due to a deficiency in our CI setup, which should be fixed in #7962 for the future.

Explanation of the error: right above this test there is a `hook` fixture which does some funky stuff, one of which is to create mock of a `pytest.Session` called `StubSession`. This uses a `self.initial_paths` attribute, which is a `Set[py.path.local]`. This cannot be changed to `pathlib.Path` since `Session` itself uses `py.path.local` here (for now at least). So to fix the mypy error you'd need to do something like this:

```py
self.initial_paths.add(py.path.local(foobar_path))
```

bluetech (approved):

PR #7953, opened 2020-10-28
Migrate test_skipping.py from testdir to pytester

cmecklenborg:
Partial fix for #7942

Converts test_skipping.py from testdir to pytester. All tests are passing locally.

### DIFF
--- a/testing/test_skipping.py
+++ b/testing/test_skipping.py
@@ -150,7 +150,7 @@ class TestEvaluation:
                     pass
         """
         )
-        item.config._hackxyz = 3
+        item.config._hackxyz = 3  # type: ignore[attr-defined]
         skipped = evaluate_skip_marks(item)
         assert skipped
         assert skipped.reason == "condition: config._hackxyz"
--- a/testing/test_skipping.py
+++ b/testing/test_skipping.py
@@ -158,7 +158,7 @@ class TestEvaluation:
 
 class TestXFail:
     @pytest.mark.parametrize("strict", [True, False])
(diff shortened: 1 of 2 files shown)


testing/test_skipping.py
```diff
@@ -148,16 +150,16 @@ def test_func(self):
...
         skipped = evaluate_skip_marks(item)
         assert skipped
         assert skipped.reason == "condition: config._hackxyz"
 
 
 class TestXFail:
     @pytest.mark.parametrize("strict", [True, False])
-    def test_xfail_simple(self, testdir, strict):
-        item = testdir.getitem(
+    def test_xfail_simple(self, pytester, strict) -> None:
```
bluetech, line 161: Would be great to catch the `pytester`s that appear in the middle of the parameter list, like here. A search/replace of ` pytest, ` with ` pytester: Pytester, ` should do the trick (also for the rest of the file). Also in this case case can annotate `strict`, like so:

```suggestion
    def test_xfail_simple(self, pytester: Pytester, strict: bool) -> None:
```

bluetech (commented):
Thanks for working on this @cmecklenborg!

This looks good to me with one comment.

nicoddemus (approved):
Thanks @cmecklenborg!

I'm leaving this approved, but please take a look at @bluetech's comment/suggestion. 👍

testing/test_skipping.py
cmecklenborg: Argh, sorry for missing those! I added typehinting for the remaining `pytester` args, as well as for the `strict` args throughout.

PR #7954, opened 2020-10-28
acceptance_test.py migrate from testdir to pytester

mmarinez:

### DIFF
--- a/changelog/7942.improvement.rst
+++ b/changelog/7942.improvement.rst
@@ -1,0 +1,1 @@
+Migrate from testdir to pytester in the acceptance_test.py file
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -9,7 +9,7 @@ import pytest
 from _pytest.compat import importlib_metadata
 from _pytest.config import ExitCode
 from _pytest.pathlib import symlink_or_skip
-from _pytest.pytester import Testdir
+from _pytest.pytester import Pytester
 
 
 def prepend_pythonpath(*dirs):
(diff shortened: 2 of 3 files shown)


testing/acceptance_test.py
```diff
@@ -20,97 +20,97 @@ def prepend_pythonpath(*dirs):
...
-        testdir.makeconftest(
+    def test_early_hook_error_issue38_1(self, pytester: Pytester) -> None:
+        pytester.makeconftest(
             """
             def pytest_sessionstart():
                 0 / 0
         """
         )
-        result = testdir.runpytest(testdir.tmpdir)
+        result = pytester.runpytest(pytester.tmpdir)
```
symonk, line 43: this is incorrect, `pytester.path` i believe.  (I won't comment on each one in this file as theres a lot - but I would recommend running the tests in this file before pushing, as well as `tox -e linting` (happens on commit if pre-commit is configured).

changelog/7942.improvement.rst
```diff
@@ -0,0 +1 @@
+Migrate from testdir to pytester in the acceptance_test.py file
```
bluetech, line 1: For internal improvements to our own testing suite, we don't add a changelog entry, because a user of the library isn't really affected by it.

bluetech (changes requested):
Thanks @mmarinez that's a good start.

In addition to @symonk's comment (replacing `pytester.tmpdir` with `pytester.path`), you'll also need to adjust some `py.path.local` APIs to `pathlib` ones. For example `path.join(...)` becomes `path.joinpath(...)`. The mypy output (found in the `linting` CI job) and test output can tell you where the adjustments are needed.

testing/acceptance_test.py
mmarinez: got it, I'll be working on the corrections

bluetech:
#8070 also does this and is more complete so I'll close this one. Thanks @mmarinez.

PR #7956, opened 2020-10-28
Fix handling recursive symlinks

### LINKED ISSUE #7951: Pytest crashes when recursive symlinks present
When I have a test directory which contains a recursive symlink (in my case this is one of the assets we are checking that our software works in such case), pytest crashes with:

```
E   OSError: [Errno 40] Too many levels of symbolic links: '/tmp/tests/recursive'
```

The /tmp/tests directory looks like the following:
```
zsolt@localhost:/tmp$ ls -l /tmp/tests/
total 4
lrwxrwxrwx 1 zsolt zsolt  9 Oct 27 15:02 recursive -> recursive
-rw-rw-r-- 1 zsolt zsolt 36 Oct 27 15:01 test_dummy.py
zsolt@localhost:/tmp$ cat /tmp/tests/test_dummy.py 

def test_dummy():
    assert True

```

It used to work with pytest 6.0.2 and now it seems to be broken on 6.1.0 and above.

I'm running pytest with `pytest /tmp/tests` command from its own, clean venv having the given pytest version (6.0.2, 6.1.0 and 6.1.1) and its dependencies only.
I'm using Linux (Ubuntu bionic), but I think this happens on every system supporting recursive symlinks.

csernazs:
When pytest was run on a directory containing a recursive symlink it failed
with ELOOP as the library was not able to determine the type of the
direntry:

src/_pytest/main.py:685: in collect
    if not direntry.is_file():
E   OSError: [Errno 40] Too many levels of symbolic links: '/home/florian/proj/pytest/tests/recursive'

This is fixed now by handling ELOOP.

Fixes #7951

### DIFF
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -9,6 +9,10 @@ import sys
 import uuid
 import warnings
 from enum import Enum
+from errno import EBADF
+from errno import ELOOP
+from errno import ENOENT
+from errno import ENOTDIR
 from functools import partial
 from os.path import expanduser
 from os.path import expandvars
@@ -36,6 +40,24 @@ LOCK_TIMEOUT = 60 * 60 * 24 * 3
 
 
 _AnyPurePath = TypeVar("_AnyPurePath", bound=PurePath)
+
+# The following function, variables and comments were
+# copied from cpython 3.9 Lib/pathlib.py file.
+
+# EBADF - guard against macOS `stat` throwing EBADF
+_IGNORED_ERRORS = (ENOENT, ENOTDIR, EBADF, ELOOP)
+
+_IGNORED_WINERRORS = (
+    21,  # ERROR_NOT_READY - drive exists but is not accessible
+    1921,  # ERROR_CANT_RESOLVE_FILENAME - fix for broken symlink pointing to itself
+)
+
+
+def _ignore_error(exception):
+    return (
+        getattr(exception, "errno", None) in _IGNORED_ERRORS
+        or getattr(exception, "winerror", None) in _IGNORED_WINERRORS
+    )
 
 
 def get_lock_path(path: _AnyPurePath) -> _AnyPurePath:
@@ -555,8 +577,23 @@ def visit(
 
     Entries at each directory level are sorted.
     """
-    entries = sorted(os.scandir(path), key=lambda entry: entry.name)
+
+    # Skip entries with symlink loops and other brokenness, so the caller doesn't
+    # have to deal with it.
+    entries = []
+    for entry in os.scandir(path):
+        try:
+            entry.is_file()
+        except OSError as err:
+            if _ignore_error(err):
+                continue
+            raise
+        entries.append(entry)
+
+    entries.sort(key=lambda entry: entry.name)
+
     yield from entries
+
     for entry in entries:
         if entry.is_dir(follow_symlinks=False) and recurse(entry):
             yield from visit(entry.path, recurse)
--- a/testing/test_collection.py
+++ b/testing/test_collection.py
@@ -1404,3 +1404,17 @@ def test_does_not_crash_on_error_from_decorated_function(testdir: Testdir) -> No
     result = testdir.runpytest()
     # Not INTERNAL_ERROR
     assert result.ret == ExitCode.INTERRUPTED
+
+
+def test_does_not_crash_on_recursive_symlink(testdir: Testdir) -> None:
+    """Regression test for an issue around recursive symlinks (#7951)."""
+    symlink_or_skip("recursive", testdir.tmpdir.join("recursive"))
+    testdir.makepyfile(
+        """
+        def test_foo(): assert True
+        """
+    )
+    result = testdir.runpytest()
+
+    assert result.ret == ExitCode.OK
+    assert result.parseoutcomes() == {"passed": 1}
--- a/testing/test_pathlib.py
+++ b/testing/test_pathlib.py
@@ -17,6 +17,8 @@ from _pytest.pathlib import import_path
 from _pytest.pathlib import ImportPathMismatchError
 from _pytest.pathlib import maybe_delete_a_numbered_dir
 from _pytest.pathlib import resolve_package_path
+from _pytest.pathlib import symlink_or_skip
+from _pytest.pathlib import visit
 
 
 class TestFNMatcherPort:
@@ -401,3 +403,14 @@ def test_commonpath() -> None:
     assert commonpath(subpath, path) == path
     assert commonpath(Path(str(path) + "suffix"), path) == path.parent
     assert commonpath(path, path.parent.parent) == path.parent.parent
+
+
+def test_visit_ignores_errors(tmpdir) -> None:
+    symlink_or_skip("recursive", tmpdir.join("recursive"))
+    tmpdir.join("foo").write_binary(b"")
+    tmpdir.join("bar").write_binary(b"")
+
+    assert [entry.name for entry in visit(tmpdir, recurse=lambda entry: False)] == [
+        "bar",
+        "foo",
+    ]
--- a/AUTHORS
+++ b/AUTHORS
@@ -314,3 +314,4 @@ Xuecong Liao
 Yoav Caspi
 Zac Hatfield-Dodds
 Zoltán Máté
+Zsolt Cserna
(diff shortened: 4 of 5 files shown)


The-Compiler:
Looks like the test fails on Windows - I guess it should probably be skipped there instead.

csernazs:
Something is not entirely correct with codecov, I'm not sure if my patch caused that drop in coverage, I've rebased my commit to master and it still appears. Could you please look at it? Thanks.

bluetech:
> Something is not entirely correct with codecov, I'm not sure if my patch caused that drop in coverage, I've rebased my commit to master and it still appears. Could you please look at it? Thanks.

The `codecov/patch` check looks at the PR diff itself to see how much of it is covered. The useful place to look at is the [Diff](https://codecov.io/gh/pytest-dev/pytest/compare/b95991aeeab28be6af94e24668414360e4d3a630...a2884aab56e457d122894d2cc4d1c1bd871d5331/diff) tab in the codecov page, which shows that the `raise` statement in the patch is not covered by any test.

If you can reasonably write a test to cover it, then that would fix it, but sometimes it's not worth the effort, in which case we can override it and merge anyway.

(As for the change itself, I need to look into it a bit before giving a proper review, which I'll do soon)

csernazs:
I did some research on how I could add a test for that bare `raise` statement. Thinking about it, what other errno can be raised from that is_file() call? It is a `DirEntry` object which was created from the results returned by `getdents()` system call (on Linux). Even if I create a directory with `0o444`, so it can be listed but the contents cannot be stat()ed, `is_file()` method returns True. This is because the entries returned by `getdents()` contains the file type (regular, socket, symlink, etc..) and it is used so not stat() is made. For a recursive symlink, I think it returns `DT_UNKNOWN` so based on posixmodule.c, cpython calls `stat()` on it which is then fails with ELOOP - this appears in strace.

Reading further, `man readdir` says:
>>>

              Currently, only some filesystems (among them: Btrfs, ext2, ext3,
              and ext4) have full support  for  returning  the  file  type  in
              d_type.   All  applications  must  properly  handle  a return of
              DT_UNKNOWN.

So it seems that it is implementation specific. I ran out of ideas adding coverage for that `raise` (mounting an nfs share is a bit overkill), but I would keep it there for completeness.

testing/test_collection.py
```diff
@@ -1404,3 +1404,18 @@ def a(): return 4
     result = testdir.runpytest()
     # Not INTERNAL_ERROR
     assert result.ret == ExitCode.INTERRUPTED
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="win32 has no recursive symlinks")
```
blueyed, line 1409: I think the error is only different / needs other handling.
This is how pathlib does it: https://github.com/python/cpython/commit/d5c120f7eb6f2a9cdab282a5d588afed307a23df

src/_pytest/main.py
```diff
@@ -676,8 +677,18 @@ def collect(self) -> Iterator[Union[nodes.Item, nodes.Collector]]:
 
                 seen_dirs: Set[py.path.local] = set()
                 for direntry in visit(str(argpath), self._recurse):
-                    if not direntry.is_file():
-                        continue
+                    try:
+                        if not direntry.is_file():
+                            continue
+                    except OSError as err:
+                        # 1921: ERROR_CANT_RESOLVE_FILENAME - fix for broken symlink pointing to itself
```
nicoddemus, line 684:
```suggestion
                        # 1921: ERROR_CANT_RESOLVE_FILENAME - fix for broken symlink pointing to itself (#7951)
```

nicoddemus (approved):
Great finding. 

Also please add yourself to `AUTHORS` @csernazs!  👍

bluetech (changes requested):
Hi @csernazs, 

I think this approach is good.

There is one more place which calls `visit()` and would require a similar treatment. Also,  while looking at this I discovered #7981 and submitted PR #7982 to fix it, but that PR would add yet another place which can trigger this exception. I think that adding the try-catch code in every instance would be too annoying, therefore I suggest the following:

- Copy [`_ignore_error()` from pathlib](https://github.com/python/cpython/blob/7d210271579ae31f43b32f73c2aff5bc4fe0d27f/Lib/pathlib.py#L37-L47) into `_pytest/pathlib.py` (with a comment indicating it was copied from pathlib as of Python 3.9).

- Modify `_pytest.pathlib.visit()` to do a pre-iteration of the `os.scandir` results, running `entry.is_file()` (while ignoring the result) and skipping the entry if the error should be ignored.

This way, we can localize the try-catch in one place and know that it doesn't happen at later stages.

WDYT?

csernazs:
I completely agree with moving the code to pathlib.py (I also wanted to do it but I couldn't find a good place for it, but pathlib.py is perfect)
I see the logic behind _ignore_errors and I can support it, to my best knowledge I thought that no others than ELOOP should be handled but re-reading stat(2) shows me that ENOTDIR and ENOENT is also possible. And on the plus side I think including _ignore_errors is a good idea as it gives the same protection as python core.

csernazs:
I've pushed the suggested changes. A few remarks:
* I kept the test in test_collection.py as it tests the original case described in the issue
* a new test added to test_pathlib.py for this case which calls visit() directly
* while copying the ignore_errors function, one typo fixed (_IGNORED_ERROS -> _IGNORED_ERRORS)

src/_pytest/pathlib.py
```diff
@@ -555,8 +577,20 @@ def visit(
 
     Entries at each directory level are sorted.
     """
-    entries = sorted(os.scandir(path), key=lambda entry: entry.name)
+
+    entries = []
+    for entry in os.scandir(path):
+        try:
```
bluetech, line 583: I would add a comment here, something like this:

> Skip entries with symlink loops and other brokenness, so the caller doesn't
> have to deal with it.

src/_pytest/pathlib.py
```diff
@@ -555,8 +577,20 @@ def visit(
...
+            entry.is_file()
+        except OSError as err:
+            if _ignore_error(err):
+                continue
```
bluetech, line 587:
```suggestion
                continue
            raise
```

bluetech (commented):
@csernazs Perfect! I left a suggestion and a comment.

testing/test_collection.py
```diff
@@ -1404,3 +1404,17 @@ def a(): return 4
...
+
+
+def test_does_not_crash_on_recursive_symlink(testdir: Testdir) -> None:
+    """Regression test for an issue around recursive symlinks (#7951)."""
+    testdir.makepyfile(
+        """
+        def test_foo(): assert True
+        """
+    )
+    os.symlink("recursive", str(testdir.tmpdir.join("recursive")))
```
bluetech, line 1416: Another comment: should use `symlink_or_skip` here (already imported), for platforms which don't support symlinks.

src/_pytest/pathlib.py
csernazs: In such case, when the directory have n entries where 1 entry is broken, the caller will not able to see any entries but the error itself. In such case, the caller won't be able to ignore or somehow workaround the error.
On the other hand, raising the error makes sense as if is_file() raises some error, probably the following open() or other file-related call will fail and it is better to report the error as soon as possible.
So made the suggested change, just wanted to comment on it.

testing/test_pathlib.py
```diff
@@ -401,3 +402,15 @@ def test_commonpath() -> None:
     assert commonpath(subpath, path) == path
     assert commonpath(Path(str(path) + "suffix"), path) == path.parent
     assert commonpath(path, path.parent.parent) == path.parent.parent
+
+
+def test_visit_ignores_errors(tmpdir) -> None:
+    tmpdir.join("foo").write_binary(b"")
+    tmpdir.join("bar").write_binary(b"")
+
+    os.symlink("recursive", str(tmpdir.join("recursive")))
```
bluetech, line 411: Sorry, looks like it's here too...

src/_pytest/pathlib.py
bluetech: I agree with both the first and second part.

bluetech (approved):
Thanks @csernazs! Will also backport this to 6.1.x.

csernazs:
Thank you guys for all the help!

PR #7959, opened 2020-10-28
Migrate test_setuponly.py from testdir to pytester

cmecklenborg:
Partial fix for #7942

Converts `test_setuponly.py` from testdir to pytester

### DIFF
--- a/testing/test_setuponly.py
+++ b/testing/test_setuponly.py
@@ -51,7 +51,7 @@ def test_show_different_scopes(pytester: Pytester, mode) -> None:
     '''
     )
 
-    result = pytester.runpytest(mode, str(p))
+    result = pytester.runpytest(mode, p)
     assert result.ret == 0
 
     result.stdout.fnmatch_lines(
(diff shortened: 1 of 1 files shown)


testing/test_setuponly.py
```diff
@@ -48,7 +51,7 @@ def test_arg1(arg_session, arg_function):
     '''
     )
 
-    result = testdir.runpytest(mode, p)
+    result = pytester.runpytest(mode, str(p))
```
bluetech, line 54: Here and in other places below, the `str()` call shouldn't be necessary. This is because `pytester.runpytest` also accepts [`os.PathLike[str]`](https://docs.python.org/3/library/os.html#os.PathLike), which in particular allows a `pathlib.Path`.

bluetech (commented):
Thanks @cmecklenborg! Just one comment about removing unneeded `str()`s, otherwise it's good to go.

bluetech (approved):

PR #7962, opened 2020-10-28
pre-commit: install typed dependencies in the mypy target

bluetech:
Otherwise, mypy doesn't know about them and their types are considered `Any`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,6 +56,9 @@ repos:
         args: []
         additional_dependencies:
           - iniconfig>=1.1.0
+          - py>=1.8.2
+          - attrs>=19.2.0
+          - packaging
 -   repo: local
     hooks:
     -   id: rst


bluetech:
If there is some way to avoid repeating the deps here, or maybe just installing all of them, that would be better. However we probably shouldn't install *pytest* itself, who knows how it might confuse things.

nicoddemus:
> If there is some way to avoid repeating the deps here, or maybe just installing all of them, that would be better.

I've wondered about this myself, perhaps @asottile can shed some light here.

nicoddemus (approved):

asottile:
> 
> 
> > If there is some way to avoid repeating the deps here, or maybe just installing all of them, that would be better.
> 
> I've wondered about this myself, perhaps @asottile can shed some light here.

there isn't a way to install from the repository under test, intentionally -- it would create an intractable caching problem

nicoddemus:
> there isn't a way to install from the repository under test, intentionally -- it would create an intractable caching problem

Oh I agree, but what I was wondering is if there's a way to load dependencies from a file (like a `requirements.txt` or something); that way I could also load some dependencies for other tools using the same file, like `tox` for example.

asottile:
> 
> 
> > there isn't a way to install from the repository under test, intentionally -- it would create an intractable caching problem
> 
> Oh I agree, but what I was wondering is if there's a way to load dependencies from a file (like a `requirements.txt` or something); that way I could also load some dependencies for other tools using the same file, like `tox` for example.

the short answer is "no", the long answer is "yes if it isn't in the repository under test, but that's not that useful" :)

nicoddemus:
I figured, thanks!

PR #7963, opened 2020-10-28
Migrate test_compat.py from testdir to pytester

cmecklenborg:
Partial fix for #7942

Converts test_compat.py from testdir to pytester


bluetech (approved):
Perfect, thanks @cmecklenborg.

PR #7976, opened 2020-10-30
#7942 refactor stepwise tests to utilize pytester

symonk:
refactor `test_stepwise.py` after my other PR merged with refactoring of the plugin.

relates to #7942


bluetech (approved):
Thanks @symonk! I like that you also annotated all of the other arguments :)

PR #7984, opened 2020-10-31
[6.1.x] Merge pull request #7979 from nicoddemus/metafunc-ref

nicoddemus:


nicoddemus:
Weird, I don't see how my patch could break with `os.PathLike` not existing in Python 3.5. 🤔

bluetech:
Happens on my backport PR as well, so not related to this specifically. Since our code on 6.1.x doesn't have any mention of `PathLike`, it must be coming from some dependency. Unless someone wants to debug this I suggest we ignore it :)

bluetech:
@nicoddemus WDYT re. ignoring the `PathLike` error on older Python 3.5 for 6.1.x? Since we don't use `PathLike` in that branch I don't think it's caused by us.

nicoddemus:
> @nicoddemus WDYT re. ignoring the PathLike error on older Python 3.5 for 6.1.x? Since we don't use PathLike in that branch I don't think it's caused by us.

Agreed. Should we hit the merge button then?

PR #7988, opened 2020-10-31
fixtures: deprecate pytest.yield_fixture()

bluetech:


src/_pytest/deprecated.py
```diff
@@ -32,6 +32,9 @@
     "Please update to the new name.",
 )
 
+YIELD_FIXTURE = PytestDeprecationWarning(
+    "@pytest.yield_fixture is deprecated.\nUse @pytest.fixture instead."
```
nicoddemus, line 36: The bit about being safe to search/replace might be useful here. 😁

nicoddemus (approved):

testing/python/fixtures.py
```diff
@@ -3636,10 +3620,10 @@ def test_1(arg1):
         """
         )
 
-    def test_yields_more_than_one(self, testdir, flavor):
+    def test_yields_more_than_one(self, testdir: Testdir) -> None:
         testdir.makepyfile(
             """
-            from test_context import fixture
+            import pytest
             @fixture(scope="module")
```
The-Compiler, line 3627: Missed one :wink:

```suggestion
            @pytest.fixture(scope="module")
```

The-Compiler (changes requested):
Nice! I didn't realize it wasn't deprecated already... Looks like it's [widely used](https://github.com/search?q=yield_fixture&type=code) but I'm still +1 for getting rid of it.

src/_pytest/deprecated.py
bluetech: Trying to keep it short, I added "they are the same".

testing/python/fixtures.py
bluetech: Oops, fixed, thanks.

asottile (approved):

nicoddemus (approved):

The-Compiler (approved):

The-Compiler:
Looks like this now conflicts with #7985 which I just merged, sorry!

PR #7997, opened 2020-11-04
#7942 test_session.py migrate from testdir to Pytester

duthades:
Partial fix for #7942

1. Migrates ```test_session.py``` from ```testdir``` to ```Pytester```.
2. Adds name to ```AUTHORS```.


bluetech (approved):
Looks great, thanks!

PR #8002, opened 2020-11-06
Add pythonenv* to gitignore

mcsitter:
GitHub Codespaces automatically create a pythonenv folder. For example: `pythonenv3.8/` which containts a virtual environment. This should not be commited by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,7 @@ issue/
 env/
 .env/
 .venv/
-pythonenv*
+/pythonenv*/
 3rdparty/
 .tox
 .cache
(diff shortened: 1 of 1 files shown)


bluetech (commented):
I don't see any harm in adding it. Just to reduce potential of this unintentionally ignoring something in the future, can reduce the scope a bit:

1. If this is a directory, add `/` at the end.
2. If it always at the top directory, add a `/` at the start.

nicoddemus (changes requested):
(Just to explicitly change the PR status as "changes requested", as per @bluetech's requests).

mcsitter:
Applied the requested changes. Thanks for approving this! This is a major quality of life improvement for me.

bluetech (approved):

PR #8003, opened 2020-11-06
test_collection.py migrate from testdir to Pytester

duthades:
Partial fix for #7942

1. Migrates test_collection.py from testdir to Pytester.
2. Some tests cannot migrated currently. They were just annotated.

Tests which were not migrated:
1. test_pytest_collect_file
    - Reason
    https://github.com/pytest-dev/pytest/blob/30287b49cd02b6a14ade1bbe9adf9711c3d1259e/src/_pytest/config/argparsing.py#L148-L157
  This function expects py.path.local

2. test__in_venv
    - Reason
    https://github.com/pytest-dev/pytest/blob/30287b49cd02b6a14ade1bbe9adf9711c3d1259e/src/_pytest/main.py#L346-L360
    This function expects py.path.local and uses function specific to py.path
3. test_fscollector_from_parent
   - Reason
    https://github.com/pytest-dev/pytest/blob/30287b49cd02b6a14ade1bbe9adf9711c3d1259e/src/_pytest/nodes.py#L486-L501
    This function expects py.path.local and uses function specific to py.path

### DIFF
--- a/testing/test_collection.py
+++ b/testing/test_collection.py
@@ -74,18 +74,20 @@ class TestCollector:
         """
         )
         cls = pytester.collect_by_name(modcol, "TestClass")
-        fn = pytester.collect_by_name(
-            pytester.collect_by_name(cls, "()"), "test_foo"  # type: ignore[arg-type]
-        )
+        assert isinstance(cls, pytest.Class)
+        instance = pytester.collect_by_name(cls, "()")
+        assert isinstance(instance, pytest.Instance)
+        fn = pytester.collect_by_name(instance, "test_foo")
+        assert isinstance(fn, pytest.Function)
 
-        parent = fn.getparent(pytest.Module)  # type: ignore[union-attr]
-        assert parent is modcol
+        module_parent = fn.getparent(pytest.Module)
+        assert module_parent is modcol
 
-        parent = fn.getparent(pytest.Function)  # type: ignore[union-attr]
-        assert parent is fn
+        function_parent = fn.getparent(pytest.Function)
+        assert function_parent is fn
 
-        parent = fn.getparent(pytest.Class)  # type: ignore[union-attr]
-        assert parent is cls
+        class_parent = fn.getparent(pytest.Class)
+        assert class_parent is cls
 
     def test_getcustomfile_roundtrip(self, pytester: Pytester) -> None:
         hello = pytester.makefile(".xxx", hello="world")
@@ -272,7 +274,7 @@ class TestCollectPluginHookRelay:
                     wascalled.append(path)
 
         testdir.makefile(".abc", "xyz")
-        pytest.main([testdir.tmpdir], plugins=[Plugin()])  # type: ignore[list-item]
+        pytest.main(testdir.tmpdir, plugins=[Plugin()])
         assert len(wascalled) == 1
         assert wascalled[0].ext == ".abc"
 
@@ -630,8 +632,8 @@ class Test_getinitialnodes:
         assert col.name == "x.py"
         assert col.parent is not None
         assert col.parent.parent is None
-        for col in col.listchain():  # type: ignore[assignment]
-            assert col.config is config
+        for parent in col.listchain():
+            assert parent.config is config
 
     def test_pkgfile(self, pytester: Pytester) -> None:
         """Verify nesting when a module is within a package.
@@ -645,14 +647,15 @@ class Test_getinitialnodes:
         with subdir.cwd():
             config = pytester.parseconfigure(x)
         col = pytester.getnode(config, x)
-        assert col.name == "x.py"  # type: ignore[union-attr]
+        assert col is not None
+        assert col.name == "x.py"
         assert isinstance(col, pytest.Module)
         assert isinstance(col.parent, pytest.Package)
         assert isinstance(col.parent.parent, pytest.Session)
         # session is batman (has no parents)
         assert col.parent.parent.parent is None
-        for col in col.listchain():  # type: ignore[assignment]
-            assert col.config is config
+        for parent in col.listchain():
+            assert parent.config is config
 
 
 class Test_genitems:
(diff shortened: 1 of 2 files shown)


testing/test_collection.py
```diff
@@ -49,31 +63,33 @@ def test_fail(): assert 0
...
             class TestClass:
                  def test_foo(self):
                      pass
         """
         )
-        cls = testdir.collect_by_name(modcol, "TestClass")
-        fn = testdir.collect_by_name(testdir.collect_by_name(cls, "()"), "test_foo")
+        cls = pytester.collect_by_name(modcol, "TestClass")
+        fn = pytester.collect_by_name(
+            pytester.collect_by_name(cls, "()"), "test_foo"  # type: ignore[arg-type]
```
bluetech, line 78: This `type: ignore` actually indicates a real issue with the typing of `Pytester.collect_by_name` -- the `modcol` argument has type `Module`, but here it is used as a general `Collector`. The implementation seems fine with this. So I suggest we fix it:

1. Change `Pytester.collect_by_name` argument `modcol` type from `Module` to `Collector`.
2. Same for `Testdir.collect_by_name`.
3. Change `Pytester._mod_collections` attribute from `Module` key to `Collector` key.

That should fix the `arg-type` issue.

testing/test_collection.py
```diff
@@ -49,31 +63,33 @@ def test_fail(): assert 0
...
+        )
 
-        parent = fn.getparent(pytest.Module)
+        parent = fn.getparent(pytest.Module)  # type: ignore[union-attr]
```
bluetech, line 81: I think we can avoid the ignores here if we add more asserts and avoiding reusing the `parent` variable name for values of different types. I suggest this:

```py
        cls = pytester.collect_by_name(modcol, "TestClass")
        assert isinstance(cls, pytest.Class)
        instance = pytester.collect_by_name(cls, "()")
        assert isinstance(instance, pytest.Instance)
        fn = pytester.collect_by_name(instance, "test_foo")
        assert isinstance(fn, pytest.Function)

        module_parent = fn.getparent(pytest.Module)
        assert module_parent is modcol

        function_parent = fn.getparent(pytest.Function)
        assert function_parent is fn

        class_parent = fn.getparent(pytest.Class)
        assert class_parent is cls
```

testing/test_collection.py
```diff
@@ -254,19 +272,19 @@ def pytest_collect_file(self, path):
                     wascalled.append(path)
 
         testdir.makefile(".abc", "xyz")
-        pytest.main([testdir.tmpdir], plugins=[Plugin()])
+        pytest.main([testdir.tmpdir], plugins=[Plugin()])  # type: ignore[list-item]
```
bluetech, line 275:
```suggestion
        pytest.main(testdir.tmpdir, plugins=[Plugin()])
```

testing/test_collection.py
```diff
@@ -602,43 +620,44 @@ def test_method(self):
...
+        with tmpdir.cwd():
+            config = pytester.parseconfigure(x)
+        col = pytester.getnode(config, x)
         assert isinstance(col, pytest.Module)
         assert col.name == "x.py"
         assert col.parent is not None
         assert col.parent.parent is None
-        for col in col.listchain():
+        for col in col.listchain():  # type: ignore[assignment]
             assert col.config is config
```
bluetech, line 634:
```suggestion
        for parent in col.listchain():
            assert parent.config is config
```

The issue here happens because the name `col` is reused -- when it gets assigned the first time mypy assigns it the type of the right-hand side, and here it doesn't match the `listchain()` type. By using a different name the issue is avoided.

testing/test_collection.py
```diff
@@ -602,43 +620,44 @@ def test_method(self):
...
-        col = testdir.getnode(config, x)
-        assert col.name == "x.py"
+        tmpdir = pytester.path
+        subdir = tmpdir.joinpath("subdir")
+        x = ensure_file(subdir / "x.py")
+        ensure_file(subdir / "__init__.py")
+        with subdir.cwd():
+            config = pytester.parseconfigure(x)
+        col = pytester.getnode(config, x)
+        assert col.name == "x.py"  # type: ignore[union-attr]
```
bluetech, line 648:
```suggestion
        assert col is not None
        assert col.name == "x.py"
```

testing/test_collection.py
```diff
@@ -602,43 +620,44 @@ def test_method(self):
...
         assert isinstance(col, pytest.Module)
         assert isinstance(col.parent, pytest.Package)
         assert isinstance(col.parent.parent, pytest.Session)
         # session is batman (has no parents)
         assert col.parent.parent.parent is None
-        for col in col.listchain():
+        for col in col.listchain():  # type: ignore[assignment]
             assert col.config is config
```
bluetech, line 655:
```suggestion
        for parent in col.listchain():
            assert parent.config is config
```

bluetech (commented):
Thanks @duthades, excellent work. I left a few comments, LGTM otherwise.

I agree about the three cases, we can keep them using `Testdir` until the underlying functionality is ported to pathlib.

testing/test_collection.py
duthades: I made this change.  ```arg-type``` issue is gone.

testing/test_collection.py
duthades: I have made this change.

testing/test_collection.py
duthades: Okay I got it. The error was coming because of reassignment. I have made this change.

testing/test_collection.py
duthades: I have made this change.

testing/test_collection.py
duthades: I have made this change.

testing/test_collection.py
duthades: I have made this change.

duthades:
@bluetech Thank you for the comments. This is the first time I am working with mypy. I read mypy documentation. Now, I have some more idea how it works. I have commited the changes you mentioned.

bluetech (approved):
Thanks @duthades!

PR #8005, opened 2020-11-07
Stop importing `pytest` to avoid upcoming import cycles

bluetech:
Don't import `pytest` from within some `_pytest` modules since an upcoming commit will import from them into `pytest`.

It would have been nice not to have to do it, so that internal plugins look more like external plugins, but with the existing layout this seems unavoidable.


RonnyPfannschmidt (approved):
Lovely, thanks

nicoddemus (approved):

PR #8008, opened 2020-11-07
Capitalize headlines

frankgerhardt:
Only 3 upper case character. But because the docs are so perfect already.


nicoddemus (approved):
Great, thank you!

PR #8014, opened 2020-11-09
assertion/rewrite: write pyc's according to PEP-552 on Python>=3.7

bluetech:
Python 3.7 changed the pyc format by adding a flags word: https://www.python.org/dev/peps/pep-0552/. Even though it is not necessary for us to match it, it is nice to be able to read pyc files we emit with standard tools for debugging the rewriter.

Update our custom pyc files to use that format. We write flags==0 meaning we still use the mtime+size format rather than the newer hash format.

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -281,12 +281,16 @@ def _write_pyc_fp(
 ) -> None:
     # Technically, we don't have to have the same pyc format as
     # (C)Python, since these "pycs" should never be seen by builtin
-    # import. However, there's little reason deviate.
+    # import. However, there's little reason to deviate.
     fp.write(importlib.util.MAGIC_NUMBER)
+    # https://www.python.org/dev/peps/pep-0552/
+    if sys.version_info >= (3, 7):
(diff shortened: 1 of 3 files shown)


src/_pytest/assertion/rewrite.py
```diff
@@ -365,21 +369,33 @@ def _read_pyc(
...
             trace(f"_read_pyc({source}): OSError {e}")
             return None
         # Check for invalid or out of date pyc file.
-        if (
-            len(data) != 12
-            or data[:4] != importlib.util.MAGIC_NUMBER
-            or struct.unpack("<LL", data[4:]) != (mtime & 0xFFFFFFFF, size & 0xFFFFFFFF)
-        ):
-            trace("_read_pyc(%s): invalid or out of date pyc" % source)
+        if len(data) != (16 if has_flags else 12):
```
nicoddemus, line 383: Nice cleanup!

nicoddemus (approved):
Great work! Just missing a CHANGELOG entry! 😁 👍

bluetech:
I added a changelog and a test to improve the coverage of all of those ifs.

PR #8017, opened 2020-11-09
Export types of builtin fixtures for type annotations

bluetech:
The first commit is a hack to fix the issue discussed here: https://github.com/pytest-dev/pytest/issues/7469#issuecomment-723304224

The second commit is a partial fix for #7469 -- only for the fixture types. If this is accepted I will work on the other items on the list, but the fixtures are the most urgent so I'm starting with them.

Please see the commit messages for the approach I took, particularly the "privacy" mechanism in the second commit.

### DIFF
--- a/changelog/7469.improvement.rst
+++ b/changelog/7469.improvement.rst
@@ -1,0 +1,23 @@
+It is now possible to construct a :class:`MonkeyPatch` object directly as ``pytest.MonkeyPatch()``,
+in cases when the :fixture:`monkeypatch` fixture cannot be used. Previously some users imported it
+from the private `_pytest.monkeypatch.MonkeyPatch` namespace.
+
+The types of builtin pytest fixtures are now exported so they may be used in type annotations of test functions.
+The newly-exported types are:
+
+- ``pytest.FixtureRequest`` for the :fixture:`request` fixture.
+- ``pytest.Cache`` for the :fixture:`cache` fixture.
+- ``pytest.CaptureFixture[str]`` for the :fixture:`capfd` and :fixture:`capsys` fixtures.
+- ``pytest.CaptureFixture[bytes]`` for the :fixture:`capfdbinary` and :fixture:`capsysbinary` fixtures.
+- ``pytest.LogCaptureFixture`` for the :fixture:`caplog` fixture.
+- ``pytest.Pytester`` for the :fixture:`pytester` fixture.
+- ``pytest.Testdir`` for the :fixture:`testdir` fixture.
+- ``pytest.TempdirFactory`` for the :fixture:`tmpdir_factory` fixture.
+- ``pytest.TempPathFactory`` for the :fixture:`tmp_path_factory` fixture.
+- ``pytest.MonkeyPatch`` for the :fixture:`monkeypatch` fixture.
+- ``pytest.WarningsRecorder`` for the :fixture:`recwarn` fixture.
+
+Constructing them is not supported (except for `MonkeyPatch`); they are only meant for use in type annotations.
+Doing so will emit a deprecation warning, and may become a hard-error in pytest 7.0.
+
+Subclassing them is also not supported. This is not currently enforced at runtime, but is detected by type-checkers such as mypy.
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -25,6 +25,7 @@ from _pytest.config import Config
 from _pytest.config import ExitCode
 from _pytest.config import hookimpl
 from _pytest.config.argparsing import Parser
+from _pytest.deprecated import check_ispytest
 from _pytest.fixtures import fixture
 from _pytest.fixtures import FixtureRequest
 from _pytest.main import Session
@@ -53,7 +54,7 @@ Signature: 8a477f597d28d172789f06886806bc55
 
 
 @final
-@attr.s
+@attr.s(init=False)
 class Cache:
     _cachedir = attr.ib(type=Path, repr=False)
     _config = attr.ib(type=Config, repr=False)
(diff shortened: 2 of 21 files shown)


nicoddemus (approved):

nicoddemus:
Unfortunately the `_ispytest` solution is very verbose requiring many changes, but I don't have a better suggestion. 👍

src/_pytest/config/__init__.py
```diff
@@ -256,6 +256,7 @@ def directory_arg(path: str, optname: str) -> str:
 
 builtin_plugins = set(default_plugins)
```
graingert, line 257:
```suggestion
 builtin_plugins = {*default_plugins, "pytester"}
```

src/_pytest/config/__init__.py
```diff
@@ -256,6 +256,7 @@ def directory_arg(path: str, optname: str) -> str:
...
 builtin_plugins.add("pytester")
+builtin_plugins.add("pytester_assertions")
```
graingert, line 259: the pytester_assertions module doesn't contain any hook functions and:

> A plugin contains one or multiple hook functions. 

so it shouldn't need to be added to the builtin_plugins

graingert, line 259:
```suggestion
builtin_plugins = {*default_plugins, "pytester"}
```

the pytester_assertions module doesn't contain any hook functions and:

> A plugin contains one or multiple hook functions. 

so it shouldn't need to be added to the builtin_plugins

src/_pytest/pytester_assertions.py
```diff
@@ -0,0 +1,56 @@
+"""Helper plugin for pytester; should not be loaded on its own."""
+# This plugin contains assertions used by pytester. pytester cannot
+# contain them itself, since it is imported by the `pytest` module,
+# hence cannot be subject to assertion rewriting, which requires a
+# module to not be already imported.
+from _pytest.pytester import HookRecorder
+from _pytest.pytester import RunResult
```
graingert, line 7: I think it might be cleaner to avoid the cyclic import here, and have `assertoutcome` take the return value of `listoutcomes`, and assert_outcomes take the return value of `parseoutcomes`

src/_pytest/cacheprovider.py
```diff
@@ -53,7 +54,7 @@
 
 
 @final
-@attr.s
+@attr.s(init=False)
```
graingert, line 57: can you do this with
```
_ispytest: bool = attr.ib(converter=check_ispytest, repr=False)
```

src/_pytest/deprecated.py
```diff
@@ -59,3 +61,10 @@
 STRICT_OPTION = PytestDeprecationWarning(
     "The --strict option is deprecated, use --strict-markers instead."
 )
+
+PRIVATE = PytestDeprecationWarning("A private pytest class or function was used.")
+
+
+def check_ispytest(ispytest: bool) -> None:
+    if not ispytest:
```
graingert, line 69: is this cleaner with a sentinel object instead?

```
class IsPytest:
    pass

IS_PYTEST = IsPytest()

def check_ispytest(ispytest: IsPytest) -> None:
    if ispytest is not IS_PYTEST:
        warn(PRIVATE, stacklevel)
```

I don't see any `*, _ispytest: bool = True` default set, so I can't see how the PytestDeprecationWarning would be triggered

changelog/7469.improvement.rst
```diff
@@ -0,0 +1,23 @@
...
+- ``pytest.Testdir`` for the :fixture:`testdir` fixture.
+- ``pytest.TempdirFactory`` for the :fixture:`tmpdir_factory` fixture.
+- ``pytest.TempPathFactory`` for the :fixture:`tmp_path_factory` fixture.
+- ``pytest.MonkeyPatch`` for the :fixture:`monkeypatch` fixture.
+- ``pytest.WarningsRecorder`` for the :fixture:`recwarn` fixture.
+
+Constructing them is not supported (except for `MonkeyPatch`); they are only meant for use in type annotations.
+Doing so will emit a deprecation warning, and may become a hard-error in pytest 7.0.
+
+Subclassing them is also not supported. This is not currently enforced at runtime, but is detected by type-checkers such as mypy.
```
graingert, line 23: it's quite clean to add a runtime deprecation warning, so might be worth doing in this PR:

```python
def __init_subclass__(cls, **kwargs):
    warnings.warn(f"subclassing {cls} is deprecated", PytestDeprecationWarning)
    super().__init_subclass__(**kwargs)
```

graingert (changes requested):
I think you're missing several `*, _ispytest: bool = False` defaults

src/_pytest/pytester.py
```diff
@@ -66,6 +69,9 @@
     import pexpect
 
 
+pytest_plugins = ["pytester_assertions"]
```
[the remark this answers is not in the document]
bluetech: It is needed for `pytester_assertions` to be loaded as a plugin and be subject to assertion rewriting.

src/_pytest/config/__init__.py
bluetech: That definition is a bit strict I guess, there is nothing actually requiring that a plugin have hooks...

In this case I need it to be a plugin so it gets rewritten.

src/_pytest/pytester_assertions.py
bluetech: I wasn't really bothered with a cyclic import in this specific case, but I it is better to avoid -- will update.

src/_pytest/cacheprovider.py
bluetech: This will actually keep a `_ispytest` attribute but I only want an `__init__` argument.

src/_pytest/deprecated.py
bluetech: Is the reason for the sentinel to make it even harder to bypass? TBH my intention is not to prevent anyone from passing `_ispytest=True` in their own code if they want to bypass the privacy check, this is pytest/python after all :) My intention is just to add a simple hurdle that says "warranty is off, this may break without notice".

changelog/7469.improvement.rst
bluetech: Interesting, I'll open an issue to track this suggestion.

bluetech:
> I think you're missing several *, _ispytest: bool = False defaults

Whoops!

Thanks for the review. I updated the PR, you can see the changes [here](https://github.com/pytest-dev/pytest/compare/ee20d13452a654cb03c0b52f900cb38f3fd7c6f6..f1e6fdcddbfe8991935685ccc5049dd957ec4382).

src/_pytest/cacheprovider.py
graingert: I really don't like having to define `__init__` methods in `@attr.s` classes:

perhaps a decorator can work here?

```python
def for_pytest_only(cls: T) -> T:
    original = cls.__init__

    @functools.wraps(original):
    def new_init(self, *args, **kwargs, *_is_pytest=False) -> None:
        if not _is_pytest:
            warnings.warn(...)
        original(*args, **kwargs)

    cls.__init__ = new_init
    return cls

T_Ctor = TypeVar["T_Ctor", bound=Callable[..., T])

def construct_from_pytest(cls: T_Ctor) -> T_Ctor:
    return functools.partial(cls, _is_pytest=True)
```


```python
@for_pytest_only
@attr.s
class Foo:
    ham = attr.s()


_Foo = construct_from_pytest(Foo)
```

graingert: It should be possible to use a `_is_pytest = attr.ib()` to skip assigning an attribute:

> To be clear: the example above is totally possible with attrs, it's just a bit more clunky.

https://github.com/python-attrs/attrs/issues/342

bluetech: This seems complicated to me. The `__init__` adds a bit of repetition but it is pretty straightforward and easy to understand.

BTW, my own *real* preference is to entirely avoid attrs/dataclasses in any public API, because then it essentially becomes a part of the contract. See also in another language: https://jakewharton.com/public-api-challenges-in-kotlin/.

graingert: > The `__init__` adds a bit of repetition

avoiding this repetition is the main purpose of `attr.s`, so I'd really rather avoid using `init=False` here

graingert: is also reduces the very heavy repetition of ` *, _ispytest: bool = False`

bluetech: @graingert The decorator seems too heavy to me, so I prefer not to have it. And init-only arguments are not supported by attrs from what I understand. So I think `init=False` is the way to go for now. So I'm marking this as resolved. But please let me know if you strongly disagree.

src/_pytest/deprecated.py
bluetech: (Marking as resolved for now based on my reply)

bluetech:
Merging per @nicoddemus's approval and replies to @graingert's comments. Let's see how the `_is_pytest` stuff works out!

PR #8018, opened 2020-11-09
Migrate from testdir to pytester

JosiasAurel:


JosiasAurel:
Remade another pull requests oncoming

PR #8038, opened 2020-11-15
Docs: Note lifetime of temporary directories

matthewhughes934:
Explanation: The default handling of these lifetimes is done in
`tmpdir.TempPathFactory.getbasetemp`, which passes `keep=3` to
`pathlib.make_numbered_dir_with_cleanup`.

GH Issue: #8036

### DIFF
--- a/src/_pytest/tmpdir.py
+++ b/src/_pytest/tmpdir.py
@@ -174,6 +174,11 @@ def tmpdir(tmp_path: Path) -> py.path.local:
     function invocation, created as a sub directory of the base temporary
     directory.
 
+    By default, a new base temporary directory is created each test session,
+    and old bases are removed after 3 sessions, to aid in debugging. If
+    ``--basetemp`` is used then it is cleared each session. See :ref:`base
+    temporary directory`.
+
     The returned object is a `py.path.local`_ path object.
 
     .. _`py.path.local`: https://py.readthedocs.io/en/latest/path.html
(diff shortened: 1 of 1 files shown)


matthewhughes934:
I wasn't sure if the comment was best placed directly in the reference or in the pydocs.

nicoddemus (approved):
Thanks!

src/_pytest/tmpdir.py
```diff
@@ -172,7 +172,9 @@ def _mk_tmp(request: FixtureRequest, factory: TempPathFactory) -> Path:
 def tmpdir(tmp_path: Path) -> py.path.local:
     """Return a temporary directory path object which is unique to each test
     function invocation, created as a sub directory of the base temporary
-    directory.
+    directory. After 3 sessions the directory will be removed by default,
```
bluetech, line 175: I think it would be good to describe the situation in a bit more detail. I suggest writing this in a separate paragraph:

```
The temporary directory is created under a base directory. By default, a
new base is created each test session, and old bases are removed after 3
sessions, to aid in debugging. If ``--basetemp`` is used then it is
cleared each session. See :ref:`base temporary directory`.
```

WDYT?

matthewhughes934: Thanks, that sounds good, particularly the extra link! I updated with 6a25660. I left out your first sentence since I thought it was sufficiently covered in the existing docs, but I can be persuaded otherwise.

bluetech (approved):
Thanks @matthewhughes934!

PR #8040, opened 2020-11-16
stop assigning nextline if its potentially not used

symonk:
only reference `nextline` inside the while loop of `LineMatcher._match_lines` rather than set it to None upfront needlessly.


symonk:
@nicoddemus / @bluetech I think we might have a bit of a environmental CI issue around `set-env` (unrelated)

symonk:
I think my previous comment has a fix merged

bluetech:
> I think my previous comment has a fix merged

Right, it should be fixed, just need to rebase on master. The change LGTM otherwise.

Issue #8042, opened 2020-11-17
--durations=0 reports each test twice


fenchu:
- [ ] a detailed description of the bug or problem you are having
with --durations=0 I get each test listed twice
```
======================================================================================================================================== slowest durations ======================================================================================================================================== 
150.30s call     tests/test_10_scim_delete.py::TestClass::test01_prereg_delete_users[test12]
21.23s call     tests/test_03_scim_insert.py::TestClass::test01_prereg_users[test12]
3.03s call     tests/test_01_services.py::TestClass::test01_services_up[test12]
3.02s call     tests/test_01_services.py::TestClass::test01_services_up[test12]
2.90s call     tests/test_11_dsfmock_delete.py::TestClass::test02_delete_dsfmock[test12]
2.90s call     tests/test_11_dsfmock_delete.py::TestClass::test02_delete_dsfmock[test12]
2.63s call     tests/test_02_dsfmock_insert.py::TestClass::test02_insert_dsfmock[test12]
2.63s call     tests/test_02_dsfmock_insert.py::TestClass::test02_insert_dsfmock[test12]
0.38s call     tests/test_03_scim_insert.py::TestClass::test01_prereg_users_from_file
0.38s call     tests/test_03_scim_insert.py::TestClass::test01_prereg_users_from_file
0.32s call     tests/test_10_scim_delete.py::TestClass::test01_prereg_users_from_file
0.32s call     tests/test_10_scim_delete.py::TestClass::test01_prereg_users_from_file
0.02s call     tests/test_02_dsfmock_insert.py::TestClass::test01_gen_users
0.02s call     tests/test_02_dsfmock_insert.py::TestClass::test01_gen_users
0.01s call     tests/test_11_dsfmock_delete.py::TestClass::test01_prereg_users_from_file
0.01s call     tests/test_11_dsfmock_delete.py::TestClass::test01_prereg_users_from_file

```
- [ ] output of `pip list` from the virtual environment you are using
```
pip freeze -l
astroid==2.4.2
atomicwrites==1.4.0
attrs==20.3.0
certifi==2020.11.8
cffi==1.14.3
chardet==3.0.4
colorama==0.4.4
cryptography==3.2.1
idna==2.10
iniconfig==1.1.1
isort==5.6.4
lazy-object-proxy==1.4.3
mccabe==0.6.1
packaging==20.4
pep8==1.7.1
pluggy==0.13.1
py==1.9.0
pycparser==2.20
PyJWT==1.7.1
[...]

PR #8043, opened 2020-11-17
refactor runner terminal summary and add more durations coverage

symonk:
Hi all, while investigating #8042 (Awaiting response from the author of that issue before I invest more time investigating as I am unable to recreate) I did a bit of tidy up in the `terminal_summary` hook of the `runner.py` plugin. 

- Improved (in my opinion) local scope var names to be a little more readable.
- replace the sorting with attrgetter vs the anonymous lambda function.
- Move to the preferred getOption(attr) calls via the config obj, rather than the option.attr access
- Pluralize the output to handle the edge case of `1 duration` vs `x durations`.
- Added some extra test coverage around `--durations`.
- slightly more readable slower than output message, including formatting to 3f (in-line with --duration-min default `0.005`.

I plan to fix the issue of 8042 if it is recreatable on a windows environment, unable to recreate so that will be a separate PR.

Thanks.


symonk:
Swapping from lambda to attrgetter is maybe over zealous for no gain really and an extra import, thoughts on that? Leaning towards putting the lambda back here

testing/acceptance_test.py
```diff
@@ -1288,3 +1290,62 @@ def test_no_brokenpipeerror_message(pytester: Pytester) -> None:
...
+        @pytest.mark.parametrize('_', range(6))
+        def test_this(_):
+            time.sleep(0.1)
+        """
+    )
+    result = pytester.runpytest("--durations=0")
+    assert result.ret == ExitCode.OK
+    # Check 6x setup & 6x teardown are hidden, but 6x call are shown
+    insertable_line = "*test_durations_zero_lists_all.py::test_this*{}*"
+    lines = [insertable_line.format(x) for x in range(6)]
```
nicoddemus, line 1347: In my experience tests that depend on a tight sleep like this one tend to be flaky; sometimes an overloaded CI machine will take a bit longer to run a setup function (CI machines tend to have shared CPUs) and then the test will fail.

Don't have a specific suggestion to improve this right now, I suggest we keep this in mind and wait and see.

nicoddemus (approved):
Thanks!

About the lambda vs operator.getter, I tend to use the latter but I think it is mainly a personal choice, so go with whatever you prefer.

testing/acceptance_test.py
symonk: Yeah I was kind of confused in how best to test it (reliably), I can increase the sleep in the test(s) themselves but at the price of 'dead time' in the test run, that said I wouldn't want to introduce something very flaky as its a massive time headache later, open to suggestions

nicoddemus: One way would be to make unit-tests instead of integration tests; that would require to refactor `pytest_terminal_summary` to extract the relevant logic about building up the list report with pre-defined durations which don't depend on sleep, and then test that.

symonk: will give it a go, any advice on how to proceed? refactoring a hook seems like the knock on effects could be collossal

nicoddemus: > will give it a go, any advice on how to proceed? refactoring a hook seems like the knock on effects could be collossal

I meant to just refactor the `pytest_terminal_hook` _implementation_ in `terminal.py`, not to change the `pytest_terminal_hook` spec itself, so the changes should be local and not affect anything else other than `terminal.py`.

src/_pytest/runner.py
```diff
@@ -64,35 +65,43 @@ def pytest_addoption(parser: Parser) -> None:
...
+    durations_list = []
+    for report_list in reporter.stats.values():
+        for report in report_list:
+            if hasattr(report, "duration"):
+                durations_list.append(report)
+    if not durations_list:
         return
-    dlist.sort(key=lambda x: x.duration, reverse=True)  # type: ignore[no-any-return]
+
+    durations_list.sort(key=attrgetter("duration"), reverse=True)
```
bluetech, line 83: I prefer the lambda only because `attrgetter` gets the field name as a string which means it cannot be typed.

src/_pytest/runner.py
```diff
@@ -64,35 +65,43 @@ def pytest_addoption(parser: Parser) -> None:
...
-            tr.write_line(
-                "(%s durations < %gs hidden.  Use -vv to show these durations.)"
-                % (len(dlist) - i, durations_min)
+        reporter.write_sep("=", "slowest %s durations" % durations)
+        durations_list = durations_list[:durations]
+
+    for index, test_report in enumerate(durations_list):
+        if verbose < 2 and test_report.duration < durations_min:
+            total = len(durations_list) - index
+            pronoun, dur = (
```
bluetech, line 94: I suggest `dur` -> `subject`

bluetech (approved):
Nice improvement @symonk. 

I agree with @nicoddemus about the `sleep`s, they will likely be flaky, in addition to slowing down the tests slightly, but if you don't get around to finding a better solution I think we can still merge, but remove/adjust the tests if they prove to be flaky.

src/_pytest/runner.py
bluetech: (I mean type *checked*)

nicoddemus:
Hi folks, anything else needed here (other than fixing the conflicts)?

symonk:
@nicoddemus / @bluetech thanks for the reviews here, ultimately I am not happy merging something tested like this into `main`, i think it will be an injustice later and create some development headache with flakyness, I plan to revisit this PR and rewrite the `pytest_terminal_summary` into something more unit-testable and write some more robust coverage here, its my current short term goal in pytest anyway.  Thanks for the reviews thus far, hopefully it won't take too long and sorry for the delay

obestwalter:
@symonk would you like to finish this up? Or maybe somebody else? Otherwise, I think we might rather want to close this for now.

bluetech:
Closing as stale per comment above.

PR #8044, opened 2020-11-17
Prefix contextmanagers with module name in doc examples

timhoffm:
AFAIK it's good practice to only `import pytest` and not individual functions (`from pytest import raises`).

Therefore code examples should have the context managers prefixed with the module name.

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -597,6 +597,7 @@ def raises(
     Use ``pytest.raises`` as a context manager, which will capture the exception of the given
     type::
 
+        >>> import pytest
         >>> with pytest.raises(ZeroDivisionError):
         ...    1/0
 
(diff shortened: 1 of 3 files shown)


timhoffm:
Um, doctest fails because `pytest` is not known there. I thought that could be fixed via the added fixture in `conftest.py`, but apparently I'm doing something wrong. A hint how to fix this is welcome.

nicoddemus:
> A hint how to fix this is welcome.

I just imported pytest explicitly and that solved it. Took the liberty of pushing to your fork. 

LGTM, thanks!

nicoddemus (approved):

nicoddemus:
Reminder to ourselves: we should squash/merge this. 👍

bluetech (approved):

PR #8048, opened 2020-11-18
Add small section on migrating from nose to pytest.

mickeypash:
This change was initiated after a discussion found [here](https://github.com/pytest-dev/pytest/discussions/8024#discussioncomment-127351).

The section currently features the `nose2pytest` tool with plans to expand
on some of the common gotchas when performing such migrations.

Questions for reviewers:
- Should I update the the CHANGELOG - I think it's not necessary.
- Should I add myself to AUTHORS - the change seems pretty small so I'll leave it for now.

PR #8051, opened 2020-11-19
Fix nose documentation

The-Compiler:
Follow-up to #8048 which seems to have been merged without the suggested changes.

cc @nicoddemus @mickeypash


nicoddemus:
Ahh thanks for catching that @The-Compiler!

nicoddemus (approved):

mickeypash:
Oh thanks! I applied the suggested changes but the force push would have overwritten them! Sorry!

The-Compiler:
No worries, no harm done :slightly_smiling_face:

PR #8052, opened 2020-11-19
permit node to warn with any warning type, not just PytestWarning

symonk:
Hi all, as per #7615 This pr enabled the item `node.warn` to handle any subclass of Warning rather than be explicitly locked to instances / subclass of PytestWarning.  I am not entirely sure on the 'use-case' for this but I'm just working on contributions and came across the issue.  Normally I wouldn't be a fan of isinstance checking and raising a ValueError under such scenario(s) but I kept the same internals as previously here because otherwise a generic TypeError is raised which offers little benefit to the  caller.

Thanks!

### DIFF
--- a/changelog/7615.improvement.rst
+++ b/changelog/7615.improvement.rst
@@ -1,1 +1,1 @@
-`node.warn` now permits any subclass of Warning, not just PytestWarnings
+:meth:`Node.warn <_pytest.nodes.Node.warn>` now permits any subclass of :class:`Warning`, not just :class:`PytestWarning <pytest.PytestWarning>`.
--- a/changelog/7615.improvement.rst
+++ b/changelog/7615.improvement.rst
@@ -1,0 +1,1 @@
+`node.warn` now permits any subclass of Warning, not just PytestWarnings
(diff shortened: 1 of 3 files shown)


RonnyPfannschmidt (approved):

symonk:
@nicoddemus can you advise on merge guidelines? Im unsure if it requires atleast 2 approvals :)

nicoddemus (approved):
Thanks @symonk!

Took the liberty of adding links to the CHANGELOG and a `..versionchanged` notice to the docstring. 👍

nicoddemus:
> @nicoddemus can you advise on merge guidelines? Im unsure if it requires atleast 2 approvals :)

We don't have a hard rule here... I think if you got at least one approval, you're free to merge the PR yourself, but often is good to get at least two or more approvals to make sure people can make further suggestions (like I just did here). Also helps to leave a quick notice like: "I will merge this later/tomorrow unless someone would like to give this another look", and see if anybody manifests. 😁 

Either way don't hesitate to ask. 👍

nicoddemus:
A quick hint: leave a phrase like "Closes #7615" in the PR description so GitHub will close the issue automatically for us when the PR gets merged. 👍

PR #8054, opened 2020-11-20
[4.6] testing: python 3.10 fix

hroncok:
This is a clean cherry-pick of d9ac2efbcdee123f73edf1829c8bd6a91be3b6d2

### DIFF
--- a/testing/test_runner.py
+++ b/testing/test_runner.py
@@ -336,8 +336,10 @@ class BaseFunctionalTests(object):
         assert reps[2].failed
         assert reps[2].when == "teardown"
         assert reps[2].longrepr.reprcrash.message in (
-            # python3 error
+            # python3 < 3.10 error
             "TypeError: teardown_method() missing 2 required positional arguments: 'y' and 'z'",
+            # python3 >= 3.10 error
+            "TypeError: TestClass.teardown_method() missing 2 required positional arguments: 'y' and 'z'",
             # python2 error
             "TypeError: teardown_method() takes exactly 4 arguments (2 given)",
         )


nicoddemus (approved):
Thanks @hroncok!

nicoddemus:
Hmmm the failures need investigation though.

testing/test_runner.py
```diff
@@ -336,10 +336,9 @@ def test_method(self):
         assert reps[2].failed
         assert reps[2].when == "teardown"
         assert reps[2].longrepr.reprcrash.message in (
-            # python3 error
             "TypeError: teardown_method() missing 2 required positional arguments: 'y' and 'z'",
-            # python2 error
-            "TypeError: teardown_method() takes exactly 4 arguments (2 given)",
```
asottile, line 342: need to keep the python2 error message in this sequence since 4.6.x still supports python 2.x

hroncok: on it

asottile (approved):

hroncok:
Still a lot of CI failures here :/

bluetech:
Unless someone volunteers to fix the 4.6.x CI, I think we can go ahead and merge this, as the failures are unrelated and the patch cannot cause any new ones.

nicoddemus:
> Unless someone volunteers to fix the 4.6.x CI, I think we can go ahead and merge this, as the failures are unrelated and the patch cannot cause any new ones.

I agree, but we will need volunteers to fix CI if we want to make a new 4.6.x release in the future.

PR #8064, opened 2020-11-23
fix mock_timing fixture name (typo) in timing.py

symonk:
Digging through the repo, apologies for nit picking :) came across a small typo for the fixture that resides in `testing/conftest.py` and corrected it.


symonk:
I highly doubt my change is the cause for the failing macos-py38 build :) I think its maybe flaky?

bluetech (approved):
Thanks!

PR #8070, opened 2020-11-25
Refactor acceptance_test to use pytester

locknic:
Partial fix for https://github.com/pytest-dev/pytest/issues/7942

- Refactors acceptance_test.py from using testdir to using pytester.
- Add myself to AUTHORS

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -89,6 +89,7 @@ Dhiren Serai
 Diego Russo
 Dmitry Dygalo
 Dmitry Pribysh
+Dominic Mortlock
 Duncan Betts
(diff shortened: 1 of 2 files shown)


bluetech (commented):
Looks great to me, thanks @locknic.

Only one request: add a `-> None` to the test function signatures, e.g.

```py
def test_config_error(self, pytester: Pytester) -> None:
```

But it's not absolutely necessary if you can't get to it.

locknic:
@bluetech Done, thanks for the review

bluetech (approved):
Perfect, thanks.

PR #8090, opened 2020-11-30
Migrate test_nose.py from testdir to pytester

cmecklenborg:
Partial fix for #7942

This PR does the following:
* Refactors test_nose.py from testdir to pytester


bluetech (approved):
Perfect, thanks!

PR #8091, opened 2020-11-30
Migrate test_parseopt.py from testdir to pytester

cmecklenborg:
Partial fix for #7942

This PR does the following:
* Refactors test_parseopt.py from testdir to pytester


bluetech (approved):
:+1:, thanks for working on this.

PR #8092, opened 2020-12-01
Migrate test_pastebin.py from testdir to pytester

cmecklenborg:
Partial fix for #7942

This PR does the following:
* Refactors test_pastebin.py from testdir to pytester


bluetech (approved):
Thanks! There were some issue with the CI but it's fine after I reran it.

PR #8122, opened 2020-12-11
Some py.path.local -> pathlib.Path

bluetech:
- Some conftest related functions
- `Config._confcutdir`
- Allow arbitrary `os.PathLike[str]` in `gethookproxy`.

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -50,9 +50,11 @@ from _pytest.compat import final
 from _pytest.compat import importlib_metadata
 from _pytest.outcomes import fail
 from _pytest.outcomes import Skipped
+from _pytest.pathlib import absolutepath
 from _pytest.pathlib import bestrelpath
 from _pytest.pathlib import import_path
 from _pytest.pathlib import ImportMode
+from _pytest.pathlib import resolve_package_path
 from _pytest.store import Store
 from _pytest.warning_types import PytestConfigWarning
 
@@ -102,9 +104,7 @@ class ExitCode(enum.IntEnum):
 
 class ConftestImportFailure(Exception):
     def __init__(
-        self,
-        path: py.path.local,
-        excinfo: Tuple[Type[Exception], Exception, TracebackType],
+        self, path: Path, excinfo: Tuple[Type[Exception], Exception, TracebackType],
     ) -> None:
         super().__init__(path, excinfo)
         self.path = path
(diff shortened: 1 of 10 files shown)


RonnyPfannschmidt (approved):
i love what you did there :+1:

nicoddemus (approved):

pllim:
Hello. We think that this PR broke `pytest-doctestplus` plugin downstream. Example log: https://github.com/astropy/pytest-doctestplus/pull/139/checks?check_run_id=1558353239 . Not sure if the breakage is intentional. Any advice would be greatly appreciated. Thanks!


cc @saimn

nicoddemus:
Hi @pllim,

This PR refactored some internal code, I see `pytest_doctestplus/plugin.py:437` calls `_getconftest_pathlist`:

```
/Users/runner/work/pytest-doctestplus/pytest-doctestplus/.tox/py38-test-pytestdev/lib/python3.8/site-packages/pytest_doctestplus/plugin.py:437: in pytest_ignore_collect
    collect_ignore = config._getconftest_pathlist("collect_ignore", path=path.dirpath())
``` 

This PR changed that `path` parameter from `py.path.local` to `Path`.

I believe changing that line to:

```python
    collect_ignore = config._getconftest_pathlist("collect_ignore", path=Path(path).parent)
```

Should fix that issue. 👍

pllim:
Hmm... now it fails here:

```
______________________ ERROR collecting tests/conftest.py ______________________
../../.tox/py38-test-pytestdev/lib/python3.8/site-packages/pytest_doctestplus/plugin.py:196: in collect
    module = self.config.pluginmanager._importconftest(
../../.tox/py38-test-pytestdev/lib/python3.8/site-packages/_pytest/config/__init__.py:560: in _importconftest
    key = conftestpath.resolve()
E   AttributeError: 'LocalPath' object has no attribute 'resolve'
```

astropy/pytest-doctestplus#140

nicoddemus:
> UPDATE: Looks like forcing self.fspath to be Path before passing it into _importconftest works but hopefully not too hacky.

No, that's the correct approach. 

Some background: we are in the process of changing internal functions to use `Path` instead of `py.path.local`, with the long term goal of eventually deprecating/removing the dependency to the `py` library. While we try to shield those changes from end users, changing the actual internal functions is unavoidable, so this unfortunately will have consequences to plugins which call internal functions (and we understand there are many situations where there are no other options other than call those internal functions).

We are sorry for the disruption that this might cause, but please bear with us. 👍

pllim:
No worries and thanks for the timely help, @nicoddemus !

PR #8125, opened 2020-12-12
tox: remove requires: pip>=20.3.1

bluetech:
Causes some [trouble in CI](https://github.com/pytest-dev/pytest/issues/8101#issuecomment-743780316) and not really needed as old pip should still work.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,6 @@ envlist =
     py37-freeze
     docs
     docs-checklinks
-requires = pip >= 20.3.1
 
 [testenv]
 commands =


nicoddemus (approved):

tox.ini
```diff
@@ -16,7 +16,6 @@ envlist =
     py37-freeze
     docs
     docs-checklinks
-requires = pip >= 20.3.1
```
antonblr, line 19: this was added to replicate CI environment locally, while working on #8119 (CI was running pip 20.3.1)

bluetech: Right, it would be interesting to figure out what this error means

> tox.exception.Error: Error: break infinite loop provisioning within /opt/hostedtoolcache/Python/3.8.6/x64/bin/python missing ['pip>=20.3.1']

but I'm taking a bit of a shortcut here to get the upcoming release going.

But even so I think we might want to leave this `requires` out; IIUC, until people get the new pip in their distro it would make the tox env creation a little longer? I haven't looked into it too much though.

antonblr: could be related  https://github.com/tox-dev/tox/issues/1518

PR #8144, opened 2020-12-14
hookspec: add pathlib.Path alternatives to py.path.local parameters in hooks

bluetech:
The first commit is just some scattered `py.path.local` -> `pathlib.Path` conversions.

The second commit adds `pathlib.Path` parameters to all hooks which take `py.path.local`, to allow hook implementers to convert to pathlib. Along with @RonnyPfannschmidt's work on the nodes (#8037) this should enable pytest itself to entirely move away from py.path internally, allowing for eventual deprecation.

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -1,5 +1,6 @@
 """Hook specifications for pytest plugins which are invoked by pytest itself
 and by builtin plugins."""
+from pathlib import Path
 from typing import Any
 from typing import Dict
 from typing import List
@@ -261,7 +262,9 @@ def pytest_collection_finish(session: "Session") -> None:
 
 
 @hookspec(firstresult=True)
-def pytest_ignore_collect(path: py.path.local, config: "Config") -> Optional[bool]:
+def pytest_ignore_collect(
+    fspath: Path, path: py.path.local, config: "Config"
+) -> Optional[bool]:
     """Return True to prevent considering this path for collection.
 
     This hook is consulted for all files and directories prior to calling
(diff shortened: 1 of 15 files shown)


src/_pytest/hookspec.py
```diff
@@ -261,27 +262,39 @@ def pytest_collection_finish(session: "Session") -> None:
...
     This hook is consulted for all files and directories prior to calling
     more specific hooks.
 
     Stops at first non-None result, see :ref:`firstresult`.
 
+    :param pathlib.Path fspath: The path to analyze.
     :param py.path.local path: The path to analyze.
     :param _pytest.config.Config config: The pytest config object.
+
+    versionchanged:: 6.3.0
```
nicoddemus, line 279:
```suggestion
    .. versionchanged:: 6.3.0
```

src/_pytest/hookspec.py
```diff
@@ -261,27 +262,39 @@ def pytest_collection_finish(session: "Session") -> None:
...
+    fspath: Path, path: py.path.local, parent: "Collector"
 ) -> "Optional[Collector]":
     """Create a Collector for the given path, or None if not relevant.
 
     The new node needs to have the specified ``parent`` as a parent.
 
+    :param pathlib.Path fspath: The path to analyze.
     :param py.path.local path: The path to collect.
+
+    versionchanged:: 6.3.0
```
nicoddemus, line 295:
```suggestion
    .. versionchanged:: 6.3.0
```

src/_pytest/hookspec.py
```diff
@@ -261,27 +262,39 @@ def pytest_collection_finish(session: "Session") -> None:
...
+        The ``fspath`` parameter was added as a pathlib.Path equivalent
```
nicoddemus, line 296: I _think_ this works:

```suggestion
        The ``fspath`` parameter was added as a :class:`pathlib.Path` equivalent
```

src/_pytest/monkeypatch.py
```diff
@@ -325,18 +324,15 @@ def syspath_prepend(self, path) -> None:
...
-    def chdir(self, path) -> None:
+    def chdir(self, path: Union[str, "os.PathLike[str]"]) -> None:
         """Change the current working directory to the specified path.
 
-        Path can be a string or a py.path.local object.
+        Path can be a string or a path object.
         """
         if self._cwd is None:
             self._cwd = os.getcwd()
         if hasattr(path, "chdir"):
```
nicoddemus, line 334: We might replace this with if/else with:

```python
Path(path).chdir()
```

nicoddemus (commented):
Cool, left a few minor comments. 👍

RonnyPfannschmidt (commented):
I just skimmed this on mobil and love what I see

We might want to add a warning for hooks using the legacy local path parameter

src/_pytest/monkeypatch.py
bluetech: Looks like `Path` doesn't have a `chdir()` method. So I just went with `os.chdir(path)` instead (i.e. removed the hasattr stuff). If something out there somehow needed that, I guess we'd find out :)

bluetech:
@nicoddemus Thanks, I addressed your comments.

@RonnyPfannschmidt My preference is to only start deprecating the `py.path` stuff once pytest itself is fully migrated, so users can do it all at once.

src/_pytest/monkeypatch.py
nicoddemus: Oh I could swear it did. 😁 

Yeah I think `os.chdir(path)` there is the right call. 👍

nicoddemus (approved):
Great work!

PR #8145, opened 2020-12-14
tests: Migrate to pytester - incremental update

antonblr:
Incremental update for #7942

Includes:
- testing/code/
- testing/examples/
- testing/io/
- testing/logging/
- testing/test_cacheprovider.py
- testing/test_monkeypatch.py
- testing/test_pluginmanager.py
- testing/test_reports.py
- testing/test_runner.py

Also migrated off of `py.path.local` where it was straight forward (and `tmpdir` -> `tmp_path`).

### DIFF
--- a/src/_pytest/reports.py
+++ b/src/_pytest/reports.py
@@ -74,6 +74,10 @@ class BaseReport:
         def __getattr__(self, key: str) -> Any:
             ...
 
+        # Allow dynamic attributes to be set in tests without typing errors.
+        def __setattr__(self, key: str, value: Any) -> None:
+            ...
+
     def toterminal(self, out: TerminalWriter) -> None:
         if hasattr(self, "node"):
             out.line(getworkerinfoline(self.node))
--- a/src/_pytest/reports.py
+++ b/src/_pytest/reports.py
@@ -74,10 +74,6 @@ class BaseReport:
         def __getattr__(self, key: str) -> Any:
             ...
 
-        # Allow dynamic attributes to be set in tests without typing errors.
-        def __setattr__(self, key: str, value: Any) -> None:
-            ...
-
     def toterminal(self, out: TerminalWriter) -> None:
         if hasattr(self, "node"):
             out.line(getworkerinfoline(self.node))
--- a/testing/test_reports.py
+++ b/testing/test_reports.py
@@ -1,4 +1,3 @@
-from pathlib import Path
 from typing import Sequence
 from typing import Union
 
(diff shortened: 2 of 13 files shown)


testing/test_reports.py
```diff
@@ -224,33 +223,31 @@ def test_extended_report_deserialization(self, testdir: Testdir) -> None:
...
         """
         )
-        reprec = testdir.inline_run()
+        reprec = pytester.inline_run()
         reports = reprec.getreports("pytest_runtest_logreport")
         assert len(reports) == 3
         test_a_call = reports[1]
-        test_a_call.path1 = testdir.tmpdir  # type: ignore[attr-defined]
-        test_a_call.path2 = Path(testdir.tmpdir)  # type: ignore[attr-defined]
+        test_a_call.path = pytester.path
```
antonblr, line 238: should I keep check for `py.path.local` as well, for regression?

src/_pytest/reports.py
```diff
@@ -74,6 +74,10 @@ def __init__(self, **kw: Any) -> None:
         def __getattr__(self, key: str) -> Any:
             ...
 
+        # Allow dynamic attributes to be set in tests without typing errors.
+        def __setattr__(self, key: str, value: Any) -> None:
+            ...
```
antonblr, line 79: there are couple of places in tests where I removed `# type: ignore[attr-defined]` - should I have kept them instead of adding `__setattr__` stub?

testing/test_reports.py
bluetech: Yes, the test seems to want to test py.path explicitly (in additional to pathlib) so we should keep at, until we get rid of py.path entirely.

src/_pytest/reports.py
bluetech: I think, since we are probably going to make the `TestReport`/`CollectReport` types public at some point, it will be better to keep the type-ignores in the tests for now, to maintain our flexibility in terms of compatibility. If we find there's a good reason to set attributes after the fact instead of only through the `__init__`, we can reconsider then.

bluetech (commented):
Great work @antonblr; looks like we're getting close!

> Also migrated off of py.path.local where it was straight forward (and tmpdir -> tmp_path).

Good idea, we'll probably have a mini-project for this conversion as well after the pytester conversion is done.

testing/test_reports.py
antonblr: added back in 5d4284f

src/_pytest/reports.py
antonblr: reverted in 5d4284f

bluetech (approved):

PR #8150, opened 2020-12-15
Use manual trigger to prepare release PRs

### LINKED ISSUE #7507: Use new GH actions manual triggers for preparing releases
As seen in: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch

It should be simple to adapt our current script to use this new mechanism instead.

nicoddemus:
Unfortunately this is untested because the workflow must first be merged
to master in order to be available for triggering.

Fix #7507

I intend to leave the other workflow in place until we can test this workflow and are satisfied with it.

### DIFF
--- a/scripts/prepare-release-pr.py
+++ b/scripts/prepare-release-pr.py
@@ -19,7 +19,6 @@ from pathlib import Path
 from subprocess import check_call
 from subprocess import check_output
 from subprocess import run
-from textwrap import dedent
 
 from colorama import Fore
 from colorama import init
@@ -127,16 +126,6 @@ def find_next_version(base_branch: str, is_major: bool) -> str:
     features = list(changelog.glob("*.feature.rst"))
     breaking = list(changelog.glob("*.breaking.rst"))
     is_feature_release = features or breaking
-
-    if is_feature_release and base_branch != "master":
-        msg = dedent(
-            f"""
-            Found features or breaking changes in `{base_branch}`, and feature releases can only be
-            created from `master`:
-        """
-        )
-        msg += "\n".join(f"* `{x.name}`" for x in sorted(features + breaking))
-        raise InvalidFeatureRelease(msg)
 
     if is_major:
         return f"{last_version[0]+1}.0.0"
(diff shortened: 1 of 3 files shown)


scripts/prepare-release-pr.py
```diff
@@ -0,0 +1,162 @@
...
+    import github3
+
+    github = github3.login(token=token)
+    owner, repo = SLUG.split("/")
+    return github.repository(owner, repo)
+
+
+def prepare_release_pr(base_branch: str, is_major: bool, token: str) -> None:
+    print()
+    print(f"Precessing release for branch {Fore.CYAN}{base_branch}")
```
bluetech, line 53:
```suggestion
    print(f"Processing release for branch {Fore.CYAN}{base_branch}")
```

scripts/prepare-release-pr.py
```diff
@@ -0,0 +1,162 @@
+"""
+This script is part of the pytest release process which is manually in the Actions
```
bluetech, line 2:
```suggestion
This script is part of the pytest release process which is triggered manually in the Actions
```

scripts/prepare-release-pr.py
```diff
@@ -0,0 +1,162 @@
...
+    valid_versions.sort()
+    last_version = valid_versions[-1]
+
+    changelog = Path("changelog")
+
+    features = list(changelog.glob("*.feature.rst"))
+    breaking = list(changelog.glob("*.breaking.rst"))
+    is_feature_release = features or breaking
+
+    if is_feature_release and base_branch != "master":
```
bluetech, line 131: BTW, our current `RELEASING.md` says that also for feature releases, we first branch and make the release from the branch, which conflicts with this. I don't really remember anymore when it got out of sync. (But this is pre-existing, not directly related to this PR)

bluetech (approved):

scripts/prepare-release-pr.py
nicoddemus: Indeed. But I will fix this now, it is just a matter of removing the check. 👍

PR #8152, opened 2020-12-15
terminal: fix "(<Skipped instance>)" skip reason in test status line

bluetech:
Noticed this issue in the new 6.2 feature. The following test when run with `pytest -v`

```py
import pytest

def test_it():
    pytest.skip()
```

would show

```
x.py::test_it SKIPPED (<Skipped instance>)                [100%]
```

### DIFF
--- a/src/_pytest/outcomes.py
+++ b/src/_pytest/outcomes.py
@@ -38,7 +38,7 @@ class OutcomeException(BaseException):
         self.pytrace = pytrace
 
     def __repr__(self) -> str:
-        if self.msg:
+        if self.msg is not None:
             return self.msg
         return f"<{self.__class__.__name__} instance>"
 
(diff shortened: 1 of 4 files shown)


src/_pytest/outcomes.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, msg: Optional[str] = None, pytrace: bool = True) -> None:
         self.pytrace = pytrace
 
     def __repr__(self) -> str:
-        if self.msg:
+        if self.msg is not None:
```
bluetech, line 41: This changes the behavior of `str(Skipped(""))`. But if an empty string is explicitly passed I think it makes sense that the `str()` is empty.

asottile (approved):

nicoddemus (approved):

PR #8159, opened 2020-12-16
Use transparent PNG for logo

### LINKED ISSUE #8158: [appearance][Dark Mode] Logo in README should have a transparent background
STR:
1. Enable Dark Mode on GitHub (https://github.com/settings/appearance)
2. Go to https://github.com/pytest-dev/pytest
3. See the pytest logo stand out with its white background while all the surroundings are dark:
pytest-readme-top

The root cause is that the image file does not have a transparent background: https://github.com/pytest-dev/pytest/blob/master/doc/en/img/pytest1.png. And I don't think anybody has its original source. So to fix this, somebody would need to do image editing or maybe image tracing — these days it sounds like a good idea to have an SVG instead.

nicoddemus:
Fix #8158


README.rst
```diff
@@ -1,6 +1,7 @@
-.. image:: https://docs.pytest.org/en/stable/_static/pytest1.png
+.. image:: https://github.com/nicoddemus/pytest/raw/transparent-logo-8158/doc/en/img/pytest_logo_curves.svg
    :target: https://docs.pytest.org/en/stable/
    :align: center
+   :height: 200
```
webknjaz, line 4: This seems to be working too so no need to edit SVG maybe.

nicoddemus: Yeah I didn't edit it. 👍

nicoddemus: Bummer `align: center` doesn't seem to work though... 😕

webknjaz: Yep, GH support for RST is quite limited.

README.rst
```diff
@@ -1,6 +1,7 @@
-.. image:: https://docs.pytest.org/en/stable/_static/pytest1.png
+.. image:: https://github.com/nicoddemus/pytest/raw/transparent-logo-8158/doc/en/img/pytest_logo_curves.svg
```
nicoddemus, line 1: This needs to be updated before merging.

README.rst
nicoddemus: If you are happy with the results, can you approve and then we can get this online? 😁

README.rst
webknjaz: Also: you may want to update the `conf.py` to use SVG and see how it goes.

webknjaz (approved):
LGTM. Feel free to merge unless you want to follow-up on the other comments (like using SVG in Sphinx too or editing SVG to abuse its way of rendering to emulate centering)

nicoddemus:
Yeah using the `svg` file on the docs looks good too:

image

webknjaz:
@nicoddemus you'll need someone else's approval because I don't have the commit bit here, hence I can't unblock it for you.

cc @The-Compiler?

nicoddemus:
> @nicoddemus you'll need someone else's approval because I don't have the commit bit here, hence I can't unblock it for you.

No worries, I can merge it as is. 👍

PR #8162, opened 2020-12-16
Remove Travis references

nicoddemus:
Since we stopped testing Python 3.5, we no longer use Travis.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,9 +20,6 @@
     :target: https://codecov.io/gh/pytest-dev/pytest
     :alt: Code coverage Status
 
-.. image:: https://travis-ci.org/pytest-dev/pytest.svg?branch=master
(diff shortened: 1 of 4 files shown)


antonblr (approved):
hey @nicoddemus - would it also make sense to get rid of "travis" in other places? Like:

https://github.com/pytest-dev/pytest/blob/master/tox.ini#L5
https://github.com/pytest-dev/pytest/blob/master/scripts/publish-gh-release-notes.py#L4
...
And, perhaps, triage issues related to it https://github.com/pytest-dev/pytest/issues?q=is%3Aissue+is%3Aopen+travis ?

nicoddemus:
@antonblr indeed, thanks. I've searched for other Travis references and removed/updated them. 👍 

I will take a look at the issues too.

PR #8166, opened 2020-12-17
Add Changelog to setup.cfg

adamchainz:
One of the main reasons I hit up a project's PyPI page is to get to its changelog. Adding a changelog in the sidebar gets a nice present icon, see e.g. https://pypi.org/project/pytest-randomly/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ classifiers =
     Topic :: Utilities
 keywords = test, unittest
 project_urls =
+    Changelog=https://docs.pytest.org/en/latest/changelog.html
     Source=https://github.com/pytest-dev/pytest
     Tracker=https://github.com/pytest-dev/pytest/issues
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ classifiers =
     Topic :: Utilities
 keywords = test, unittest
 project_urls =
-    Changelog=https://docs.pytest.org/en/latest/changelog.html
+    Changelog=https://docs.pytest.org/en/stable/changelog.html
     Source=https://github.com/pytest-dev/pytest
     Tracker=https://github.com/pytest-dev/pytest/issues
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,6 +27,7 @@ classifiers =
 keywords = test, unittest
 project_urls =
     Changelog=https://docs.pytest.org/en/stable/changelog.html
+    Twitter=https://twitter.com/pytestdotorg
     Source=https://github.com/pytest-dev/pytest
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -26,6 +26,7 @@ classifiers =
     Topic :: Utilities
 keywords = test, unittest
 project_urls =
+    Changelog=https://docs.pytest.org/en/latest/changelog.html
```
bluetech, line 29: Perhaps `/stable` would be better? Not sure...

bluetech (approved):
Thanks, looks handy to me.

setup.cfg
adamchainz: As I understand, the 'latest' page always includes the entire changelog.

nicoddemus: The latest page on PyPI is about the last release, correct? If so the correct changelog would be `stable`: https://docs.pytest.org/en/stable/changelog.html

adamchainz, line 29:
```suggestion
    Changelog=https://docs.pytest.org/en/stable/changelog.html
```

adamchainz: Okay fair

setup.cfg
```diff
@@ -26,6 +26,7 @@ classifiers =
     Topic :: Utilities
 keywords = test, unittest
 project_urls =
+    Changelog=https://docs.pytest.org/en/stable/changelog.html
```
graingert, line 29: there's a full list of all the special icons warehouse supports:

https://github.com/pypa/warehouse/blob/851e4c37bc1a9d1ef299e9300c7aa3ffb7cb09d5/warehouse/templates/packaging/detail.html#L20-L58

we may want to add some more eg https://twitter.com/pytestdotorg

graingert (approved):

setup.cfg
nicoddemus: Should it be then: `twitter=https://twitter.com/pytestdotorg`?

Could you please add an inline suggestion if you are sure about the syntax @graingert ?

nicoddemus (approved):
Thanks @adamchainz!

setup.cfg
graingert:
```suggestion
    Changelog=https://docs.pytest.org/en/stable/changelog.html
    Twitter=https://twitter.com/pytestdotorg
```

@nicoddemus do we want the coverage on here? or anything else from that list?

nicoddemus: IMHO this is good enough... no need to add coverage/CI information here, I think.

nicoddemus:
Backport: https://github.com/pytest-dev/pytest/pull/8167

PR #8167, opened 2020-12-17
[6.2.x] Add Changelog to setup.cfg (#8166)

nicoddemus:


nicoddemus:
Hmm pypy3 failure is unrelated:

```
Run actions/setup-python@v2
  with:
    python-version: pypy3
    token: ***
Error: IS_WINDOWS is not defined
```

Seems like something is wrong with the Action? 🤔

Issue #7679, opened 2020-08-24
Include code that runs in runpytest_subprocess in coverage


bluetech:
In #7651 we had some code which is only covered by a subprocess test and is somewhat hard to cover inprocess.

It *seems* that the code running in the subprocess is not covered. @nicoddemus mentioned that some coverage configuration is needed to get this to work: https://coverage.readthedocs.io/en/stable/subprocess.html

asottile:
one way I've done this before is via `coverage-enable-subprocess`

essentially:
1. install `coverage-enable-subprocess` under test
2. https://github.com/asottile/setuptools-golang/blob/a6eaed0801255d4b2b7fb9e964126bce2747a0a9/tox.ini#L6
3. https://github.com/asottile/setuptools-golang/blob/a6eaed0801255d4b2b7fb9e964126bce2747a0a9/.coveragerc#L5

antonblr:
Closing via #8188 - lines which have been shown as [uncovered](https://codecov.io/gh/pytest-dev/pytest/compare/d426a79a90351dff0492fbd40404b1256b24f91f...0d5a65091d45e04662837ace13dd88dba954ea8a/diff) in #7651 are covered now.

PR #8168, opened 2020-12-17
tests: Migrate to pytester - final update

antonblr:
This should take care of #7942  completely, with a few exceptions: 

* `testing/test_pytester.py` - two tests were not migrated, they seem to be needed for the time being:
  * [testing/test_pytester.py::test_pytester_addopts_before_testdir](https://github.com/pytest-dev/pytest/blob/bd894e3065ba6fa13327ad5dfc94f2b6208cf0ff/testing/test_pytester.py#L618)
  * [testing/test_pytester.py::test_testtmproot](https://github.com/pytest-dev/pytest/blob/bd894e3065ba6fa13327ad5dfc94f2b6208cf0ff/testing/test_pytester.py#L814)
    Other `testdir` tests were renamed/updated to use pytester (code coverage could drop for Testdir proxy functions).

* `testing/test_collection.py` - removed `testdir` usages and added `py.path.local` conversion where required. I should be easy to fix as part of `py.path.local` -> `pathlib` migration.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -123,7 +123,8 @@ jobs:
       with:
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python }}
-      uses: actions/setup-python@v2
+      # https://github.com/actions/setup-python/issues/171
+      uses: actions/setup-python@v2.1.4
       with:
         python-version: ${{ matrix.python }}
     - name: Install dependencies
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -123,8 +123,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python }}
-      # https://github.com/actions/setup-python/issues/171
-      uses: actions/setup-python@v2.1.4
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
     - name: Install dependencies
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -528,7 +528,7 @@ class FSCollector(Collector):
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.gethookproxy(fspath)
 
-    def isinitpath(self, path: "os.PathLike[str]") -> bool:
+    def isinitpath(self, path: Union[str, "os.PathLike[str]"]) -> bool:
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.isinitpath(path)
 
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -528,7 +528,7 @@ class FSCollector(Collector):
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.gethookproxy(fspath)
 
-    def isinitpath(self, path: py.path.local) -> bool:
+    def isinitpath(self, path: "os.PathLike[str]") -> bool:
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.isinitpath(path)
 
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -660,7 +660,7 @@ class Package(Module):
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.gethookproxy(fspath)
 
-    def isinitpath(self, path: "os.PathLike[str]") -> bool:
+    def isinitpath(self, path: Union[str, "os.PathLike[str]"]) -> bool:
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.isinitpath(path)
 
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -660,7 +660,7 @@ class Package(Module):
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.gethookproxy(fspath)
 
-    def isinitpath(self, path: py.path.local) -> bool:
+    def isinitpath(self, path: "os.PathLike[str]") -> bool:
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.isinitpath(path)
 
--- a/testing/test_debugging.py
+++ b/testing/test_debugging.py
@@ -21,10 +21,11 @@ _ENVIRON_PYTHONBREAKPOINT = os.environ.get("PYTHONBREAKPOINT", "")
 
 
 @pytest.fixture(autouse=True)
-def pdb_env(request, monkeypatch: MonkeyPatch):
+def pdb_env(request):
     if "pytester" in request.fixturenames:
         # Disable pdb++ with inner tests.
-        monkeypatch.setenv("PDBPP_HIJACK_PDB", "0")
+        pytester = request.getfixturevalue("pytester")
+        pytester._monkeypatch.setenv("PDBPP_HIJACK_PDB", "0")
 
 
 def runpdb_and_get_report(pytester: Pytester, source: str):
--- a/testing/test_debugging.py
+++ b/testing/test_debugging.py
@@ -21,11 +21,10 @@ _ENVIRON_PYTHONBREAKPOINT = os.environ.get("PYTHONBREAKPOINT", "")
 
 
 @pytest.fixture(autouse=True)
-def pdb_env(request):
+def pdb_env(request, monkeypatch: MonkeyPatch):
     if "pytester" in request.fixturenames:
         # Disable pdb++ with inner tests.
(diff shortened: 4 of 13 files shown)


src/_pytest/nodes.py
```diff
@@ -528,7 +528,7 @@ def gethookproxy(self, fspath: "os.PathLike[str]"):
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.gethookproxy(fspath)
 
-    def isinitpath(self, path: py.path.local) -> bool:
+    def isinitpath(self, path: "os.PathLike[str]") -> bool:
```
antonblr, line 531: Needed for `testing/deprecated_test.py` migration.

testing/test_junitxml.py
```diff
@@ -4,51 +4,60 @@
...
     return the parsed ``DomNode`` of the root xml node.
 
     The ``family`` parameter is used to configure the ``junit_family`` of the written report.
     "xunit2" is also automatically validated against the schema.
     """
 
-    def run(*args, family="xunit1"):
+    def run(
+        *args: Union[str, "os.PathLike[str]"], family: Optional[str] = "xunit1",
+    ) -> Tuple[RunResult, "DomNode"]:
```
antonblr, line 49: would love to have some 👀 on type annotation here as I'm still learning

testing/test_pytester.py
```diff
@@ -323,9 +325,9 @@ def test_remove_added(self) -> None:
         snapshot.restore()
         assert sys.modules == original
 
-    def test_add_removed(self, monkeypatch) -> None:
+    def test_add_removed(self, monkeypatch: MonkeyPatch) -> None:
         assert self.key not in sys.modules
-        monkeypatch.setitem(sys.modules, self.key, "something")
+        monkeypatch.setitem(sys.modules, self.key, self.mod)
```
antonblr, line 330: had to cast into `ModuleType` to avoid `Cannot infer type argument 2 of "setitem" of "MonkeyPatch"` mypy error.

.github/workflows/main.yml
```diff
@@ -123,7 +123,8 @@ jobs:
       with:
         fetch-depth: 0
     - name: Set up Python ${{ matrix.python }}
-      uses: actions/setup-python@v2
+      # https://github.com/actions/setup-python/issues/171
+      uses: actions/setup-python@v2.1.4
```
antonblr, line 127: a workaround for `Error: IS_WINDOWS is not defined` to get builds green ->   https://github.com/actions/setup-python/issues/171

testing/test_pytester.py
```diff
@@ -419,66 +422,64 @@ def test_preserve_container(self, monkeypatch, path_type) -> None:
...
                 \""".format(os.getpid())
             )
-            assert testdir.runpytest(testfile).ret == 0
+            assert pytester.runpytest(testfile).ret == 0
         """
     )
-    result = testdir.runpytest_subprocess(
-        "-p", "pytester", "--runpytest", "subprocess", testfile
-    )
+    result = pytester.runpytest("-p", "pytester", "--runpytest", "subprocess", testfile)
```
antonblr, line 445: I think this should be run via `pytester.runpytest()` rather then `pytester.runpytest_subprocess()` to utilize `"--runpytest", "subprocess"` args? 

Added coverage suggests the same -> [src/_pytest/pytester.py#L1145](https://codecov.io/gh/pytest-dev/pytest/src/9bdd7512db9844af81ac5c8601f8a58e3e92086d/src/_pytest/pytester.py#L1145) (Codecov link)

src/_pytest/nodes.py
bluetech, line 531: Right, I forgot these existed when I changed `Session`. I would match the `Session` signature exactly:

```suggestion
    def isinitpath(self, path: Union[str, "os.PathLike[str]"]) -> bool:
```

src/_pytest/python.py
```diff
@@ -660,7 +660,7 @@ def gethookproxy(self, fspath: "os.PathLike[str]"):
         warnings.warn(FSCOLLECTOR_GETHOOKPROXY_ISINITPATH, stacklevel=2)
         return self.session.gethookproxy(fspath)
 
-    def isinitpath(self, path: py.path.local) -> bool:
+    def isinitpath(self, path: "os.PathLike[str]") -> bool:
```
bluetech, line 663:
```suggestion
    def isinitpath(self, path: Union[str, "os.PathLike[str]"]) -> bool:
```

testing/test_debugging.py
```diff
@@ -21,11 +21,10 @@
 
 
 @pytest.fixture(autouse=True)
-def pdb_env(request):
+def pdb_env(request, monkeypatch: MonkeyPatch):
```
bluetech, line 24: I *suspect* the idea here was to hook into the `pytester`'s monkeypatch lifetime but haven't checked it. I suggest keeping the idea though just to be safe. The `getfixturevalue` would change to `"pytester"` and `pytester.monkeypatch` would change to `pytester._monkeypatch` (with underscore).

testing/test_junitxml.py
bluetech: LGTM!

testing/test_junitxml.py
```diff
@@ -4,51 +4,60 @@
...
+def schema() -> xmlschema.XMLSchema:
     """Return an xmlschema.XMLSchema object for the junit-10.xsd file."""
     fn = Path(__file__).parent / "example_scripts/junit-10.xsd"
     with fn.open() as f:
         return xmlschema.XMLSchema(f)
 
 
 @pytest.fixture
-def run_and_parse(testdir, schema):
+def run_and_parse(pytester: Pytester, schema: xmlschema.XMLSchema) -> T:
```
bluetech, line 39: The `T` here I think is not what we want. In general a generic in return position which is not bound by any inputs is probably not useful, because it does not provide any type safety, the caller can define it to be whatever it wants.

I think `typing.Callable` is not able to type callables with args and kwargs so it'd be hard to type strictly.  [Callable protocol](https://mypy.readthedocs.io/en/stable/protocols.html#callback-protocols) can do it but they require newer Python version than pytest supports. So I suggest just using a callable class like this:

```py
class RunAndParse:
    def __init__(self, pytester: Pytester, schema: xmlschema.XmlSchema) -> None:
        self.pytester = pytester
        self.schema = schema

    def __call__(
        self, *args: str, family: Optional[str] = "xunit1"
    ) -> Tuple[RunResult, "DomNode"]:
        if family:
            args = ("-o", "junit_family=" + family) + args
        xml_path = self.pytester.path.joinpath("junit.xml")
        result = self.pytester.runpytest("--junitxml=%s" % xml_path, *args)
        if family == "xunit2":
            with xml_path.open() as f:
                self.schema.validate(f)
        xmldoc = minidom.parse(str(xml_path))
        return result, DomNode(xmldoc)


@pytest.fixture
def run_and_parse(pytester: Pytester, schema: xmlschema.XmlSchema) -> RunAndParse:
    """Fixture that returns a function that can be used to execute pytest and
    return the parsed ``DomNode`` of the root xml node.

    The ``family`` parameter is used to configure the ``junit_family`` of the written report.
    "xunit2" is also automatically validated against the schema.
    """
    return RunAndParse(pytester, schema)
```

Then you can also type the `run_and_parse` fixture arguments in tests as `run_and_parse: RunAndParse`.

testing/test_pytester.py
```diff
@@ -68,55 +68,55 @@ class rep2:
...
     """
     )
-    result = testdir.runpytest()
+    result = pytester.runpytest()
     result.assert_outcomes(passed=1)
 
 
-def test_testdir_with_doctest(testdir):
-    """Check that testdir can be used within doctests.
+def test_pytester_with_doctest(pytester: Pytester):
```
bluetech, line 95:
```suggestion
def test_pytester_with_doctest(pytester: Pytester) -> None:
```

testing/test_pytester.py
bluetech: :+1: 

It would be a bit more straightforward to read if the `ModuleType("something")` was inlined directly instead of using the `self.mod` field. It's repetitious but that's not so bad in tests IMO.

testing/test_pytester.py
bluetech: Right, I think it makes sense.

Actually, `pytester.runpytest` translates to either `runpytest_subprocess` or `runpytest_inline` depending on the `--runpytest` argument (which is what is being tested here...). So might even want to use `runpytest_inline` directly here.

The coverage thing I think is just because we don't collect coverage from subprocesses. There's an issue for this #7679.

bluetech (commented):
Great work! LGTM with a few comments.

testing/test_junitxml.py
asottile: can `family` be `None`? -- otherwise it should just be `: str` -- I made a video on this topic: [python typing: Optional is not optional!](https://youtu.be/7XZYlDpSxOU?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY)

antonblr: ha, nice video @asottile (I click like 5 times) - I see what you put into "optional" meaning. But yes, `family` is optional *and* could be `None` as well 
https://github.com/pytest-dev/pytest/blob/9bdd7512db9844af81ac5c8601f8a58e3e92086d/testing/test_junitxml.py#L1305

antonblr:
Thanks @bluetech, great feedback! I think I've addressed everything - 196b173c8a86833b96f90128a6cc9928e17b6c23

bluetech (approved):

PR #8174, opened 2020-12-19
More py.path -> pathlib conversions

bluetech:
After this there's not a lot left, mostly the nodes' `fspath` and some `tmpdir`'s in testing.

I tried to split the changes to standalone commits, the last commit however is a bit larger.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -43,6 +43,8 @@ from _pytest._io.saferepr import safeformat
 from _pytest._io.saferepr import saferepr
 from _pytest.compat import final
 from _pytest.compat import get_real_func
+from _pytest.pathlib import absolutepath
+from _pytest.pathlib import bestrelpath
 
 if TYPE_CHECKING:
     from typing_extensions import Literal
@@ -78,16 +80,16 @@ class Code:
         return self.raw.co_name
 
     @property
-    def path(self) -> Union[py.path.local, str]:
+    def path(self) -> Union[Path, str]:
         """Return a path object pointing to source code, or an ``str`` in
         case of ``OSError`` / non-existing file."""
         if not self.raw.co_filename:
             return ""
         try:
-            p = py.path.local(self.raw.co_filename)
+            p = absolutepath(self.raw.co_filename)
             # maybe don't try this checking
-            if not p.check():
-                raise OSError("py.path check failed.")
+            if not p.exists():
+                raise OSError("path check failed.")
             return p
         except OSError:
             # XXX maybe try harder like the weird logic
@@ -223,7 +225,7 @@ class TracebackEntry:
         return source.getstatement(self.lineno)
 
     @property
-    def path(self) -> Union[py.path.local, str]:
+    def path(self) -> Union[Path, str]:
         """Path to the source code."""
         return self.frame.code.path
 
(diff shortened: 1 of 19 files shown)


src/_pytest/_code/code.py
```diff
@@ -223,7 +225,7 @@ def statement(self) -> "Source":
         return source.getstatement(self.lineno)
 
     @property
-    def path(self) -> Union[py.path.local, str]:
```
nicoddemus, line 226: I think a `trivial` changelog entry might be good, in case there are plugin authors using the `code` methods that have changed.

src/_pytest/_code/code.py
```diff
@@ -353,17 +355,19 @@ def cut(
         for x in self:
             code = x.frame.code
             codepath = code.path
+            if path is not None and codepath != path:
```
nicoddemus, line 358: 👍 to breaking the long conditional.

src/_pytest/doctest.py
```diff
@@ -120,32 +121,32 @@ def pytest_unconfigure() -> None:
 
 
 def pytest_collect_file(
-    path: py.path.local, parent: Collector,
+    fspath: Path, path: py.path.local, parent: Collector,
```
nicoddemus, line 124: We don't use `path` anymore, we can drop it from the arguments.

```suggestion
    fspath: Path, parent: Collector
```

src/_pytest/python.py
```diff
@@ -205,13 +206,13 @@ def pytest_collect_file(
...
-def path_matches_patterns(path: py.path.local, patterns: Iterable[str]) -> bool:
+def path_matches_patterns(path: Path, patterns: Iterable[str]) -> bool:
     """Return whether path matches any of the patterns in the list of globs given."""
-    return any(path.fnmatch(pattern) for pattern in patterns)
+    return any(fnmatch_ex(pattern, path) for pattern in patterns)
 
 
-def pytest_pycollect_makemodule(path: py.path.local, parent) -> "Module":
-    if path.basename == "__init__.py":
+def pytest_pycollect_makemodule(fspath: Path, path: py.path.local, parent) -> "Module":
```
nicoddemus, line 214:
```suggestion
def pytest_pycollect_makemodule(fspath: Path, parent) -> "Module":
```

src/_pytest/terminal.py
```diff
@@ -285,30 +285,23 @@ class WarningReport:
...
+    :ivar tuple fslocation:
         File system location of the source of the warning (see ``get_location``).
     """
 
     message = attr.ib(type=str)
     nodeid = attr.ib(type=Optional[str], default=None)
-    fslocation = attr.ib(
-        type=Optional[Union[Tuple[str, int], py.path.local]], default=None
-    )
+    fslocation = attr.ib(type=Optional[Tuple[str, int]], default=None)
```
nicoddemus, line 294: 👍

nicoddemus (approved):
LGTM, great work!

src/_pytest/_code/code.py
bluetech: Right, added.

src/_pytest/doctest.py
bluetech: It's actually still used, just well hidden with the `path`/`fspath` naming mixing up.

PR #8188, opened 2020-12-23
coverage: Include code that runs in subprocesses

antonblr:
Should address #7679

[coverage_enable_subprocess](https://pypi.org/project/coverage_enable_subprocess/) emits `$TOP` variable during the tests run, which can be used to ensure that all coverage data files  are in one place, for cases like when  tests need to `cd` and run many processes in parallel.

`+0.41%` of [coverage](https://codecov.io/gh/pytest-dev/pytest/tree/subprocess-coverage/src/_pytest)

Also, it might make sense to run `Report Coverage` as a separate job at the very end -> to report all or none of coverage to Codecov (e.g. if one of the master builds with coverage fail the coverage graph will have a dip)


.coveragerc
```diff
@@ -10,6 +10,10 @@ include =
   *\Lib\site-packages\pytest.py
 parallel = 1
 branch = 1
+# From https://pypi.org/project/coverage_enable_subprocess/
+# $TOP env var is emitted during the tests run
```
nicoddemus, line 14: Out of curiosity, who sets `TOP`? I don't see it in `coverage_enable_subprocess`' sources, is it defined by the system?

nicoddemus (approved):
The coverage increase is good, just wondering how the `$TOP` variable is defined and by who.

.coveragerc
antonblr: let me look into this - this one was left from my initial iteration, it could be that the absolute path to coverage file in CI is the culprit.

tox.ini
```diff
@@ -30,7 +30,7 @@ setenv =
     # "tox -e py37-coverage".
     coverage: _PYTEST_TOX_COVERAGE_RUN=coverage run -m
     coverage: _PYTEST_TOX_EXTRA_DEP=coverage-enable-subprocess
-    coverage: COVERAGE_FILE={toxinidir}/.coverage
```
asottile, line 33: I think this was correct already 🤔 -- iirc `coverage` picks up the `COVERAGE_FILE` environment variable

it's unclear to me why there's custom code in github actions for this though 🤔

antonblr:
@nicoddemus - I'm glad you were more curious. Pytest coverage already setup per https://coverage.readthedocs.io/en/stable/subprocess.html, except that CI setup didn't specify absolute paths to coverage config/data files (like in `tox.ini`). 

To be more explicit about that we need to provide a (absolute) path to `.coverage` file I've added `COVERAGE_DATA_PATH` with `data_file = ${COVERAGE_DATA_PATH-.}/.coverage` entry to `.coveragerc`

Ideally CI should run `tox -e ...-coverage` to avoid discrepancies/duplication, but that's would require some reorganization (e.g. we don't want upload coverage locally).

tox.ini
antonblr: Yes, the dev env was already setup correctly. This discrepancy, I guess, came out from the fact that we want to report coverage in a separate GHA step? Plus, we don't need upload coverage locally?

bluetech:
Thanks for looking at this @antonblr.

> Ideally CI should run tox -e ...-coverage to avoid discrepancies/duplication, but that's would require some reorganization (e.g. we don't want upload coverage locally).

I think it would be a nice extra in this PR to unify tox and CI if possible, seems like it's just duplicated in two places now.

If we change the "Test with coverage" in CI to run `"tox -e ${{ matrix.tox_env }}-coverage"` (I added the `-coverage` at the end) and remove its `env` block entirely, would it not be sufficient? From what I can see it would end up doing the same, and the upload step could remain separate, but I may be missing something.

antonblr:
>If we change the "Test with coverage" in CI to run "tox -e ${{ matrix.tox_env }}-coverage" (I added the -coverage at the end) and remove its env block entirely, would it not be sufficient? From what I can see it would end up doing the same, and the upload step could remain separate, but I may be missing something.

Yeah, sort of - need to update [report-coverage.sh](https://github.com/pytest-dev/pytest/blob/master/scripts/report-coverage.sh) to only perform `coverage xml` && Codecov upload.
 Plus, if we want `Report coverage` in a separate from the tests step for clarity (by "report" I mean `coverage combine && coverage report`, not the upload), we would need to modify tox envs to split these steps.

I'll take a look into into it.

antonblr:
@bluetech / @nicoddemus please have another look. 
* Using `tox -e ${{ matrix.tox_env }}-coverage` (caveat: coverage is reported in the same step)
* Removed `Prepare coverage token` - Codecov allow tokenless uploads for public repos now 
* `report-coverage.sh` -> `upload-coverage.sh` and:
  * do not run `coverage combine && coverage report`, it is already combined and reported in `tox -e ...-coverage`
  * upload only if `github.repository == 'pytest-dev/pytest'` (this includes PRs from forks)

BTW - do we need `-F GHA` still? From the history it look like this was necessary when moving from Travis.

bluetech (approved):
`+4 -51` is great, excellent work @antonblr.

PR #8195, opened 2020-12-27
Add missing space in '--version' help message

christophebedard:
Previously:

```
  -V, --version         display pytest version and information about plugins.When given twice, also
                        display information about plugins.
```

### DIFF
--- a/src/_pytest/helpconfig.py
+++ b/src/_pytest/helpconfig.py
@@ -51,7 +51,7 @@ def pytest_addoption(parser: Parser) -> None:
         action="count",
         default=0,
         dest="version",
-        help="display pytest version and information about plugins."
+        help="display pytest version and information about plugins. "
         "When given twice, also display information about plugins.",
     )
     group._addoption(
--- a/testing/test_helpconfig.py
+++ b/testing/test_helpconfig.py
@@ -28,6 +28,9 @@ def test_help(pytester: Pytester) -> None:
                                 For example: -m 'mark1 and not mark2'.
         reporting:
           --durations=N *
+          -V, --version         display pytest version and information about plugins.
+                                When given twice, also display information about
+                                plugins.
         *setup.cfg*
         *minversion*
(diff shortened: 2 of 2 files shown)


src/_pytest/helpconfig.py
```diff
@@ -51,7 +51,7 @@ def pytest_addoption(parser: Parser) -> None:
         action="count",
         default=0,
         dest="version",
-        help="display pytest version and information about plugins."
+        help="display pytest version and information about plugins. "
```
symonk, line 54: Hi @christophebedard Thanks a lot for taking the time to fix the spacing here, looks good to me.  If you want to add a pytester based test for the `-h` output here if you so desire.

symonk (approved):
Thanks!

src/_pytest/helpconfig.py
christophebedard: I've been trying to add a test for it, e.g.

```diff
diff --git a/testing/test_helpconfig.py b/testing/test_helpconfig.py
index 9a433b1b1..bd06411b2 100644
--- a/testing/test_helpconfig.py
+++ b/testing/test_helpconfig.py
@@ -32,6 +32,7 @@ def test_help(pytester: Pytester) -> None:
         *minversion*
         *to see*markers*pytest --markers*
         *to see*fixtures*pytest --fixtures*
+          -V, --version         display pytest version and information about plugins. When given twice, also display information about plugins.
     """
     )

```

but without writing something that's kinda overkill I'm having a hard time trying to deal with the fact that the help message can be split over multiple lines :laughing:

christophebedard: > or can I assume that the output is always the same (i.e. same maximum width) when testing?

seems like this should be good

asottile (approved):
test is maybe overkill but change looks good --thanks for the fix!

bluetech (approved):

PR #8209, opened 2020-12-30
pytest-dev#8204 migrate some tests to tmp_path fixture

sousajf1:
migrating some tests from tmpdir to tmp_path fixture.

### DIFF
--- a/testing/test_argcomplete.py
+++ b/testing/test_argcomplete.py
@@ -3,6 +3,7 @@ import sys
 from pathlib import Path
 
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
 
 # Test for _argcomplete but not specific for any application.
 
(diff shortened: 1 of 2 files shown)


testing/test_argcomplete.py
```diff
@@ -65,16 +66,16 @@ def __call__(self, prefix, **kwargs):
...
     @pytest.mark.skipif("sys.platform in ('win32', 'darwin')")
-    def test_compare_with_compgen(self, tmpdir):
+    def test_compare_with_compgen(self, tmp_path: Path) -> None:
         from _pytest._argcomplete import FastFilesCompleter
 
         ffc = FastFilesCompleter()
         fc = FilesCompleter()
 
-        with tmpdir.as_cwd():
+        with tmp_path.cwd() as cwd:
```
bluetech, line 75: The new code here doesn't do the same as the previous. The previous `as_cwd()` changes the current directory for the duration of the `with` block. I'm not sure what the `with tmp_path.cwd()` does but it doesn't change the directory.

I suggest adding a `monkeypatch` fixture to the test, and replacing the `with` with `monkeypatch.chdir(tmp_path)` (and removing one level of indentation from all of the following lines).

bluetech (changes requested):
That's a good start @sousajf1, just one thing to fix.

testing/test_argcomplete.py
sousajf1: I wasn't sure if the change dir was really needed, and since all the tests passed i thought it was just syntatic sugar for the creation of "data". I've changed it adding the monkeypatch as you suggested :)

bluetech (approved):
Thanks @sousajf1!

PR #8216, opened 2021-01-01
DOC: Update multiple references to testdir to pytester

xuhdev:
In https://docs.pytest.org/en/stable/reference.html#testdir, it is
suggested:

> New code should avoid using testdir in favor of pytester.

Multiple spots in the documents still use testdir and they can be quite
confusing (especially the plugin writing guide).


bluetech (approved):
Thanks for catching this.

PR #8218, opened 2021-01-01
Misc small code improvements

bluetech:
These are few small commits with misc improvements, adding type annotations and such, that I gathered while looking at other stuff. Please see the commits.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -16,6 +16,7 @@ from typing import Generic
 from typing import Iterable
 from typing import Iterator
 from typing import List
+from typing import MutableMapping
 from typing import Optional
 from typing import overload
 from typing import Sequence
@@ -525,9 +526,10 @@ class FixtureRequest:
         return self._pyfuncitem.fspath  # type: ignore
 
     @property
-    def keywords(self):
+    def keywords(self) -> MutableMapping[str, Any]:
         """Keywords/markers dictionary for the underlying node."""
-        return self.node.keywords
+        node: nodes.Node = self.node
+        return node.keywords
 
     @property
     def session(self) -> "Session":
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -607,14 +607,11 @@ class FixtureRequest:
     def _get_fixturestack(self) -> List["FixtureDef[Any]"]:
         current = self
         values: List[FixtureDef[Any]] = []
-        while 1:
-            fixturedef = getattr(current, "_fixturedef", None)
-            if fixturedef is None:
-                values.reverse()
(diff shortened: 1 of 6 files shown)


src/_pytest/fixtures.py
```diff
@@ -607,14 +609,11 @@ def _get_active_fixturedef(
...
-            fixturedef = getattr(current, "_fixturedef", None)
-            if fixturedef is None:
-                values.reverse()
-                return values
-            values.append(fixturedef)
-            assert isinstance(current, SubRequest)
+        while isinstance(current, SubRequest):
+            values.append(current._fixturedef)  # type: ignore[has-type]
             current = current._parent_request
+        values.reverse()
```
Zac-HD, line 615:
```suggestion
        assert current is None
        values.reverse()
```

Ensures that we don't get weird behaviour if a non-`SubRequest` ever makes it onto the fixture stack.

src/_pytest/fixtures.py
```diff
@@ -525,9 +526,10 @@ def fspath(self) -> py.path.local:
         return self._pyfuncitem.fspath  # type: ignore
 
     @property
-    def keywords(self):
+    def keywords(self) -> MutableMapping[str, Any]:
         """Keywords/markers dictionary for the underlying node."""
-        return self.node.keywords
+        node: nodes.Node = self.node
+        return node.keywords
```
Zac-HD, line 532: I think adding `def node(self) -> nodes.Node:` above would satisfy typecheckers, and I'd prefer it to modifying this function body.

Zac-HD (approved):
Two small comments, LGTM overall 😄

RonnyPfannschmidt (approved):

src/_pytest/fixtures.py
bluetech: I haven't fully figured out what the `SubRequest` stuff does exactly (quite a spaghetti code there) but your proposed assert doesn't work -- the `_parent_request` eventually reaches a non-`SubRequest` `FixtureRequest` (which `SubRequest` is a subclass of...), not `None`.

src/_pytest/fixtures.py
bluetech: Right, I was tempted to add it but didn't because

1. `nodes.Node` itself is not a type we intend to make public, but could use `Union[Item, Collector]` instead

2. Even `Union[Item, Collector]` would not be great - often the user does `request.node` in a known context and would assume the `node` is a specific node type (like `Function`) and then access it that way, but if we make the type `Union[Item, Collector]` it would suddenly generate a type error and require an assert. So for now I just keep it `Any`.

nicoddemus (approved):
Great!

PR #8219, opened 2021-01-01
runner: refactor SetupState

bluetech:
This PR refactors `SetupState` which handles the setting up and tearing down of nodes as they are executed. I believe it makes the code simpler and easier to understand, and also stricter. I also added a docstring trying to explain how it works a bit.

You can read the commits, if you want to see the progression and justification for each step, or just read the PR diff for the final result (though maybe use patience diff, the github diff somewhat botches it).

Because this makes things stricter, in particular the requirements that finalizers be added only once a node is present in the stack, this may introduce some regressions in plugins which do tricky stuff, or even user code (though in this case I don't think so). Of course I am ready to handle them on a case by case basis if they occur.

### DIFF
--- a/src/_pytest/runner.py
+++ b/src/_pytest/runner.py
@@ -438,9 +438,6 @@ class SetupState:
 
     def _pop_and_teardown(self) -> None:
         colitem = self.stack.pop()
-        self._teardown_with_finalization(colitem)
-
-    def _callfinalizers(self, colitem: Node) -> None:
         finalizers = self._finalizers.pop(colitem, None)
         exc = None
         while finalizers:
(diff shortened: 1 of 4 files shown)


src/_pytest/runner.py
```diff
@@ -396,14 +396,75 @@ def pytest_make_collect_report(collector: Collector) -> CollectReport:
...
 class SetupState:
-    """Shared state for setting up/tearing down test items or collectors."""
+    """Shared state for setting up/tearing down test items or collectors
+    in a session.
+
+    Suppose we have a collection tree as follows:
+
+    <Session session>
+        <Module mod1>
+            <Function test1>
```
nicoddemus, line 406: Probably better to change the name of `test1` and `test2` to `item1` and `item2`, to better reflect the rest of the text.

nicoddemus (approved):
Great work @bluetech!

Sorry for the delay on this, I missed the notification and only found this now. 👍

RonnyPfannschmidt (approved):
dito here, this one is lovely, thanks!

src/_pytest/runner.py
bluetech: Oops, fixed.

Thanks for the review!

PR #8226, opened 2021-01-05
[6.2.x] doc: Add note about training early bird discount

nicoddemus:


RonnyPfannschmidt (approved):

The-Compiler:
Thanks @RonnyPfannschmidt and @nicoddemus! :heart:

The-Compiler:
Hm, looks like this didn't make it to https://docs.pytest.org/en/stable/ - any idea why?

nicoddemus:
Seems `stable` refers to the last tag always:

image

```
λ git log -1 6.2.1
commit d3971c30f4d0f1890a372af3b98de41ee555dcb3 (tag: 6.2.1)
Author: pytest bot <pytestbot@gmail.com>
Date:   Tue Dec 15 13:06:34 2020 +0000

    Prepare release version 6.2.1

```

Which makes sense, as `6.2.x` has not been released yet... if you want to get this up ASAP I think it is OK to make a new release, we only have a single bug fix in `6.2.x`, but that's OK (we might have missed some backports though).

The-Compiler:
Meh... if we want to not have outdated information on the website, we'd then need another release on January 15th to remove the note, and another one some ~2 weeks later to remove the box entirely... I guess that's a bit overkill, so let's keep it as-is.

PR #8255, opened 2021-01-18
Explain how to create binary files in the doc of `pytest.makefile`.

### LINKED ISSUE #8215: Let pytest.makefile support writing bytes
`pytest.makefile` doesn't support writing bytes like `pytester.makefile(ext='.bin', my_file=b'binary content')`.

The following piece of code should produce `test_a.bin` with the binary content written to it:

```python
pytest_plugins = ('pytester',)


def test_a(pytester):
    "Test pytester.makefile."

    pytester.makefile(ext='.bin', my_file=b'binary content')
```

#### Alternative Solutions

Alternative would be take the return value of `makefile` and use `write_bytes`.

xuhdev:
Close #8215


xuhdev:
cc @asottile

nicoddemus:
Thanks @xuhdev!

Took the liberty of moving the example below in its own section, seem better that way. 👍

nicoddemus (approved):

PR #8260, opened 2021-01-20
Always handle faulthandler stderr even if already enabled

### LINKED ISSUE #8258: Faulthandler not enabled when pytest run under Python Dev Mode (-X dev)
# Description
I use [Python Dev Mode](https://docs.python.org/3/library/devmode.html#devmode) as a handy shortcut for enabling faulthandler, enabling debug mode of asyncio, logging resource leaks, and other goodies. Unfortunately, pytest's faulthandler plugin doesn't install itself if something else has already enabled the faulthandler. This means the output of the faulthandler is swallowed unless pytest is run with `-s`

# Minimal example
Here is a file with 2 tests, one which fails with an asyncio error when run under dev mode, and another which occasionally seg faults. I would like to see the asyncio error when it doesn't segfault, and the stack trace when it does.
```python
import ctypes
import asyncio
import threading
import time

def test_wrong_thread():
    loop = asyncio.get_event_loop()
    t = threading.Thread(target=loop.run_forever)
    t.setDaemon(True)
    t.start()
    loop.call_soon(print)

def test_segfault():
    if time.time() % 1 < 0.3:
        ctypes.string_at(0)
```
Running
```
$ python -X dev -m pytest
```
Gives the error from asyncio:
```
RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one
```
But when it segfaults the stack trace is swallowed unless run with `-s`
# Suggestion
Either unconditionally enabling faulthandler in the plugin (regardless of whether it was already enabled) or adding an option to do so would fix this, I am happy to provide a PR if this is a valid solution.
This naive solution fixes things for my example:
```diff
--- faulthandler.py	2021-01-20 09:42:57.611168438 +0000
+++ faulthandler_patch.py	2021-01-20 09:43:30.767501381 +0000
@@ -25,7 +25,7 @@
 def pytest_configure(config: Config) -> None:
     import faulthandler
 
-    if not faulthandler.is_enabled():
+    if True:
         # faulthhandler is not enabled, so install plugin that does the actual work
         # of enabling faulthandler before each test executes.
         config.pluginmanager.register(FaultHandlerHooks(), "faulthandler-hooks")
```

# Environment
```
$ pip list
Package       Version
------------- ----------
appdirs       1.4.3
attrs         20.3.0
CacheControl  0.12.6
certifi       2019.11.28
chardet       3.0.4
colorama      0.4.3
contextlib2   0.6.0
distlib       0.3.0
distro        1.4.0
html5lib      1.0.1
idna          2.8
iniconfig     1.1.1
ipaddr        2.2.0
lockfile      0.12.2
msgpack       0.6.2
packaging     20.8
pep517        0.8.2
pip           20.0.2
pkg-resources 0.0.0
pluggy        0.13.1
progress      1.5
py            1.10.0
pyparsing     2.4.7
pytest        6.2.1
pytoml        0.1.21
requests      2.22.0
retrying      1.3.3
setuptools    44.0.0
six           1.14.0
toml          0.10.2
urllib3       1.25.8
webencodings  0.5.1
wheel         0.34.2
```

nicoddemus:
It seems the code that would not install pytest's faulthandler support if it was already enabled is not really needed at all, and even detrimental when using `python -X dev -m pytest` to run Python in "dev" mode.

Also simplified the plugin by removing the hook class, now the hooks will always be active so there's no need to delay the hook definitions anymore.

Fix #8258

@thomascobb would you mind testing this branch? I've already tested with the example you posted in #8258, but it would be great if you could make sure this works with your real use case.

You can install this branch locally using:

```
pip install git+https://github.com/nicoddemus/pytest.git@faulthandler-mode-X-8258
```

### DIFF
--- a/src/_pytest/faulthandler.py
+++ b/src/_pytest/faulthandler.py
@@ -25,92 +25,72 @@ def pytest_addoption(parser: Parser) -> None:
 def pytest_configure(config: Config) -> None:
     import faulthandler
 
-    if not faulthandler.is_enabled():
-        # faulthhandler is not enabled, so install plugin that does the actual work
-        # of enabling faulthandler before each test executes.
-        config.pluginmanager.register(FaultHandlerHooks(), "faulthandler-hooks")
-    else:
-        # Do not handle dumping to stderr if faulthandler is already enabled, so warn
-        # users that the option is being ignored.
-        timeout = FaultHandlerHooks.get_timeout_config_value(config)
-        if timeout > 0:
-            config.issue_config_time_warning(
-                pytest.PytestConfigWarning(
-                    "faulthandler module enabled before pytest configuration step, "
-                    "'faulthandler_timeout' option ignored"
-                ),
-                stacklevel=2,
-            )
+    stderr_fd_copy = os.dup(get_stderr_fileno())
+    config._store[fault_handler_stderr_key] = open(stderr_fd_copy, "w")
+    faulthandler.enable(file=config._store[fault_handler_stderr_key])
 
 
-class FaultHandlerHooks:
-    """Implements hooks that will actually install fault handler before tests execute,
-    as well as correctly handle pdb and internal errors."""
+def pytest_unconfigure(config: Config) -> None:
+    import faulthandler
 
-    def pytest_configure(self, config: Config) -> None:
-        import faulthandler
-
-        stderr_fd_copy = os.dup(self._get_stderr_fileno())
-        config._store[fault_handler_stderr_key] = open(stderr_fd_copy, "w")
-        faulthandler.enable(file=config._store[fault_handler_stderr_key])
-
-    def pytest_unconfigure(self, config: Config) -> None:
-        import faulthandler
-
-        faulthandler.disable()
-        # close our dup file installed during pytest_configure
-        # re-enable the faulthandler, attaching it to the default sys.stderr
-        # so we can see crashes after pytest has finished, usually during
-        # garbage collection during interpreter shutdown
+    faulthandler.disable()
+    # Close the dup file installed during pytest_configure.
+    if fault_handler_stderr_key in config._store:
         config._store[fault_handler_stderr_key].close()
         del config._store[fault_handler_stderr_key]
-        faulthandler.enable(file=self._get_stderr_fileno())
+    # Re-enable the faulthandler, attaching it to the default sys.stderr
+    # so we can see crashes after pytest has finished, usually during
+    # garbage collection during interpreter shutdown.
+    faulthandler.enable(file=get_stderr_fileno())
 
-    @staticmethod
-    def _get_stderr_fileno():
+
+def get_stderr_fileno() -> int:
+    try:
+        fileno = sys.stderr.fileno()
+        # The Twisted Logger will return an invalid file descriptor since it is not backed
+        # by an FD. So, let's also forward this to the same code path as with pytest-xdist.
+        if fileno == -1:
+            raise AttributeError()
+        return fileno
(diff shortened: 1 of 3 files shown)


thomascobb:
I've tested and it works for me, thanks!

src/_pytest/faulthandler.py
```diff
@@ -25,92 +25,72 @@ def pytest_addoption(parser: Parser) -> None:
...
-        # re-enable the faulthandler, attaching it to the default sys.stderr
-        # so we can see crashes after pytest has finished, usually during
-        # garbage collection during interpreter shutdown
+    faulthandler.disable()
+    # Close the dup file installed during pytest_configure.
+    if fault_handler_stderr_key in config._store:
         config._store[fault_handler_stderr_key].close()
         del config._store[fault_handler_stderr_key]
-        faulthandler.enable(file=self._get_stderr_fileno())
+    # Re-enable the faulthandler, attaching it to the default sys.stderr
```
bluetech, line 41: So IIUC this will enable faulthandler for the host process after pytest is done even if it was not enabled before, right? I think maybe we only want to do this part if faulthandler was enabled before `pytest_configure`, just so we leave things as we got them?

bluetech (approved):
One comment, otherwise LGTM, and also simplifies things besides fixing the bug.

src/_pytest/faulthandler.py
nicoddemus: Definitely, good catch. 👍

nicoddemus: Hmmm on second thought not sure. 

The comment says: "so we can see crashes after pytest has finished", so this will not happen when executing `pytest` normally, where `faulthandler` was not enabled before.

I suggest we leave it like this, seems the more useful setting. What do you think?

bluetech: I agree it's useful, but IMO "after pytest has finished" is out of our domain. We should perhaps encourage users to run pytest with `-X dev`, maybe even do it automatically when the `pytest` entry point is used (I might open an issue to discuss this :thinking: ). But if someone is using `pytest.main()` I think we shouldn't touch it (that is, restore to previous value).

nicoddemus: > I agree it's useful, but IMO "after pytest has finished" is out of our domain.

I personally think practicability wins here, but no worries, I will apply the requested change.

nicoddemus: Hmm I realized `test_crash_near_exit` is meant to test the behavior of catching crashes during interpreter shutdown:

```python
    def test_crash_near_exit(pytester: Pytester) -> None:
        """Test that fault handler displays crashes that happen even after
        pytest is exiting (for example, when the interpreter is shutting down)."""
        pytester.makepyfile(
            """
        import faulthandler
        import atexit
        def test_ok():
            atexit.register(faulthandler._sigabrt)
        """
        )
        result = pytester.runpytest_subprocess()
        result.stderr.fnmatch_lines(["*Fatal Python error*"])
```

Applying the requested change makes the test fail (as expected). Really think this is the way to go @bluetech?

bluetech: Yes -- IMO like campers say, "leave it as you found it". Given that `pytest.main` is a public API, we shouldn't assume that we are the complete owners of the process after we finish. Ideally we would not even assume it *while* we are running but that is not really possible unfortunately (due to things like replacing `sys.stdout` and other global resources).

On the other hand when we are run from the cli then we *can* assume that so we can enable whatever we want then, but needs some discussion.

nicoddemus: Fair enough, done. 👍

PR #8274, opened 2021-01-25
Fix an example codeblock so that it clearly shows distinct instances

philsnow:
The description before this says "each test has a unique instance of the class", but the codeblock shows two instances both with memory address `0xdeadbeef`.


nicoddemus:
Thanks @philsnow!  

Unfortunately that bit of documentation is automatically generated by [`regendoc`](https://github.com/pytest-dev/regendoc), which replaces all memory addresses by the constant `0xdeadbeef`.

I'm closing this in favor of https://github.com/pytest-dev/regendoc/issues/12.

PR #8276, opened 2021-01-25
Mention that class variables are shared between tests

### LINKED ISSUE #8252: Docs: Add details on possible leakage when using test classes
Hi! I just ran into a similar problem to [this one](https://github.com/pytest-dev/pytest-mock/issues/84), where the result of tests depended on the order in which they were run. The root cause of this was that I defined attributes on the `TestClass`'s level and mocked methods of these attributes. Eventually I found out that this could be overcome by using the `setup_method` of the class instead of class attributes. When going over the docs for [grouping multiple tests in a class](https://docs.pytest.org/en/latest/getting-started.html#group-multiple-tests-in-a-class), I found no hint of this possible problem. Others might run into this problem as well and there already exists proper documentation of the solution [here](https://docs.pytest.org/en/stable/xunit_setup.html#class-level-setup-teardown).

May I suggest to replace the following sentences by a more accurate documentation? As there seem to be cases in which it is not 100% true, I find it possibly misleading (e.g. when combining pytest with mocking).

> Something to be aware of when grouping tests inside classes is that each test has a unique instance of the class. Having each test share the same class instance would be very detrimental to test isolation and would promote poor test practices.

christian-steinmeyer:
closes #8252

### DIFF
--- a/doc/en/getting-started.rst
+++ b/doc/en/getting-started.rst
@@ -210,7 +210,7 @@ This is outlined below:
     FAILED test_class_demo.py::TestClassDemoInstance::test_two - assert 0
     2 failed in 0.12s
 
-Note, that attributes added at class level (e.g. outside of the ``setup_method(self)``), will be shared between tests.
+Note that attributes added at class level are *class attributes*, so they will be shared between tests.
 
 Request a unique temporary directory for functional tests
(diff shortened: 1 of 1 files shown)


nicoddemus (approved):
Thanks!

PR #8296, opened 2021-01-29
Remove some no longer needed type-ignores

bluetech:
Due to the recent upgrade to mypy 0.800.


nicoddemus (approved):

The-Compiler:
What do you think about setting `warn_unused_ignores = True` for mypy in `setup.cfg`? It'd mean upgrades of mypy aren't automatic anymore, but we then wouldn't have ignores around for longer than needed.

bluetech:
Unfortunately it's not possible because some unused-ignores are python-version and platform dependent. I.e. you run on Linux Python 3.9 you don't get it, but on Windows 3.6 you do. It is possible to tell mypy to use a specific platform, but then it will stop checking these parts (for developers running on the relevant version/platform), so up to now we didn't do it.

PR #8294, opened 2021-01-29
Doc: Clarify pytester.run

xuhdev:
The document of the stdin parameter in `pytester.run` is confusing. Also
added some aesthetic updates along the way.

PR #8298, opened 2021-01-29
Type annotation polishing for symbols around Pytester.run

xuhdev:
Hopefully these will help document readers understand pertinent methods
and constants better.

Following up #8294

### DIFF
--- a/src/_pytest/pytester.py
+++ b/src/_pytest/pytester.py
@@ -42,6 +42,8 @@ from _pytest import timing
 from _pytest._code import Source
 from _pytest.capture import _get_multicapture
 from _pytest.compat import final
+from _pytest.compat import NOTSET
+from _pytest.compat import NotSetType
 from _pytest.config import _PluggyPlugin
 from _pytest.config import Config
 from _pytest.config import ExitCode
@@ -653,7 +655,7 @@ class Pytester:
 
     __test__ = False
 
-    CLOSE_STDIN: "Final[type]" = object
+    CLOSE_STDIN: "Final" = NOTSET
 
     class TimeoutExpired(Exception):
         pass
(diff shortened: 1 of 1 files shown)


xuhdev:
One thing I don't understand is why `CLOSE_STDIN = object`. Perhaps it was meant to be `object()`?

src/_pytest/pytester.py
```diff
@@ -1529,7 +1525,7 @@ class Testdir:
 
     __test__ = False
 
-    CLOSE_STDIN = Pytester.CLOSE_STDIN
+    CLOSE_STDIN: "Final[type]" = Pytester.CLOSE_STDIN
```
bluetech, line 1528: I *think* (haven't verified) that can be just the following and mypy will infer the reset:

```suggestion
    CLOSE_STDIN: "Final" = Pytester.CLOSE_STDIN
```

In which case, it'd be nice to add a `Final` for the two other constants below while you're at it.

src/_pytest/pytester.py
```diff
@@ -1695,8 +1691,8 @@ def collect_by_name(
     def popen(
         self,
         cmdargs,
-        stdout: Union[int, TextIO] = subprocess.PIPE,
-        stderr: Union[int, TextIO] = subprocess.PIPE,
+        stdout=subprocess.PIPE,
```
bluetech, line 1694: Was the above incorrect?

bluetech (commented):
Thanks @xuhdev, nice changes.

> One thing I don't understand is why CLOSE_STDIN = object. Perhaps it was meant to be object()?

Yes, just the `object` type is odd, probably was meant to be `object()`. But `object()` also is not very good for typing. For a previous case I did this:

https://github.com/pytest-dev/pytest/blob/275174e14795314f36b068ea757eaf032a3f8dc1/src/_pytest/compat.py#L34-L40

Maybe do the same for `CLOSE_STDIN`?

src/_pytest/pytester.py
xuhdev: It depends on what it means to be incorrect. I removed the annotation here because this is the only method in `Testdir` that has some parameters annotated. And it makes sense because the docstring pretty much just says "please read the Pytester counterpart," and by removing the annotations makes the generated document easier to read. Would you agree? I can revert this part of the change if it is intended so.

bluetech: Generally if the type annotation is correct, I'd keep it in place. But I'm OK with either way.

src/_pytest/pytester.py
xuhdev: Thanks, this makes sense.

bluetech (approved):
Thanks!

PR #8316, opened 2021-02-04
doc: Remove past training

The-Compiler:
For the curious, it ended up as quite a success with 6 people (though obviously remote-only) :smile:


Zac-HD (approved):
Glad it went well 😁

nicoddemus (approved):
Cool!

PR #8319, opened 2021-02-05
Type annotation for request.param

paxcodes:
- [n/a] Include documentation when adding new features.
- [n/a] Include new tests or update existing tests when applicable.

____

Context: Under Pylance's 'strict' mode I'm getting `reportUnknownMemberType`. See EricTraut's [comment](https://github.com/microsoft/pyright/discussions/1451).

I'm hoping this PR would fix it.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -758,7 +758,7 @@ class SubRequest(FixtureRequest):
         self,
         request: "FixtureRequest",
         scope: "_Scope",
-        param,
+        param: Optional[Any],
         param_index: int,
         fixturedef: "FixtureDef[object]",
         *,


src/_pytest/fixtures.py
```diff
@@ -758,7 +758,7 @@ def __init__(
         self,
         request: "FixtureRequest",
         scope: "_Scope",
-        param,
+        param: Optional[Any],
```
bluetech, line 761: I don't think `Optional` is correct here, there is no `None` here (unless it's the actual param value).

One other complication is that `param` can be `NOTSET` and then `self.param` is only set conditionally. There is currently no way to represent this in Python (happily). This is the reason it isn't annotated.

But overall I'm fine with `request.param` just being `Any` for now, until we can clean this up. So if you remove the `Optional` I'd be OK.

bluetech (changes requested):
Thanks @paxcodes, see comment below.

bluetech (approved):

PR #8320, opened 2021-02-05
Update plugin-list every Sunday instead of everyday

nicoddemus:
Every day seems a bit excessive lately, let's make it a bit less frequent.


bluetech (approved):
I would do it every week even...

nicoddemus:
> I would do it every week even...

Agreed, updated. 👍

PR #8322, opened 2021-02-05
doc: Remove confusing fixture sentence

The-Compiler:
There is no previous `test_ehlo` example (it follows much later) - and the same thing is described further down in ""Requesting" fixtures" already.


nicoddemus (approved):
Thanks!

PR #8354, opened 2021-02-19
Remove a redundant paragraph

feanil:
doc This paragraph looks like it is a more verbose version of the sentence right above it.  Removing it doesn't reduce the amount of information here but does make the section flow a little better.


bluetech (approved):
I agree.

cc @SalmonMode as the original author in case we're missing something.

SalmonMode:
Seems fine without it to me

bluetech:
OK, thanks @feanil!

PR #8419, opened 2021-03-09
clean up mkdtemp usage

asottile:
Committed via https://github.com/asottile/all-repos

(I felt that these were slightly better than the previous code, if not -- feel free to close 😄)


RonnyPfannschmidt (approved):
It's better than before, maybe tmpdir could /should be used and monkeypatch for the cwd

nicoddemus (approved):

PR #8429, opened 2021-03-10
Improve pytest.approx error messages readability (Pull request)

### LINKED ISSUE #8335: Improve pytest.approx error messages readability
Hey all. I think that as important as getting `PASSED` on our tests, it is important to get useful messages when the tests `FAIL`. That being said...

While comparing numerical data with `pytest.approx`, I find the messages a bit hard to read when we are comparing big enough numpy arrays. Example:

```python
def test_bla():
    import numpy as np
    a = np.linspace(0, 100, 20)
    b = np.linspace(0, 100, 20)
    a[10] += 0.5
    assert a == pytest.approx(b)
```

The error message:   
```
>       assert a == pytest.approx(b)
E       assert array([  0.        ,   5.26315789,  10.52631579,  15.78947368,\n        21.05263158,  26.31578947,  31.57894737,  36.84210526,\n        42.10526316,  47.36842105,  53.13157895,  57.89473684,\n        63.15789474,  68.42105263,  73.68421053,  78.94736842,\n        84.21052632,  89.47368421,  94.73684211, 100.        ]) == approx([0.0 ± 1.0e-12, 5.2631578947368425 ± 5.3e-06, 10.526315789473685 ± 1.1e-05, 15.789473684210527 ± 1.6e-05, 21.05263157894737 ± 2.1e-05, 26.315789473684212 ± 2.6e-05, 31.578947368421055 ± 3.2e-05, 36.8421052631579 ± 3.7e-05, 42.10526315789474 ± 4.2e-05, 47.36842105263158 ± 4.7e-05, 52.631578947368425 ± 5.3e-05, 57.89473684210527 ± 5.8e-05, 63.15789473684211 ± 6.3e-05, 68.42105263157896 ± 6.8e-05, 73.6842105263158 ± 7.4e-05, 78.94736842105263 ± 7.9e-05, 84.21052631578948 ± 8.4e-05, 89.47368421052633 ± 8.9e-05, 94.73684210526316 ± 9.5e-05, 100.0 ± 1.0e-04])
```

This is the main example I want to solve, although I've implemented a solution for other cases.

Would be nice to have something like this:
```
>       assert a == pytest.approx(b)
E       assert comparison failed for 1 values:
E         Index | Obtained           | Expected                    
E         (10,) | 53.131578947368425 | 52.631578947368425 ± 5.3e-05
```

#### ~Alternative~ Solutions
I haven't tried any workaround. But I implemented a solution, using `pytest_assertrepr_compare`. It is in this gist: https://gist.github.com/tarcisiofischer/cab56a5eb65aaa03bdbb4d206f1d49f0

Examples:
```
>       assert [1, 2, 3, 4] == pytest.approx([1, 3, 3, 5])
E       assert comparison failed for 2 values:
E         Index | Obtained | Expected   
E         1     | 2        | 3 ± 3.0e-06
E         3     | 4        | 5 ± 5.0e-06
```

```
>       assert 1.0 == pytest.approx(2.0)
E       assert comparison failed
E         Obtained: 1.0
E         Expected: 2.0 ± 2.0e-06
```

```
>       assert {'a': 1, 'b': 3} == pytest.approx({'a': 1.7, 'b': 3})
E       assert comparison failed for 1 values:
E         Index | Obtained | Expected     
E         a     | 1        | 1.7 ± 1.7e-06
```

A big one (sorry for the dump)
```
>       assert np.array([
            [
                [1.1987311, 12412342.3],
                [3.214143244, 1423412423415.677]
            ], [
                [1, 2],
                [3, 219371297321973]
            ]
        ]) == pytest.approx(
            np.array([
                [
                    [1.12313, 12412342.3],
                    [3.214143244, 534523542345.677]
                ], [
                    [1, 2],
                    [3, 7]
                ]
            ])
        )
E       assert comparison failed for 3 values:
E         Index     | Obtained          | Expected                  
E         (0, 0, 0) | 1.1987311         | 1.12313 ± 1.1e-06         
E         (0, 1, 1) | 1423412423415.677 | 534523542345.677 ± 5.3e+05
E         (1, 1, 1) | 219371297321973.0 | 7.0 ± 7.0e-06
```

#### Additional context
Just to mention, I didn't find any similar issue, so I'm starting a new one. The following issue has similar title, but it is completely different: https://github.com/pytest-dev/pytest/issues/6985

Please, let me know if you guys like my proposal, and feel free to suggest improvements - I'll be happy to hear ideas.
Also, feel free to disagree and close this proposal, so I will just move my code to my project's `conftest.py` ;)

tarcisiofischer:
closes #8335

### DIFF
--- a/changelog/8335.improvement.rst
+++ b/changelog/8335.improvement.rst
@@ -1,7 +1,7 @@
-Improved pytest.approx assertion messages
+Improved :func:`pytest.approx` assertion messages for sequences of numbers.
 
 The assertion messages now dumps a table with the index and the error of each diff.
-Example:
+Example::
 
     >       assert [1, 2, 3, 4] == pytest.approx([1, 3, 3, 5])
     E       assert comparison failed for 2 values:
--- a/changelog/8335.improvement.rst
+++ b/changelog/8335.improvement.rst
@@ -1,0 +1,10 @@
+Improved pytest.approx assertion messages
+
+The assertion messages now dumps a table with the index and the error of each diff.
+Example:
+
+    >       assert [1, 2, 3, 4] == pytest.approx([1, 3, 3, 5])
+    E       assert comparison failed for 2 values:
+    E         Index | Obtained | Expected
+    E         1     | 2        | 3 ± 3.0e-06
+    E         3     | 4        | 5 ± 5.0e-06
--- a/changelog/8335.improvement.rst
+++ b/changelog/8335.improvement.rst
@@ -6,5 +6,5 @@ Example::
     >       assert [1, 2, 3, 4] == pytest.approx([1, 3, 3, 5])
     E       assert comparison failed for 2 values:
     E         Index | Obtained | Expected
-    E         1     | 2        | 3 ± 3.0e-06
-    E         3     | 4        | 5 ± 5.0e-06
+    E         1     | 2        | 3 +- 3.0e-06
+    E         3     | 4        | 5 +- 5.0e-06
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1706,7 +1706,14 @@ class FunctionDefinition(Function):
 
 @hookimpl(tryfirst=True)
 def pytest_assertrepr_compare(config, op, left, right):
-    """pytest.approx message handler"""
+    """Improves pytest.approx assertion messages.
+
+    Note:
+    Since the 'assertion' plugin is always installed by default on pytest, and it is installed before this one,
+    it'll always get preferred. It's implementation has a code that handles assertion messages,
+    which changes the message for "high" verbosity level, regardless of the lhs and rhs types,
+    so the 'tryfirst' is necessary here so that we can handle the message even with verbosity level > 0.
+    """
     from _pytest.python_api import ApproxBase
 
     if op != "==":
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1702,3 +1702,24 @@ class FunctionDefinition(Function):
         raise RuntimeError("function definitions are not supposed to be run as tests")
 
     setup = runtest
+
+
+@hookimpl(tryfirst=True)
+def pytest_assertrepr_compare(config, op, left, right):
+    """
+    pytest.approx message handler
+    """
+    from _pytest.python_api import ApproxBase
+
+    if op != "==":
+        return None
+
+    if not isinstance(left, ApproxBase) and not isinstance(right, ApproxBase):
+        return None
+
+    # Although the common order should be obtained == expected, this ensures both ways
+    approx_side = left if isinstance(left, ApproxBase) else right
+    other_side = right if isinstance(left, ApproxBase) else left
+
+    verbosity_level = config.getoption("verbose")
+    return approx_side.repr_compare(other_side, verbosity_level)
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1702,29 +1702,3 @@ class FunctionDefinition(Function):
         raise RuntimeError("function definitions are not supposed to be run as tests")
 
     setup = runtest
-
-
-@hookimpl(tryfirst=True)
-def pytest_assertrepr_compare(config, op, left, right):
-    """Improves pytest.approx assertion messages.
-
-    Note:
-    Since the 'assertion' plugin is always installed by default on pytest, and it is installed before this one,
-    it'll always get preferred. Its implementation has code that handles assertion messages,
-    which changes the message for "high" verbosity level, regardless of the lhs and rhs types,
-    so the 'tryfirst' is necessary here so that we can handle the message even with verbosity level > 0.
-    """
-    from _pytest.python_api import ApproxBase
-
-    if op != "==":
-        return None
-
-    if not isinstance(left, ApproxBase) and not isinstance(right, ApproxBase):
-        return None
-
-    # Although the common order should be obtained == expected, this ensures both ways
-    approx_side = left if isinstance(left, ApproxBase) else right
-    other_side = right if isinstance(left, ApproxBase) else left
-
-    verbosity_level = config.getoption("verbose")
-    return approx_side.repr_compare(other_side, verbosity_level)
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1710,7 +1710,7 @@ def pytest_assertrepr_compare(config, op, left, right):
 
     Note:
     Since the 'assertion' plugin is always installed by default on pytest, and it is installed before this one,
-    it'll always get preferred. It's implementation has a code that handles assertion messages,
+    it'll always get preferred. Its implementation has code that handles assertion messages,
     which changes the message for "high" verbosity level, regardless of the lhs and rhs types,
     so the 'tryfirst' is necessary here so that we can handle the message even with verbosity level > 0.
     """
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1706,9 +1706,7 @@ class FunctionDefinition(Function):
 
 @hookimpl(tryfirst=True)
 def pytest_assertrepr_compare(config, op, left, right):
-    """
-    pytest.approx message handler
-    """
+    """pytest.approx message handler"""
     from _pytest.python_api import ApproxBase
 
     if op != "==":
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -162,7 +162,7 @@ class ApproxNumpy(ApproxBase):
 
         if np_array_shape != other_side.shape:
             return [
-                "Impossible to compare lists with different shapes.",
+                "Impossible to compare arrays with different shapes.",
                 f"Shapes: {np_array_shape} and {other_side.shape}",
             ]
 
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -1,7 +1,5 @@
 import math
 import pprint
-from collections.abc import Iterable
-from collections.abc import Mapping
 from collections.abc import Sized
 from decimal import Decimal
 from numbers import Complex
@@ -10,7 +8,9 @@ from typing import Any
 from typing import Callable
 from typing import cast
 from typing import Generic
+from typing import Iterable
 from typing import List
+from typing import Mapping
 from typing import Optional
 from typing import overload
 from typing import Pattern
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -11,6 +11,7 @@ from typing import Any
 from typing import Callable
 from typing import cast
 from typing import Generic
+from typing import List
 from typing import Optional
 from typing import overload
 from typing import Pattern
@@ -41,13 +42,13 @@ def _non_numeric_type_error(value, at: Optional[str]) -> TypeError:
 
 def _generate_approx_error_message(
     full_object: object,
-    message_data: typing.List[typing.List[str]],
+    message_data: List[List[str]],
     number_of_elements: int,
-    different_ids: typing.List[typing.Any],
+    different_ids: List[Any],
     max_abs_diff: float,
     max_rel_diff: float,
     verbosity_level: int,
-) -> typing.List[str]:
+) -> List[str]:
     message_data.insert(0, ["Index", "Obtained", "Expected"])
     max_sizes = [0, 0, 0]
     for index, obtained, expected in message_data:
@@ -91,7 +92,7 @@ class ApproxBase:
     def __repr__(self) -> str:
         raise NotImplementedError
 
-    def repr_compare(self, other_side, verbosity_level) -> typing.List[str]:
+    def repr_compare(self, other_side: "ApproxBase", verbosity_level: int) -> List[str]:
         return [
             "comparison failed",
             f"Obtained: {other_side}",
@@ -143,7 +144,9 @@ class ApproxNumpy(ApproxBase):
         list_scalars = _recursive_list_map(self._approx_scalar, self.expected.tolist())
         return f"approx({list_scalars!r})"
 
-    def repr_compare(self, other_side, verbosity_level) -> typing.List[str]:
+    def repr_compare(
+        self, other_side: "ApproxNumpy", verbosity_level: int
+    ) -> List[str]:
         import itertools
         import math
 
@@ -238,7 +241,9 @@ class ApproxMapping(ApproxBase):
             {k: self._approx_scalar(v) for k, v in self.expected.items()}
         )
 
-    def repr_compare(self, other_side, verbosity_level) -> typing.List[str]:
+    def repr_compare(
+        self, other_side: "ApproxMapping", verbosity_level: int
+    ) -> List[str]:
         import math
 
         approx_side_as_map = {
@@ -309,7 +314,9 @@ class ApproxSequencelike(ApproxBase):
             seq_type(self._approx_scalar(x) for x in self.expected)
         )
 
-    def repr_compare(self, other_side, verbosity_level) -> typing.List[str]:
+    def repr_compare(
+        self, other_side: "ApproxSequencelike", verbosity_level: int
+    ) -> List[str]:
         import math
         import numpy as np
 
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -55,7 +55,7 @@ def _generate_approx_error_message(
         max_sizes[1] = max(max_sizes[1], len(obtained))
         max_sizes[2] = max(max_sizes[2], len(expected))
     err_msg_as_list = [
-        f"comparison failed. Mismatched elements: {len(different_ids)} / {number_of_elements}):",
+        f"comparison failed. Mismatched elements: {len(different_ids)} / {number_of_elements}:",
         f"Max absolute difference: {max_abs_diff}",
         f"Max relative difference: {max_rel_diff}",
     ] + [
@@ -958,20 +958,3 @@ class RaisesContext(Generic[_E]):
         if self.match_expr is not None:
             self.excinfo.match(self.match_expr)
         return True
-
-
-def pytest_assertrepr_compare(config, op, left, right):
-    from _pytest.python_api import ApproxBase
-
-    if op != "==":
-        return None
-
-    if not isinstance(left, ApproxBase) and not isinstance(right, ApproxBase):
-        return None
-
-    # Although the common order should be obtained == expected, this ensures both ways
-    approx_side = left if isinstance(left, ApproxBase) else right
-    other_side = right if isinstance(left, ApproxBase) else left
-
-    verbosity_level = config.getoption("verbose")
-    return approx_side.repr_compare(other_side, verbosity_level)
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -144,9 +144,15 @@ class ApproxNumpy(ApproxBase):
         import itertools
         import math
 
-        def _get_index_from_list(list, index):
-            value = list
-            for i in index:
+        def _get_value_from_nested_list(
+            nested_list: List[Any], nd_index: Tuple[Any, ...]
+        ) -> Any:
+            """
+            Helper function to get the value out of a nested list, given an n-dimensional index.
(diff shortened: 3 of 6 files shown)


src/_pytest/python.py
```diff
@@ -1702,3 +1702,22 @@ def runtest(self) -> None:
         raise RuntimeError("function definitions are not supposed to be run as tests")
 
     setup = runtest
+
+
+@hookimpl(tryfirst=True)
```
tarcisiofischer, line 1707: I don't think it's obvious why this is necessary, so I'll try to make a quick explanation:

Since the `assertion` plugin is always installed by default, and it is installed before this one, he'll always get preferred. It's implementation has [this code](https://github.com/pytest-dev/pytest/blob/bfbf733a3ed2f58a71570aa6e950444f3c5b0ea1/src/_pytest/assertion/util.py#L192-L193), which changes the message for "high" verbosity level, regardless of the lhs and rhs types, so I need this `tryfirst` here so that I can handle the message even with verbosity level > 0...

testing/python/approx.py
```diff
@@ -43,7 +44,220 @@ def report_failure(self, out, test, example, got):
...
+                r"assert comparison failed. Mismatched elements: 20 / 20:",
+                rf"  Max absolute difference: {SOME_FLOAT}",
+                rf"  Max relative difference: {SOME_FLOAT}",
+                r"  Index \| Obtained\s+\| Expected",
+                rf"  \(0,\)\s+\| {SOME_FLOAT} \| {SOME_FLOAT} ± {SOME_FLOAT}",
+                rf"  \(1,\)\s+\| {SOME_FLOAT} \| {SOME_FLOAT} ± {SOME_FLOAT}",
+                rf"  \(2,\)\s+\| {SOME_FLOAT} \| {SOME_FLOAT} ± {SOME_FLOAT}",
+                rf"  \(3,\)\s+\| {SOME_FLOAT} \| {SOME_FLOAT} ± {SOME_FLOAT}...",
+                "",
+                r"\s*...Full output truncated \(17 lines hidden\), use '-vv' to show",
```
tarcisiofischer, line 236: This "Full output truncated" message is not mine. It's sad that I don't seem to have full control over the verbosity level due to [this code](https://github.com/pytest-dev/pytest/blob/bfbf733a3ed2f58a71570aa6e950444f3c5b0ea1/src/_pytest/assertion/__init__.py#L146)

Don't know how to proceed here.

tarcisiofischer:
I believe my tests are not running, since they need `numpy` to run.
I believe they are being skip on CI. I don't know how to run with `numpy` installed on CI.

nicoddemus:
Thanks for the PR! 

> I believe my tests are not running, since they need numpy to run.

Why do you comment that, is it because you didn't see they running, or are you assuming they are not running?

The `windows-py37` and `ubuntu-py37` execute the `py37-numpy` tox env, so they should be running the test suite with numpy installed: https://github.com/pytest-dev/pytest/blob/main/.github/workflows/main.yml#L55

src/_pytest/python.py
nicoddemus: Thanks for the explanation (this would be a good addition to the docstring of the function btw).

This makes me wonder if we should just bite the bullet and add this logic to `assertion.pytest_assertrepr_compare`.

@RonnyPfannschmidt @bluetech opinions?

src/_pytest/python_api.py
```diff
@@ -38,6 +39,36 @@ def _non_numeric_type_error(value, at: Optional[str]) -> TypeError:
     )
 
 
+def _generate_approx_error_message(
+    full_object: object,
+    message_data: typing.List[typing.List[str]],
```
nicoddemus, line 44: Please import directly from `typing` instead of importing it:

```suggestion
    message_data: List[List[str]],
```

Same for all the other type annotations.

src/_pytest/python_api.py
```diff
@@ -60,6 +91,13 @@ def __init__(self, expected, rel=None, abs=None, nan_ok: bool = False) -> None:
     def __repr__(self) -> str:
         raise NotImplementedError
 
+    def repr_compare(self, other_side, verbosity_level) -> typing.List[str]:
```
nicoddemus, line 94: Is it possible to type annotate this method fully?

changelog/8335.improvement.rst
```diff
@@ -0,0 +1,10 @@
+Improved pytest.approx assertion messages
```
nicoddemus, line 1:
```suggestion
Improved :func:`pytest.approx` assertion messages for sequences of numbers.
```

changelog/8335.improvement.rst
```diff
@@ -0,0 +1,10 @@
...
+
+The assertion messages now dumps a table with the index and the error of each diff.
+Example:
```
nicoddemus, line 4:
```suggestion

Example::
```

changelog/8335.improvement.rst
```diff
@@ -0,0 +1,10 @@
...
+
+    >       assert [1, 2, 3, 4] == pytest.approx([1, 3, 3, 5])
+    E       assert comparison failed for 2 values:
+    E         Index | Obtained | Expected
+    E         1     | 2        | 3 ± 3.0e-06
```
nicoddemus, line 9: Unfortunately the `±` breaks for me on Windows. I suggest we replace it with `+-` otherwise the docs won't build locally on Windows, although this doesn't seem to be a problem on Linux.

```suggestion
    E         1     | 2        | 3 +- 3.0e-06
```

changelog/8335.improvement.rst
```diff
@@ -0,0 +1,10 @@
...
+    E         3     | 4        | 5 ± 5.0e-06
```
nicoddemus, line 10:
```suggestion
    E         3     | 4        | 5 +- 5.0e-06
```

testing/python/approx.py
nicoddemus: I think this is fine... pytest intentionally truncates the output generated by the hooks, so each hook doesn't have to explicitly deal with that.

I propose to leave it like this, unless you want to propose changing that behavior; if so please open another issue so it can be discussed. 👍

testing/python/approx.py
```diff
@@ -43,7 +44,220 @@ def report_failure(self, out, test, example, got):
     return MyDocTestRunner()
 
 
+@contextmanager
+def temporary_verbosity(config, verbosity=0):
+    original_verbosity = config.getoption("verbose")
+    config.option.verbose = verbosity
+    yield
```
nicoddemus, line 51: For extra safety I suggest using `try/finally` around this `yield`.

nicoddemus (changes requested):
Thanks a lot for submitting this PR @tarcisiofischer. 👍

tarcisiofischer:
> Thanks for the PR!
> 
> > I believe my tests are not running, since they need numpy to run.
> 
> Why do you comment that, is it because you didn't see they running, or are you assuming they are not running?
> 
> The `windows-py37` and `ubuntu-py37` execute the `py37-numpy` tox env, so they should be running the test suite with numpy installed: https://github.com/pytest-dev/pytest/blob/main/.github/workflows/main.yml#L55

I didn't saw they running. But I guess you are right. Thanks!

changelog/8335.improvement.rst
tarcisiofischer: I find it weird, because this symbol was already being used on the previous assertion message format.
Should I change anyway? It'll "break" the convention with the previous implemented message, I believe ([See this](https://github.com/pytest-dev/pytest/blob/main/src/_pytest/python_api.py#L235))

testing/python/approx.py
tarcisiofischer: I'll just leave it like this. Doesn't seem a big deal. I usually use `-vv` when I want to see the full output anyway

src/_pytest/python_api.py
tarcisiofischer: Yes. Will do.

changelog/8335.improvement.rst
tarcisiofischer: Ah, you meant only for the docs. Sorry. Will do.

src/_pytest/python.py
tarcisiofischer: I added the docs. Will keep the conversation, thought, just to see if people want to discuss further.

changelog/8335.improvement.rst
nicoddemus: Exactly, sorry I should have been more explicit. 😁

src/_pytest/python_api.py
```diff
@@ -145,6 +240,47 @@ def __repr__(self) -> str:
             {k: self._approx_scalar(v) for k, v in self.expected.items()}
         )
 
+    def repr_compare(
+        self, other_side: Dict[Any, float], verbosity_level: int
```
bluetech, line 244:
```suggestion
        self, other_side: Mapping[object, float], verbosity_level: int
```

src/_pytest/python_api.py
```diff
@@ -177,6 +313,51 @@ def __repr__(self) -> str:
...
 
+    def repr_compare(
+        self, other_side: Sequence[float], verbosity_level: int
+    ) -> List[str]:
+        import math
+        import numpy as np
+
+        if len(self.expected) != len(other_side):
+            return [
+                "Impossible to compare lists with different sizes.",
```
bluetech, line 324: It would be good to utilize the length-comparison part of `_compare_eq_sequence` here. I guess it would need to be split from there.

(You don't have to do this, maybe add a TODO)

src/_pytest/python_api.py
```diff
@@ -60,6 +93,13 @@ def __init__(self, expected, rel=None, abs=None, nan_ok: bool = False) -> None:
     def __repr__(self) -> str:
         raise NotImplementedError
 
+    def repr_compare(self, other_side: Any, verbosity_level: int) -> List[str]:
```
bluetech, line 96:
```suggestion
    def repr_compare(self, other_side: object, verbosity_level: int) -> List[str]:
```

src/_pytest/python.py
bluetech: > This makes me wonder if we should just bite the bullet and add this logic to assertion.pytest_assertrepr_compare.

Given that `approx` is built into pytest, and is not in a plugin, I think it does make sense. Currently the hook is defined in the `python` plugin which is not really related.

However, if it's too much work it's OK to leave as-is for now, maybe add a TODO.

testing/python/approx.py
```diff
@@ -43,7 +44,256 @@ def report_failure(self, out, test, example, got):
...
+
 class TestApprox:
+    def test_error_messages(self, assert_approx_raises_regex):
+        np = pytest.importorskip("numpy")
+
+        assert_approx_raises_regex(
+            2.0,
+            1.0,
+            [
+                "assert comparison failed",
```
bluetech, line 102: I believe the the purpose of the "assert" here is to show the fully resolved `<left> <op> <right>`, e.g.

```
x = 1
y = 2
assert x == y
```

shows

```
x.py:4: in test_x
    assert x == y
E   assert 1 == 2
```

So I think "assert comparison failed" is not what it's *supposed* to show.

bluetech (commented):
Thanks @tarcisiofischer, this is a nice improvement!

I left some comments. In general I think that if possible, it would be good to integrate with the existing assert repr infrastructure, because implementing `pytest_assertrepr_compare` requires a wholesale replacement which can lead to inconsistencies and duplication and such. But if it's too much work that's OK.

src/_pytest/python_api.py
```diff
@@ -38,6 +41,36 @@ def _non_numeric_type_error(value, at: Optional[str]) -> TypeError:
...
+        max_sizes[1] = max(max_sizes[1], len(obtained))
+        max_sizes[2] = max(max_sizes[2], len(expected))
+    err_msg_as_list = [
+        f"comparison failed. Mismatched elements: {len(different_ids)} / {number_of_elements}:",
+        f"Max absolute difference: {max_abs_diff}",
+        f"Max relative difference: {max_rel_diff}",
+    ] + [
+        f"{indexes:<{max_sizes[0]}} | {obtained:<{max_sizes[1]}} | {expected:<{max_sizes[2]}}"
+        for indexes, obtained, expected in (
+            message_data[:10] if verbosity_level <= 0 else message_data
```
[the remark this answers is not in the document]
tarcisiofischer: Yeah, you got a point. Actually, I implemented this before I knew about the automatic-truncation from the pytest code. I guess I shouldn't handle the verbosity level at all, and let the pytest do it's work.

Will remove all the verbosity level related code.

testing/python/approx.py
tarcisiofischer: Hmm, I don't follow you on this one, sorry. You mean to change the test or the assertion message? :thinking:

tarcisiofischer:
> Thanks @tarcisiofischer, this is a nice improvement!
> 
> I left some comments. In general I think that if possible, it would be good to integrate with the existing assert repr infrastructure, because implementing `pytest_assertrepr_compare` requires a wholesale replacement which can lead to inconsistencies and duplication and such. But if it's too much work that's OK.

(Offtopic: I hate that GitHub don't let me "reply" directly on the message, so I have to "Quote" a whole new message just to reply :face_with_thermometer: )

Note that I can't use the `repr` infrastructure directly, because, there, I don't have information on both lhs on rhs, so I can't show the pretty messages due to the lack of information. [See this comment](https://github.com/pytest-dev/pytest/issues/8335#issuecomment-777064362) and [this reply](https://github.com/pytest-dev/pytest/issues/8335#issuecomment-777493168). Feel free to prove me wrong, though. I don't really like having to deal with those `pytest_assertrepr_compare` stuff, tbh :sweat_smile:

Perhaps it'll take me a while longer, but I want to give it a shot and try to move my code to inside `assertion.pytest_assertrepr_compare`, then. What you think?

bluetech:
> Perhaps it'll take me a while longer, but I want to give it a shot and try to move my code to inside assertion.pytest_assertrepr_compare, then. What you think?

Thanks, I know a lot of comments, some of which conflicting, can be a bit frustrating.

> Note that I can't use the repr infrastructure directly, because, there, I don't have information on both lhs on rhs, so I can't show the pretty messages due to the lack of information.

I was thinking more to hook up to the `assertrepr_compare` default implementation directly (found in `src/_pytest/assertion/utils.py`). First it formats the `assert ...` line with the `repr`s (this is what I referred to in the "assert comparison failed" comment).  This part I think can stay as-is. Then it delegates further explanation to `_compare_eq_any`. So what you want to do is modify `_compare_eq_any` to handle the approx types specially, there you can format any way you want, and you have both sides. You can keep the bulk of the logic in the `Approx*` classes as in your current PR if you prefer, but it will be dispatched from a `_compare_eq_approx` function or such. I hope it makes sense.

bluetech:
> I was thinking more to hook up to the assertrepr_compare default implementation directly

This is some confusing choice of words. No hooks involved here, I mean "modify directly".

testing/python/approx.py
tarcisiofischer: (got it now)

tarcisiofischer:
> > Perhaps it'll take me a while longer, but I want to give it a shot and try to move my code to inside assertion.pytest_assertrepr_compare, then. What you think?
> 
> Thanks, I know a lot of comments, some of which conflicting, can be a bit frustrating.
> 
> > Note that I can't use the repr infrastructure directly, because, there, I don't have information on both lhs on rhs, so I can't show the pretty messages due to the lack of information.
> 
> I was thinking more to hook up to the `assertrepr_compare` default implementation directly (found in `src/_pytest/assertion/utils.py`). First it formats the `assert ...` line with the `repr`s (this is what I referred to in the "assert comparison failed" comment). This part I think can stay as-is. Then it delegates further explanation to `_compare_eq_any`. So what you want to do is modify `_compare_eq_any` to handle the approx types specially, there you can format any way you want, and you have both sides. You can keep the bulk of the logic in the `Approx*` classes as in your current PR if you prefer, but it will be dispatched from a `_compare_eq_approx` function or such. I hope it makes sense.

Hey! Sorry for the delay... Been a little busy and a bit stressed :P Anyway... I think I made the changes you suggested, please, feel free to take a look and point out anything wrong. I cannot run on CI (although it passes on my machine hehe).

Thanks in advance! :)

tarcisiofischer:
@bluetech @nicoddemus Need some help here... Don't know why `py310` is failing some tests :thinking:
Doesn't seem related to my changes, and I'm unsure if I need to do something. If you guys have any clue, I'd appreciate...

bluetech:
> Need some help here... Don't know why py310 is failing some tests thinking

This is not your fault - currently broken in `main`. It doesn't block the merging :)

tarcisiofischer:
> > Need some help here... Don't know why py310 is failing some tests thinking
> 
> This is not your fault - currently broken in `main`. It doesn't block the merging :)

Ready to review/merge, then :)

src/_pytest/python_api.py
```diff
@@ -60,6 +88,13 @@ def __init__(self, expected, rel=None, abs=None, nan_ok: bool = False) -> None:
     def __repr__(self) -> str:
         raise NotImplementedError
 
+    def repr_compare(self, other_side: Any) -> List[str]:
```
bluetech, line 91: Let's make these private -- `ApproxBase` instances are public but we don't want `repr_compare` to be construed as public.

```suggestion
    def _repr_compare(self, other_side: Any) -> List[str]:
```

src/_pytest/python_api.py
```diff
@@ -105,6 +140,60 @@ def __repr__(self) -> str:
         list_scalars = _recursive_list_map(self._approx_scalar, self.expected.tolist())
         return f"approx({list_scalars!r})"
 
+    def repr_compare(self, other_side: "ndarray") -> List[str]:
+        import itertools
+        import math
+
+        def _get_index_from_list(list, index):
```
bluetech, line 147: Can you add type annotations this?

bluetech (approved):
Thanks @tarcisiofischer, this looks pretty nice to me.

That said, I myself have never used `pytest.approx` so I don't feel comfortable giving a final approval - I'll defer to @nicoddemus for that :)

src/_pytest/python_api.py
tarcisiofischer: But then, `util.py`'s `_compare_eq_any` would call this method, which is private :(

asottile:
pre-commit.ci run

(sorry, trying to debug https://github.com/pre-commit-ci/issues/issues/62)

src/_pytest/python_api.py
nicoddemus: That's not a problem; pytest's calling its own internal code is fine, what we don't want is for external users calling (and depending on) this internal API. 👍

src/_pytest/python_api.py
```diff
@@ -105,6 +140,66 @@ def __repr__(self) -> str:
         list_scalars = _recursive_list_map(self._approx_scalar, self.expected.tolist())
         return f"approx({list_scalars!r})"
 
+    def repr_compare(self, other_side: "ndarray") -> List[str]:
+        import itertools
+        import math
+
+        def _get_value_from_nested_list(
```
nicoddemus, line 147:
```suggestion
        def get_value_from_nested_list(
```

It is a local method, no need for the `_`. 👍

src/_pytest/python_api.py
```diff
@@ -145,6 +240,44 @@ def __repr__(self) -> str:
             {k: self._approx_scalar(v) for k, v in self.expected.items()}
         )
 
+    def repr_compare(self, other_side: Mapping[object, float]) -> List[str]:
```
nicoddemus, line 243: Same here to make this private. 👍

nicoddemus (changes requested):
Hey Pescador! 😁 

This looks great, let's just do the two minor changes left and we can merge. Let me know if you are short on time and I can apply them myself.

Thanks again for working on this, we definitely appreciate this!

src/_pytest/python_api.py
tarcisiofischer: sgtm, then

nicoddemus (approved):

nicoddemus:
Thanks again @tarcisiofischer!

PR #8433, opened 2021-03-11
Added how-to and reference directories.

evildmp:
Moved various documents into subdirectories, how-to and
reference.

Updated multiple links to use `:ref:` instead of `:doc:`,
meaning that files can henceforth be moved around without
breaking references.


nicoddemus (approved):
Looking good!

PR #8434, opened 2021-03-12
Split fixtures documentation into separate documents.

evildmp:
Sections have been moved to:

* reference/fixtures.rst
* how-to/fixtures.rst
* fixtures.rst

according to their function. Further refinement and rewriting will be required.

Some material has been moved to a new "Anatomy of a test" document, in
anticipation that material from other sections will also find a natural
home there later.

Removed several unneeded reference targets from fixtures documentation.


bluetech (approved):
Looks good to me, comments (not specific to this PR, and not meant to block this PR):

1. In the contents page, the "Reference guides" section appears before the "Explanation" section. I think maybe it wants to be the other way around? But OK if it's intentional.

2. Maybe "Explanation" should be "Explanations"?

3. The "pytest fixtures: explicit, modular, scalable" title seems out of place now, because it is not followed by any text. If we want to keep it for sentimental reasons, should rename "About fixtures" to that. If not, we should change it to something else.

evildmp:
> 1. In the contents page, the "Reference guides" section appears before the "Explanation" section. I think maybe it wants to be the other way around? But OK if it's intentional.

It's intended, but you're right that there's a question about. In the scheme https://documentation.divio.com it makes sense to think of the four components in the order _tutorial_ > _how_-to > _reference_ > _explanation_. That's not because they follow each other in a time sequence (only _tutorial_ has any sense in time ordering) but because having started with _tutorial_, there are logical connections or overlaps in nature between each one and the next. 

It doesn't matter too much, in other words. I like to keep explanation last though.

> 2. Maybe "Explanation" should be "Explanations"?

I think _Explanation_ is more correct.

> 3. The "pytest fixtures: explicit, modular, scalable" title seems out of place now, because it is not followed by any text. If we want to keep it for sentimental reasons, should rename "About fixtures" to that. If not, we should change it to something else.

Agreed; if it wants to go somewhere, it should be in the explanation part. But as a title, probably it should just go. We wouldn't keep code for sentimental reasons.

bluetech:
Got it, thanks!

nicoddemus (approved):

PR #8441, opened 2021-03-13
Moved more sections from reference to how-to.

evildmp:
* re-ordered and grouped how-to guides in how-to landing page
* retitled moved documents appropriately
* separated writing plugins/writing hook functions into two documents


doc/en/explanation/fixtures.rst
```diff
@@ -6,6 +6,7 @@ About fixtures
 .. seealso:: :ref:`how-to-fixtures`
 .. seealso:: :ref:`Fixtures reference <reference-fixtures>`
 
+pytest fixtures designed to be explicit, modular and scalable.
```
bluetech, line 9:
```suggestion
pytest fixtures are designed to be explicit, modular and scalable.
```

doc/en/how-to/cache.rst
```diff
@@ -2,8 +2,8 @@
 .. _cache:
 
 
-Cache: working with cross-testrun state
-=======================================
+How to re-run failed tests with the cache plugin
```
bluetech, line 5: The `--last-failed` functionality is certainly what the cache plugin is most used for, but as the previous title says, it is a more general system for keeping state across test runs/invocations. I don't know how long the titles can be, but it would be ~nice if that aspect can be preserved.

doc/en/how-to/index.rst
```diff
@@ -3,19 +3,58 @@
...
    mark
-   monkeypatch
+   parametrize
    tmpdir
-   capture
+   monkeypatch
+   doctest
+   cache
+
+pytest output and outcomes
```
bluetech, line 22: Maybe instead of "pytest", just "Tests"?

doc/en/how-to/index.rst
```diff
@@ -3,19 +3,58 @@
...
+   writing_hook_functions
+
+pytest and other test systems
+-----------------------------
+
+.. toctree::
+   :maxdepth: 1
+
+   existingtestsuite
    nose
```
bluetech, line 50: If the order is meaningful, I would move nose below unittest, as unittest is much more common.

bluetech (approved):
Thanks, LGTM, see a few minor comments.

nicoddemus:
Doc building has failed: 

```
checking consistency... /home/runner/work/pytest/pytest/doc/en/plugin_list.rst: WARNING: document isn't included in any toctree
```

Can you take a look at that @evildmp ?

evildmp:
@nicoddemus I don't understand that error @nicoddemus ... there shouldn't be any file at doc/en/plugin_list.rst. There isn't in https://github.com/evildmp/pytest/tree/evolutionary-documentation-restructure/doc/en for example, and it doesn't exist there locally.

nicoddemus:
That file is generated automatically by this cron job: https://github.com/pytest-dev/pytest/blob/main/.github/workflows/update-plugin-list.yml

bluetech:
In particular the path this auto-generated file is written to is determined here: https://github.com/pytest-dev/pytest/blob/cdbeb03aefde2f5ad50492a43ae54c4dce375070/scripts/update-plugin-list.py#L77

So if the file is moved during the restructuring, the script needs to be adjusted with the new path as well. Then it should be fine.

evildmp:
@bluetech I thought I had fixed this here: https://github.com/pytest-dev/pytest/pull/8441/files#diff-92e37f9ac8790b36b5e61bc942ca13939f4dacc25bca5f22719d8d0542cf9546R77 but it's still reporting the same error: https://github.com/pytest-dev/pytest/pull/8441/checks?check_run_id=2108393443#step:5:236.

I get the same error locally by the way. I can see how it might be possible that I have managed to get something cached in my local environment that is causing this, but harder to understand how it could happen here. Any ideas?

bluetech:
That looks good, I think that you only need to remove the file `doc/en/plugin_list.rst`. In fact to save some time I pushed a commit to this PR which does this, will merge once it's green.

PR #8453, opened 2021-03-16
Fix value in error message about negative relative tolerance

encukou:
Hello!
A static code analyzer informed us that the wrong variable is used for the "relative tolerance can't be negative" error message in `pytest.approx`.
I don't know how to best write a test for this, since error messages usually aren't covered by tests.

A reproducer:
```
from pytest import approx
0 == approx(1, rel=-3)
```
gives:
```
ValueError: relative tolerance can't be negative: 1e-12
```
but should give:
```
ValueError: relative tolerance can't be negative: -3
```

(There's also no documentation; I don't think this is docs-worthy.)

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -317,7 +317,7 @@ class ApproxScalar(ApproxBase):
 
         if relative_tolerance < 0:
             raise ValueError(
-                f"relative tolerance can't be negative: {absolute_tolerance}"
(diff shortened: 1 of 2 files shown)


kalekundert:
Thanks for finding and fixing this!  I think it'd probably be good to add a test, something along the lines of:
```
with pytest.raises(ValueError, match='-3'):
    0 == approx(1, rel=-3)
```

kalekundert (approved):

PR #8454, opened 2021-03-16
some small quality of life changes

symonk:
Small one, just tidied up a few places while scanning through some code, making intent slightly clearer in places, documenting things slightly better in places and reducing the iterations when normalizing `Mark|MarkDecorator` instances


src/_pytest/mark/__init__.py
```diff
@@ -245,24 +240,25 @@ def deselect_by_mark(items: "List[Item]", config: Config) -> None:
...
+    deselected: List[Item] = []
     for item in items:
-        if expression.evaluate(MarkMatcher.from_item(item)):
-            remaining.append(item)
-        else:
-            deselected.append(item)
-
+        (
+            remaining if expr.evaluate(MarkMatcher.from_item(item)) else deselected
+        ).append(item)
```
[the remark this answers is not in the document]
symonk: agreed, got a bit carried away there

src/_pytest/mark/structures.py
```diff
@@ -335,7 +332,7 @@ def markname(self) -> str:
         return self.name  # for backward-compat (2.4.1 had this attr)
 
     def __repr__(self) -> str:
-        return f"<MarkDecorator {self.mark!r}>"
+        return f"<MarkDecorator {repr(self.mark)}>"
```
symonk, line 335: Any advice on how to unit test this, all that comes to mind is something like:

```
assert repr(MarkDecorator(mark="foobar", _ispytest=True)) == "MarkDecorator(mark='foobar')"
```

Im surprised codecov is complaining as I only changed the `!r` to call `repr` directly here, maybe its only on new diffs :)

The-Compiler: FWIW I don't see how this is better anyways :slightly_smiling_face:

symonk: It looks to me like this __repr__ is not used and the one provided by I assume `@attr.s` (not familiar with that lib yet) is used?

The-Compiler: What I meant is that I don't see the benefit of `{repr(self.mark)}` over `{self.mark!r}`, given that `!r` is the usual way to get a repr in f-strings (and `str.format()`).

But you're right - attrs autogenerates a `__repr__` here, so this isn't needed at all. Looks like it was added when this class was originally written in 2009 (dcf194ebb8877da46af6fc97ce494e64dd1f766e), but then never removed when attrs was introduced in 2017 (07b2b18a01a3ba12757a4eb1c0e082922f5299e5).

src/_pytest/mark/structures.py
```diff
@@ -377,18 +371,21 @@ def get_unpacked_marks(obj) -> List[Mark]:
...
 
 def normalize_mark_list(mark_list: Iterable[Union[Mark, MarkDecorator]]) -> List[Mark]:
-    """Normalize marker decorating helpers to mark objects.
+    """
+    Normalize an iterable of Mark or MarkDecorator objects into a list of marks
+    by retrieving the `mark` attribute on MarkDecorator instances.
 
-    :type List[Union[Mark, Markdecorator]] mark_list:
-    :rtype: List[Mark]
+    :param mark_list: Iterable[Union[Mark, MarkDecorator]] to normalize
```
nicoddemus, line 378:
```suggestion
    :param mark_list: marks to normalize
```

However I think describing the parameters here is a bit redundant and could be removed. 👍

nicoddemus (approved):
Thanks @symonk!

nicoddemus:
Gentle ping @RonnyPfannschmidt @The-Compiler for a final review. 👍

RonnyPfannschmidt (approved):

The-Compiler:
Looks good! I personally would still prefer `{obj!r}` over `{repr(obj)}` in f-strings (for the same reason that I'd use `{num:02}` over `{str(num).zfill(2)}`, for example) - but I don't think that should block the PR.

symonk:
@nicoddemus apologies only looping back to some old PR's now, happy to change whatever here to move it on 👍

nicoddemus:
Thanks @symonk for the ping!

Perhaps we just need to rebase this and get @The-Compiler's final approve?

The-Compiler:
No need to wait for me, my feedback above wasn't blocking 🙂

nicoddemus:
Ahh ok, gonna merge as is then. 👍

PR #8472, opened 2021-03-20
testing: stop relying on comparing to py.path in fnmatcher tests

bluetech:
This set of tests works by checking the expected result, and checking that py.path gets the same result.

Remove the latter part, as preparation for deprecating py.path.

### DIFF
--- a/testing/test_pathlib.py
+++ b/testing/test_pathlib.py
@@ -8,7 +8,6 @@ from types import ModuleType
 from typing import Generator
 
 import pytest
-from _pytest.compat import legacy_path
 from _pytest.monkeypatch import MonkeyPatch
 from _pytest.pathlib import bestrelpath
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (commented):
im torn, while this doesn't add value aanymore as py.path will goo, keeping it until the drop will keep us from accidental deviation

nicoddemus:
> im torn, while this doesn't add value aanymore as py.path will goo, keeping it until the drop will keep us from accidental deviation

I understand the sentiment, but I think we have some good tests in there to make sure we are not deviating. Even if we end up deviating, it might be better to deviate a little now than we drop py.path support.

So I'm +0.5 on it. 👍

RonnyPfannschmidt (approved):

nicoddemus (approved):

PR #8477, opened 2021-03-21
Restructured 'How to invoke pytest' guide

evildmp:


symonk:
Hi @evildmp I suspect this will clash slightly with a current PR from @nicoddemus as I believe `--resultlog` is deprecated

nicoddemus (approved):
Thanks!

PR #8505, opened 2021-03-31
Fix plugin-list label in the docs

nicoddemus:
Last time I "fixed" this, I left a `\` at the start of the string to avoid an initial newline, but didn't realize it was a raw string.

This should fix it now for good (the extra newline at the start doesn't really hurt).


The-Compiler (approved):
Thanks!

PR #8537, opened 2021-04-10
fix #8464 wrong root dir when -c is passed

### LINKED ISSUE #8464: -c flag produces the wrong root dir
### Scenario tested

```
tests/
       test.py
pytest.ini
pytest2.ini
```

The contents of `test.py`

```py
from pathlib import Path

def test_root(pytestconfig):
    assert Path(pytestconfig.rootpath).resolve().absolute() == Path('.').resolve().absolute()

```
`pytest.ini` and `pytest2.ini` are the same

+ When running  `pytest tests/` the test is **passing** (correct rootdir is printed):  `rootdir: D:\pytest, configfile: pytest.ini`
+ When running `pytest tests/ -c pytest2.ini` the test is **failing** (wrong rootdir and config are reported) :  `rootdir: D:\pytest\tests, configfile: ..\pytest2.ini`

Please let me know if I can submit a fix for this :)

elbehery95:
Fix for https://github.com/pytest-dev/pytest/issues/8464 including a doc update

@nicoddemus  let me know your thoughts and feedback

### DIFF
--- a/changelog/8464.bugfix.rst
+++ b/changelog/8464.bugfix.rst
@@ -1,0 +1,1 @@
+``-c`` no longer yield wrong rootdir.
--- a/doc/en/reference/customize.rst
+++ b/doc/en/reference/customize.rst
@@ -160,9 +160,11 @@ Here is the algorithm which finds the rootdir from ``args``:
   ``setup.cfg`` in each of the specified ``args`` and upwards. If one is
   matched, it becomes the ``configfile`` and its directory becomes the ``rootdir``.
 
-- If no ``configfile`` was found, use the already determined common ancestor as root
+- If no ``configfile`` was found and no configuration argument is passed, use the already determined common ancestor as root
   directory. This allows the use of pytest in structures that are not part of
   a package and don't have any particular configuration file.
+
+- If no ``configfile`` was found and configuration argument is passed, use the configuration argument directory as root directory.
 
 If no ``args`` are given, pytest collects test below the current working
 directory and also starts determining the ``rootdir`` from there.
--- a/doc/en/reference/customize.rst
+++ b/doc/en/reference/customize.rst
@@ -145,6 +145,8 @@ Finding the ``rootdir``
 
 Here is the algorithm which finds the rootdir from ``args``:
 
+- If ``-c`` is passed in the command-line, use that as configuration file, and its directory as ``rootdir``.
+
 - Determine the common ancestor directory for the specified ``args`` that are
   recognised as paths that exist in the file system. If no such paths are
   found, the common ancestor directory is set to the current working directory.
@@ -163,8 +165,6 @@ Here is the algorithm which finds the rootdir from ``args``:
 - If no ``configfile`` was found and no configuration argument is passed, use the already determined common ancestor as root
   directory. This allows the use of pytest in structures that are not part of
   a package and don't have any particular configuration file.
-
-- If no ``configfile`` was found and configuration argument is passed, use the configuration argument directory as root directory.
 
 If no ``args`` are given, pytest collects test below the current working
 directory and also starts determining the ``rootdir`` from there.
--- a/src/_pytest/config/findpaths.py
+++ b/src/_pytest/config/findpaths.py
@@ -176,7 +176,11 @@ def determine_setup(
         inipath: Optional[Path] = inipath_
         inicfg = load_config_dict_from_file(inipath_) or {}
         if rootdir_cmd_arg is None:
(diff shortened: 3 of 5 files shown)


doc/en/reference/customize.rst
```diff
@@ -160,10 +160,12 @@ Here is the algorithm which finds the rootdir from ``args``:
   ``setup.cfg`` in each of the specified ``args`` and upwards. If one is
   matched, it becomes the ``configfile`` and its directory becomes the ``rootdir``.
 
-- If no ``configfile`` was found, use the already determined common ancestor as root
+- If no ``configfile`` was found and no configuration argument is passed, use the already determined common ancestor as root
   directory. This allows the use of pytest in structures that are not part of
   a package and don't have any particular configuration file.
 
+- If no ``configfile`` was found and configuration argument is passed, use the configuration argument directory as root directory.
```
nicoddemus, line 167: Let's move this as a first item in this list, and reword it a bit:

```
- If ``-c`` is passed, use that as configuration file and its directory as ``rootdir``.
```

src/_pytest/config/findpaths.py
```diff
@@ -176,7 +176,11 @@ def determine_setup(
         inipath: Optional[Path] = inipath_
         inicfg = load_config_dict_from_file(inipath_) or {}
         if rootdir_cmd_arg is None:
-            rootdir = get_common_ancestor(dirs)
```
nicoddemus, line 179: Actually if ``inifile`` is given, we should use its directory unconditionally as ``rootdir``, so this line should be:

```python
if rootdir_cmd_arg is None:
    rootdir = inipath.parent
```

testing/test_config.py
```diff
@@ -1405,6 +1405,51 @@ def test_with_specific_inifile(
         assert inipath == p
         assert ini_config == {"x": "10"}
 
+    @pytest.mark.parametrize(
+        "tree,ini_file,args, expected_root",
```
nicoddemus, line 1409: Let's use spaces here for readability:

```suggestion
        "tree, ini_file, args, expected_root",
```

testing/test_config.py
```diff
@@ -1405,6 +1405,51 @@ def test_with_specific_inifile(
...
+            pytest.param(
+                ["tests/a/test1", "tests/b/test2"],
+                "tests/a/cfg.ini",
+                ["tests"],
+                "tests/a",
+                id="nested",
+            ),
+        ],
+    )
+    def test_with_args_and_specific_inifile(
```
nicoddemus, line 1434: Thanks for being through here, but I think we only need a test similar to the one posted in the original issue:

```python
def test_explicit_config_file_sets_rootdir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
    tests_dir = tmp_path / "tests"
    tests_dir.mkdir()

    monkeypatch.chdir(tmp_path)

    # No config file is explicitly given: rootdir is determined to be cwd.
    rootpath, found_inipath, *_ = determine_setup(None, [str(tests_dir)])
    assert rootpath == tmp_path
    assert found_inipath is None

    # Config file is explicitly given: rootdir is determined to be inifile's directory.
    inipath = (tmp_path / "pytest.ini")
    inipath.touch()
    rootpath, found_inipath, *_ = determine_setup(str(inipath), [str(tests_dir)])
    assert rootpath == tmp_path
    assert found_inipath == inipath
```

This should be enough to enforce the correct behavior.

nicoddemus (changes requested):
Thanks a lot @elbehery95 for the PR!

However I think I wasn't very clear, but the semantics here should be a bit different: if an inifile is passed in the command-line, then we should always use its directory as rootdir. Please see my comments. 👍

src/_pytest/config/findpaths.py
elbehery95: Thanks totally agree

testing/test_config.py
elbehery95: Much cleaner and on-point, replaced  `test_with_args_and_specific_inifile` with `test_explicit_config_file_sets_rootdir` thanks so much for clarification

changelog/8464.bugfix.rst
```diff
@@ -0,0 +1 @@
+``-c`` no longer yield wrong rootdir.
```
nicoddemus, line 1:
```suggestion
``-c <config file>`` now also properly defines ``rootdir`` as the directory that contains ``<config file>``.
```

nicoddemus (approved):
Thanks a ton @elbehery95 for the contribution!

PR #8540, opened 2021-04-11
Fix assertion rewriting on Python 3.10

hauntsaninja:
Fixes https://github.com/pytest-dev/pytest/issues/8539

This seems to have been the result of https://bugs.python.org/issue43798

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -684,12 +684,9 @@ class AssertionRewriter(ast.NodeVisitor):
         if not mod.body:
             # Nothing to do.
             return
+
         # Insert some special imports at the top of the module but after any
         # docstrings and __future__ imports.
-        aliases = [
-            ast.alias("builtins", "@py_builtins"),
-            ast.alias("_pytest.assertion.rewrite", "@pytest_ar"),
-        ]
         doc = getattr(mod, "docstring", None)
         expect_docstring = doc is None
         if doc is not None and self.is_rewrite_disabled(doc):
@@ -721,6 +718,19 @@ class AssertionRewriter(ast.NodeVisitor):
             lineno = item.decorator_list[0].lineno
         else:
             lineno = item.lineno
+        if sys.version_info >= (3, 10):
+            aliases = [
+                ast.alias("builtins", "@py_builtins", lineno=lineno, col_offset=0),
+                ast.alias(
+                    "_pytest.assertion.rewrite", "@pytest_ar",
+                    lineno=lineno, col_offset=0
+                ),
+            ]
+        else:
+            aliases = [
+                ast.alias("builtins", "@py_builtins"),
+                ast.alias("_pytest.assertion.rewrite", "@pytest_ar"),
+            ]
         imports = [
             ast.Import([alias], lineno=lineno, col_offset=0) for alias in aliases
         ]
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -685,8 +685,8 @@ class AssertionRewriter(ast.NodeVisitor):
             # Nothing to do.
             return
 
-        # Insert some special imports at the top of the module but after any
-        # docstrings and __future__ imports.
+        # We'll insert some special imports at the top of the module, but after any
+        # docstrings and __future__ imports, so first figure out where that is.
         doc = getattr(mod, "docstring", None)
         expect_docstring = doc is None
         if doc is not None and self.is_rewrite_disabled(doc):
@@ -718,6 +718,7 @@ class AssertionRewriter(ast.NodeVisitor):
             lineno = item.decorator_list[0].lineno
         else:
             lineno = item.lineno
+        # Now actually insert the special imports.
         if sys.version_info >= (3, 10):
             aliases = [
                 ast.alias("builtins", "@py_builtins", lineno=lineno, col_offset=0),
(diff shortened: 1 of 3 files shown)


hauntsaninja:
Hmm, looks like the test is failing on master as well. Looks like it's https://github.com/benjaminp/six/pull/352

src/_pytest/assertion/rewrite.py
```diff
@@ -684,12 +684,9 @@ def run(self, mod: ast.Module) -> None:
         if not mod.body:
             # Nothing to do.
             return
+
         # Insert some special imports at the top of the module but after any
         # docstrings and __future__ imports.
```
asottile, line 689: this comment has been orphaned

hauntsaninja: That comment is still relevant in its current place! L690-L720 is figuring out the "after any docstrings and __future__ imports part".

asottile: it's 40 lines away from the code that's relevant to the comment (the "special imports")

hauntsaninja: Okay, made a change! :-)

asottile (approved):

The-Compiler:
Could someone elaborate on how I can actually reproduce this? I'm trying to fix the testsuite on Python 3.10, but I don't see this issue happening at all.

asottile:
> 
> 
> Could someone elaborate on how I can actually reproduce this? I'm trying to fix the testsuite on Python 3.10, but I don't see this issue happening at all.

for example: https://github.com/asottile/flake8-match

```
git clone git@github.com:asottile/flake8-match
sudo add-apt-repository ppa:deadsnakes/nightly
sudo apt install python3.10-dev
tox -e py310
```

The-Compiler:
Ah, I see - I was running 3.10.0a7 (because that's what @Zac-HD mentioned in the issue), but this is actually a [very recent change](https://github.com/python/cpython/pull/25324). I can reproduce it when installing `python3.10-dev` via pyenv, which gives me "Python 3.10.0a7+ (heads/master:d9ba9de, Apr 14 2021, 16:49:31)".

This also explains why this isn't needed on CI - it will be once beta 1 is out.

nicoddemus (approved):
Thanks a lot @hauntsaninja!

nicoddemus:
Should have squashed, my bad! 🤦‍♂️

hroncok:
Will this not break actual alpha 7?

nicoddemus:
Hmm I suppose, good point. Should we revert this and wait for beta 1?

hroncok:
I think we can make the version check more specific.

hroncok:
I cannot trigger a problem on Python 3.10.0a7, so maybe this is fine.

The-Compiler:
FYI, GitHub Actions seems to have pushed a [rather weird](https://github.com/actions/setup-python/issues/207) "beta 1" release which isn't actually beta 1 - so if we're seeing an influx of new reports about this, that'd be why (as `python: 3.10-dev` now uses that affected release rather than Alpha 7).

nicoddemus:
Backport: https://github.com/pytest-dev/pytest/pull/8630

AlirezaTheH:
I'm still having this issue on Python `3.10.4`, pytest `7.1.2`!

asottile:
@AlirezaTheH please open a new issue with a reproduction

parsanoori:
Still have this problem:
```
git clone https://gitlab.com/nanuchi/gitlab-cicd-crash-course.git
cd gitlab-cicd-crash-course
make test
```
The output is:
```
. src/.venv/bin/activate \
&& pytest -v
========================================================= test session starts =========================================================
platform linux -- Python 3.10.4, pytest-6.2.2, py-1.11.0, pluggy-0.13.1 -- /home/parsa/gits/gitlab-cicd-crash-course/src/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/parsa/gits/gitlab-cicd-crash-course
collected 0 items / 1 error                                                                                                           

=============================================================== ERRORS ================================================================
____________________________________________________ ERROR collecting test session ____________________________________________________
/usr/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1050: in _gcd_import
    ???
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
<frozen importlib._bootstrap>:1006: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:688: in _load_unlocked
    ???
src/.venv/lib/python3.10/site-packages/_pytest/assertion/rewrite.py:161: in exec_module
    source_stat, co = _rewrite_test(fn, self.config)
src/.venv/lib/python3.10/site-packages/_pytest/assertion/rewrite.py:356: in _rewrite_test
    co = compile(tree, fn_, "exec", dont_inherit=True)
E   TypeError: required field "lineno" missing from alias
======================================================= short test summary info =======================================================
ERROR  - TypeError: required field "lineno" missing from alias
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
========================================================== 1 error in 0.23s ===========================================================
make: *** [makefile:56: test] Error 2
```
Related info:
```
python3 --version # the command
Python 3.10.4 # the output


python3 -m pytest --version # the command
pytest 7.1.2 # the output
```

The-Compiler:
@parsanoori as seen in the output header, you're actually running pytest 6.2.2 in the virtualenv there.

parsanoori:
It's the git maintainer's problem.
Thanks.

PR #8550, opened 2021-04-14
Add tmp_path_factory methods to the docs

albertvillanova:
Add the methods of tmp_path_factory to the docs.

Currently, only the class description appears in the docs.


bluetech:
Whoops, true!

I see it's also missing for `tmpdir_factory` (just below), can you add it there as well?

bluetech (approved):
Thank you!

PR #8551, opened 2021-04-14
doc: Update open trainings

The-Compiler:
The training is sold out with 10 people (yay!), so there's a second date for it now.

Is there a better way we can keep pytest.org up to date with those changes? From what I understand, it only gets updated when we make a new release from `main`, which hasn't happened this year so far? It's kinda troublesome we still advertise a training from February (with "Early-bird discount available until January 15th.") there... 

I struggle to think of a better solution though... Can we (and do we want to) reconfgure RTD so it shows "latest" by default instead, or would that be too confusing if we had breaking changes?

### DIFF
--- a/doc/en/index.rst
+++ b/doc/en/index.rst
@@ -2,7 +2,7 @@
 
 .. sidebar:: Next Open Trainings
 
-   - `Professionelles Testen für Python mit pytest <https://www.enterpy.de/lecture_compact1.php?id=12713>`_ (German), part of the enterPy conference, April 22nd, remote.
+   - `Professionelles Testen für Python mit pytest <https://www.enterpy.de/lecture_compact1.php?id=12713>`_ (German), part of the enterPy conference, April 22nd (sold out) and May 20th, remote.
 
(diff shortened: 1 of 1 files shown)


bluetech (approved):

bluetech:
> I struggle to think of a better solution though... Can we (and do we want to) reconfgure RTD so it shows "latest" by default instead, or would that be too confusing if we had breaking changes?

It used to be pointed at latest but we changed it to stable for this reason.

I don't have any bright ideas besides cherry-picking these over to the stable branch (can merge without waiting for approval) and point the RTD `stable` at the stable *branch* (rather than release) if that's even possible.

The-Compiler:
I can take a look but I assume someone would need to give me access? I just created an account over there, `The-Compiler`.

nicoddemus:
> I just created an account over there, The-Compiler.

Done! 👍

The-Compiler:
Taking the freedom to merge this as the failures are unrelated. Will follow up with a backport too.

PR #8553, opened 2021-04-14
ci: Mark Python 3.10 as experimental

The-Compiler:
It currently breaks all builds and even after #8540 it looks like tests
still fail because of https://github.com/benjaminp/six/issues/341

As a hotfix, mark it as experimental until everything is fixed, so that
our CI isn't all red.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,6 +18,7 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
+    continue-on-error: "${{ matrix.experimental == true }}"
 
     strategy:
       fail-fast: false
@@ -73,6 +74,7 @@ jobs:
             python: "3.10-dev"
             os: windows-latest
             tox_env: "py310-xdist"
+            experimental: true
 
           - name: "ubuntu-py36"
             python: "3.6"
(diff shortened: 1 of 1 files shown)


bluetech (approved):

The-Compiler:
Hm, looks like CI still is marked as failed. Not sure what I'm missing?

nicoddemus:
Not sure we need this: I did research this a bit before (can't find links now), but seems the official GH Actions stance on "mark build as allowed to fail" is to make the build not required, which we already do:

image


The effect is that even though 3.10 has failed here, the "merge" button is still enabled (in the screenshot above if `ubuntu-py36` failed, then the merge button wouldn't be enabled).

The-Compiler:
GitHub [documents it](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run) as:

> You can allow specific jobs in a job matrix to fail without failing the workflow run.

And indeed looking at the icon at the top left, the [workflow run passed](https://github.com/pytest-dev/pytest/actions/runs/747725468) - but it doesn't seem to affect the PRs being marked as failure. So yeah, let's close this, as it doesn't really help much after all.

PR #8555, opened 2021-04-14
Fix Python 3.10 test issues

The-Compiler:
Potential alternative to #8553. Fixes the Python 3.10 tests for me, without #8540. See #8546, though I wouldn't close it yet, as we might want to change the enum reprs instead?

### DIFF
--- a/testing/python/metafunc.py
+++ b/testing/python/metafunc.py
@@ -448,7 +448,10 @@ class TestMetafunc:
         enum = pytest.importorskip("enum")
         e = enum.Enum("Foo", "one, two")
         result = idmaker(("a", "b"), [pytest.param(e.one, e.two)])
-        assert result == ["Foo.one-Foo.two"]
+        if sys.version_info[:2] >= (3, 10):
+            assert result == ["one-two"]
+        else:
+            assert result == ["Foo.one-Foo.two"]
 
     def test_idmaker_idfn(self) -> None:
         """#351"""
--- a/testing/test_pytester.py
+++ b/testing/test_pytester.py
@@ -744,10 +744,16 @@ def test_run_result_repr() -> None:
 
     # known exit code
     r = pytester_mod.RunResult(1, outlines, errlines, duration=0.5)
-    assert (
-        repr(r) == "<RunResult ret=ExitCode.TESTS_FAILED len(stdout.lines)=3"
-        " len(stderr.lines)=4 duration=0.50s>"
-    )
+    if sys.version_info[:2] >= (3, 10):
+        assert repr(r) == (
+            "<RunResult ret=TESTS_FAILED len(stdout.lines)=3"
+            " len(stderr.lines)=4 duration=0.50s>"
+        )
+    else:
+        assert repr(r) == (
+            "<RunResult ret=ExitCode.TESTS_FAILED len(stdout.lines)=3"
+            " len(stderr.lines)=4 duration=0.50s>"
+        )
 
     # unknown exit code: just the number
     r = pytester_mod.RunResult(99, outlines, errlines, duration=0.5)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,14 @@ filterwarnings = [
     "default:invalid escape sequence:DeprecationWarning",
     # ignore use of unregistered marks, because we use many to test the implementation
     "ignore::_pytest.warning_types.PytestUnknownMarkWarning",
+    # https://github.com/benjaminp/six/issues/341
+    "ignore:_SixMetaPathImporter\\.exec_module\\(\\) not found; falling back to load_module\\(\\):ImportWarning",
+    # https://github.com/benjaminp/six/pull/352
+    "ignore:_SixMetaPathImporter\\.find_spec\\(\\) not found; falling back to find_module\\(\\):ImportWarning",
+    # https://github.com/pypa/setuptools/pull/2517
+    "ignore:VendorImporter\\.find_spec\\(\\) not found; falling back to find_module\\(\\):ImportWarning",
+    # https://github.com/pytest-dev/execnet/pull/127
+    "ignore:isSet\\(\\) is deprecated, use is_set\\(\\) instead:DeprecationWarning",
 ]
 pytester_example_dir = "testing/example_scripts"
 markers = [
--- a/testing/acceptance_test.py
+++ b/testing/acceptance_test.py
@@ -1173,7 +1173,7 @@ def test_usage_error_code(pytester: Pytester) -> None:
     assert result.ret == ExitCode.USAGE_ERROR
 
 
-@pytest.mark.filterwarnings("default")
+@pytest.mark.filterwarnings("default::pytest.PytestUnhandledCoroutineWarning")
 def test_warn_on_async_function(pytester: Pytester) -> None:
     # In the below we .close() the coroutine only to avoid
     # "RuntimeWarning: coroutine 'test_2' was never awaited"
@@ -1206,7 +1206,7 @@ def test_warn_on_async_function(pytester: Pytester) -> None:
     )
 
 
-@pytest.mark.filterwarnings("default")
+@pytest.mark.filterwarnings("default::pytest.PytestUnhandledCoroutineWarning")
 def test_warn_on_async_gen_function(pytester: Pytester) -> None:
     pytester.makepyfile(
         test_async="""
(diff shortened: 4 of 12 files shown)


The-Compiler:
Huh. `ubuntu-py310` works, but on `windows-py310`, all tests using `@pytest.mark.filterwarnings("default")` (I think?) still fail. I'm stumped.

testing/python/metafunc.py
```diff
@@ -448,7 +448,10 @@ def test_idmaker_enum(self) -> None:
         enum = pytest.importorskip("enum")
         e = enum.Enum("Foo", "one, two")
         result = idmaker(("a", "b"), [pytest.param(e.one, e.two)])
-        assert result == ["Foo.one-Foo.two"]
+        assert result in [
```
nicoddemus, line 451: I suggest to break this using an if:

```python
if sys.version_info[:2] >= (3, 10):
    assert result == ["one-two"]
else:
    assert result == ["Foo.one-Foo.two"]
```

Having explicit `version_info` checks makes it much easier to find dead code once we drop Python versions. 👍

testing/test_pytester.py
```diff
@@ -744,10 +744,14 @@ def test_run_result_repr() -> None:
 
     # known exit code
     r = pytester_mod.RunResult(1, outlines, errlines, duration=0.5)
-    assert (
-        repr(r) == "<RunResult ret=ExitCode.TESTS_FAILED len(stdout.lines)=3"
-        " len(stderr.lines)=4 duration=0.50s>"
-    )
+    expected = [
```
nicoddemus, line 747: Ditto here as previous comment.

The-Compiler:
Fair point @nicoddemus - but do you have any idea what's up with the failures on Windows?

nicoddemus:
> Fair point @nicoddemus - but do you have any idea what's up with the failures on Windows?

Ahh sorry, missed those comments. Not at a quick glance, will try to test this locally when I get the chance. 👍

testing/test_pytester.py
bluetech, line 747: I suggest we just make it consistent here -- make it use one or the other (the new one looks better to me)

testing/python/metafunc.py
bluetech, line 451: I described my preferred solution [here](https://github.com/pytest-dev/pytest/issues/8546#issuecomment-818662849), though shouldn't hold up this PR.

testing/test_pytester.py
nicoddemus: I suppose we definitely will search for "3.10" when dropping that version, so in the end doesn't really matter I guess.

nicoddemus:
OK did some debugging (sorry for the delay) and here is what I found.

First I could not reproduce the tests running them in isolation, I needed to do a binary search and to find the minimal two tests that when executed in order, would fail the second:

```python
import pytest

pytest.main([
	"testing/test_nose.py::test_raises",
	"testing/acceptance_test.py::test_warn_on_async_function", 
	"-x", 
	"-v"],
)
```

Running this script causes `test_warn_on_async_function` to fail, which it doesn't when run in isolation.

Looking at `test_nose.py::test_raises`, we don't see anything special in there, so I suspected that just by importing `nose` we were triggering the problem. I changed the test to:

```python
def test_raises(pytester: Pytester) -> None:
    pytester.makepyfile("import nose")
```

And indeed the problem happens, so importing `nose` is triggering the extra warning in `test_warn_on_async_function`.

Why it fails? Well it changes the warning filtering to `default`:

```python
@pytest.mark.filterwarnings("default")
def test_warn_on_async_function(pytester: Pytester) -> None:
```

And it checks the number of warnings exactly:

```python
	result.stdout.fnmatch_lines(
        [
            "test_async.py::test_1",
            "test_async.py::test_2",
            "test_async.py::test_3",
            "*async def functions are not natively supported*",
            "*3 skipped, 3 warnings in*",
        ]
    )
```

So when `nose` is imported, an extra warning is triggered, which causes that check to fail (expected 3 warnings, got 4).

Why this is happening only on Windows? I suspect there's some Windows-specific code which causes that warning to be triggered, but I didn't research it further.

Why is the warning leaking from one test to the other? Not sure too.

---

What to do?

I think the test is a little brittle, it expects only 3 warnings even though it changed the warnings filter to `default`, so any extra warnings raised in the future would cause it to fail. I think we don't need to check the exact number of warnings here, just checking the messages is enough here, so perhaps this:

```python
	result.stdout.fnmatch_lines(
        [
            "test_async.py::test_1",
            "test_async.py::test_2",
            "test_async.py::test_3",
            "*async def functions are not natively supported*",
            "*3 skipped, * warnings in*",
        ]
    )
```

?

Didn't check the other tests specifically, but I suspect it is the same problem.

bluetech:
> What to do?

I agree with your suggestion.

The-Compiler:
Sounds good @nicoddemus. Still busy with my pytest training, but I plan to pick this up tomorrow or on Friday. I think it's really important to have a passing CI again before everyone starts ignoring (or being confused by) the commit status.

nicoddemus:
Definitely, thanks for tackling this @The-Compiler!

The-Compiler:
Looks like there are way more tests failing though, sometimes we're e.g. expecting a certain order in the warning output, but the [new warnings](https://github.com/pytest-dev/pytest/pull/8555/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) mess everything up. Still no idea why that's only the case on Windows, though.

Will need to dig into it some more I guess, but not today.

The-Compiler:
This should be ready for another round of review now - and hopefully finally result in a green CI everywhere, fingers crossed! :crossed_fingers: 

A couple of comments:

- Agreed with making the version checks more explicit, based on @nicoddemus' suggestions.
- I disagree with @nicoddemus' proposal above on how to handle warnings in tests. I don't think we should write our integration tests in a way that they are less strict and pass even if other warnings occur - in fact, quite the opposite! If unexpected warnings occur, the tests should fail. So instead of changing the matching, I decided to adjust the `filterwarnings` marks so they don't unconditionally set `default` (or `always`) for *all* warnings (which would ignore our existing filters in `pyproject.toml`, and also ignore the default `error` filter, i.e. potentially hide real issues!). Instead, they now only affect the warning class the test in question actually uses.
- `test_collect_symlink_dir` failed for me on Windows (which needs special privileges to create symlinks). Not sure how it ever could've run in earlier versions, as it didn't use `symlink_or_skip` like all the others. Maybe it just runs on CI because the CI has those privileges or something. I didn't investigate further, as this PR already took a lot more work than expected :sweat_smile: 
- The beta 1 brought a couple of new issues, which now should all be fixed as well. One of them is https://github.com/pytest-dev/execnet/pull/127, the other is changed output for SyntaxErrors in skip/xfail expressions:

```
_______________________ ERROR at setup of test_nameerror _______________________
name 'asd' is not defined

Error evaluating 'skipif' condition
    asd
NameError: name 'asd' is not defined
________________________ ERROR at setup of test_syntax _________________________
invalid syntax. Perhaps you forgot a comma? (<xfail condition>, line 1)

Error evaluating 'xfail' condition
    syntax error
     ^
SyntaxError: invalid syntax
```

(the "During handling of the above exception, another exception occurred:" and everything above it is new).

nicoddemus (approved):
Awesome, thanks!

The-Compiler:
Taking the freedom to merge this, since it's approved and I'd really like to see passing PR status checks again. @bluetech if you feel strongly about changing pytest to always output consistent enum repr's, I suggest you open a new issue. I can see where you're coming from, but personally I think it's too minor to worry about it, as those repr's aren't typically exposed to users.

hroncok:
Note that this was reverted in upstream.

https://mail.python.org/archives/list/python-dev@python.org/message/LSTMFAPSPD3BGZ4D6HQFODXZVB3PLYKF/

The Python version with the reverted thing is 3.10.0b4. We can adapt the if to say 3.11, or we can revert it for now, or we can adapt the if to be more complex (and check for the actual pre-release).

hroncok:
When I said *this*, I've meant just the *Adjust enum reprs for Python 3.10* commit here, will edit my comment.

nicoddemus:
>  We can adapt the if to say 3.11, or we can revert it for now, or we can adapt the if to be more complex (and check for the actual pre-release).

I vote to adapt the `if` to use `3.11`.

hroncok:
On it.

hroncok:
https://github.com/pytest-dev/pytest/pull/8895

PR #8557, opened 2021-04-15
Backport training update

The-Compiler:
Next I plan to:

- Set the `6.2.x` branch as an active version on RTD (like `4.6.x` already is)
- Set the `6.2.x` version as default instead of `stable`
- Make a reminder to revert back to `stable` once we have a new release out

I don't see a way to tell RTD to use the latest `6.*.*` branch, but that's as close as we can get.

Does that seem right?

### DIFF
--- a/doc/en/index.rst
+++ b/doc/en/index.rst
@@ -2,7 +2,7 @@
 
 .. sidebar:: Next Open Trainings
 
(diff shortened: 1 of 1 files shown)


nicoddemus:
> Does that seem right?

It does, thanks for looking it up!

nicoddemus (approved):

bluetech (approved):

PR #8564, opened 2021-04-19
Fix some typos, remove redundant words and escapes

eumiro:
Fix some typos, remove redundant words.
Remove redundant escapes from a regex.


The-Compiler (approved):
That's a lot of "that that"s! 😉

PR #8570, opened 2021-04-20
Logging docs: fix reference to pytest.LogCaptureFixture.

soxofaan:
Small PR to fix reference to `LogCaptureFixture` from https://docs.pytest.org/en/stable/logging.html#caplog-fixture


bluetech (approved):
Thanks!

PR #8575, opened 2021-04-21
Remove "upload_sphinx" section from setup.cfg

domdfcoding:
The `Sphinx-PyPI-upload` tool the section corresponds to hasn't been updated since 2009, and the new (c. 2018) PyPI doesn't support documentation uploads anyway.

Therefore I think the section is redundant and safe to remove.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,9 +82,6 @@ source_dir = doc/en/
 build_dir = doc/build
 all_files = 1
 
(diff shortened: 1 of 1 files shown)


nicoddemus:
Thanks! Indeed readthedocs host the documentation for us, so this is unnecessary. 👍

nicoddemus (approved):

PR #8595, opened 2021-04-26
"fix" a couple of http -> https redirects

nschloe:
Found with
```
urli-fix . -a http: -i pytest
```
(https://github.com/nschloe/urli)


bluetech (approved):
Nice, thanks!

PR #8610, opened 2021-04-29
Expand command line argument examples with validation examples

rvandegrift:
This shows how to pass more detailed error messages back to the user when
they've provided an incorrect value for a custom command line option.


bluetech (approved):
Thank you, that's a nice addition.

PR #8611, opened 2021-04-30
make numpy example in doctest to work

gcamargo1:


The-Compiler:
Thanks for the contribution! However, this snipped was moved to `doc/en/how-to/doctest.rst` in the `main` branch, which means your PR leads to conflicts - did you perhaps accidentally base it on an older version of the sources?

PR #8614, opened 2021-04-30
docfix: tolerance setting in unittest.TestCase.assertAlmost

### LINKED ISSUE #8587: approx() documentation page incorrectly compares with unittest as of 3.2
Hello!
While navigating on the docs, I found that the `approx()` documentation, while comparing its usage to the analogous function in `unittest`, says:

> unittest.TestCase.assertAlmostEqual(a, b): True if a and b are within an absolute tolerance of 1e-7. No relative tolerance is considered and the absolute tolerance cannot be changed, so this function is not appropriate for very large or very small numbers.

This statement is not true as of unittest version 3.2, as:

> Changed in version 3.2: [...] Added the delta keyword argument.

which lets a user choose the desired level of tolerance.

I am not a `pytest` user (yet), so I don't know what I could change that statement to, just pointing that out to someone who does.

rahul-kumi:
Closes #8587

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -640,10 +640,9 @@ def approx(expected, rel=None, abs=None, nan_ok: bool = False) -> ApproxBase:
 
     - ``unittest.TestCase.assertAlmostEqual(a, b)``: True if ``a`` and ``b``
       are within an absolute tolerance of ``1e-7``.  No relative tolerance is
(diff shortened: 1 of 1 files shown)


rahul-kumi:
@Zac-HD, can you please reveiw this PR.

Zac-HD (approved):

The-Compiler:
Thanks @rahul-kumi and @Zac-HD!

PR #8615, opened 2021-04-30
Suggest numpy testing module on pytest approx docs

### LINKED ISSUE #8337: Recommend `np.testing` module instead of `pytest.approx` with Numpy arrays
Via https://github.com/pytest-dev/pytest/issues/8335#issuecomment-777359893, this issue proposes that we should deprecate `pytest.approx()` for `numpy.ndarray`s in favor of the `numpy.testing` module.  That module is custom-built with lots of useful assertion helpers for Numpy, while `pytest.approx()` only incidentally copes with multi-dimensional arrays and the risk of treating arrays as truthy booleans in assertions.

I particularly like this because we can very easily [check if something is an array without importing Numpy](https://twitter.com/jakevdp/status/1130983439862181888), and teaching people about the `numpy.testing` module is likely to lead to better tests 🙂

tarcisiofischer:
closes #8337 

I was in doubt if this change needs a `changelog`, and, if it needs, if I should add  `trivial` or `docs` (because it is both, IMO)


changelog/8337.trivial.rst
```diff
@@ -0,0 +1 @@
+Recommend np.testing module on :doc:`pytest.approx` documentation.
```
nicoddemus, line 1:
```suggestion
Recommend `numpy.testing <https://numpy.org/doc/stable/reference/routines.testing.html>`__ module on :doc:`pytest.approx` documentation.
```

This file also should be named `8337.doc.rst`. 👍

nicoddemus (changes requested):
Thanks @tarcisiofischer!

Small suggestion on the changelog, other than that LGTM!

tarcisiofischer:
Is there any way I can see the generated docs before merging? I didn't check the produced HTML file :thinking:

nicoddemus:
> Is there any way I can see the generated docs before merging? I didn't check the produced HTML file 🤔

There's a readthedocs build that you can see the generated HTML:

https://pytest--8615.org.readthedocs.build/en/8615/

image

tarcisiofischer:
LGTM. Ready to merge.

nicoddemus (approved):

nicoddemus:
FYI I've fixed the changelog and moved the suggestion to a "Note" box at the end of the text, which I believe it is better. 👍

nicoddemus:
Note: we should squash this IMHO.

src/_pytest/python_api.py
```diff
@@ -478,6 +478,14 @@ def approx(expected, rel=None, abs=None, nan_ok: bool = False) -> ApproxBase:
       special case that you explicitly specify an absolute tolerance but not a
       relative tolerance, only the absolute tolerance is considered.
 
+    .. note::
+
+        Although ``approx`` can handle numpy arrays, you may also want to check
+        `numpy.testing`__ module, which have many possibilities of comparing
+        floating point numbers, if you are comparing two numpy arrays of floats.
```
Zac-HD, line 485:
```suggestion
        ``approx`` can handle numpy arrays, but we recommend the
        specialised test helpers in `numpy.testing`__ if you need
        support for comparisons, NaNs, or ULP-based tolerances.
```

This is pretty pedantic, sorry, but the Numpy helpers support non-floats too (e.g. integers).  I also think it's worth spelling out what `numpy.testing` does that `pytest.approx` can't.

nicoddemus: Good point, applied!

Waiting for final approval before hitting merge. 👍

tarcisiofischer: Seems reasonable to me!

Zac-HD (approved):

PR #8620, opened 2021-05-02
doc: plugin link on readme.rst fixed

freelancing-solutions:
Close https://github.com/pytest-dev/pytest/issues/8619.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -162,6 +162,7 @@ Josh Karpel
 Joshua Bronson
 Jurko Gospodnetić
 Justyna Janczyszyn
+Justice Ndou
 Kale Kundert
 Kamran Ahmad
 Karl O. Pinc
--- a/README.rst
+++ b/README.rst
@@ -93,7 +93,7 @@ Features
 
 - Python 3.6+ and PyPy3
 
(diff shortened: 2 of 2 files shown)


freelancing-solutions:
document fix , 
Plugin list link is broken
a working link has been relocated to :
[plugin list ](https://docs.pytest.org/en/latest/reference/plugin_list.html)

nicoddemus:
Thanks @freelancing-solutions. 

While usually we point to the `stable` docs, I think it makes sense here to point to `latest` instead.

The-Compiler (commented):
Agreed that it makes more sense to point to `latest` here, especially because that page doesn't exist in `stable` at all.

However, I guess the version switch is a bit worrying though (if a user clicks through to the plugin list and then uses the docs menu, they'll end up with potentially "wrong" (i.e. too new) documentation, without noticing.

Not sure what to do here.

nicoddemus:
> However, I guess the version switch is a bit worrying though (if a user clicks through to the plugin list and then uses the docs menu, they'll end up with potentially "wrong" (i.e. too new) documentation, without noticing.

But we don't include the `README.rst` file in the docs, so the only way for the user to use this link is through GitHub, unless I'm missing something. 🤔

The-Compiler:
Ah, right, I missed that! Let's merge this then.

The-Compiler (approved):

Issue #8606, opened 2021-04-28
Verbosity and --fixtures / --fixtures-per-test


The-Compiler:
Something that's usually interesting to know when checking what fixtures exist in a project is where exactly they are defined. In other words, imagine I'm looking at a test, and I want to know where those fixtures are coming from.

Intuitively, I'd run with `--fixtures-per-test`, which shows me the fixture names and docstrings, but I still don't know where they come from. I was today years old when I learned that adding `--verbose` will show that information.

I think there's nothing particularly verbose about it, though! It's just one line, which will usually fit on the same line as the fixture name anyways! However, something that's *very* verbose (sometimes) are the docstrings...

What about:

- Always showing the path to the fixtures
- Only showing the first line of the docstrings unless `--verbose` is given?

nicoddemus:
Definitely 👍 

I don't use that option often, but using it now your proposal makes total sense (specially showing the first line of the docstring... I really don't need the full docs each time `monkeypatch` shows up 😁).

symonk:
I like the idea 👍

The-Compiler:
Fixed via #8626 - thanks @rahul-kumi!

PR #8626, opened 2021-05-03
add feature to view fixture source location in invocations with --fixtures-per-test option

rahul-kumi:
This implements the proposal in issue #8606

The following are the changes made to the behaviour:
- Always showing the path to the fixtures when using the `--fixtures-per-test` option
- Only showing the first line of the docstrings unless `--verbose` is given


rahul-kumi:
I have not made an entry into `changelog` and `documentation`.
I have put them on hold to get your reviews on these change and add them.

@The-Compiler, @nicoddemus Please review the PR and let me know if there are changes to be made.

src/_pytest/python.py
```diff
@@ -1428,15 +1428,13 @@ def write_fixture(fixture_def: fixtures.FixtureDef[object]) -> None:
         argname = fixture_def.argname
         if verbose <= 0 and argname.startswith("_"):
             return
-        if verbose > 0:
+        else:
             bestrel = get_best_relpath(fixture_def.func)
             funcargspec = f"{argname} -- {bestrel}"
```
The-Compiler, line 1433: IMHO this gets just a tad more readable if you [eschew the extraneous else](https://www.youtube.com/watch?t=289&v=JVVMMULwR4s&feature=youtu.be) :wink:

```suggestion
        bestrel = get_best_relpath(fixture_def.func)
        funcargspec = f"{argname} -- {bestrel}"
```

testing/python/show_fixtures_per_test.py
```diff
@@ -172,7 +170,6 @@ def foo():
     pytester.maketxtfile(
         """
         >>> 1 + 1
-        2
```
The-Compiler, line 175: seems like an unintended/unrelated change?

The-Compiler (changes requested):
Yay, thank you! :tada: Looks good overall, I commented on two small things.

rahul-kumi:
@The-Compiler, Thanks for the review.
I made the suggested changes.

Can you please take a look and let me know if documentation and changelog for this change can be added?

The-Compiler (commented):
Thanks! One thing I noticed just now is that it kind of breaks with docstrings where the summary spans across multiple lines, like e.g. with `mocked_urlopen`:

https://github.com/pytest-dev/pytest/blob/3a2fd9630547e224bea6e024a4f6852761235c61/testing/test_pastebin.py#L128-L132

This now displays as:

```
-------------------- fixtures used by test_create_new_paste --------------------
------------------------ (testing/test_pastebin.py:163) ------------------------
mocked_urlopen -- testing/test_pastebin.py:129
    Monkeypatch the actual urlopen calls done by the internal plugin
monkeypatch -- src/_pytest/monkeypatch.py:29
    A convenient fixture for monkey-patching.
pastebin -- testing/test_pastebin.py:88
    no docstring available
```

so ideally, we'd instead split on the first empty line. Perhaps something like `.split('\n\n')` would already be good enough?

Also, it looks like this PR only adjusts `--fixtures-per-test` - I think it'd be good to adjust `--fixtures` in the same way as well.

rahul-kumi:
@The-Compiler,
Use of the `.split('\n\n')` seems to be a good idea to tackle cases where the docstring sentence is spread across different lines. Introducing this would mean that we would be returning the first paragraph of the docstring if it is paragraphed else the entire content.

Hope this behaviour is acceptable.

rahul-kumi:
@The-Compiler,
>so ideally, we'd instead split on the first empty line. Perhaps something like .split('\n\n') would already be good enough?

This is addressed in this commit: c0dc982

>I think it'd be good to adjust --fixtures in the same way as well.

This is addressed in this commit: 02f026f

The-Compiler (changes requested):
Thanks for the update! Also a great idea to add some test for the docstring parsing.

One thing I didn't realize before looking at the updated diff is that `--fixtures --verbose` actually did show the path in a different color:

image

I think this helps readability a bit, and IMHO it'd be good to keep that. What do you think?

Other than that, this will still need a changelog entry, but seems fine otherwise. I don't think any other changes to the documentation are needed, but maybe I missed something.

rahul-kumi:
@The-Compiler, I have made the suggested changes.
Can you please review the same and let me know if there's anything else that has to be improved.

The-Compiler:
I've seen your review request and will take another look, but it'll probably be somewhen next week until I get around to it.

src/_pytest/python.py
```diff
@@ -1508,18 +1508,17 @@ def _showfixtures_main(config: Config, session: Session) -> None:
                 tw.line()
                 tw.sep("-", f"fixtures defined from {module}")
                 currentmodule = module
-        if verbose <= 0 and argname[0] == "_":
+        if verbose <= 0 and argname.startswith("_"):
```
The-Compiler, line 1511: Seems like an unrelated and potentially unneeded change? (but I'm okay with both variants).

src/_pytest/python.py
```diff
@@ -1508,18 +1508,17 @@ def _showfixtures_main(config: Config, session: Session) -> None:
...
             continue
-        tw.write(argname, green=True)
+        tw.write(f"{argname}", green=True)
```
The-Compiler, line 1513: Is this intended?

The-Compiler (changes requested):
Thanks, the output looks great now! I commented on two changes where I'm unsure what your intentions were - after that I think this should be ready to merge!

src/_pytest/python.py
rahul-kumi: I wanted to keep this consistent, noticed a similar check taking the `startswith` form. So made this change.

src/_pytest/python.py
rahul-kumi: Yes this inclusion is intended.

The-Compiler: Alright - I don't really see the point of using an f-string here (with only one single string variable in it), but this is too minor to block this PR.

The-Compiler (approved):
Thanks again! Makes this output much more useful, in my opinion!

src/_pytest/python.py
```diff
@@ -1428,15 +1428,15 @@ def write_fixture(fixture_def: fixtures.FixtureDef[object]) -> None:
...
         if verbose <= 0 and argname.startswith("_"):
             return
-        if verbose > 0:
-            bestrel = get_best_relpath(fixture_def.func)
-            funcargspec = f"{argname} -- {bestrel}"
-        else:
-            funcargspec = argname
-        tw.line(funcargspec, green=True)
+        bestrel = get_best_relpath(fixture_def.func)
+        tw.write(f"{argname}", green=True)
```
rahul-kumi, line 1432: @The-Compiler, I noticed your suggestion to use `argname` in this line as a direct variable reference instead of using it as a formatted string. I am not able to find that comment to reference it.

I think that would be a good thing to do and keeps things consistent. Please let me know if this small change can go in as a separate PR.

PR #8628, opened 2021-05-04
Allow prereleases from prepare-release workflow

### LINKED ISSUE #7551: Update release-on-comment script to handle release candidates
As discussed in https://github.com/pytest-dev/pytest/pull/7548#issuecomment-664988552

nicoddemus:
Unfortunately untested, as it is only available once it lands on `main`.

Fix #7551

### DIFF
--- a/.github/workflows/prepare-release-pr.yml
+++ b/.github/workflows/prepare-release-pr.yml
@@ -11,6 +11,10 @@ on:
         description: 'Major release? (yes/no)'
         required: true
         default: 'no'
+      prerelease:
+        description: 'Prerelease (ex: rc1). Leave empty if not a pre-release.'
+        required: true
+        default: ''
 
 # Set permissions at the job level.
 permissions: {}
@@ -41,9 +45,9 @@ jobs:
     - name: Prepare release PR (minor/patch release)
       if: github.event.inputs.major == 'no'
       run: |
-        tox -e prepare-release-pr -- ${{ github.event.inputs.branch }} ${{ github.token }}
+        tox -e prepare-release-pr -- ${{ github.event.inputs.branch }} ${{ github.token }} --prerelease '${{ github.event.inputs.prerelease }}'
 
     - name: Prepare release PR (major release)
       if: github.event.inputs.major == 'yes'
       run: |
-        tox -e prepare-release-pr -- ${{ github.event.inputs.branch }} ${{ github.token }} --major
+        tox -e prepare-release-pr -- ${{ github.event.inputs.branch }} ${{ github.token }} --major --prerelease '${{ github.event.inputs.prerelease }}'
--- a/scripts/prepare-release-pr.py
+++ b/scripts/prepare-release-pr.py
@@ -46,14 +46,16 @@ def login(token: str) -> Repository:
     return github.repository(owner, repo)
 
 
-def prepare_release_pr(base_branch: str, is_major: bool, token: str) -> None:
+def prepare_release_pr(
+    base_branch: str, is_major: bool, token: str, prerelease: str
+) -> None:
     print()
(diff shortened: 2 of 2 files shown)


.github/workflows/prepare-release-pr.yml
```diff
@@ -11,6 +11,10 @@ on:
         description: 'Major release? (yes/no)'
         required: true
         default: 'no'
+      prerelease:
+        description: 'Prerelease (ex: "rc1", including the quotes)'
+        required: true
+        default: '""'
```
asottile, line 17: the quoting on this seems weird -- I would expect `default: ''` to be better (coupled with a suggested change below)

.github/workflows/prepare-release-pr.yml
```diff
@@ -34,9 +38,9 @@ jobs:
     - name: Prepare release PR (minor/patch release)
       if: github.event.inputs.major == 'no'
       run: |
-        tox -e prepare-release-pr -- ${{ github.event.inputs.branch }} ${{ secrets.chatops }}
+        tox -e prepare-release-pr -- ${{ github.event.inputs.branch }} ${{ secrets.chatops }} --prerelease ${{ github.event.inputs.prerelease }}
```
asottile, line 41: and `--prerelease '${{ github.event.inputs.prerelease }}'`

The-Compiler:
@nicoddemus What are your plans on this? If we want a 7.0.0rc1 (or 6.3.0rc1), we should probably get this in first.

nicoddemus:
Argh completely missed this one.

I'm not sure I will be able to get to this today (I'll do my best), but if anybody wants to take this please go ahead. 👍

.github/workflows/prepare-release-pr.yml
nicoddemus: Done, thanks!

.github/workflows/prepare-release-pr.yml
nicoddemus: Done, thanks!

nicoddemus:
Applied @asottile's suggestions. 👍

.github/workflows/prepare-release-pr.yml
```diff
@@ -11,6 +11,10 @@ on:
         description: 'Major release? (yes/no)'
         required: true
         default: 'no'
+      prerelease:
+        description: 'Prerelease (ex: "rc1", including the quotes)'
```
The-Compiler, line 15: Is the "including the quotes" still accurate now?

nicoddemus: Nope, thanks for catching it. 👍

nicoddemus: Fixed!

The-Compiler (approved):
LGTM now! Fingers crossed that it won't fail in some way which is totally obvious in hindsight but we all missed somehow :see_no_evil:

nicoddemus:
> Fingers crossed that it won't fail in some way which is totally obvious in hindsight but we all missed somehow

Heh ditto! 😁

The-Compiler:
Hmm wait, I think this is only a part of the equation, though? At least according to [RELEASING.rst](https://github.com/pytest-dev/pytest/blob/main/RELEASING.rst) we're still triggering @pytestbot via a release comment (rather than triggering the workflow manually), so [release-on-comment.py](https://github.com/pytest-dev/pytest/blob/main/scripts/release-on-comment.py) would need an update to support something like a `@pytestbot please prepare [major] prerelease from ...` too, no?

nicoddemus:
> t least according to RELEASING.rst we're still triggering @pytestbot via a release comment (rather than triggering the workflow manually)

Those are not up to date, we have made the last two releases (or so) with this new workflow, I think we can remove the `release-on-comment` one as obsolete. 👍

PR #8647, opened 2021-05-08
[6.1.x] Merge pull request #8540 from hauntsaninja/assert310

erikkemperman:
(cherry picked from commit af31c60db1dc4e513f47aabf8f6e844b23afd35f)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -260,6 +260,7 @@ Sankt Petersbug
 Segev Finer
 Serhii Mozghovyi
 Seth Junot
+Shantanu Jain
 Simon Gomizelj
 Simon Kerr
 Skylar Downes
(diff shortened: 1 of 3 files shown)


asottile:
sorry, we don't backport to releases prior to the latest minor version -- this has been released in 6.2.4

erikkemperman:
Fair enough. I thought I’d give it a shot — I’m running into this bug with mypy’s unit tests, and I expect they will want to continue supporting python 3.5 and thus depend on pytest<6.2.0...

asottile:
presumably you only need pytest<6.2.0 for python3.5 and can use modern pytest elsewhere?

erikkemperman:
Thanks for taking a look!

Issue #8482, opened 2021-03-23
`pytest` hangs when `RecursionError` happens for `MagicMock`


jiasli:
A demo for an infinite recursion:

```py
# test_func.py

from unittest import mock
depth = 0


def func(obj):
    global depth
    depth = depth + 1
    print(depth)
    if depth == 1000:
        raise Exception
    return func(obj.attr)  # Infinite loop


def test_func():
    func(mock.MagicMock())


test_func()
```

```
$ python test_func.py
...
983
984
985
Traceback (most recent call last):
  File "test_func.py", line 20, in <module>
    test_func()
  ...
    if type(value) is cls:
RecursionError: maximum recursion depth exceeded while calling a Python object
```

But `pytest` just hangs:

```
$ pytest test_func.py --capture no
...
940
941
942
```

This happens on both Linux and Windows.

```
> pytest -V
pytest 6.2.2
```

The-Compiler:
Looks like it hangs here:

```
Current thread 0x00007f34402db740 (most recent call first):
  File "/usr/lib/python3.9/unittest/mock.py", line 2436 in __init__
  File "/usr/lib/python3.9/unittest/mock.py", line 1131 in _increment_mock_call
  File "/usr/lib/python3.9/unittest/mock.py", line 1092 in __call__
  File "?", line 1 in <module>
  File ".../pytest/src/_pytest/_code/code.py", line 166 in eval
  File ".../pytest/src/_pytest/_code/code.py", line 425 in recursionindex
  File ".../pytest/src/_pytest/_code/code.py", line 884 in _truncate_recursive_traceback
  File ".../pytest/src/_pytest/_code/code.py", line 851 in repr_traceback
  File ".../pytest/src/_pytest/_code/code.py", line 923 in repr_excinfo
  File ".../pytest/src/_pytest/_code/code.py", line 665 in getrepr
  File ".../pytest/src/_pytest/nodes.py", line 437 in _repr_failure_py
  File ".../pytest/src/_pytest/nodes.py", line 509 in repr_failure
  File ".../pytest/src/_pytest/runner.py", line 395 in pytest_make_collect_report
  File ".../pytest/.venv/lib/python3.9/site-packages/pluggy/callers.py", line 187 in _multicall
  File ".../pytest/.venv/lib/python3.9/site-packages/pluggy/manager.py", line 84 in <lambda>
[...]

PR #8651, opened 2021-05-09
Implement a linear algorithm in the `recursionindex` function

kri-k:
See https://github.com/pytest-dev/pytest/issues/8482#issuecomment-835841190

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -1,3 +1,4 @@
+import collections
 import inspect
 import re
 import sys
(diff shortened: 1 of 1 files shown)


bluetech:
Hmm, that's interesting.

If I understand correctly, the algorithm relies on the fact that the last frame is part of the loop:

> Finding beginning of the loop: such n that frames[n] equals to the last_frame

But I don't think this holds. Suppose the loop is:

    a -> b -> c -> a -> b -> c -> ....

Now suppose that `a`, before calling `b`, does some other work, i.e. the call stack progresses like this:

```
a
a -> aa
a -> aa -> aaa
a -> aa
a
a -> b
<loop>
```

Then it might happen that the recursion limit is hit right when we're at `aa` or `aaa` which are not part of the loop.

kri-k:
> Now suppose that `a`, before calling `b`, does some other work
>...
> Then it might happen that the recursion limit is hit right when we're at `aa` or `aaa` which are not part of the loop.

If I understand correctly, you mean that if traceback will have the following frames `a -> aa -> aaa` when `RecursionError` happens, then the proposed algorithm will not be able to find the cycle.

It seems to me that this behavior is correct, because at the time of the exception raised, the program has not yet reached the execution of an infinite loop. So, we should not show warning `!!! Recursion detected (same locals & position)` (also, maybe we should clarify this message and write about _infinite_ recursion?)

And, in this situation, I believe, we can't predict existence of the loop `a -> b -> c -> a ...` without explicitly further code execution :)

bluetech:
I'm thinking of something like this:

```py
def a():
  aaa(20)
  b()

def aaa(n):
   if n == 0:
       return
   aaa(n - 1)

def b():
   c()

def c():
   a()
```

The loop here is `a -> b -> c -> a`, but the `RecursionError` is very likely to hit at something like `a -> b -> c -> a -> ... -> a -> aaa -> aaa -> aaa -> aaa`, and the `last_frame` here is not part of the loop.

I didn't try it though so might be missing some detail.

kri-k:
> The loop here is `a -> b -> c -> a`, but the `RecursionError` is very likely to hit at something like `a -> b -> c -> a -> ... -> a -> aaa -> aaa -> aaa -> aaa`, and the `last_frame` here is not part of the loop.

Oh, I get it, thanks.

This particular example will not work because algorithm will group all frames by [position](https://github.com/pytest-dev/pytest/pull/8651/files#diff-ac2099a172465905236568f8c175328fb75ebc1b80ce39f4799e61373d355418R422) first.

But you're right, here is another example that fails my code:
```python
def b(n):
    a(n)  # position A

def a(n):
    if n == 0:
        b(1)
        b(0)
    if n == 3:
        return
    b(n + 1)
```

Here, for position `A`, we can have the following list of frames: `a(0) -> a(0) -> ... -> a(0) -> a(1) -> a(2) -> a(3)`.

PR #8658, opened 2021-05-10
Replaced if-else statements with ternary expression

zkneupper:
* Replaced if-else statements with ternary expressions where appropriate


The-Compiler (approved):
Looks great and I don't spot any behavioral changes. Will wait a bit before merging though, in case someone else wants to take a look as well.

RonnyPfannschmidt (approved):
looks good to me :+1:

The-Compiler:
Thanks @zkneupper!

PR #8659, opened 2021-05-11
Open log file in context manager

zkneupper:
* Open the log file safely using a context manager (using a with block).
  * This will make sure that the file closed even if an exception is raised


The-Compiler (approved):
Good catch, thanks!

PR #8660, opened 2021-05-11
Replace some for loops, and other minor changes

zkneupper:
* Replace for loop using the `in` operator
* Replace for loop with a generator expression inside any()
* Replace for loop with a dictionary comprehension
* Use list comprehension"
* Simplify arguments for range()


The-Compiler (approved):
Looks great and I don't spot any behavioral changes. Will wait a bit before merging though, in case someone else wants to take a look as well.

Issue #8663, opened 2021-05-12
Python 3.10: TestTrialUnittest.test_trial_error fails due to DeprecationWarning in twisted


hroncok:
This is Python 3.10.0b1 on Fedora Linux.

On branch main @ 045ad5ac2d0bf6d321e93e2f3d9dc429301900bf:

```
$ tox -e py310-unittestextras -- testing/test_unittest.py
...
py310-unittestextras inst-nodeps: ...pytest/.tox/.tmp/package/3/pytest-6.3.0.dev440+g045ad5ac2.tar.gz
py310-unittestextras installed: argcomplete==1.12.3,asynctest==0.13.0,attrs==21.2.0,Automat==20.2.0,certifi==2020.12.5,chardet==4.0.0,constantly==15.1.0,elementpath==2.2.2,hyperlink==21.0.0,hypothesis==6.12.0,idna==2.10,incremental==21.3.0,iniconfig==1.1.1,mock==4.0.3,nose==1.3.7,packaging==20.9,pluggy==0.13.1,py==1.10.0,Pygments==2.9.0,pyparsing==2.4.7,pytest @ file://...pytest/.tox/.tmp/package/3/pytest-6.3.0.dev440%2Bg045ad5ac2.tar.gz,requests==2.25.1,six==1.16.0,sortedcontainers==2.3.0,toml==0.10.2,Twisted==21.2.0,urllib3==1.26.4,xmlschema==1.6.2,zope.interface==5.4.0
py310-unittestextras run-test-pre: PYTHONHASHSEED='761603216'
py310-unittestextras run-test: commands[0] | pytest testing/test_unittest.py
============================= test session starts ==============================
platform linux -- Python 3.10.0b1, pytest-6.3.0.dev440+g045ad5ac2, py-1.10.0, pluggy-0.13.1
cachedir: .tox/py310-unittestextras/.pytest_cache
rootdir: ...pytest, configfile: pyproject.toml
plugins: hypothesis-6.12.0
collected 69 items

testing/test_unittest.py ........................F...................... [ 68%]
....................s.                                                   [100%]

=================================== FAILURES ===================================
______________________ TestTrialUnittest.test_trial_error ______________________

self = <test_unittest.TestTrialUnittest object at 0x7f61f20d9ae0>
pytester = <Pytester PosixPath('/tmp/pytest-of-churchyard/pytest-6/test_trial_error0')>

    def test_trial_error(self, pytester: Pytester) -> None:
[...]

PR #8664, opened 2021-05-12
Ignore DeprecationWarnings in test_trial_error

hroncok:
Fixes https://github.com/pytest-dev/pytest/issues/8663

### DIFF
--- a/testing/test_unittest.py
+++ b/testing/test_unittest.py
@@ -570,7 +570,9 @@ class TestTrialUnittest:
                 # will crash both at test time and at teardown
         """
         )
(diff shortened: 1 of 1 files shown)


Zac-HD (approved):
Thanks @hroncok, looks good to me 😄

bluetech (approved):
Looks like it's still not fixed on [twisted master](https://github.com/twisted/twisted/blob/e0c069b429a66343e3e6c7ef19df512710f4775c/src/twisted/python/threadable.py#L104), they should probably fix it, however no reason for us to fail on twisted deprecation warnings.

PR #8674, opened 2021-05-14
fix: type `Markable` is reduced to `Any` by mypy

terencehonles:
The following code: 

https://github.com/pytest-dev/pytest/blob/364bbe42fc02d435e6d94d13ad3accc7e00875fd/src/_pytest/mark/structures.py#L269

when using `reveal_type` with mypy resolves to `Any` which is not what we want, but I believe the type variable can probably be simplified to the following and still be correct:

```python
Markable = TypeVar("Markable", bound=Callable)
```

The previous definition defined the return of the callable and did accept any class. I don't believe we care about the return of the callable and classes are also callable so this should be sufficient and still correct.

### DIFF
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -3,6 +3,7 @@ import inspect
 import warnings
 from typing import Any
 from typing import Callable
+from typing import cast
 from typing import Collection
 from typing import Iterable
 from typing import Iterator
@@ -264,9 +265,9 @@ class Mark:
 
 
 # A generic parameter designating an object to which a Mark may
-# be applied -- a test function (callable) or class.
+# be applied -- a test function (callable) or class (also callable).
 # Note: a lambda is not allowed, but this can't be represented.
-Markable = TypeVar("Markable", bound=Union[Callable[..., object], type])
+Markable = TypeVar("Markable", bound=Callable[..., Any])
 
 
 @attr.s(init=False, auto_attribs=True)
@@ -344,25 +345,22 @@ class MarkDecorator:
         mark = Mark(self.name, args, kwargs, _ispytest=True)
         return MarkDecorator(self.mark.combined_with(mark), _ispytest=True)
 
-    # Type ignored because the overloads overlap with an incompatible
-    # return type. Not much we can do about that. Thankfully mypy picks
-    # the first match so it works out even if we break the rules.
     @overload
-    def __call__(self, arg: Markable) -> Markable:  # type: ignore[misc]
+    def __call__(self, arg: Markable) -> Markable:
         pass
 
     @overload
-    def __call__(self, *args: object, **kwargs: object) -> "MarkDecorator":
+    def __call__(self, *args: Any, **kwargs: Any) -> "MarkDecorator":
         pass
 
-    def __call__(self, *args: object, **kwargs: object):
+    def __call__(self, *args: Any, **kwargs: Any) -> Union[Markable, "MarkDecorator"]:
         """Call the MarkDecorator."""
         if args and not kwargs:
             func = args[0]
             is_class = inspect.isclass(func)
             if len(args) == 1 and (istestfunc(func) or is_class):
                 store_mark(func, self.mark)
-                return func
+                return cast(Markable, func)
         return self.with_args(*args, **kwargs)
 
 


terencehonles:
@bluetech thoughts? It looks like you added this code. You can check that this resolves to Any with mypy:

```python
from typing import Callable, TypeVar, Union

Markable = TypeVar("Markable", bound=Union[Callable[..., object], type])
reveal_type(Markable)
```

terencehonles:
Sorry, I probably should have just opened this as an issue. This might be an issue with the overload for @parameterize that I'm seeing. This doesn't seem to be accurate if I expand my test case:

```python
def markable(arg: Callable[[Dict[str, str]], Path]) -> None:                                                                 
    pass

Markable = TypeVar("Markable", bound=Union[Callable[..., object], type])

def test(m: Markable) -> Markable:                                            
    return m                                                               

reveal_type(test(markable))  #    def markable(arg: Callable[[Dict[str, str]], Path]):
```

Which is mostly correct, but missing the return type

terencehonles:
It looks like that is just because the return type is `-> None` so mypy omits it when revealing. So I'll look into the overload, but this looks like it _can_ be simplified and the return type can be widened to Any.

terencehonles:
It looks like it is probably the ignored code:

https://github.com/pytest-dev/pytest/blob/364bbe42fc02d435e6d94d13ad3accc7e00875fd/src/_pytest/mark/structures.py#L347-L352

I was able to add the following to the end of this PR and check that it resolved the types correctly:

```python
if TYPE_CHECKING:
    from typing import Dict

    @MARK_GEN.parametrize('test', [1])
    def testing(x: Dict[str, int]) -> List[int]:
        return list(x.values())
    reveal_type(testing)
```

```
src/_pytest/mark/structures.py:589: note: Revealed type is 'def (x: builtins.dict[builtins.str, builtins.int]) -> builtins.list[builtins.int]'
```

terencehonles:
Well I'm still having issues, and for me it looks like it's related to using a named tuple and I'm working on reporting that to mypy, but I believe it's probably because of this https://github.com/python/typing/issues/253#issuecomment-235358878

nicoddemus:
Thanks @terencehonles!

Leaving this to our in-house typing expert @bluetech. 😁

src/_pytest/mark/structures.py
```diff
@@ -266,7 +267,7 @@ def combined_with(self, other: "Mark") -> "Mark":
 # A generic parameter designating an object to which a Mark may
 # be applied -- a test function (callable) or class.
```
bluetech, line 268:
```suggestion
# be applied -- a test function (callable) or class (also callable).
```

src/_pytest/mark/structures.py
```diff
@@ -266,7 +267,7 @@ def combined_with(self, other: "Mark") -> "Mark":
...
 # Note: a lambda is not allowed, but this can't be represented.
-Markable = TypeVar("Markable", bound=Union[Callable[..., object], type])
+Markable = TypeVar("Markable", bound=Callable[..., Any])
```
bluetech, line 270: Any reason to change the `object` to `Any` here?

src/_pytest/mark/structures.py
```diff
@@ -344,25 +345,22 @@ def with_args(self, *args: object, **kwargs: object) -> "MarkDecorator":
         mark = Mark(self.name, args, kwargs, _ispytest=True)
         return MarkDecorator(self.mark.combined_with(mark), _ispytest=True)
 
-    # Type ignored because the overloads overlap with an incompatible
-    # return type. Not much we can do about that. Thankfully mypy picks
-    # the first match so it works out even if we break the rules.
     @overload
-    def __call__(self, arg: Markable) -> Markable:  # type: ignore[misc]
```
bluetech, line 351: I wonder how come this error doesn't happen anymore. The overloads still overlap with an incompatible return type.

src/_pytest/mark/structures.py
```diff
@@ -344,25 +345,22 @@ def with_args(self, *args: object, **kwargs: object) -> "MarkDecorator":
...
+    def __call__(self, arg: Markable) -> Markable:
         pass
 
     @overload
-    def __call__(self, *args: object, **kwargs: object) -> "MarkDecorator":
+    def __call__(self, *args: Any, **kwargs: Any) -> "MarkDecorator":
```
bluetech, line 353: Any reason to change `object` to `Any` here?

src/_pytest/mark/structures.py
```diff
@@ -344,25 +345,22 @@ def with_args(self, *args: object, **kwargs: object) -> "MarkDecorator":
...
         pass
 
-    def __call__(self, *args: object, **kwargs: object):
+    def __call__(self, *args: Any, **kwargs: Any) -> Union[Markable, "MarkDecorator"]:
```
bluetech, line 356: `Markable` is an unbound `TypeVar` here, which usually doesn't make sense. Since the typing in covered by the overloads I chose to omit the return type here.

bluetech (commented):
Hi @terencehonles,

Trying to follow the issue here, I'm not quite sure I got it. Can you maybe add a "test case" to the file `tests/typing_checks.py`? This is just a file which mypy checks. A "test case" in this file is simply a piece of code that would fail to type-check before the change, and type-check correctly after. It is not executed.

IIUC, your change makes `Markable` just a `Callable` which allows it to unify with the other overload and thus somehow make mypy work, however I think that `ignore[misc]` is actually correct and if it *doesn't* trigger then something is off...

src/_pytest/mark/structures.py
terencehonles: I just double checked an `None` is also an instance of `Any`, but it seems odd to specify `object` when you mean you don't care. I can leave it as is if `object` is preferred and it doesn't trip up type checkers.

src/_pytest/mark/structures.py
terencehonles: I believe it changed when I changed the below to use `Any` instead of `object`, but I can double check that and see if I may have bumped the mypy version I was testing with.

src/_pytest/mark/structures.py
terencehonles: As mentioned above https://github.com/pytest-dev/pytest/pull/8674#discussion_r632996374, IIRC changing this to `Any` fixed how mypy was resolving the `@overloads` (but I can double check it wasn't the `Union` below)

src/_pytest/mark/structures.py
terencehonles: Is it unbound? I figured mypy would use the overload, but I can adjust this so there is less overlap. I can force non keyword arguments but without knowing at least one required keyword argument I don't know how to tell mypy that at least one has to be provided. to force the `MarkDecorator` path.

src/_pytest/mark/structures.py
terencehonles: Yup, reverting this line only causes mypy to complain about the overlap again, I'm using:

mypy 0.812
Python 3.8.9

src/_pytest/mark/structures.py
bluetech: I'll try to explain:

Both types are compatible with everything, i.e. `x: object = <anything>` and `x: Any = <anything>` both work. The difference is that `x: Any` allows doing anything *with* the object, while `x: object` allows *nothing*. So `object` is completely safe - if you want to do anything with it besides generic object capabilities like `str(x)`, you have to dynamically check its type with `isinstance(x, <type>)`. `Any` however is completely unsafe, and strict typing would conceptually disallow it.

So when to use which? If pytest returns some unknown value to the user, I don't want to force them to do `isinstance` checks, which are an headache, so returning `Any` makes sense. But if pytest *receives* such a value, it should use the safe variant.

In this specific case, the return type is just used in a `TypeVar` which is bound to some concrete value anyway, so whether it's `object` or `Any` shouldn't actually matter. So using the safer `object` is nicer.

src/_pytest/mark/structures.py
terencehonles: Just to follow up on the last part of your comment... As you have it mypy (possibly any type checker) will read this as `-> Any`. I had thought as you are on a different codebase and I noticed that mypy wasn't checking the function at all (I had left off the argument signature too), I then added the argument signature and tested that I could return anything even outside the bounds of the `@overload`s. This might have been fixed in mypy, but it's best to be explicit about your return type. When I had added the return annotation it required casting the return in the `Markable` path since it was now checking the return was not `Any`.

bluetech: When a function has overloads defined, the typing of the implementation function is only used for the implementation -- it is ignored in favor of the overloads everywhere else.

src/_pytest/mark/structures.py
terencehonles: To follow up more after reviewing the test file you linked to and the issues it links to, it looks like you've seen that using `object` is problematic for mypy (i.e. https://github.com/pytest-dev/pytest/pull/7496) and I'd probably suggest converting everything using `object` to `Any` to not try to convince `mypy` to associate a type since that's what you're saying when you use `object` (and I'm not sure why mypy doesn't treat `object` the same way it treats `Any`)

terencehonles:
Thanks for the review @bluetech, I'm responding to your overall comment, but replies to the line specific comments are inline. 

> Trying to follow the issue here, I'm not quite sure I got it. Can you maybe add a "test case" to the file `tests/typing_checks.py`? This is just a file which mypy checks. A "test case" in this file is simply a piece of code that would fail to type-check before the change, and type-check correctly after. It is not executed.

Thanks for pointing me to this file, it was useful to see the other issues that had come up and as mentioned inline I think we should probably be changing any use of `object` to `Any` as mentioned in this comment https://github.com/pytest-dev/pytest/pull/8674#discussion_r632998160

As far as adding a test case, I understand the PR had a few comment that probably made it harder to follow, but it looks like the bug I was seeing is in mypy as mentioned in this comment https://github.com/pytest-dev/pytest/pull/8674#issuecomment-840948058 and not in pytest. Putting the example code at the bottom of the file without my changes:

```python
if TYPE_CHECKING:
    from typing import Dict

    @MARK_GEN.parametrize('test', [1])
    def testing(x: Dict[str, int]) -> List[int]:
        return list(x.values())
    reveal_type(testing)
```

It properly revealed the correct types. So this change has been reduced to not fixing a bug in pytest, but simplifying the type variables and removing the mypy error.

> IIUC, your change makes `Markable` just a `Callable` which allows it to unify with the other overload and thus somehow make mypy work, however I think that `ignore[misc]` is actually correct and if it _doesn't_ trigger then something is off...

Yes, I think this is mostly correct, but I'm not sure I agree about the last part :sweat_smile:, looking for possible documentation I believe the difference is mentioned here https://mypy.readthedocs.io/en/stable/more_types.html#type-checking-the-variants

> So in this example, the `int` argument in the first variant is a subtype of the `object` argument in the second, yet the ``int return type is not a subtype of `str`. Both conditions are true, so mypy will correctly flag `unsafe_func` as being unsafe.

src/_pytest/mark/structures.py
terencehonles: Yes, that's correct, but shouldn't the function return be type checked so that future editors don't accidentally make a mistake?

src/_pytest/mark/structures.py
terencehonles: I'm not sure I agree with which to use based on what I have seen. And based on my overall comment https://github.com/pytest-dev/pytest/pull/8674#issuecomment-841715808 mypy _does_ read into `object` too much.

I see you're basically describing what's here https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object, but what you are trying to do is more akin to TypeScripts `unknown` and `any`. Although I understand what you are trying to achieve, I don't think that always works in practice, or at least it touches more rough edges of mypy. I definitely agree with what you're trying to accomplish, but I think since mypy isn't there yet and type checking is still relatively young using `Any` is safer (rough edges not for the users) than using `object` unless there is a PEP which will either add an `unknown` type a la TypeScript.

src/_pytest/mark/structures.py
bluetech: I think best we can do is add a comment.

src/_pytest/mark/structures.py
bluetech: OK, so using `Any` here is OK if it fixes some problem.

bluetech:
@terencehonles So what's left from me is:

- I'd still like to see the last unresolved comment fixed (removing the return type from the implemention `def`)
- If you could add a regression test (doesn't type check before, type checks after) to the `tests/typing_checks.py` file that would also be great.

bluetech:
I'll close this now since I'm not sure it's an improvement, but feel free to reopen if you come back to it.

PR #8695, opened 2021-05-23
argparsing: export Parser and OptionGroup for typing purposes

bluetech:
`Parser` is used by many plugins and custom hooks. `OptionGroup` is
exposed by the `parser.addgroup` API.

The constructors of both are marked private, they are not meant to be
constructed directly.

Refs #7469.


nicoddemus (approved):
Great work! 🚀

PR #8704, opened 2021-05-26
Make the test class instance behaviour clearer

offbyone:
In the "Getting Started" doc, the test class instance example for
instance sharing doesn't actually demonstrate anything about the
reinstantiation of the class. This change shows clearly how the instance
data isn't retained between test runs.


nicoddemus (approved):
Hi @offbyone, thanks for the contribution!

PR #8705, opened 2021-05-26
Fix cyclic links in goodpractices.rst

MapleCCC:
It makes more sense to not link to the section itself, which causes a confusing cycle.


nicoddemus (approved):
Definitely, thanks for catching that!

PR #8708, opened 2021-05-27
Update index.rst

MapleCCC:


MapleCCC:
Oh, wait, I did not read carefully and ended up opening a pull request against the `6.2.x` branch. Should we also port the change to the `main` branch? I am not very clear which branch is the upstream-y development branch here.

The-Compiler:
Yup, I noticed this just after merging as well but didn't have the time to comment right away. From a quick look it looks like those changes are already in main? Doesn't hurt to have them here as well though, given that the default docs are based on this branch.

PR #8725, opened 2021-06-03
Add CHANGELOG entry for #8227

nicoddemus:
This was missing in the original PR.


asottile:
actually, we should be able to revert this now -- the traceback frames have been fixed upstream -- I'll prep a PR with a revert

asottile:
here's the revert: #8729

PR #8726, opened 2021-06-03
Stop using exception.message

hexagonrecursion:
https://bugzilla.mozilla.org/show_bug.cgi?id=1713825#c0

> In PEP 352, exception.message was deprecated (now, you can get the string from of an exception by simply doing str(exception). As of Python 3.0, exception.message was dropped, and attempting to access it causes an error.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,6 +25,7 @@ Andrea Cimatoribus
 Andreas Motl
 Andreas Zeidler
 Andrew Shapton
+Andrey Bienkowski
 Andrey Paramonov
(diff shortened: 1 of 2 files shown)


The-Compiler:
@hexagonrecursion Did you intend to close this?

graingert:
it looks like the exception actually assigns the deprecated `self.message` attr:

https://github.com/pytest-dev/pytest/blob/3c451751af3f53fdb90aa6ac4afa2329222d54d2/src/_pytest/mark/expression.py#L57-L69

PR #8749, opened 2021-06-10
updated type hints for caplog.at_level to match caplog.set_level

beniwohli:
The underlying logging API accepts the log level both as int and
as string.


RonnyPfannschmidt (approved):

nicoddemus (approved):
Thanks @beniwohli!

PR #8752, opened 2021-06-11
pygments themes are customizable

### LINKED ISSUE #7132: Ability to choose syntax highlighting style by configuration
This issue is somewhat related to #6906.
Syntax highlighting is a great feature that I have been waiting for a long time. Thank you for that!
However, the pygment style that is used by default does not really match my terminal color configuration. Colors have poor contrast, and in the end, I cannot really benefit from syntax highlighting.

I would love to be able to choose the pygments theme from CLI options, or in a configuration file.

Another idea would be to configure the pygments theme with environment variables. This is because, pytest configuration generally depends on the project you are testing, and the configuration file is shipped in the project repository. However colors configuration will be dependent on the developper setup and should probably better remain unversioned. Environment variables allows that.

terminal

What do you think?

azmeuk:
This patchs allows users to set two environment variables (*PYTEST_THEME* and *PYTEST_THEME_MODE*) to customize the pygments theme to use. It fixes #7132

I know some people were interested in [rich](https://github.com/willmcgugan/rich) and suggested to wait and see, but this solution is so easy and unintrusive I think it might be worth right now.

What do you think?

### DIFF
--- a/changelog/7132.feature.rst
+++ b/changelog/7132.feature.rst
@@ -1,0 +1,1 @@
+Added two environment variables :envvar:`PYTEST_THEME` and :envvar:`PYTEST_THEME_MODE` to let the users customize the pygments theme used.
--- a/doc/en/reference/customize.rst
+++ b/doc/en/reference/customize.rst
@@ -239,3 +239,11 @@ Builtin configuration file options
 ----------------------------------------------
 
 For the full list of options consult the :ref:`reference documentation <ini options ref>`.
+
+Syntax highlighting theme customization
+---------------------------------------
+
+The syntax highlighting themes used by pytest can be customized using two environment variables:
+
+- :envvar:`PYTEST_THEME` sets a `pygment style <https://pygments.org/docs/styles/>`_ to use.
+- :envvar:`PYTEST_THEME_MODE` sets this style to *light* or *dark*.
--- a/doc/en/reference/reference.rst
+++ b/doc/en/reference/reference.rst
@@ -1075,6 +1075,14 @@ Contains comma-separated list of modules that should be loaded as plugins:
 
     export PYTEST_PLUGINS=mymodule.plugin,xdist
 
+.. envvar:: PYTEST_THEME
+
+Sets a `pygment style <https://pygments.org/docs/styles/>`_ to use for the code output.
+
+.. envvar:: PYTEST_THEME_MODE
+
+Sets the :envvar:`PYTEST_THEME` to be either *dark* or *light*.
+
 .. envvar:: PY_COLORS
 
 When set to ``1``, pytest will use color in terminal output.
--- a/src/_pytest/_io/terminalwriter.py
+++ b/src/_pytest/_io/terminalwriter.py
@@ -195,16 +195,39 @@ class TerminalWriter:
 
     def _highlight(self, source: str) -> str:
         """Highlight the given source code if we have markup support."""
+        from _pytest.config.exceptions import UsageError
+
         if not self.hasmarkup or not self.code_highlight:
             return source
         try:
             from pygments.formatters.terminal import TerminalFormatter
             from pygments.lexers.python import PythonLexer
             from pygments import highlight
+            import pygments.util
         except ImportError:
             return source
         else:
-            highlighted: str = highlight(
-                source, PythonLexer(), TerminalFormatter(bg="dark")
-            )
-            return highlighted
+            try:
+                highlighted: str = highlight(
+                    source,
+                    PythonLexer(),
+                    TerminalFormatter(
+                        bg=os.getenv("PYTEST_THEME_MODE", "dark"),
+                        style=os.getenv("PYTEST_THEME"),
+                    ),
+                )
+                return highlighted
+            except pygments.util.ClassNotFound:
+                raise UsageError(
+                    "PYTEST_THEME environment variable had an invalid value: '{}'. "
+                    "Only valid pygment styles are allowed.".format(
+                        os.getenv("PYTEST_THEME")
+                    )
+                )
+            except pygments.util.OptionError:
+                raise UsageError(
+                    "PYTEST_THEME_MODE environment variable had an invalid value: '{}'. "
+                    "The only allowed values are 'dark' and 'light'.".format(
+                        os.getenv("PYTEST_THEME_MODE")
+                    )
+                )
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -2408,6 +2408,60 @@ class TestCodeHighlight:
             )
         )
 
+    def test_code_highlight_custom_theme(
+        self, pytester: Pytester, color_mapping, monkeypatch: MonkeyPatch
+    ) -> None:
+        pytester.makepyfile(
+            """
+            def test_foo():
+                assert 1 == 10
+        """
+        )
+        monkeypatch.setenv("PYTEST_THEME", "solarized-dark")
+        monkeypatch.setenv("PYTEST_THEME_MODE", "dark")
+        result = pytester.runpytest("--color=yes")
+        result.stdout.fnmatch_lines(
+            color_mapping.format_for_fnmatch(
+                [
+                    "    {kw}def{hl-reset} {function}test_foo{hl-reset}():",
+                    ">       {kw}assert{hl-reset} {number}1{hl-reset} == {number}10{hl-reset}",
+                    "{bold}{red}E       assert 1 == 10{reset}",
+                ]
+            )
+        )
+
+    def test_code_highlight_invalid_theme(
+        self, pytester: Pytester, color_mapping, monkeypatch: MonkeyPatch
+    ) -> None:
+        pytester.makepyfile(
+            """
+            def test_foo():
+                assert 1 == 10
+        """
+        )
+        monkeypatch.setenv("PYTEST_THEME", "invalid")
+        result = pytester.runpytest_subprocess("--color=yes")
+        result.stderr.fnmatch_lines(
+            "ERROR: PYTEST_THEME environment variable had an invalid value: 'invalid'. "
+            "Only valid pygment styles are allowed."
+        )
+
+    def test_code_highlight_invalid_theme_mode(
+        self, pytester: Pytester, color_mapping, monkeypatch: MonkeyPatch
+    ) -> None:
+        pytester.makepyfile(
+            """
+            def test_foo():
+                assert 1 == 10
+        """
+        )
+        monkeypatch.setenv("PYTEST_THEME_MODE", "invalid")
+        result = pytester.runpytest_subprocess("--color=yes")
+        result.stderr.fnmatch_lines(
+            "ERROR: PYTEST_THEME_MODE environment variable had an invalid value: 'invalid'. "
+            "The only allowed values are 'dark' and 'light'."
+        )
+
 
 def test_raw_skip_reason_skipped() -> None:
     report = SimpleNamespace()
--- a/AUTHORS
+++ b/AUTHORS
@@ -106,6 +106,7 @@ Edson Tadeu M. Manoel
 Eduardo Schettino
 Eli Boyarski
 Elizaveta Shashkova
+Éloi Rivard
 Endre Galaczi
 Eric Hunsberger
 Eric Liu


azmeuk:
I can. What do you expect? Catching the pygments exception and raise a custom exception with a nicer error message when the environment variables have invalid values?

doc/en/reference/customize.rst
```diff
@@ -239,3 +239,11 @@ Builtin configuration file options
...
 
 For the full list of options consult the :ref:`reference documentation <ini options ref>`.
+
+Syntax highlighting theme customization
+---------------------------------------
+
+The syntax highlighting themes used by pytest can be customized using two environment variables:
+
+- **PYTEST_THEME** sets a `pygment style <https://pygments.org/docs/styles/>`_ to use.
+- **PYTEST_THEME_MODE** sets this style to *light* or *dark*
```
nicoddemus, line 249:
```suggestion
- **PYTEST_THEME_MODE** sets this style to *light* or *dark*.
```

changelog/7132.feature.rst
```diff
@@ -0,0 +1 @@
+Added two environment variables ``PYTEST_THEME`` and ``PYTEST_THEME_MODE`` to let the users customize the pygments theme used.
```
nicoddemus, line 1: I think it would be nice to add a link to the docs here (possibly just change the env vars to `:envvar:PYTEST_THEME`, I think).

nicoddemus (changes requested):
Thanks a lot @azmeuk!
 
Please also add the new variables to the `reference.rst` file, in the "Environment Variables" section.

azmeuk:
I added some tests and applied your suggestions.

src/_pytest/_io/terminalwriter.py
```diff
@@ -195,16 +195,34 @@ def _write_source(self, lines: Sequence[str], indents: Sequence[str] = ()) -> No
 
     def _highlight(self, source: str) -> str:
         """Highlight the given source code if we have markup support."""
+        from _pytest.config.exceptions import UsageError
```
azmeuk, line 198: I let this import inside the function to avoid a circular import error.

src/_pytest/_io/terminalwriter.py
```diff
@@ -195,16 +195,35 @@ def _write_source(self, lines: Sequence[str], indents: Sequence[str] = ()) -> No
...
+                    PythonLexer(),
+                    TerminalFormatter(
+                        bg=os.getenv("PYTEST_THEME_MODE", "dark"),
+                        style=os.getenv("PYTEST_THEME"),
+                    ),
+                )
+                return highlighted
+            except pygments.util.ClassNotFound:
+                raise UsageError(
+                    "PYTEST_THEME environment variable had an invalid value. "
```
nicoddemus, line 222: One more suggestion: could we include the value of `PYTEST_THEME` into the message? This often helps when debugging errors/tracebacks from environments other than yours.

src/_pytest/_io/terminalwriter.py
```diff
@@ -195,16 +195,35 @@ def _write_source(self, lines: Sequence[str], indents: Sequence[str] = ()) -> No
...
+                    "Only valid pygment styles are allowed."
+                )
+            except pygments.util.OptionError:
+                raise UsageError(
+                    "PYTEST_THEME_MODE environment variable had an invalid value. "
```
nicoddemus, line 227: Same suggestion here to include the value of the environment variable into the message. 👍

nicoddemus (approved):
Thanks @azmeuk! Looks great!

Left two other suggestions, otherwise LGTM. 👍

azmeuk:
Done :)

The-Compiler (changes requested):
I don't want to bikeshed this, but I wonder why it should be an environment variable, when (almost) all other things are either ini settings or commandline arguments?

We already have `--code-highlight` from #6934, so why not `--code-highlight-theme` and `--code-highlight-theme-mode` or so?

nicoddemus:
I'm 👍 on configuring this as env vars... my thinking is that this is something you might want change just for yourself because of preference/color blindness issues, and this is not easily done with config vars (you can of course use command-line vars indirectly with `PYTEST_ADDOPTS`).

azmeuk:
I suggested an environment variable because of [this comment](https://github.com/pytest-dev/pytest/issues/6984#issuecomment-778825560), and because each contributor of a given project may have its own terminal configuration, and need its own color customization.

I am open to add/replace a command line argument though.

RonnyPfannschmidt:
unless pytest gets someething nice to unify config by env var, config file and cli argument im of the oppinion that this particular one is best placed in a env var, accompanied with some google search firendly docs for people that need it

azmeuk:
It looks like the failing tests are unrelated with this patch.

RonnyPfannschmidt (approved):

nicoddemus:
@The-Compiler do you agree with the reasoning for the environment variables? If so could you merge it then? Thanks! 🙇

The-Compiler:
Does this really work? I don't see a change in output when I set `PYTEST_THEME`, and the [TerminalFormatter](https://pygments.org/docs/formatters/#TerminalFormatter) we use doesn't seem to have a `style` option at all.

nicoddemus:
@azmeuk did you test this yourself? Can you post some screenshots please?

azmeuk:
> Does this really work? I don't see a change in output when I set `PYTEST_THEME`, and the [TerminalFormatter](https://pygments.org/docs/formatters/#TerminalFormatter) we use doesn't seem to have a `style` option at all.

I think you are right, I did not tested with other pygments themes. This [piece of documentation](https://github.com/pygments/pygments/blob/faf69c028f0efa7798a938163e8432d838acfda2/pygments/formatter.py#L31) indeed tells that `TerminalFormatter` does not use the `style` parameter. I will checks this out.

RonnyPfannschmidt:
https://github.com/pygments/pygments/blob/faf69c028f0efa7798a938163e8432d838acfda2/pygments/formatters/terminal256.py#L100 will use the color style, 

https://github.com/pygments/pygments/blob/faf69c028f0efa7798a938163e8432d838acfda2/pygments/formatters/terminal.py#L25-L58 will use a color scheme argument

i believe its very safe to usually pick the 256 formatter, lets figure how to correctly pick

RonnyPfannschmidt:
pygments itself does a explicit choice - https://github.com/pygments/pygments/blob/faf69c028f0efa7798a938163e8432d838acfda2/pygments/cmdline.py#L390-L395

i wonder if the color sceme can be worked into the details

azmeuk:
> i wonder if the color sceme can be worked into the details

What do you mean? Use the same trick as pygments to guess if a TerminalFormatter or a Terminal256Formatter should be used?

RonnyPfannschmidt:
@azmeuk no, i meant trying to figure of color schemes could be sensibly mapped to the low color terminal (or alternatively at least having color scemes for the visually imparied

azmeuk:
Now rereading my PR I guess it does not really have sense anymore unless we indeed use `Terminal256Formatter` instead of `TerminalFormatter`. Is it ok for you if I do the move?

RonnyPfannschmidt:
i'm going to be happy with formatter change

i believe for full compat we should do the same check pygments does (and print a warning if a color style is passed but not supported due to lack of 256 color terminals)

a followup then could be to provide a color sceme for the color-bind either by pygments directly or by addding one and upstreaming later

The-Compiler:
We're not really in "easy and unintrusive" terrain at that point anymore, IMHO... Even pygmentize gets this wrong, checking if 256 is in `$TERM` is by no means sufficient to check if the terminal supports 256 colors.

Case in point, here's the pygments default output with my terminal and `TerminalFormatter`:

image

and here when it switches to 256 color mode:

image

So this would really make things less readable (and uniform) for many setups, rather than more so.

FWIW I still think the right solution to this is the user configuring the colors in their terminal (and sticking with the current 16-color formatter), rather than adding 256-color support just so that we can use pygment themes. Right now, the output looks like other colorized output (say, `ls` and such) in a terminal, while with hand-picked/hardcoded color themes, there's a much bigger chance for conflicts with how the rest of the terminal output looks.

azmeuk:
I have a custom set of terminal colors (solarized), every terminal utility displays well except pytest.
I think we could meet in the middle:
- If `PYTEST_THEME` is defined, use `Terminal256Color` or `TerminalTrueColorFormatter with the passed theme,
- Else use `TerminalFormatter`,
What do you think @The-Compiler?

The-Compiler:
I guess that'd work indeed!

azmeuk:
Ok here it is.
Here are some tests with the solarized, emacs, and tango themes.

solarized-dark
emacs
tango

testing/test_terminal.py
```diff
@@ -2408,6 +2408,43 @@ def test_foo():
...
+        pytester.makepyfile(
+            """
+            def test_foo():
+                assert 1 == 10
+        """
+        )
+        monkeypatch.setenv("PYTEST_THEME", "solarized-dark")
+        result = pytester.runpytest("--color=yes")
+        result.stdout.fnmatch_lines(
+            color_mapping.format_for_fnmatch(
```
RonnyPfannschmidt, line 2423: i think we can either drop call  to the color mapping formatter or add a mode where we can enter actual expected numbers for a color scheme (to read better than the raw escape sequences)

doc/en/reference/customize.rst
```diff
@@ -239,3 +239,9 @@ Builtin configuration file options
 ----------------------------------------------
 
 For the full list of options consult the :ref:`reference documentation <ini options ref>`.
+
+Syntax highlighting theme customization
+---------------------------------------
+
+The syntax highlighting themes used by pytest can be customized using the :envvar:`PYTEST_THEME`.
+Its value can be any valid `pygment style <https://pygments.org/docs/styles/>`_.
```
hugovk, line 247:
```suggestion
Its value can be any valid `Pygments style <https://pygments.org/docs/styles/>`_.
```

doc/en/reference/reference.rst
```diff
@@ -1032,6 +1032,10 @@ Contains comma-separated list of modules that should be loaded as plugins:
 
     export PYTEST_PLUGINS=mymodule.plugin,xdist
 
+.. envvar:: PYTEST_THEME
+
+Sets a `pygment style <https://pygments.org/docs/styles/>`_ to use for the code output.
```
hugovk, line 1037:
```suggestion
Sets a `Pygments style <https://pygments.org/docs/styles/>`_ to use for the code output.
```

testing/test_terminal.py
```diff
@@ -2408,6 +2408,43 @@ def test_foo():
...
+            """
+            def test_foo():
+                assert 1 == 10
+        """
+        )
+        monkeypatch.setenv("PYTEST_THEME", "invalid")
+        result = pytester.runpytest("--color=yes")
+        result.stderr.fnmatch_lines(
+            "ERROR: PYTEST_THEME environment variable had an invalid value: 'invalid'. "
+            "Only valid pygment styles are allowed."
```
hugovk, line 2445:
```suggestion
            "Only valid Pygments styles are allowed."
```

src/_pytest/_io/terminalwriter.py
```diff
@@ -195,16 +195,34 @@ def _write_source(self, lines: Sequence[str], indents: Sequence[str] = ()) -> No
...
+                    formatter = TerminalTrueColorFormatter(
+                        style=os.getenv("PYTEST_THEME") or "default"
+                    )
+                else:
+                    formatter = TerminalFormatter(bg="dark")
+
+            except pygments.util.ClassNotFound:
+                raise UsageError(
+                    "PYTEST_THEME environment variable had an invalid value: '{}'. "
+                    "Only valid pygment styles are allowed.".format(
```
hugovk, line 222:
```suggestion
                    "Only valid Pygments styles are allowed.".format(
```

hugovk (commented):
Just a little typo.

azmeuk:
@RonnyPfannschmidt @hugovk Thank you for your feedbacks, they are fixed.

azmeuk:
@The-Compiler Is this PR acceptable to you now?

testing/conftest.py
```diff
@@ -136,6 +136,23 @@ def pytester(pytester: Pytester, monkeypatch: MonkeyPatch) -> Pytester:
     return pytester
 
 
+class _ColorMapping:
```
The-Compiler, line 139: You could declare `COLORS: Dict[str, str]` here, so that mypy knows that it exists.

testing/conftest.py
```diff
@@ -136,6 +136,23 @@ def pytester(pytester: Pytester, monkeypatch: MonkeyPatch) -> Pytester:
...
+
+    @classmethod
+    def format_for_fnmatch(cls, lines: List[str]) -> List[str]:
+        """Replace color names for use with LineMatcher.fnmatch_lines"""
+        return [line.format(**cls.COLORS).replace("[", "[[]") for line in lines]
+
+    @classmethod
+    def format_for_rematch(cls, lines: List[str]) -> List[str]:
+        """Replace color names for use with LineMatcher.re_match_lines"""
+        return [line.format(**cls.RE_COLORS) for line in lines]
```
The-Compiler, line 153: Might want to create `RE_COLORS` here (or in an `__init__`, but I think it's fine to do it every time the function is called). That way, you don't need to duplicate `RE_COLORS = {k: re.escape(v) for k, v in COLORS.items()}` for all subclasses, and mypy will be happy as well.

The-Compiler (changes requested):
Thanks for the update! Seems fine now, other than the [linting issues](https://results.pre-commit.ci/run/github/37489525/1623793629.DSRkRswNTduNuLir8BnprA). I added a couple of comments for the mypy ones.

azmeuk:
Here it is!

azmeuk:
@The-Compiler Ping! Would you consider accepting this PR now?

The-Compiler:
While test failures are unrelated, the [CI is still failing](https://results.pre-commit.ci/run/github/37489525/1623922295.nGxIv4bPRuqGJp538MioEg).

azmeuk:
Ok. This should be fixed now.

The-Compiler (approved):
Thanks! Looks good now, but I'm still confused about the test failures. I thought they were only something unrelated being flaky, but I can see them consistently happen in your branch, but apparently [not on the main branch](https://github.com/pytest-dev/pytest/commits/main). However, I'm unable to reproduce locally... anyone has an idea what's going on there?

azmeuk:
I rebased against the main branch, just in case.

azmeuk:
Different tests are failing each time, but the common thing is `ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>`. Not sure where it comes from.

The-Compiler:
I've been trying to reproduce with Python 3.9 in a Ubuntu 20.04 Docker container, but still no luck, unfortunately... Maybe someone else has an idea what could be going on.

azmeuk:
@nicoddemus @RonnyPfannschmidt do you maybe have any clue?

azmeuk:
I just tried -but failed- to reproduce the errors with [act](https://github.com/nektos/act).

nicoddemus:
> @nicoddemus @RonnyPfannschmidt do you maybe have any clue?

Sorry no clue :( 

Have you tried rebasing on the latest `main`?

azmeuk:
> Have you tried rebasing on the latest main?

Yep.

I found that the errors are raised randomly on different other tests due to `test_code_highlight_invalid_theme`. I suppose trying to access a non existing theme in pygments has side effects that make some other tests fails, depending on the order the are executed.

Here I am looking for unclosed file descriptors and warnings...

nicoddemus:
> I found that the errors are raised randomly on different other tests due to test_code_highlight_invalid_theme. I suppose trying to access a non existing theme in pygments has side effects that make some other tests fail, depending on the order the are executed.

If this is happening in a few tests, we can use `pytester.runpytest_subprocess()`, which will ensure no state will leak.

azmeuk:
> If this is happening in a few tests, we can use `pytester.runpytest_subprocess()`, which will ensure no state will leak.

:heart: :heart: :heart:
 
Tests are green now, this patch should be mergeable!

nicoddemus:
Woot! Thanks for the patience and following through with the PR! ❤️

azmeuk:
@nicoddemus @RonnyPfannschmidt @The-Compiler anyone interested in pressing the merge button?

nicoddemus:
Thanks!

PR #8754, opened 2021-06-12
Remove outdated docs about pytest.warns and DeprecatedWarning

nicoddemus:
Since #2908, the user doesn't need to set warning filters to capture
`DeprecationWarning` with `pytest.warns`.

Fix #8666

### DIFF
--- a/src/_pytest/tmpdir.py
+++ b/src/_pytest/tmpdir.py
@@ -1,6 +1,7 @@
 """Support for providing temporary directories to test functions."""
 import os
 import re
+import sys
 import tempfile
 from pathlib import Path
 from typing import Optional
@@ -130,9 +131,9 @@ class TempPathFactory:
             # Also, to keep things private, fixup any world-readable temp
             # rootdir's permissions. Historically 0o755 was used, so we can't
             # just error out on this, at least for a while.
-            if hasattr(os, "getuid"):
-                rootdir_stat = rootdir.stat()
+            if sys.platform != "win32":
                 uid = os.getuid()
+                rootdir_stat = rootdir.stat()
                 # getuid shouldn't fail, but cpython defines such a case.
                 # Let's hope for the best.
                 if uid != -1:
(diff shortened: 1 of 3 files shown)


testing/test_warnings.py
```diff
@@ -516,6 +516,24 @@ def test_hidden_by_system(self, pytester: Pytester, monkeypatch) -> None:
         result = pytester.runpytest_subprocess()
         assert WARNINGS_SUMMARY_HEADER not in result.stdout.str()
 
+    def test_recwarn(self, pytester: Pytester, monkeypatch) -> None:
+        """
+        Check that recwarn can capture DeprecationWarning by default
+        without custom filterwarnings (see #8666).
+        """
+        pytester.makepyfile(
```
asottile, line 524: I think this test should be inside test_recwarn.py?

also I don't think `pytester` is needed here -- this can be tested directly (?)

the testsuite is failing because this test function name matches `test_recwarn.py`

nicoddemus: Good points, all done.

I initially used pytester to ensure our own warning filters wouldn't interfere with the test, but in the end I'm not sure it is needed after all.

src/_pytest/tmpdir.py
```diff
@@ -131,8 +131,8 @@ def getbasetemp(self) -> Path:
             # rootdir's permissions. Historically 0o755 was used, so we can't
             # just error out on this, at least for a while.
             if hasattr(os, "getuid"):
+                uid = os.getuid()  # type:ignore[attr-defined]
```
nicoddemus, line 134: This was done in a separate commit (84c5fe4d440faa70b69e618779828af55436b569).

asottile: I think the `hasattr` check should be `if sys.platform != 'win32'`  -- this matches the guard from typeshed: https://github.com/python/typeshed/blob/48a346920bbc24165bd9e9cf29a2b5140eae2abc/stdlib/os/__init__.pyi#L388

asottile (approved):

src/_pytest/tmpdir.py
nicoddemus: Agree, thanks. Updated. 👍

nicoddemus:
Backport: https://github.com/pytest-dev/pytest/pull/8755

PR #8761, opened 2021-06-14
Add pytest.version_tuple

nicoddemus:
This adds `pytest.version_tuple`, which makes it simpler for users to do something depending on the pytest version, such as declaring hooks which are introduced in later versions.

This feature was added originally in https://github.com/pypa/setuptools_scm/pull/475.

Followup to https://github.com/pytest-dev/pytest/pull/7605.


RonnyPfannschmidt (approved):
:chart_with_upwards_trend:

PR #8792, opened 2021-06-23
Fix API links in 'How to capture warnings' documentation page

jrbourbeau:
I noticed a few places where a leading `:` was missing for the `:func:` sphinx directive which caused the link to the corresponding API docs to not render correctly. This PR adds the leading `:` to ensure links are rendered. 

Current rendering:

Screen Shot 2021-06-23 at 1 54 42 PM

With the changes in this PR:

Screen Shot 2021-06-23 at 1 54 48 PM


nicoddemus (approved):
Thanks!

PR #8795, opened 2021-06-25
Adjust releasing infra/docs

The-Compiler:
Follow-up for #8150, see #7507

There, @nicoddemus said:

> I intend to leave the other workflow in place until we can test this workflow and are satisfied with it.

No idea if the releasing instructions are correct like this (we still manually create the maintenance branch instead of automating it, right?).


RELEASING.rst
```diff
@@ -14,60 +14,90 @@ Preparing: Automatic Method
...
+     git branch 7.1.x upstream/main
+     git push upstream 7.1.x
 
-1. Create a new maintenance branch from ``main``::
+Then, trigger the workflow with the following inputs:
 
-        git fetch --all
-        git branch 5.2.x upstream/main
-        git push upstream 5.2.x
+   - branch: **7.1.x**
```
nicoddemus, line 46: This is not rendering correctly:

image

I think it needs to be dedented?

RELEASING.rst
```diff
@@ -14,60 +14,90 @@ Preparing: Automatic Method
...
+   - major release: **no**
+   - prerelease: empty
+
+Or via the commandline::
```
nicoddemus, line 50:
```suggestion
Or via the commandline using `GitHub's cli <https://github.com/cli/cli>`__::
```

nicoddemus (approved):
LGTM!

Left minor comments. 👍

PR #8800, opened 2021-06-27
doc: Adding exit code for pytest usage examples (#8790)

Srip:
Fix https://github.com/pytest-dev/pytest/issues/8790

### DIFF
--- a/doc/en/how-to/usage.rst
+++ b/doc/en/how-to/usage.rst
@@ -168,8 +168,7 @@ You can invoke ``pytest`` from Python code directly:
 
 .. code-block:: python
 
-    if __name__ == "__main__":
-        sys.exit(pytest.main())
+    retcode = pytest.main()
 
 this acts as if you would call "pytest" from the command line.
 It will not raise ``SystemExit`` but return the exitcode instead.
@@ -177,8 +176,7 @@ You can pass in options and arguments:
 
 .. code-block:: python
 
-    if __name__ == "__main__":
-        sys.exit(pytest.main(["-x", "mytestdir"]))
+    retcode = pytest.main(["-x", "mytestdir"])
 
 You can specify additional plugins to ``pytest.main``:
 
--- a/doc/en/how-to/usage.rst
+++ b/doc/en/how-to/usage.rst
@@ -168,7 +168,8 @@ You can invoke ``pytest`` from Python code directly:
 
 .. code-block:: python
 
-    pytest.main()
+    if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


doc/en/how-to/usage.rst
```diff
@@ -168,15 +168,17 @@ You can invoke ``pytest`` from Python code directly:
 
 .. code-block:: python
 
-    pytest.main()
```
nicoddemus, line 171: In this snippet I think it makes more sense to have this instead:

```python
retcode = pytest.main()
```

Here we shouldn't assume that this will be called as the "main" part of a script (might be confusing).

doc/en/how-to/usage.rst
```diff
@@ -168,15 +168,17 @@ You can invoke ``pytest`` from Python code directly:
...
+        sys.exit(pytest.main())
 
 this acts as if you would call "pytest" from the command line.
 It will not raise ``SystemExit`` but return the exitcode instead.
 You can pass in options and arguments:
 
 .. code-block:: python
 
-    pytest.main(["-x", "mytestdir"])
+    if __name__ == "__main__":
```
nicoddemus, line 180: Same here:

```python
retcode = pytest.main(["-x", "mytestdir"])
```

doc/en/how-to/usage.rst
```diff
@@ -191,7 +193,8 @@ You can specify additional plugins to ``pytest.main``:
             print("*** test run reporting finishing")
 
 
-    pytest.main(["-qq"], plugins=[MyPlugin()])
+    if __name__ == "__main__":
```
nicoddemus, line 196: 👍

doc/en/how-to/usage.rst
```diff
@@ -168,15 +168,17 @@ You can invoke ``pytest`` from Python code directly:
 
 .. code-block:: python
 
-    pytest.main()
+    if __name__ == "__main__":
+        sys.exit(pytest.main())
 
 this acts as if you would call "pytest" from the command line.
 It will not raise ``SystemExit`` but return the exitcode instead.
```
nicoddemus, line 175: While at it lets add some useful links. 👍 

```suggestion
It will not raise :class:`SystemExit` but return the :ref:`exit code <exit-codes>` instead.
```

nicoddemus (changes requested):
Thanks @Srip, we appreciate the contribution!

Left a few minor suggestions, please take a look. 👍

nicoddemus (approved):
Thanks a lot!

PR #8806, opened 2021-06-28
Fix "external plugins" link

brylie:
The "external plugins" link was pointing to a deprecated Heroku app. This PR uses the correct plugin link to save the reader an extra click.


The-Compiler:
Already seems fixed in the `main` branch to which the docs will point to as soon as 7.0 is released (very soon!), but in the meantime, I guess this doesn't hurt either.

The-Compiler (approved):

PR #8810, opened 2021-06-28
Opt-out from regendoc for problematic example

The-Compiler:
See #8807

### DIFF
--- a/doc/en/how-to/skipping.rst
+++ b/doc/en/how-to/skipping.rst
@@ -369,6 +369,9 @@ Here is a simple test file with the several usages:
 
 Running it with the report-on-xfail option gives this output:
 
+.. FIXME: Use $ instead of ! again to reenable regendoc once it's fixed:
+   https://github.com/pytest-dev/pytest/issues/8807
+
 .. code-block:: pytest
 
     ! pytest -rx xfail_demo.py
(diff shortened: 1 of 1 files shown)


doc/en/how-to/skipping.rst
```diff
@@ -371,7 +371,7 @@ Running it with the report-on-xfail option gives this output:
 
 .. code-block:: pytest
 
-    example $ pytest -rx xfail_demo.py
+    ! pytest -rx xfail_demo.py
```
nicoddemus, line 374: Can you perhaps add a comment outside the block, pointing to a relevant issue in regendoc (and mention this there)? Just so we can eventually change this back to `$` once regendoc gets fixed. 👍

nicoddemus (approved):

doc/en/how-to/skipping.rst
The-Compiler: IMHO, those kind of FIXME comments as part of code always get lost/outdated anyways - given that we already have #8807 and the commit message points to it, that might be enough already?

nicoddemus: No worries, but personally I see value in them: whenever I encounter a "TODO" with an issue link, I always click on it and see if it has been resolved and the TODO warrants a review. I've found many "forgotten" TODOs that way.

Plus when someone sees that weird `!` there, a link will promptly explain why that's there.

But as I said, I don't feel too strongly either, so feel free to leave it out if you think it is not necessary. 👍

The-Compiler: Ok, done!

PR #8811, opened 2021-06-28
Adjust fixme comments

The-Compiler:
Minor change but probably worth getting into 7.0 to be technically correct :sweat_smile:


nicoddemus (approved):
Well spotted. 👍

PR #8813, opened 2021-06-28
ci: Persist git credentials for prepare-release-pr

The-Compiler:
See #8599, follow-up to #8681

The token is needed so we can use `git push`.


nicoddemus (approved):
Makes sense, thanks!

The-Compiler:
I'll wait for @bluetech to chime in, won't get around to finishing this today anyways.

bluetech (approved):
Makes sense, sorry for missing it!

The-Compiler:
Don't worry about it, always hard with stuff which can't be tested!

PR #8826, opened 2021-06-29
closes #8824 Changelog rewording for 7.0

### LINKED ISSUE #8824: Changelog rewording for 7.0
See the comments in https://github.com/pytest-dev/pytest/pull/8814/files#diff-e02b5aa2806cf9086c18c4d38b880df457d9313bf02ba108d58e67336abbab66 - those should be fixed in the individual `changelog/*.rst` files.

Ignore the first comment about the title, that's tracked in #8817.

ericluoliu:


changelog/8509.improvement.rst
```diff
@@ -1,5 +1,5 @@
-Fixed issue where `TestCase.setUpClass` is not called when a test has `/` in its name since pytest 6.2.0.
+Fixed issue where :method:`unittest.TestCase.setUpClass` is not called when a test has `/` in its name since pytest 6.2.0.
```
The-Compiler, line 1: My bad, this should be 

```suggestion
Fixed issue where :meth:`unittest.TestCase.setUpClass` is not called when a test has `/` in its name since pytest 6.2.0.
```

changelog/8248.trivial.rst
```diff
@@ -1 +1 @@
-Internal Restructure: let python.PyObjMixing inherit from nodes.Node to carry over typing information.
+Internal Restructure: let ``python.PyObjMixing`` inherit from ``nodes.Node`` to carry over typing information.
```
The-Compiler, line 1: Just noticed this should actually be

```suggestion
Internal Restructure: let ``python.PyObjMixin`` inherit from ``nodes.Node`` to carry over typing information.
```

changelog/8761.feature.rst
```diff
@@ -1 +1 @@
-New :ref:`version-tuple` attribute, which makes it simpler for users to do something depending on the pytest version (such as declaring hooks which are introduced in later versions).
+New :ref:`version_tuple` attribute, which makes it simpler for users to do something depending on the pytest version (such as declaring hooks which are introduced in later versions).
```
The-Compiler, line 1: I'm not sure which one is correct here - @nicoddemus @RonnyPfannschmidt?

changelog/8645.improvement.rst
```diff
@@ -1,4 +1,4 @@
-Reducing confusion from `pytest.warns(None)` by:
+Reducing confusion from ``pytest.warns(None)`` by:
 
-- Allowing no arguments to be passed in order to catch any exception (no argument defaults to `Warning`).
-- Emit a deprecation warning if passed `None`.
+- Allowing no arguments to be passed in order to catch any exception (no argument defaults to :class:`Warning`).
+- Emit a deprecation warning if passed ``None``.
```
The-Compiler, line 4: This should probably be split off to a separate `.deprecation.rst` file? Or perhaps the whole thing should go under deprecations rather than improvements? @Zac-HD @olgarithms what do you think?

The-Compiler (changes requested):
Great work, thanks! Just a couple of small remaining changes, as well as some remaining questions to other maintainers/contributors about things I'm not sure about.

As for dropping #8411, you'll need to remove the `changelog/8411.trivial.rst` file entirely (e.g. via `git rm`). We also agreed to move #8503 from internal to bugfix, which you can do by renaming `8503.trivial.rst` to `8503.bugfix.rst` (e.g. via `git mv`).

changelog/8509.improvement.rst
ericluoliu: Cool, thank you so much for the pointers! I made another commit. Let me know when you have resolved changelog/8761.feature.rst and changelog/8645.improvement.rst and I'll make those changes ASAP

changelog/8761.feature.rst
The-Compiler: Ah, it looks like after resolving the previous build failure, Sphinx now fails with:

```
/home/runner/work/pytest/pytest/doc/en/_changelog_towncrier_draft.rst:114: WARNING: undefined label: version_tuple
```

here, so I was wrong and `version-tuple` was correct here. I thought it was referring to the variable `version_tuple`, but it actually refers to [a `version-tuple` label](https://github.com/pytest-dev/pytest/commit/7eb0792cba262ef2e659525e2e9f93742169efab#diff-039da146358bc54358e2b627fe9544ac13412c5f4bcc913277b45ac5256e8c61R239) added to the docs.

changelog/8509.improvement.rst
```diff
@@ -1,5 +1,5 @@
-Fixed issue where `TestCase.setUpClass` is not called when a test has `/` in its name since pytest 6.2.0.
+Fixed issue where :meth:`unittest.TestCase.setUpClass` is not called when a test has `/` in its name since pytest 6.2.0.
```
Zac-HD, line 1:
```suggestion
Fixed issue where :meth:`unittest.TestCase.setUpClass` is not called when a test has ``/`` in its name since pytest 6.2.0.
```

changelog/8645.improvement.rst
Zac-HD: Hmm.  I think it probably makes most sense to move this entirely to a `.deprecation.rst` file, with a message like 

```rst
:func:`pytest.warns(None) <pytest.warns>` is now deprecated because many people used
it to mean "this code does not emit warnings", but it actually had the effect of 
checking that the code emits at least one warning of any type - like ``pytest.warns()``
or ``pytest.warns(Warning)``.
```

changelog/8509.improvement.rst
nicoddemus: Btw, no need to worry about \` vs \`` because we configure the docs so they are equivalent:

https://github.com/pytest-dev/pytest/blob/501637547ecefa584db3793f71f1863da5ffc25f/doc/en/conf.py#L102-L103

The-Compiler: IMHO we should still be consistent though - right now we use an odd mix of both...

nicoddemus: Agree, just mentioning that is not necessary to go hunting for those, unless inclined to spend time on it for consistency reasons, because otherwise the end result is the same.

changelog/8447.deprecation.rst
```diff
@@ -1,4 +1,4 @@
-Defining a custom pytest node type which is both an item and a collector now issues a warning.
+Defining a custom pytest node type which is both an :class:`pytest.Item <Item>` and a :class:pytest.Collector <Collector>` now issues a warning.
```
The-Compiler, line 1:
```suggestion
Defining a custom pytest node type which is both an :class:`pytest.Item <Item>` and a :class:`pytest.Collector <Collector>` now issues a warning.
```

Last one hopefully! 🙂

The-Compiler (changes requested):
One issue left from a quick look, afterwards I think this should be ready!

changelog/8645.improvement.rst
olgarithms: Happy with this decision too!

The-Compiler (approved):
Looks great now - thanks a lot @ericluoliu, one step closer to 7.0 again! :tada:

Issue #8819, opened 2021-06-29
Add missing sys import in usage.rst example


The-Compiler:
Oh we missed a `import sys` in the code above it (I can't suggest a change because it is not part of the diff :/)

_Originally posted by @nicoddemus in https://github.com/pytest-dev/pytest/pull/8814#discussion_r660069659_

The-Compiler:
@Srip would you like to open a follow-up PR for this one?

PR #8827, opened 2021-06-29
doc: Adding the missing sys import (#8819)

Srip:
Fixing the missing import statement in usage.rst file


The-Compiler (approved):
Thanks for the follow-up, @Srip! :+1:

Issue #8823, opened 2021-06-29
Filter [bot] names in release automation


The-Compiler:
Not related to the release itself, but we should probably remove `[bot]` names here.

_Originally posted by @The-Compiler in https://github.com/pytest-dev/pytest/pull/8814#r660067534_

PR #8828, opened 2021-06-29
Issue #8823 - remove bots from contributors list

ericluoliu:


scripts/release.py
```diff
@@ -22,6 +22,11 @@ def announce(version):
 
     contributors = set(stdout.splitlines())
 
+    # remove strings within contributors that have substring "[bot]"
+    for name in contributors:
```
nicoddemus, line 26:
```suggestion
    for name in list(contributors):
```

Otherwise we will change the size of the set during iteration and raise an error.

nicoddemus: But probably it is easier to update line 23 to:

```python
contributors = set(x for x in stdout.splitlines() if "[bot]" not in x)
```

ericluoliu: Yeah I agree. I think the one-liner works well and looks nicer. Should I make a new commit?

The-Compiler: We typically squash PRs into a single commit when merging, so feel free to just push a second commit to the same branch (which will automatically update the PR).

As an aside, I'd probably use `name` instead of `x`, and `if not name.endswith("[bot]")` as I think it's always used as a suffix.

The-Compiler (approved):
Looks great now, thanks! :tada: One step closer to the 7.0 release!

PR #8830, opened 2021-06-30
Fix docs URL in unregistered marks warning message

### LINKED ISSUE #8787: Wrong link on warning message for custom mark
When I try to use custome marker ex. "@pytest.mark.webtest" after test finishes i get warning:
"PytestUnknownMarkWarning: Unknown pytest.mark.uploadtenantcontent - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html"

But this link: https://docs.pytest.org/en/latest/mark.html does not exists, i get 404 
i think correct link should be: https://docs.pytest.org/en/6.2.x/mark.html
or maybe: https://docs.pytest.org/en/6.2.x/example/markers.html

pytest: 6.2.4
OS: MacOS Catalina

pip list:
attrs           20.3.0
boto3           1.17.82
botocore        1.20.82
iniconfig       1.1.1
jmespath        0.10.0
mstest          0.1.0
packaging       20.9
pip             21.0.1
pluggy          0.13.1
py              1.10.0
pyparsing       2.4.7
pytest          6.2.4
pytest-mock     3.6.1
python-dateutil 2.8.1
s3transfer      0.4.2
setuptools      54.1.2
six             1.16.0
urllib3         1.26.5
wheel           0.36.2

DrGFreeman:
Fixes #8787

Fix URL of documentation link in message of warning issued on unregisterd marks from https://docs.pytest.org/en/latest/mark.html (broken link) to https://docs.pytest.org/en/latest/how-to/mark.html.

### DIFF
--- a/changelog/8787.bugfix.rst
+++ b/changelog/8787.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed URL of documentation link in message of warning issued on unregisterd marks.
+Fixed URL of documentation link in message of warning issued on unregistered marks.
--- a/changelog/8787.bugfix.rst
+++ b/changelog/8787.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed URL of documentation link in message of warning issued on unregisterd marks.
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -530,7 +530,7 @@ class MarkGenerator:
                 warnings.warn(
                     "Unknown pytest.mark.%s - is this a typo?  You can register "
                     "custom marks to avoid this warning - for details, see "
-                    "https://docs.pytest.org/en/latest/how-to/mark.html" % name,
+                    "https://docs.pytest.org/en/stable/how-to/mark.html" % name,
                     PytestUnknownMarkWarning,
                     2,
                 )
--- a/src/_pytest/mark/structures.py
+++ b/src/_pytest/mark/structures.py
@@ -530,7 +530,7 @@ class MarkGenerator:
                 warnings.warn(
                     "Unknown pytest.mark.%s - is this a typo?  You can register "
                     "custom marks to avoid this warning - for details, see "
-                    "https://docs.pytest.org/en/stable/mark.html" % name,
+                    "https://docs.pytest.org/en/latest/how-to/mark.html" % name,
                     PytestUnknownMarkWarning,
                     2,
                 )
--- a/AUTHORS
+++ b/AUTHORS
@@ -163,6 +163,7 @@ Jordan Speicher
 Joseph Hunkeler
 Josh Karpel
 Joshua Bronson
+Julien de la Bruère-Terreault
 Jurko Gospodnetić
 Justyna Janczyszyn
 Justice Ndou


changelog/8787.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed URL of documentation link in message of warning issued on unregisterd marks.
```
The-Compiler, line 1:
```suggestion
Fixed URL of documentation link in message of warning issued on unregistered marks.
```

src/_pytest/mark/structures.py
```diff
@@ -530,7 +530,7 @@ def __getattr__(self, name: str) -> MarkDecorator:
                 warnings.warn(
                     "Unknown pytest.mark.%s - is this a typo?  You can register "
                     "custom marks to avoid this warning - for details, see "
-                    "https://docs.pytest.org/en/stable/mark.html" % name,
+                    "https://docs.pytest.org/en/latest/how-to/mark.html" % name,
```
The-Compiler, line 533: Note how this already points to `stable/` (with a link that works correctly) rather than `latest/`, see 93d2ccbfb7b99755e8f4a2a3947c466949ee3c46. So I'm not sure where @bsypniewski got the `latest` link from...

I believe we should continue to point to `stable/` here, but to `stable/how-to/mark.html` which will then work correctly once 7.0.0 is out.

What do others think?

The-Compiler (changes requested):

src/_pytest/mark/structures.py
The-Compiler: Actually, that's a problem for far more than just this link... I opened #8831 to track this. As for this one, I believe the old URL is actually correct for now.

DrGFreeman: > I believe we should continue to point to stable/ here, but to stable/how-to/mark.html which will then work correctly once 7.0.0 is out.

Makes sense. I found the new page searching in `latest/` but the link should point to `stable/` like before.

changelog/8787.bugfix.rst
DrGFreeman: Fixed in https://github.com/pytest-dev/pytest/pull/8830/commits/9afc1af093ddfd7e42c9a0a653bfa0ef8297356f.

src/_pytest/mark/structures.py
DrGFreeman: Link updated in https://github.com/pytest-dev/pytest/pull/8830/commits/820d56c6a8957164ab5872879d3b411b209701f3.

DrGFreeman:
Thanks for the feedback @The-Compiler. I made the two changes requested. Shall I squash the commits?

The-Compiler:
No need for squashing, we typically do that when merging anyways. I'd rather fix everything together, as right now this breaks the (previously working) link for people running from git - thus, it'd be good to have everything ready and merge it only shortly before 7.0.0 is released - what do you think?

The-Compiler:
Superseded by #8858 where I'm updating all the links - thanks for the contribution even if it didn't end up being merged! :+1:

PR #8837, opened 2021-06-30
Change deprecation warnings into errors for 7.0 release

### LINKED ISSUE #7363: Change deprecation warnings into errors for 8.0
Ref: https://github.com/pytest-dev/pytest/issues/5584, https://github.com/pytest-dev/pytest/pull/7362

Note: originally this was meant to be integrated in 7.0 in #8837, but we decided to postpone it as per https://github.com/pytest-dev/pytest/pull/8837#issuecomment-872138897.

### LINKED ISSUE #8833: 7.0 and test_deprecation_warning_as_error
https://github.com/pytest-dev/pytest/blob/9cdb7223b5f465d6511e2da1885ad37701eb8d02/testing/test_warnings.py#L520-L529

what's the plan there for 7.0? @nicoddemus?

nicoddemus:
Fix #7363
Close #8833

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,6 +58,7 @@ repos:
           - py>=1.8.2
           - attrs>=19.2.0
           - packaging
+          - types-atomicwrites
           - types-toml
           - types-pkg_resources
 -   repo: local
--- a/changelog/7363.breaking.rst
+++ b/changelog/7363.breaking.rst
@@ -1,0 +1,23 @@
+**PytestDeprecationWarning are now errors by default.**
+
+Following our plan to remove deprecated features with as little disruption as
+possible, all warnings of type ``PytestDeprecationWarning`` now generate errors
+instead of warning messages.
+
+**The affected features will be effectively removed in pytest 7.1**, so please consult the
+`Deprecations and Removals <https://docs.pytest.org/en/stable/deprecations.html>`__
+section in the docs for directions on how to update existing code.
+
+In the pytest ``7.0.X`` series, it is possible to change the errors back into warnings as a
+stopgap measure by adding this to your ``pytest.ini`` file:
+
+.. code-block:: ini
+
+    [pytest]
+    filterwarnings =
+        ignore::pytest.PytestDeprecationWarning
+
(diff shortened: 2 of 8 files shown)


The-Compiler:
Ah, I see - but with this, I suppose some of my suggestions for moving things from 7.0 to 8.0 in #8834 and #8826 are actually wrong? Could you please take a look at those?

The-Compiler:
Oh, also, how does this work with warnings which weren't part of 6.3.0, but only introduced in this release? Looking at [deprecated.py](https://github.com/pytest-dev/pytest/blame/main/src/_pytest/deprecated.py) and the [changelogs](https://github.com/pytest-dev/pytest/tree/main/changelog) we seem to have many new deprecations which haven't been in a release yet?

nicoddemus:
> Ah, I see - but with this, I suppose some of my suggestions for moving things from 7.0 to 8.0 in #8834 and #8826 are actually wrong? Could you please take a look at those?

Yes, those all should be changed back to 7.0, as that's when they turned into errors.

> Looking at deprecated.py and the changelogs we seem to have many new deprecations which haven't been in a release yet?

Bummer, yes you are right. Unfortunately we took so long to make a new release (with things landing on `main`) that we are not following through our policy of showing deprecations for two minor releases before turning them into errors.

1. We revert the deprecations altogether?
2. We don't turn those warnings into errors for 7.0, instead postponing that to 8.0?

I think 2) is the simplest one (much less work), and it just means some warnings will be visible longer, which I think is preferable than turning them into errors right away.


@RonnyPfannschmidt @bluetech @asottile what do you guys think?

bluetech:
I'd go with 2 -- no reason to revert them, can just not turn them into errors.

The-Compiler:
I'd prefer 2 as well, especially because this major release is a bit special - it's only really a major release because of how much stuff accumulated. I'd rather keep 7.0 a rather painless release for people (reducing the pain to whatever will perhaps surface due to all our internal changes), and have the intended breakage in a 8.0 where we are a bit more flexible and hopefully not as pressured by not having had a release in a long time.

nicoddemus:
Great, closing this then! I have updated the other issues as appropriate. 👍

PR #8840, opened 2021-06-30
Fix nonpython.rst usage of deprecated Node.fspath

### LINKED ISSUE #8821: Fix fspath warnings in nonpython.rst example
Looks like something we should fix in the example

_Originally posted by @The-Compiler in https://github.com/pytest-dev/pytest/pull/8814#discussion_r660071924_

nicoddemus:
Fix #8821

Not sure this is the best solution, but at least that's how we do it in `DoctestItem`.

### DIFF
--- a/doc/en/example/nonpython/conftest.py
+++ b/doc/en/example/nonpython/conftest.py
@@ -1,5 +1,6 @@
 # content of conftest.py
 import pytest
+from _pytest.compat import legacy_path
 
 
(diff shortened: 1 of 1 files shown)


The-Compiler:
I have some thoughts on this - I'll post them over at #8821 since they're not directly related to the PR.

nicoddemus:
As we discussed in #8821, let's defer the warning to a later version.

PR #8874, opened 2021-07-08
Docs: monkeypatch.setenv does not accept bool

adamjstewart:
The [`monkeypatch` docs](https://docs.pytest.org/en/6.2.x/reference.html?highlight=monkeypatch#pytest.monkeypatch.monkeypatch) show an example of passing `prepend=False` to `monkeypatch.setenv`. However, the [`monkeypatch.setenv` docs](https://docs.pytest.org/en/6.2.x/reference.html?highlight=monkeypatch#pytest.MonkeyPatch.setenv) clearly state that `prepend` is of type `str` or `None`. This PR fixes the `monkeypatch` docs.

### DIFF
--- a/doc/en/builtin.rst
+++ b/doc/en/builtin.rst
@@ -132,7 +132,7 @@ For information about fixtures, see :ref:`fixtures`. To see a complete list of a
             monkeypatch.delattr(obj, name, raising=True)
             monkeypatch.setitem(mapping, name, value)
             monkeypatch.delitem(obj, name, raising=True)
-            monkeypatch.setenv(name, value, prepend=False)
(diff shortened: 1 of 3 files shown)


The-Compiler (changes requested):
Good catch! It looks like the same example is in `doc/en/builtin.rst` and `doc/en/how-to/monkeypatch.rst` too - could you adjust them as well?

I also wonder if we should use `prepend=None` instead, given that those examples mostly seem to show the default signature.

adamjstewart:
Done and done 😄

nicoddemus (approved):

The-Compiler (approved):

PR #8895, opened 2021-07-12
Adjust enum reprs conditional from Python 3.10 to 3.11

hroncok:
I don't think a changelog fragment is necessary: It changes tests only, no chnagelog fragment was even in the original commit that introduced the condiitonal.

### DIFF
--- a/testing/python/metafunc.py
+++ b/testing/python/metafunc.py
@@ -448,7 +448,7 @@ class TestMetafunc:
         enum = pytest.importorskip("enum")
         e = enum.Enum("Foo", "one, two")
         result = idmaker(("a", "b"), [pytest.param(e.one, e.two)])
-        if sys.version_info[:2] >= (3, 10):
+        if sys.version_info[:2] >= (3, 11):
             assert result == ["one-two"]
         else:
             assert result == ["Foo.one-Foo.two"]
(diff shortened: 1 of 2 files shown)


The-Compiler:
FWIW I'd prefer a revert of those changes - according to the [discussion](https://mail.python.org/archives/list/python-dev@python.org/thread/ZMC67QA2JVQJSWSFWRS6IM6ZX4EK277G/#LSTMFAPSPD3BGZ4D6HQFODXZVB3PLYKF), there are potentially incompatible changes planned for 3.11, and there will be a PEP written first - so at this point, it seems very speculative to me to already have our tests handle the new format when we have no idea how it'll end up looking like.

hroncok:
Fine by me ;)

hroncok:
Alternative approach (full revert) in https://github.com/pytest-dev/pytest/pull/8896

nicoddemus:
> FWIW I'd prefer a revert of those changes - according to the discussion, there are potentially incompatible changes planned for 3.11, and there will be a PEP written first - so at this point, it seems very speculative to me to already have our tests handle the new format when we have no idea how it'll end up looking like.

Agreed, thanks for pointing that out. I thought the changes would land in 3.11, so why not just adjust the code? But given more discussion will certainly follow, then I agree perfectly.

nicoddemus:
Closing in favor of https://github.com/pytest-dev/pytest/pull/8896.

PR #8896, opened 2021-07-12
Revert "Adjust enum reprs for Python 3.10"

hroncok:
I don't think a changelog fragment is necessary: It changes tests only, no chnagelog fragment was even in the original commit that introduced the conditional.

Alternative to https://github.com/pytest-dev/pytest/pull/8895


nicoddemus (approved):
Thanks!

The-Compiler (approved):

PR #8904, opened 2021-07-13
Add types-atomicwrites to mypy's pre-commit configuration

nicoddemus:
This is a Windows only dependency, so the problem only shows up when trying to commit on Windows.


nicoddemus:
Going ahead and merging this as it is harmless. 👍

PR #8905, opened 2021-07-13
rename tmpdir.rst to tmp_path.rst

### LINKED ISSUE #8897: rename tmpdir.rst to tmp_path.rst
but also with a permanent 308 redirect from tmpdir to tmp_path so the link juice is conserved

currently the tmp_path docs are hosted at: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture

but really it would be better to use https://docs.pytest.org/en/6.2.x/tmp_path.html#the-tmp-path-fixture

graingert:
fixes #8897

### DIFF
--- a/doc/en/contents.rst
+++ b/doc/en/contents.rst
@@ -28,7 +28,7 @@ How-to guides
    how-to/fixtures
    how-to/mark
    how-to/parametrize
-   how-to/tmpdir
+   how-to/tmp_path
    how-to/monkeypatch
    how-to/doctest
    how-to/cache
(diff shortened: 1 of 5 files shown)


nicoddemus (approved):
Thanks!

nicoddemus (changes requested):
Ahh seems missed to update the doctree:

```
checking consistency... /home/runner/work/pytest/pytest/doc/en/how-to/tmp_path.rst: WARNING: document isn't included in any toctree
```

nicoddemus (approved):

nicoddemus:
Redirect configured:

image

PR #8858, opened 2021-07-06
Adjust doc links for new scheme

The-Compiler:
Closes #8831 and supersedes #8830.

Some notes:

1) Since we have redirects in place now (#8471), we don't strictly need to do this (except for a couple of cases where the contents moved, I'll comment on those below). However, I still think it's good if we point to the "canonical" location of the docs.
2) Those links won't work right now, unless you replace `/stable/` by `/latest/`. They will however work once 7.0 is released and RTD regenerates the stable docs.

I'd probably hold off merging this shortly before the 7.0 release (perhaps even the final one rather than the prerelease?) so that we don't break links for people trying the rc or running from git before 7.0 is out.

cc @DrGFreeman

PR #8929, opened 2021-07-21
Fix dead link in README for 'assert statements' #8926

theshortman:


The-Compiler:
The link is correct, but only will work properly once the 7.0 release is out. See #8858 for some context. Thanks for the contribution, though!

PR #8945, opened 2021-07-26
Fix Dead link in README for 'assert statements'

eamanu:
Closes: #8926

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,7 +79,7 @@ Due to ``pytest``'s detailed assertion introspection, only plain ``assert`` stat
 Features
 --------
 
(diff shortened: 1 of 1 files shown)


nicoddemus:
Thanks!

The new `how-to` URL is actually correct in `main` (which is the `latest` docs), but not yet in `6.2.x` (which is the `stable` docs).

We can merge this though because we have a redirect `/assert.html -> /how-to/assert.html` configured, so the fixed URL works in both `latest` and `stable`. 

After 7.0.0 we can revert it, because then `/how-to/assert.html` will be valid in both `latest` and `stable`.

nicoddemus (approved):

eamanu:
Thanks @nicoddemus

The-Compiler:
FYI, reverted in #9369.

PR #8955, opened 2021-07-29
Making `--debug` more configurable for the pytest user

### LINKED ISSUE #8954: make `--debug` more configurable via `--debug-file` or `--debug-append`
Hi, it is probably common (enough) that when using `xdist` people are running pytest multiple time(s), I think @nicoddemus even mentioned he does this, we do in work similar like:

```python
# multiple jenkins stages & pytest invocations are involved in one 'CI run'
pytest -m parallel and not isolated --debug
pytest -m isolated --debug
```

By default `--debug` opens the `pytestdebug.log` file with `w` which means collecting artifacts later to assist with debugging when using a CI system like jenkins for example is often problematic as you will only receive the isolated runs debug logs here

I propose one of the following:

- `--debug-file=debug1.log` (I think keeping the same root dir here but allowing unique names could be beneficial)
- `--debug-append` which _appends_ (`a`) to the file if it already exists

Thoughts? I will PR if you think its got any merit

symonk:
Hi all, hope you are well,

This patch adds capabilities to provide a customised file name to `--debug` which routes the debug logs into a file of the users choice.  I have proposed this because we use multiple pytest executions as part of one jenkins job and find we have a wealth of internal plugins, running each run with `--debug` and artifacting the log is useful for us on a per run basis, we run parallel tests with xdist (want the debug log) then isolated sequential tests in a single process (want the debug log) and by default, this overwrites the pytestdebug.log.  see #8954 for some of my reasoning.

This remains the same functionality to most users in the sense that:

There are now 3 options to `--debug`:

 - `--debug` in isolation will result in a `pytestdebug.log` file in the root directory as prior
 - `--debug` omitted from the command line (ini args etc) will result in no log file written
 - `--debug mylog` will result in `mylog.log` holding the debug logs

Notes:
 - `.log` is suffixed to the user file if it does not end in `.log` just to keep things relatively similar to the original file here
 - unit / integration tests added
 - caveat: This is now user input and we open with `w`, so there is a chance someone passes a file that exists and we truncate and write logs to it 🤔 (not really sure how to solve that one, we don't want to append and can't check if it exists do nothing as overwriting it is the norm)

Let me know what you think,  thanks!

Close #8954


symonk:
windows `b` comes before `f` but not quite the same on unix* i guess, I'm a windows user at home so never spotted this

src/_pytest/helpconfig.py
```diff
@@ -114,17 +120,30 @@ def pytest_cmdline_parse():
...
             debugfile.close()
-            sys.stderr.write("wrote pytestdebug information to %s\n" % debugfile.name)
+            sys.stderr.write("wrote pytest debug information to %s\n" % debugfile.name)
             config.trace.root.setwriter(None)
             undo_tracing()
 
         config.add_cleanup(unset_tracing)
 
 
+def _resolve_debug_log_name(file_name: str) -> str:
```
nicoddemus, line 134: As I mentioned, I don't think we need this.

nicoddemus (commented):
Looks good, thanks! Left a few comments, please take a look.

testing/test_config.py
```diff
@@ -2042,3 +2043,58 @@ def test_parse_warning_filter_failure(arg: str) -> None:
 
     with pytest.raises(warnings._OptionError):
         parse_warning_filter(arg, escape=True)
+
+
+def test_debug_file_name_is_suffixed_with_dot_log() -> None:
```
nicoddemus, line 2048: Personally I would group those new tests into a class:

```python
class TestDebugOption:

    def test_suffixed_with_dot_log() -> None:
        ...
```

But that's just a matter of choice I guess.

nicoddemus (changes requested):
Good work, thanks @symonk!

symonk:
> Good work, thanks @symonk!

Thanks again @nicoddemus for the review, should be sorted now, appreciate it.

nicoddemus (approved):
Thanks!

PR #8962, opened 2021-07-31
Add examples of parametrizing classes and all tests in a module

### LINKED ISSUE #8947: DOC: pytest.mark.parametrize on class definition should be documented

Using `pytest.mark.parametrize` on class definition should be documented. It is mentioned in https://stackoverflow.com/questions/17434031/parametrizing-test-classes-in-pytest/ (the second answer) and it works, but I cannot find it hinted at https://github.com/pytest-dev/pytest/blob/main/doc/en/example/parametrize.rst . All the existing example involving classes are way too complicated for the particular usage I am looking for.

```python
import pytest


@pytest.mark.parametrize('x', [1, 2])
class TestMe:
    def test_1(self, x):
        print(f'test 1: {x}')

    def test_2(self, x):
        print(f'test 2: {x}')
```

gives me correctly:

```
test 1: 1
test 1: 2
test 2: 1
test 2: 2
```

But since it is not documented, is it only working by accident? Or it is actually supported?

Document it if it is actually a supported usage.

#### Alternative Solutions

Disallow it if it is not actually a supported usage.

#### Additional context

astropy/astropy#11972

cc @WilliamJamieson

eamanu:
This PR add examples of the parametrize in classes definition.

Closes #8947

### DIFF
--- a/doc/en/how-to/parametrize.rst
+++ b/doc/en/how-to/parametrize.rst
@@ -110,7 +110,41 @@ the simple test function.  And as usual with test function arguments,
 you can see the ``input`` and ``output`` values in the traceback.
 
 Note that you could also use the parametrize marker on a class or a module
-(see :ref:`mark`) which would invoke several functions with the argument sets.
+(see :ref:`mark`) which would invoke several functions with the argument sets,
+for instance:
+
+
+.. code-block:: python
+
+    import pytest
+
+
+    @pytest.mark.parametrize("n,expected", [(1, 2), (3, 4)])
+    class TestClass:
+        def test_simple_case(self, n, expected):
+            assert n + 1 == expected
+
+        def test_weird_simple_case(self, n, expected):
+            assert (n * 1) + 1 == expected
+
+
+You can assign the global variable **pytestmark**:
+
+
+.. code-block:: python
+
+    import pytest
+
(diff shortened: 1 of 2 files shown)


doc/en/how-to/parametrize.rst
```diff
@@ -110,7 +110,41 @@ the simple test function.  And as usual with test function arguments,
...
+    @pytest.mark.parametrize("n,expected", [(1, 2), (3, 4)])
+    class TestClass:
+        def test_simple_case(self, n, expected):
+            assert n + 1 == expected
+
+        def test_weird_simple_case(self, n, expected):
+            assert (n * 1) + 1 == expected
+
+
+You can assign the global variable **pytestmark**:
```
nicoddemus, line 131:
```suggestion
To parametrize all tests in a module, you can assign to the :globalvar:`pytestmark` global variable:
```

nicoddemus (approved):
Thanks @eamanu, looks great!

doc/en/how-to/parametrize.rst
nicoddemus: Went ahead and applied that suggestion myself so we can merge this. 👍

symonk (approved):
Nice one @eamanu - thanks a lot!

eamanu:
\o/

pllim:
Thanks! Didn't know about the module level one. TIL. 🙇‍♀️

PR #8963, opened 2021-07-31
typeshed `os.PathLike` is now `@runtime_checkable`, remove mypy ignore

symonk:
`os.PathLike` is now `@runtime_checkable` in typeshed, so it is my understanding we no longer need to suppress `os.PathLike` isinstance checks.

https://github.com/python/typeshed/blob/e1b9ab33727ddbb2bab21c0a1c161d9a0b2c7c2f/stdlib/os/__init__.pyi#L311-L313

```py
@runtime_checkable
class PathLike(Protocol[_AnyStr_co]):
    def __fspath__(self) -> _AnyStr_co: ...
```

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -305,9 +305,7 @@ def _prepareconfig(
 ) -> "Config":
     if args is None:
         args = sys.argv[1:]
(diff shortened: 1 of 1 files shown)


bluetech (approved):
Thanks @symonk!

nicoddemus (approved):

PR #8967, opened 2021-08-01
`pytest_assertion_pass` is no longer considered `experimental`

symonk:
Good morning,

I think the hook has been around for long enough without any real issues coming up (relatively minor stuff from a quick issue tracker search) so I've just gone ahead and removed the documentation with regards to `pytest_assertion_pass`.

 - Does something like this need a change log entry?
 - I think it's only docs that need updated here? There exists a `PytestExperimentalApiWarning` but I couldn't find any code connected to this particular hook relating to that.

Thanks.


nicoddemus:
> Does something like this need a change log entry?

I think so!

symonk:
cool, I've added a changelog here, I'm wondering do I need to do anything around the `ini` options for: `enable_assertion_pass_hook`, I'm not familiar with 'experimental' features, I'm guessing this hook should become completely part of the ecosystem if it moves from an experimental state (rather than flagged opt-in)? guidance would be great there, thanks again

nicoddemus:
Let's keep the option, because that hook has a noticeable cost even if you are not using it.

nicoddemus (approved):

nicoddemus:
(Make sure to squash/merge the PR 👍 )

PR #9017, opened 2021-08-18
Docs: refer to non-deprecated unittest alias

hugovk:
> *Deprecated since version 3.2:* `assertRegexpMatches` and `assertRaisesRegexp` have been renamed to `assertRegex()` and `assertRaisesRegex()`.

https://docs.python.org/3/library/unittest.html#deprecated-aliases


RonnyPfannschmidt (approved):
:+1: thanks

nicoddemus (approved):

PR #9053, opened 2021-08-28
As discussed in Discord, move 3.10 changelog to trivial

nicoddemus:
This was backported to `6.2.x`, but we agreed this should
actually a trivial change and go into `6.2.5`, instead of a new `6.3.0`.


The-Compiler (approved):
Sounds good to me, given that #8494 really only has internal/tooling changes - thus it seems reasonable to mark it as trivial.

Also we released 6.2.4 with "Fixed assertion rewriting on Python 3.10." (#8539), implying that it already supported Python 3.10 - so this hardly seems like a new feature from an user's perspective anyways.

nicoddemus:
Backport: https://github.com/pytest-dev/pytest/pull/9056

PR #9061, opened 2021-08-30
fix: prevent approx from being used without a comparison

kalekundert:
Recently it was brought to my attention that some of the top search-engine hits for `pytest.approx` use the function without actually comparing it to anything.  For example, the [following snippet](https://www.programcreek.com/python/example/90870/pytest.approx) is currently the second result for "pytest approx" in duck duck go:
```python
def test_text_classification_unified_information(notebooks, tmp):
    notebook_path = notebooks["tc_unified_information"]
    pm.execute_notebook(
        notebook_path,
        OUTPUT_NOTEBOOK,
        kernel_name=KERNEL_NAME,
        parameters=dict(
            DATA_FOLDER=tmp,
            BERT_CACHE_DIR=tmp,
            BATCH_SIZE=32,
            BATCH_SIZE_PRED=512,
            NUM_EPOCHS=1,
            TEST=True,
            QUICK_RUN=True,
        ),
    )
    result = sb.read_notebook(OUTPUT_NOTEBOOK).scraps.data_dict
    assert pytest.approx(result["accuracy"], 0.93, abs=ABS_TOL)
    assert pytest.approx(result["precision"], 0.93, abs=ABS_TOL)
    assert pytest.approx(result["recall"], 0.93, abs=ABS_TOL)
    assert pytest.approx(result["f1"], 0.93, abs=ABS_TOL)
```
Note that none of those assertions actually do anything, because the `approx` object unconditionally evaluates to true (and `0.93` is being interpreted as the relative tolerance; not as the expected value).  But the tests still pass, which is not good.  This is apparently a common mistake; in addition to the snippet above, I was able to find several other examples the same bug with just a cursory search.

This PR will cause these tests to fail by implementing `approx.__bool__()` to raise an `AssertionError` that briefly explains how to correctly use `approx`.


src/_pytest/python_api.py
```diff
@@ -100,6 +100,9 @@ def __eq__(self, actual) -> bool:
             a == self._approx_scalar(x) for a, x in self._yield_comparisons(actual)
         )
 
+    def __bool__(self):
+        raise AssertionError("use approx() like this: `assert a == approx(b)`")
```
nicoddemus, line 104: I think we can add a bit more context:

```suggestion
        raise AssertionError("approx() is not supported in a boolean context.\nUse approx() like this: `assert a == approx(b)`")
```

nicoddemus (changes requested):
Good catch, thanks for the PR!

Besides my other suggestion, we just need a `9061.improvement.rst` file with something like:


```rst
Using :func:`pytest.approx` in a boolean context now raises an error hinting at the proper usage.

It is apparently common for users to mistakenly use ``pytest.approx`` like this:

.. code-block:: python

    assert pytest.approx(result["accuracy"], 0.93, abs=ABS_TOL)

While the correct usage is:

.. code-block:: python

    assert result["accuracy"] == pytest.approx(0.93, abs=ABS_TOL) 

The new error message helps catching those mistakes.
```

src/_pytest/python_api.py
kalekundert: I like that better

src/_pytest/python_api.py
```diff
@@ -100,6 +100,11 @@ def __eq__(self, actual) -> bool:
             a == self._approx_scalar(x) for a, x in self._yield_comparisons(actual)
         )
 
+    def __bool__(self):
+        raise AssertionError(
+            "approx() is not supported in a boolean context.\nDid you mean: `assert a == approx(b)`"
```
nicoddemus, line 105:
```suggestion
            "approx() is not supported in a boolean context.\nDid you mean: `assert a == approx(b)`?"
```

nicoddemus (approved):
Awesome, thanks @kalekundert!

kalekundert:
Ok, I think it's ready to merge.  Thanks for getting to this so fast!

src/_pytest/python_api.py
nicoddemus: Oh just realized, perhaps this should be `breaking`? It will break test suites which are misusing it.

changelog/9061.improvement.rst
```diff
@@ -0,0 +1,15 @@
+Using :func:`pytest.approx` in a boolean context now raises an error hinting at the proper usage.
```
nicoddemus, line 1: Oh just realized, perhaps this should be `breaking`? It will break test suites which are misusing it.

kalekundert: I wouldn't consider it breaking, because those test suites are broken already and the documented behavior of `approx()` is unchanged.  But I can see it both ways.

nicoddemus: Yeah I understand your point, but under our definition it would be a "breaking" change because it will break the test suites, even if the test suites are incorrect, so it is our policy to make it prominently in the CHANGELOG (We have done the same in the past, where we make an incorrect behavior an error but test suites were passing otherwise).

kalekundert: Ok, I'll change the file name, one sec...

PR #9070, opened 2021-09-02
Include figures in PDF docs

andrewdotn:
The PDF documentation on readthedocs was missing the figures in the
fixtures reference chapter. This PR uses a Sphinx plugin that
automatically converts the checked-in SVG files to the PDF input files
that LaTeX requires.

The SVG-to-PDF conversion is done by inkscape, which gave the best
conversion among the tools I tried. However, it [does not yet
understand][href-bug] that you can write a plain `href` instead of
`xlink:href` in svg files, so I’ve had to edit the SVG files
accordingly.

[href-bug]: https://github.com/TeX-Live/luatex.git

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,6 +7,10 @@ python:
       - method: pip
         path: .
 
+build:
+  apt_packages:
+    - inkscape
+
 formats:
   - epub
   - pdf
--- a/doc/en/conf.py
+++ b/doc/en/conf.py
@@ -55,6 +55,7 @@ extensions = [
     "sphinx.ext.viewcode",
     "sphinx_removed_in",
     "sphinxcontrib_trio",
+    "sphinxcontrib.inkscapeconverter",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/doc/en/example/fixtures/fixture_availability.svg
+++ b/doc/en/example/fixtures/fixture_availability.svg
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="572" height="542">
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="572" height="542">
     <style>
         text {
             font-family: 'Consolas', 'Menlo', 'DejaVu Sans Mono', 'Bitstream Vera Sans Mono', monospace;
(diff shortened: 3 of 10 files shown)


nicoddemus:
Hi @andrewdotn,

Thanks for tackling this issue.

Unfortunately I can't see the generated PDF files (the RTD build for PRs doesn't seem to generate them). Can you attach a generated PDF file so we can double check it before merging? Thanks!

andrewdotn:
Sure, here you go: [pytest.pdf](https://github.com/pytest-dev/pytest/files/7106488/pytest.pdf)

This was built in the `readthedocs/build:latest` docker container, so the file *should* be the same as what rtd will build. But I did have to manually do several setup steps, because I’m not aware of a script that goes directly from a `.readthedocs.yml` file to built documentation. As far as I can tell, the only thing that really knows how to do that is the big rtd django app itself. So there is a small chance that some difference will cause a problem, but I will definitely check the readthedocs site after merge to be sure it worked :)

On PDF generation in PRs, [readthedocs says](https://docs.readthedocs.io/en/stable/pull-requests.html#limitations) that, during builds from pull requests, “Additional formats like PDF and Epub aren’t built to produce results quicker.”

nicoddemus:
Awesome, thanks!

The figures look fine to me. 👍 

However the "contents" page is blank to me:

image


While the PDF from readthedocs is showing it fine:

image


Any idea what could have caused this?

andrewdotn:
Ah, yes, I know what happened:
  - You have to run LaTeX multiple times to get the TOC/index populated, because the first run writes out page number references to disk, and subsequent runs read those files.
  - There is a script that will do this automatically, but by default it gives up if any errors are encountered
  - There are multiple errors in the current sphinx-generated LaTeX file related to characters which are not in the default font. Some are emoji, and some are Cyrillic or CJK in the list of plugin descriptions. The Cyrillic is easy to fix (switch `latex_engine` from `pdflatex` to `xelatex`) but I thought it best to keep this PR focused on images. The emoji are trickier and [the only reasonable way I know to support them](https://tex.stackexchange.com/a/572220/2141) requires a newer version of `lualatex` than what comes with the Ubuntu 18.04 of the readthedocs docker images. That version of `lualatex` will take care of CJK characters easily too.
  - The version I sent you was a clean build with only one run LaTeX due to pre-existing errors, hence, no TOC
  - readthedocs still ends up with a TOC because it passes `-f` flag to `latexmk`

Here’s an updated [pytest.pdf](https://github.com/pytest-dev/pytest/files/7107122/pytest.pdf) created with the `-f` flag as readthedocs does.

nicoddemus (approved):
Thanks @andrewdotn!

Issue #7857, opened 2020-10-05
pytest-localserver not migrated from bitbucket?


oz123:
The documentation still links to it, but the original repository is gone.
Can you please restore the repository?

The-Compiler:
AFAIK it was a mercurial repository which was deleted by Bitbucket, so I don't think there's much we can do (other than upload a local copy to GitHub, if someone has one)

oz123:
I have a relatively an old copy of the repository which can be migrated to git (up to version ~0.3). Would like me to push it here?
Maybe someone has a newer copy_

bluetech:
The repositories have been archived, here is for example the old pytest one: https://bitbucket-archive.softwareheritage.org/projects/py/pytest-dev/pytest.html

The-Compiler:
yep, looks like it's here: https://bitbucket-archive.softwareheritage.org/projects/py/pytest-dev/pytest-localserver.html

There are also some other pytest projects there which didn't seem to make it (`curl-report`, maybe `pep8`, `quickcheck`, maybe more?). Dunno what we should do with those, given that nobody seems to care about them much... Maybe import to GitHub and set as archived?

bluetech:
If they had been under pytest-dev in bitbucket then that sounds like a good idea to me.

They would need to be converted from mercurial to git however.

nicoddemus:
> Maybe import to GitHub and set as archived?

Agreed. 👍 

We need some volunteer(s) to: migrate from Hg to Git, import/add to GitHub (just the repository, we can skip issues/PRs for now) and mark as archived.

diazona:
It would be great to get this going so that pytest-localserver could be updated with a newer certificate, at least. I've been getting errors of the form `ssl.SSLError: [SSL: CA_MD_TOO_WEAK] ca md too weak` with recent Python (3.8.11), related to the certificate included in pytest-localserver.

I can help with a one-time conversion, if that's what it takes, but at the moment I don't think I can make any long-term commitment to help maintain the project, so I don't want to step on the toes of someone who could do that.

kloczek:
So what happens with `pytest-localserver`?
Is it discontinued and whatever was using that module should be updated?
Any replacement?

diazona:
I converted the repository to Git using https://github.com/frej/fast-export and uploaded it to https://github.com/diazona/pytest-localserver. (The conversion _seems_ to have worked although I haven't gone through the history to verify it.)

[...]

PR #9092, opened 2021-09-14
Remove obsolete projects page from docs

nicoddemus:
The projects page unfortunately contains a lot of old projects, and some information there appears to be stale. 

It was created at a time pytest was starting to be used, but seems pointless now as pytest is the most popular testing framework, so listing a few dozen projects which use it seems pointless.

Rather than updating it, I think we should remove it because this has the chance to become stale again soon, and not sure how valuable it is anyway.

Decided to open this PR directly instead of an issue for discussion first.

(Triggered by #7857).


bluetech (approved):
Makes sense to me

PR #9104, opened 2021-09-19
Refactor change in PR#7687

alexharv074:
In https://github.com/pytest-dev/pytest/pull/7687 a bugfix was added
after NOTSET was changed to a singleton enum, which ended up breaking
ID generation when it checked for isinstance(Enum).

That change, however, did not update the code path in the same ID
generation when used as an idfn function.

This adjusts the logic to handle both cases.

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -1311,6 +1311,10 @@ def _idval(
     nodeid: Optional[str],
     config: Optional[Config],
 ) -> str:
+    if val is NOTSET:
+        # Fallback to default. Note that NOTSET is an enum.Enum.
+        return str(argname) + str(idx)
+
     if idfn:
         try:
             generated_id = idfn(val)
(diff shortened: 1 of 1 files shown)


alexharv074:
This PR is raised for discussion as much as a request to merge. As a full disclosure, I did not fully understand what this code I edited is doing, and exactly the nature of the bug I am fixing. Rather, I made an educated guess after seeing what was happening inside the debugger.

nicoddemus:
Hi @alexharv074,

Can you explain why the new code is better? They seem equivalent to me, with the exception that yours has a bit of extra duplication.

bluetech:
@alexharv074 I think a regression test (fails before your change, passes after) would help, both in understanding the issue and in making sure it won't repeat again.

Zac-HD:
Closing this PR as it's been inactive for almost two years.

PR #9112, opened 2021-09-22
scripts/update-plugin-list: escape summary for special RST chars

bluetech:
A new plugin has this summary:

    Continiously runs pytest on changes in *.py files

The `*` is interpreted as a special character and fails the CI.

Add some rudimentary escaping to hopefully prevent this.


nicoddemus (approved):
Thanks!

PR #9131, opened 2021-09-28
Update pastebin URL from bpaste to bpa.st

jeffra:
This fixes an issue that I am seeing when adding `--pastebin=failed` to pytest that results in a `bad response: HTTP Error 405: Not Allowed` error. Small repro below:

My environment:
* python 3.9.5 and also tried 3.6.10
* pytest 6.2.5
* Ubuntu 20.04.2 LTS and 18.04.5 LTS

test_fail.py:
```python
import pytest
def test_fail():
    assert False
```

cmd: `pytest --pastebin=failed --no-header test_fail.py`
output: 
```
================================= test session starts =================================
collected 1 item

test_fail.py F                                                                  [100%]

====================================== FAILURES =======================================
______________________________________ test_fail ______________________________________
tmpdir = local('/tmp/pytest-of-jerasley/pytest-13/test_fail0')

    def test_fail(tmpdir):
>       assert False
E       assert False

test_fail.py:4: AssertionError
======================== Sending information to Paste Service =========================
test_fail.py:4: AssertionError --> bad response: HTTP Error 405: Not Allowed
=============================== short test summary info ===============================
FAILED test_fail.py::test_fail - assert False
================================== 1 failed in 0.61s ==================================
```

It appears the bpaste URL has changed and is redirected to bpa.st. The redirect I think is causing issues with urllib. After updating the pastebin URL to https://bpa.st the results are the following:

```
================================= test session starts =================================
collected 1 item

test_fail.py F                                                                  [100%]

====================================== FAILURES =======================================
______________________________________ test_fail ______________________________________
tmpdir = local('/tmp/pytest-of-jerasley/pytest-14/test_fail0')

    def test_fail(tmpdir):
>       assert False
E       assert False

test_fail.py:4: AssertionError
======================== Sending information to Paste Service =========================
test_fail.py:4: AssertionError --> https://bpa.st/show/QEGQ
=============================== short test summary info ===============================
FAILED test_fail.py::test_fail - assert False
```

### DIFF
--- a/changelog/9131.bugfix.rst
+++ b/changelog/9131.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the URL used by ``--pastebin`` to use `bpa.st <http://bpa.st>`__.
--- a/src/_pytest/pastebin.py
+++ b/src/_pytest/pastebin.py
@@ -77,7 +77,7 @@ def create_new_paste(contents: Union[str, bytes]) -> str:
     from urllib.parse import urlencode
 
     params = {"code": contents, "lexer": "text", "expiry": "1week"}
-    url = "https://bpaste.net"
+    url = "https://bpa.st"
     try:
         response: str = (
             urlopen(url, data=urlencode(params).encode("ascii")).read().decode("utf-8")
--- a/testing/test_pastebin.py
+++ b/testing/test_pastebin.py
@@ -161,12 +161,12 @@ class TestPaste:
 
     def test_create_new_paste(self, pastebin, mocked_urlopen) -> None:
         result = pastebin.create_new_paste(b"full-paste-contents")
-        assert result == "https://bpaste.net/show/3c0c6750bd"
+        assert result == "https://bpa.st/show/3c0c6750bd"
         assert len(mocked_urlopen) == 1
         url, data = mocked_urlopen[0]
         assert type(data) is bytes
         lexer = "text"
-        assert url == "https://bpaste.net"
+        assert url == "https://bpa.st"
         assert "lexer=%s" % lexer in data.decode()
         assert "code=full-paste-contents" in data.decode()
         assert "expiry=1week" in data.decode()


nicoddemus (changes requested):
I personally did not know the URL has changed, thanks (it is not working for me at the moment, but I suppose that's transitory).

Could you please also add a CHANGELOG entry? Add `changelog/9131.bugfix.rst`, with the contents:

```rst
Fixed the URL used by ``--pastebin`` to use `bpa.st <http://bpa.st>`__.
```

jeffra:
> I wonder when it become broken. Do you use `--pastebin` regularly? When was the last time it worked?
> 
> You can guess where I'm going with this. The best way to know that a feature can be removed is for it to break silently and then see if anyone notices. Never let a good breakage go to waste :)
> 
> It always seemed odd to me that this particular plugin is in core, the functionality is pretty obscure. Also that it hardcodes a specific provider. Who is the owner of `bpaste.net`? This service has no privacy policy or security policy or any kind of attribution for that matter (whois is privacy protected). It's not unreasonable to imagine people submit private/sensitive test results there.

This is actually my first time using the `--pastebin` feature. I also thought it was a bit odd it was hard-coded to one provider haha :) I was just trying to figure out a good way to save the output of individual test failures that may be quite long in our case. Often if more than 1 test fails it becomes a bit cumbersome to find the relevant logs for each failure (see example: https://github.com/microsoft/DeepSpeed/runs/3738493444). I figured if we could simply add --pastebin it would be easy to click on the individual test failures to see the relevant logs one by one.

The-Compiler:
I think @supakeen (a regular and op in the `#python` IRC channel) runs bpaste and might want to comment. It's a pretty standard pastebin in various open source spaces as far as I'm aware, so probably not a bad choice.

Looks like the switch happened in c076f4e789ad8fa69560b21e40aa5cd49452ac01 by @RonnyPfannschmidt, and the original implementation comes from c17a09adafeb8bc36b70f182c904ce7bcfa0e97e back in 2009.

supakeen:
Hey there, thanks for tagging me. Being unable to paste with the old-style API on bpaste.net right now is due to my server configuration being broken. It's supposed to let a direct POST to bpaste.net through and not rewrite it to bpa.st. Either solution is fine for me (adjusting domain to bpa.st) or keeping it as is until I fix the server configuration.

I'll update here once the latter is done. My bad, backwards compatibility is a big thing for pinnwand (and the domains I host it on) so this is silly from me.

supakeen:
I've updated the configuration for the webservers, a direct post to bpaste.net is now passed onto the backend without a 302 or 405. This PR is still fine of course.

As a side note, pytest is using the super old endpoint from the LodgeIt days (paste.pocoo.org, etc) which is fine and supported but it'd be nicer to use the v1 api. This way you can provide users the removal link directly so they can undo pastes and you get the full URL back (plus useful errors):

https://pinnwand.readthedocs.io/en/latest/usage.html#v1

@jeffra could you try out both your old testcase and with the PR applied? Here's my output.

```
€ cat test_fail.py 
import pytest

def test_fail():
    assert False
€ venv/bin/pytest --pastebin=failed test_fail.py
===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.9.5, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- /home/user/venv/bin/python3.9
rootdir: /home/user
collected 1 item                                                                                                                                                                              

test_fail.py F                                                                                                                                                                          [100%]

========================================================================================== FAILURES ===========================================================================================
__________________________________________________________________________________________ test_fail __________________________________________________________________________________________

    def test_fail():
>       assert False
E       assert False

test_fail.py:4: AssertionError
============================================================================ Sending information to Paste Service =============================================================================
test_fail.py:4: AssertionError --> https://bpaste.net/show/SYJQ
=================================================================================== short test summary info ===================================================================================
FAILED test_fail.py::test_fail - assert False
====================================================================================== 1 failed in 0.24s ======================================================================================
```

nicoddemus (approved):
LGTM.

@bluetech do you still have concerns about the service?

bluetech (approved):
The PR LGTM, even though the old URL now works, IIUC the new URL is the preferred one.

Regarding the plugin itself, I might open a separate issue some time, but it should not hold up this PR.

nicoddemus:
Thanks everyone!

PR #9132, opened 2021-09-29
doc/en/reference/plugin_list.rst typos fix

axiaoxin:
fixed typos in doc/en/reference/plugin_list.rst


bluetech:
Thanks @axiaoxin. The plugin list is automatically generated from PyPI, so if we merge this, it will just be overridden on the next update. If you'd like, you can submit the fixes to each individual package (usually the description is in `setup.py` or `setup.cfg`), and once they make a release with the fix, it will be picked up in the plugin list.

PR #9140, opened 2021-09-30
Fix an issue with importlib spec_from_file_location

### LINKED ISSUE #8918: pytest._pathlib.import_path should pass absolute paths on Python>=3.8 <3.10
In https://bugs.python.org/issue44070, `spec_from_file_location()` has changed its behaviour so that the responsibility of passing absolute paths is on the caller between 3.8 and 3.10

So something like this is needed:

```python
    if (3, 8) <= sys.version_info[:2] < (3, 10) and not os.path.isabs(fn):
        # module.__spec__.__file__ is supposed to be absolute in py3.8+
        # importlib.util.spec_from_file_location does this automatically from
        # 3.10+
        # This was backported to 3.8 and 3.9, but then reverted in 3.8.11 and
        # 3.9.6
        # See https://twistedmatrix.com/trac/ticket/10230
        # and https://bugs.python.org/issue44070
        fn = os.path.join(os.getcwd(), fn)
```
https://github.com/pytest-dev/pytest/blob/d200598de9588fe6d7a68bddc3e638afce80f153/src/_pytest/pathlib.py#L500

sgaist:
As explained in issue #8918, in https://bugs.python.org/issue44070, spec_from_file_location() has changed its behaviour so that the responsibility of passing absolute paths is on the caller between 3.8 and 3.10.

Since the code already uses Path, I choose to use resolve to get the full path in place of the suggested solution.

closes #8918

### DIFF
--- a/src/_pytest/pathlib.py
+++ b/src/_pytest/pathlib.py
@@ -497,6 +497,16 @@ def import_path(
             if spec is not None:
                 break
         else:
+            if (3, 8) <= sys.version_info[:2] < (3, 10) and not path.is_absolute():
+                # module.__spec__.__file__ is supposed to be absolute in py3.8+
+                # importlib.util.spec_from_file_location does this automatically from
+                # 3.10+
+                # This was backported to 3.8 and 3.9, but then reverted in 3.8.11 and
+                # 3.9.6
+                # See https://twistedmatrix.com/trac/ticket/10230
+                # and https://bugs.python.org/issue44070
+                path = path.resolve()
+
             spec = importlib.util.spec_from_file_location(module_name, str(path))
 
         if spec is None:


sgaist:
I am wondering whether this should also have an entry in the change log since it should not trigger any behaviour change.

bluetech:
The issue is not very clear to me, so I can't give a proper review. I think it would help to demonstrate an actual breakage due to this, otherwise maybe better to just leave it as-is? I mean, it has worked thus far with no complaints, and this code deals with a triple edge-case (`importlib` import mode, `sys.meta_path` is empty, `path` is relative) which is not covered by any tests.

nicoddemus:
In the end this patch is not needed because the paths given to `import_path` are always absolute anyway. This can be verified because the new line was never hit by our tests, but I also added `assert self.path.is_absolute()` to `PyCollector.__init__` to verify that.

So closing this as "not necessary".

Nevertheless thanks @sgaist for the contribution!

PR #9147, opened 2021-10-01
main: avoid Path(Path(...)) calls, they're slow

bluetech:
See for reference #9125. From my measurements it is not *that* extreme, however I can confirm it shaves off ~5s from pandas collection time.

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -538,14 +538,18 @@ class Session(nodes.FSCollector):
     pytest_collectreport = pytest_runtest_logreport
 
     def isinitpath(self, path: Union[str, "os.PathLike[str]"]) -> bool:
-        return Path(path) in self._initialpaths
+        # Optimization: Path(Path(...)) is much slower than isinstance.
+        path_ = path if isinstance(path, Path) else Path(path)
+        return path_ in self._initialpaths
 
     def gethookproxy(self, fspath: "os.PathLike[str]"):
+        # Optimization: Path(Path(...)) is much slower than isinstance.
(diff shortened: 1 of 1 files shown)


src/_pytest/main.py
```diff
@@ -538,14 +538,18 @@ def pytest_runtest_logreport(
...
 
     def isinitpath(self, path: Union[str, "os.PathLike[str]"]) -> bool:
-        return Path(path) in self._initialpaths
+        # Optimization: Path(Path(...)) is much slower than isinstance.
+        path_ = path if isinstance(path, Path) else Path(path)
+        return path_ in self._initialpaths
 
     def gethookproxy(self, fspath: "os.PathLike[str]"):
+        # Optimization: Path(Path(...)) is much slower than isinstance.
+        path = fspath if isinstance(fspath, Path) else Path(fspath)
```
nicoddemus, line 547: Perhaps create a `make_path` function (or something like that)? I understand only two places do this right now, however having a function brings some advantages:

1. We have place to put that comment about optimization.
2. Encourages others in the future to use the function, instead of copying/pasting the same two lines.

nicoddemus (approved):

src/_pytest/main.py
bluetech: I thought about it, but decided against it:

1. I think in most cases it is not performance critical, so would just obfuscate the code
2. The function call itself might have some overhead

So I just went with repeating in these two places. However, if there is ever [a third](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) we probably should.

PR #9149, opened 2021-10-01
Updated logging.py for #9146

### LINKED ISSUE #9146: Documentation for caplog `set_level()` should make it clearer it's a threshold

The [documentation for `caplog.set_level()`](https://docs.pytest.org/en/stable/reference.html#pytest.LogCaptureFixture.set_level) says
> Set the level of a logger for the duration of a test.

This does not make it clear that it's setting the _threshold_ level for capturing logging.

The Python logging documentation makes this clear for `logging.setLevel()`:
> Sets the threshold for this logger to level. Logging messages which are less severe than level will be ignored; ...

`caplog.set_level()` ought to do the same.

#### Additional context

The only reason I know it's a threshold is because I tried running `pytest -rA` with different levels set in caplog and the output contained messages at that level and above

Vijay-Arora:
Close #9146

### DIFF
--- a/changelog/9146.doc.rst
+++ b/changelog/9146.doc.rst
@@ -1,1 +1,1 @@
-Documentation improvement for #9146
+Improve Documentation for `caplog.set_level`.
--- a/changelog/9146.doc.rst
+++ b/changelog/9146.doc.rst
@@ -1,0 +1,1 @@
+Documentation improvement for #9146
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -447,8 +447,8 @@ class LogCaptureFixture:
         self.handler.reset()
 
     def set_level(self, level: Union[int, str], logger: Optional[str] = None) -> None:
-        """Sets the threshold for this logger to level. Logging messages which are less 
-        severe than level will be ignored; logging messages which have severity level or 
+        """Sets the threshold for this logger to level. Logging messages which are less
+        severe than level will be ignored; logging messages which have severity level or
         higher will be emitted by whichever handler or handlers service this logger, unless
         a handler’s level has been set to a higher severity level than level.
 
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -447,12 +447,9 @@ class LogCaptureFixture:
         self.handler.reset()
 
     def set_level(self, level: Union[int, str], logger: Optional[str] = None) -> None:
-        """Sets the threshold for this logger to level.
+        """Set the threshold level of a logger for the duration of a test.
 
-        Logging messages which are less severe than level will be ignored; logging messages
-        which have severity level or higher will be emitted by whichever handler or handlers
-        service this logger, unless a handler’s level has been set to a higher severity level
-        than `level`.
+        Logging messages which are less severe than this level will not be captured.
 
         .. versionchanged:: 3.4
             The levels of the loggers changed by this function will be
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -447,11 +447,11 @@ class LogCaptureFixture:
         self.handler.reset()
 
     def set_level(self, level: Union[int, str], logger: Optional[str] = None) -> None:
-        """Sets the threshold for this logger to level. 
-        
-        Logging messages which are less severe than level will be ignored; logging messages 
-        which have severity level or higher will be emitted by whichever handler or handlers 
-        service this logger, unless a handler’s level has been set to a higher severity level 
+        """Sets the threshold for this logger to level.
+
+        Logging messages which are less severe than level will be ignored; logging messages
+        which have severity level or higher will be emitted by whichever handler or handlers
+        service this logger, unless a handler’s level has been set to a higher severity level
         than `level`.
 
         .. versionchanged:: 3.4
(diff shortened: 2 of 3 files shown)


nicoddemus:
Tip: you can add `Closes #9146` to the PR description, which serves both to link the PR to the issue, and GitHub will close the issue automatically after the PR is merged (just added it myself). 👍

changelog/9146.doc.rst
```diff
@@ -0,0 +1 @@
+Documentation improvement for #9146
```
nicoddemus, line 1: No need to add the issue number to the description, this will be added automatically later. Better to describe the change itself.

```suggestion
Improve documentation for `caplog.set_level`.
```

src/_pytest/logging.py
```diff
@@ -447,7 +447,10 @@ def clear(self) -> None:
         self.handler.reset()
 
     def set_level(self, level: Union[int, str], logger: Optional[str] = None) -> None:
-        """Set the level of a logger for the duration of a test.
+        """Sets the threshold for this logger to level. Logging messages which are less
+        severe than level will be ignored; logging messages which have severity level or
+        higher will be emitted by whichever handler or handlers service this logger, unless
+        a handler’s level has been set to a higher severity level than level.
```
nicoddemus, line 453:
```suggestion
        """
        Sets the threshold for this logger to level. 
        
        Logging messages which are less severe than level will be ignored; logging messages 
        which have this severity level or higher will be emitted by whichever handler or handlers service
        this logger, unless a handler’s level has been set to a higher severity level than level.
```

Vijay-Arora:
Thanks! But I see there are some checks which I see failed for eg. At least 1 approving review is required by reviewers with write access. What should I do in this case? Should I update the PR or let it be?

nicoddemus (changes requested):
Thanks @Vijay-Arora! Left a few request for changes, please take a look.

cc @MetRonnie

changelog/9146.doc.rst
Vijay-Arora: Thank you!

Vijay-Arora:
> Thanks @Vijay-Arora! Left a few request for changes, please take a look.
> 
> cc @MetRonnie

Hey @nicoddemus, should I go ahead and make the requested changes and open a new PR for this?

nicoddemus:
> Hey @nicoddemus, should I go ahead and make the requested changes and open a new PR for this? 

Just push the commits to the same branch and the PR will update automatically.

> Also, I see some failures, do I need to make some changes for that too?

Hmm ignore them for now, they don't seem related to your PR, I will investigate those later.

Vijay-Arora:
> > Hey @nicoddemus, should I go ahead and make the requested changes and open a new PR for this?
> 
> Just push new commits to the same branch and the PR will update automatically.
> 
> > Also, I see some failures, do I need to make some changes for that too?
> 
> Hmm ignore them for now, they don't seem related to your PR, I will investigate those later.

Cool, thanks, Making the changes now.

src/_pytest/logging.py
```diff
@@ -447,7 +447,12 @@ def clear(self) -> None:
         self.handler.reset()
 
     def set_level(self, level: Union[int, str], logger: Optional[str] = None) -> None:
-        """Set the level of a logger for the duration of a test.
+        """Sets the threshold for this logger to level.
```
MetRonnie, line 450: In my opinion, the original tagline was slightly clearer, but just needed to mention it's a threshold.
```suggestion
        """Set the threshold level of a logger for the duration of a test.
```

src/_pytest/logging.py
```diff
@@ -447,7 +447,12 @@ def clear(self) -> None:
...
+
+        Logging messages which are less severe than level will be ignored; logging messages
+        which have severity level or higher will be emitted by whichever handler or handlers
+        service this logger, unless a handler’s level has been set to a higher severity level
+        than `level`.
```
MetRonnie, line 455: I'm not sure some of these details from the logging module make sense in the context of caplog. My suggestion is keep it short and simple 😃 
```suggestion
        Logging messages which are less severe than this level will not be captured.
```

MetRonnie (commented):
Thanks for taking this on! As the author of the original issue, I've added my thoughts

src/_pytest/logging.py
nicoddemus: Agree, good points. @Vijay-Arora could you implement the change plase?

MetRonnie:
@nicoddemus I do wonder if the docstring for [`at_level`](https://docs.pytest.org/en/stable/reference.html#pytest.LogCaptureFixture.at_level) should be updated at the same time, but I don't know whether the behaviour is also a threshold or not, I've never used it.

src/_pytest/logging.py
Vijay-Arora: Hi @nicoddemus and @MetRonnie , Thanks for the feedback; I'll definitely make the changes.

ziebam:
@MetRonnie Good catch. As far as I understand it, the behavior is the same for [`at_level`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.LogCaptureFixture.at_level) as it also utilizes the [`setLevel`](https://docs.python.org/3/library/logging.html#logging.Handler.setLevel) method, therefore an update for its docs seems reasonable.

@Vijay-Arora Would you be interested in driving this one home? 😄

src/_pytest/logging.py
thisch: @Vijay-Arora could you implement the change please?

Zac-HD (approved):
Let's get this merged!

PR #9150, opened 2021-10-01
Add note about fixture teardown and SIGTERM and SIGQUIT signals

### LINKED ISSUE #9141: Make it clear in the documentation that fixtures are not suitable for managing resource lifecycles
Partitioned off of https://github.com/pytest-dev/pytest/issues/5243

Proposal: Make it clear in the pytest documentation that users are responsible for resource lifecycle management. 

Currently, [there are misleading examples](https://docs.pytest.org/en/latest/how-to/fixtures.html#parametrizing-fixtures) such as:

```
@pytest.fixture(scope="module", params=["smtp.gmail.com", "mail.python.org"])
def smtp_connection(request):
    smtp_connection = smtplib.SMTP(request.param, 587, timeout=5)
    yield smtp_connection
    print("finalizing {}".format(smtp_connection))
    smtp_connection.close()
```

that suggest otherwise. The above fixture from the documentation is highly suggestive that pytest believes fixtures are appropriate to use for resource management; however, [pytest is only a test runner and not a resource manager](https://github.com/pytest-dev/pytest/issues/5243#issuecomment-931629005) according to @asottile

A documentation change that makes it clear that pytest fixtures are not fit for the purpose of handling resource lifecycles in all cases (user beware). There are likely other places in the documentation that require review concerning this point.

Also, it would be useful to all users to document common ways a user can handle resource lifecycles, especially for external resources that could be leaked, that are known to work smoothly with pytest's framework.

#### Alternative Solutions
None

#### Additional context
Identified during discussions in:

- https://github.com/pytest-dev/pytest/issues/5243

nicoddemus:
Close #9141

cc @gsmethells

### DIFF
--- a/doc/en/explanation/fixtures.rst
+++ b/doc/en/explanation/fixtures.rst
@@ -156,3 +156,20 @@ Another good approach is by adding the data files in the ``tests`` folder.
 There are also community plugins available to help to manage this aspect of
 testing, e.g. `pytest-datadir <https://pypi.org/project/pytest-datadir/>`__
 and `pytest-datafiles <https://pypi.org/project/pytest-datafiles/>`__.
+
+.. _fixtures-signal-cleanup:
+
+A note about fixture cleanup
+----------------------------
+
+pytest does not do any special processing for :data:`SIGTERM <signal.SIGTERM>` and
+:data:`SIGQUIT <signal.SIGQUIT>` signals (:data:`SIGINT <signal.SIGINT>` is handled naturally
+by the Python runtime via :class:`KeyboardInterrupt`), so fixtures that manage external resources which are important
+to be cleared when the Python process is terminated (by those signals) might leak resources.
+
+The reason pytest does not handle those signals to perform fixture cleanup is that signal handlers are global,
+and changing them might interfere with the code under execution.
+
+If fixtures in your suite need special care regarding termination in those scenarios,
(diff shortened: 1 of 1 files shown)


gsmethells:
@nicoddemus LGTM 👍🏻

doc/en/explanation/fixtures.rst
```diff
@@ -156,3 +156,20 @@ Another good approach is by adding the data files in the ``tests`` folder.
...
+pytest does not do any special processing for :data:`SIGTERM <signal.SIGTERM>` and
+:data:`SIGQUIT <signal.SIGQUIT>` signals (:data:`SIGINT <signal.SIGINT>` is handled naturally
+by the Python runtime via :class:`KeyboardInterrupt`), so fixtures that manage external resources which are important
+to be cleared when the Python process is terminated (by those signals) might leak resources.
+
+The reason pytest does not handle those signals to perform fixture cleanup is that signal handlers are global,
+and changing them might interfere with the code under execution.
+
+If fixtures in your suite need special care regarding termination in those scenarios,
+see `this comment <https://github.com/pytest-dev/pytest/issues/5243#issuecomment-491522595>`__ in the issue
```
nicoddemus, line 174: The idea is to change this to a link to a plugin if one gets published.

bluetech (approved):

PR #9152, opened 2021-10-02
mark/structures: slightly optimize some functions

bluetech:
`normalize_mark_list` shows up in pandas collection profiles. It's simple enough to improve.


RonnyPfannschmidt (approved):
looks good,
should we plan to add some micro benchmarks for this type of thing?

bluetech:
>  should we plan to add some micro benchmarks for this type of thing?

Will be nice to set up some benchmark suite which can be periodically run to find regressions, or to evaluate perf changes. I might work on it when I have some time.

PR #9158, opened 2021-10-03
replace hardcoded urls to docs.pytest.org with internal crossrefs, add crossrefs to numpy docs, use sphinx pep role

hoefling:
This PR replaces all hardcoded links to https://docs.pytest.org with internal crossreferences (with missing labels being added). Also, hardcoded links to `numpy` docs were replaced by intersphinx crossrefs. Lastly, all hardcoded references to PEP pages were replaced by usages of Sphinx' `:pep:` role.

This is the second PR in series, the first one (#9041) replaced hardcoded URLs to stdlib. Also probably #9081 is partially addressed by this.

Apologies for the PR size. This was mostly a semi-automated search-and-replace, with manually fixing broken links afterwards.

### DIFF
--- a/doc/en/announce/release-2.1.0.rst
+++ b/doc/en/announce/release-2.1.0.rst
@@ -11,8 +11,8 @@ The single biggest news about this release are **perfected assertions**
 courtesy of Benjamin Peterson.  You can now safely use ``assert``
 statements in test modules without having to worry about side effects
 or python optimization ("-OO") options.  This is achieved by rewriting
-assert statements in test modules upon import, using a PEP302 hook.
-See https://docs.pytest.org/en/stable/how-to/assert.html for
+assert statements in test modules upon import, using a :pep:`302` hook.
+See :doc:`../how-to/assert` for
 detailed information.  The work has been partly sponsored by my company,
 merlinux GmbH.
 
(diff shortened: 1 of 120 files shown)


Zac-HD (commented):
I think we should consider leaving the `announce` changes for later, since they might be used as plain text as well as built with Sphinx.

Otherwise looks good to me 👍

hoefling:
@Zac-HD sure thing, reverted the changes in `announce` dir!

Zac-HD (approved):
Looks like a solid improvement on the status quo, so let's merge!  

Thanks again @hoefling, we really appreciate your work on this - docs and cleanup are often neglected, but are absolutely essential for users 🥰

Zac-HD:
If you're after a next step, maybe `extlinks` for issue links (etc)?  [Here's the setup we use for Hypothesis](https://github.com/HypothesisWorks/hypothesis/blob/710699c3cdbfcb3a342de08f426c1c8906c66cba/hypothesis-python/docs/conf.py#L97-L112).

hoefling:
@Zac-HD yep, this is definitely something I‘d like to tackle on next. Although I haven‘t used `extlinks` extensively so far, so will look into it first - maybe this extension should also get a checking posttransform similar to what was proposed for `intersphinx` in Sphinx lately.

PR #9166, opened 2021-10-05
nodes: micro-optimize _imply_path

bluetech:
This reduces pandas collection time by a couple of seconds  (94081513 function calls -> 88398561 function calls).

### DIFF
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -93,24 +93,38 @@ def iterparentnodeids(nodeid: str) -> Iterator[str]:
         yield nodeid
 
 
+def _check_path(path: Path, fspath: LEGACY_PATH) -> None:
+    if Path(fspath) != path:
+        raise ValueError(
+            f"Path({fspath!r}) != {path!r}\n"
+            "if both path and fspath are given they need to be equal"
+        )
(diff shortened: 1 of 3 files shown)


src/_pytest/nodes.py
```diff
@@ -93,26 +93,40 @@ def iterparentnodeids(nodeid: str) -> Iterator[str]:
...
             fspath = legacy_path(path)
         return path, fspath
-
     else:
         assert fspath is not None
         return Path(fspath), fspath
 
 
+# Optimization: use _imply_path_only over _imply_path when only need Path.
+#               This is to avoid `legacy_path(path)` which is suprisngly heavy.
```
nicoddemus, line 119:
```suggestion
#               This is to avoid `legacy_path(path)` which is surprisingly heavy.
```

nicoddemus (approved):

src/_pytest/nodes.py
bluetech: Fixed, thanks. I also added a bit to a test which would hopefully fix the missing coverage.

RonnyPfannschmidt (approved):
fabulous  :rocket:

PR #9170, opened 2021-10-06
Rewrite only modules part of the pytest plugin

gaborbernat:
The idea is to only rewrite packages that contain pytest plugins and not other packages in the same namespace. 

PS. The tests are failing but will fix it if we think this is a good direction to go in. 

Could fix https://github.com/pytest-dev/pytest/issues/2419


nicoddemus:
Ahh I see, thanks for the clarification! 🙇

gaborbernat:
I'll close this for now.

PR #9241, opened 2021-10-27
Make plugin list in PDF docs readable

andrewdotn:
The current PDF docs attempt to format the list of all plugins as a
table, without any word-wrapping of the plugin description. That results
in almost all the information getting cut off. This PR formats the same
information into more of a paragraph format for the PDF, with nothing
cut off.

### DIFF
--- a/scripts/update-plugin-list.py
+++ b/scripts/update-plugin-list.py
@@ -1,7 +1,8 @@
 import datetime
 import pathlib
 import re
-from textwrap import dedent, indent
+from textwrap import dedent
+from textwrap import indent
 
 import packaging.version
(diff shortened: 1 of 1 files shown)


wimglenn:
@andrewdotn  I wonder if you have any ideas to make this table render better? This is what it looks like to me in Chrome browser / macOS - "last release" column way too narrow, and there's a lot of wrapping with hyphenation characters for some reason rendering as a box. Maybe the "status" and/or "requires" column(s) could be dropped, they don't have much useful info and it would allow the others to be wider.

Screen Shot 2021-12-01 at 12 29 36 AM

PR #9243, opened 2021-10-27
minor tweaks for terminal.py

symonk:
Some minor tweaks to the `terminal.py` module, namely:

 - Removal of unused args for `mywriter` and `_locationline` (tags and lineno respectively), neither appear used
 - Some extra typing in places
 - Change `getreports` to use a list comp
 - Simplify `_getfailureheadline`

edit: Other merged PRs have done some of the same work here, so its quite miniscule now


src/_pytest/terminal.py
```diff
@@ -231,7 +231,7 @@ def pytest_configure(config: Config) -> None:
     config.pluginmanager.register(reporter, "terminalreporter")
     if config.option.debug or config.option.traceconfig:
 
-        def mywriter(tags, args):
```
bluetech, line 234: The `tags` arg is probably required because that's how `setprocessor` calls it.

src/_pytest/terminal.py
```diff
@@ -511,7 +511,9 @@ def pytest_runtest_logstart(
         # Ensure that the path is printed before the
         # 1st test of a module starts running.
         if self.showlongtestinfo:
-            line = self._locationline(nodeid, *location)
```
bluetech, line 514: I guess the idea was to pass the entire location, even if one part of it is currently unused?

src/_pytest/terminal.py
```diff
@@ -879,16 +883,13 @@ def mkrel(nodeid):
             if self.verbosity >= 2 and nodeid.split("::")[0] != fspath.replace(
                 "\\", nodes.SEP
             ):
-                res += " <- " + bestrelpath(self.startpath, fspath)
+                res += " <- " + bestrelpath(self.startpath, Path(fspath))
```
bluetech, line 886: Note: this is actually a bug fix, we are working on it in #9238

symonk: interesting, thanks

src/_pytest/terminal.py
symonk: ah good spot which is pluggy-land I think, will revert :+1:

src/_pytest/terminal.py
symonk: Yeah I wasen't 100% since the `lineno` was never used later, would it be preferred to keep passing `*location` here even if `lineno` is completely unused? happy to revert that too

src/_pytest/terminal.py
bluetech: #9238 was merged; can you rebase?

src/_pytest/terminal.py
```diff
@@ -879,16 +881,13 @@ def mkrel(nodeid):
...
         else:
             res = "[location]"
         return res + " "
 
-    def _getfailureheadline(self, rep):
-        head_line = rep.head_line
```
bluetech, line 888: Looking again, this code looks like it was written this way so that the `head_line` property is only evaluated once. Probably doesn't matter, but I'll keep it this way.

src/_pytest/terminal.py
```diff
@@ -879,16 +881,13 @@ def mkrel(nodeid):
...
-        if head_line:
-            return head_line
-        return "test session"  # XXX?
```
bluetech, line 891: We should keep this XXX comment - I don't know what it refers to but it would probably help *someone* looking at this code.

symonk:
Thanks @bluetech between some reverts & the other PR outlined this ended up being tiny, sorry to take your time for such trivial changes in the end.  Appreciate it as always

bluetech (approved):

PR #9248, opened 2021-10-29
doc: support sphinx 5

bluetech:
Sphinx 4 (or releases since) provides two nice things (see [changelog](https://www.sphinx-doc.org/en/master/changes.html)):

- It supports "canonical" names for types. I'm not sure exactly what it is, but makes a lot more internal type annotations references resolve in the API Reference. Previously sphinx was confused by the `pytest.*` reexports from `_pytest`.
- ~~`python_use_unqualified_type_names` config which makes the API Reference signatures a little less noisy.~~

### DIFF
--- a/doc/en/requirements.txt
+++ b/doc/en/requirements.txt
@@ -2,9 +2,6 @@ pallets-sphinx-themes
 pluggy>=1.0
 pygments-pytest>=2.2.0
 sphinx-removed-in>=0.2.0
-sphinx>=3.1,<4
+sphinx>=5,<6
 sphinxcontrib-trio
 sphinxcontrib-svg2pdfconverter
-
-# XXX: sphinx<4 is broken with latest jinja2
-jinja2<3.1
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -20,6 +20,16 @@ from typing import Union
 import attr
 import py
 
+# fmt: off
+# Workaround for https://github.com/sphinx-doc/sphinx/issues/10351.
+# If `overload` is imported from `compat` instead of from `typing`,
+# Sphinx doesn't recognize it as `overload` and the API docs for
+# overloaded functions look good again. But type checkers handle
+# it fine.
+# fmt: on
+if True:
+    from typing import overload as overload
+
 if TYPE_CHECKING:
     from typing_extensions import Final
 
@@ -345,7 +355,6 @@ else:
 if sys.version_info >= (3, 8):
     from functools import cached_property as cached_property
 else:
-    from typing import overload
     from typing import Type
 
     class cached_property(Generic[_S, _T]):
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -37,6 +37,8 @@ from _pytest.terminal import TerminalReporter
 
 if TYPE_CHECKING:
     logging_StreamHandler = logging.StreamHandler[StringIO]
+
+    from typing_extensions import Literal
 else:
     logging_StreamHandler = logging.StreamHandler
 
@@ -382,20 +384,19 @@ class LogCaptureFixture:
 
     @property
     def handler(self) -> LogCaptureHandler:
-        """Get the logging handler used by the fixture.
-
-        :rtype: LogCaptureHandler
-        """
+        """Get the logging handler used by the fixture."""
         return self._item.stash[caplog_handler_key]
 
-    def get_records(self, when: str) -> List[logging.LogRecord]:
+    def get_records(
+        self, when: "Literal['setup', 'call', 'teardown']"
+    ) -> List[logging.LogRecord]:
         """Get the logging records for one of the possible test phases.
 
-        :param str when:
-            Which test phase to obtain the records from. Valid values are: "setup", "call" and "teardown".
+        :param when:
+            Which test phase to obtain the records from.
+            Valid values are: "setup", "call" and "teardown".
 
         :returns: The list of captured records at the given stage.
-        :rtype: List[logging.LogRecord]
 
         .. versionadded:: 3.4
         """
@@ -452,8 +453,8 @@ class LogCaptureFixture:
             The levels of the loggers changed by this function will be
             restored to their initial values at the end of the test.
 
-        :param int level: The level.
-        :param str logger: The logger to update. If not given, the root logger.
+        :param level: The level.
+        :param logger: The logger to update. If not given, the root logger.
         """
         logger_obj = logging.getLogger(logger)
         # Save the original log-level to restore it during teardown.
@@ -471,8 +472,8 @@ class LogCaptureFixture:
         the end of the 'with' statement the level is restored to its original
         value.
 
-        :param int level: The level.
-        :param str logger: The logger to update. If not given, the root logger.
+        :param level: The level.
+        :param logger: The logger to update. If not given, the root logger.
         """
         logger_obj = logging.getLogger(logger)
         orig_level = logger_obj.level
--- a/doc/en/conf.py
+++ b/doc/en/conf.py
@@ -38,6 +38,7 @@ release = ".".join(version.split(".")[:2])
 
 autodoc_member_order = "bysource"
 autodoc_typehints = "description"
+autodoc_typehints_description_target = "documented"
 todo_include_todos = 1
 
 latex_engine = "lualatex"
--- a/doc/en/conf.py
+++ b/doc/en/conf.py
@@ -162,11 +162,11 @@ linkcheck_workers = 5
 
 _repo = "https://github.com/pytest-dev/pytest"
 extlinks = {
-    "bpo": ("https://bugs.python.org/issue%s", "bpo-"),
-    "pypi": ("https://pypi.org/project/%s/", ""),
-    "issue": (f"{_repo}/issues/%s", "issue #"),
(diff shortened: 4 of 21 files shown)


hoefling:
@bluetech 
> Hmm just noticed a regression of sorts - now sphinx shows a bunch of private constructors (e.g. [here](https://pytest--9248.org.readthedocs.build/en/9248/reference/reference.html#pytest.Cache) - search for `_ispytest` in the API Reference) where previously it didn't.

That's because [Sphinx added new flag for rendering typehints](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints_description_target) which is set to `"all"`, so suddenly all arguments start to be documented :-D Setting it to `autodoc_typehints_description_target = "documented"` and explicitly stating what init parameters are documented in the class docstring should help. Example: this diff

```diff
--- a/doc/en/conf.py
+++ b/doc/en/conf.py
@@ -38,6 +38,7 @@ release = ".".join(version.split(".")[:2])
 
 autodoc_member_order = "bysource"
 autodoc_typehints = "description"
+autodoc_typehints_description_target = "documented"
 todo_include_todos = 1
 
 latex_engine = "lualatex"
--- a/doc/en/reference/reference.rst
+++ b/doc/en/reference/reference.rst
@@ -347,7 +347,7 @@ Under the hood, the cache plugin uses the simple
 
 ``config.cache`` is an instance of :class:`pytest.Cache`:
 
-.. autoclass:: pytest.Cache()
+.. autoclass:: pytest.Cache(cachedir, config)
    :members:
 
 
--- a/src/_pytest/cacheprovider.py
+++ b/src/_pytest/cacheprovider.py
@@ -55,6 +55,10 @@ Signature: 8a477f597d28d172789f06886806bc55
 @final
 @attr.s(init=False, auto_attribs=True)
 class Cache:
+    """
+    :param Path cachedir:
+    :param Config config:
+    """
     _cachedir: Path = attr.ib(repr=False)
     _config: Config = attr.ib(repr=False)
```
renders
image

hoefling:
Same is with fixtures, e.g. if I create an incomplete info field list for `capsys` via
```diff
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -906,6 +906,9 @@ def capsys(request: SubRequest) -> Generator[CaptureFixture[str], None, None]:
     The captured output is made available via ``capsys.readouterr()`` method
     calls, which return a ``(out, err)`` namedtuple.
     ``out`` and ``err`` will be ``text`` objects.
+
+    :return: an instance of :class:`CaptureFixture[str] <.CaptureFixture>`.
+    :rtype: :class:`CaptureFixture[str] <.CaptureFixture>`
     """
     capman = request.config.pluginmanager.getplugin("capturemanager")
     capture_fixture = CaptureFixture[str](SysCapture, request, _ispytest=True)
```
(both fields are not needed and set just for demo) renders

image

What's not nice is that the info field list is part of the docstring, so the usage example is shifted down - the example could be moved to the docstring though and placed where desired.

bluetech:
@hoefling Thanks for looking at this.

> Setting it to autodoc_typehints_description_target = "documented" and explicitly stating what init parameters are documented in the class docstring should help.

Seems like we're going to have to do this. It would take some work -- often we omit `:param:` for "obvious" parameters/short functions.

---

Another issue with sphinx 4 I noticed is that it now always(?) renders the full signatures of overloaded functions, like `pytest.raise` and others. These were the reason we switched to `autodoc_typehints = "description"` in the first place (see #8405) so that's also something we need to fix. Maybe the right solution it to just not use autodoc for those.

hoefling:
> Seems like we're going to have to do this. It would take some work -- often we omit `:param:` for "obvious" parameters/short functions.

This is something I could tackle, using https://docs.pytest.org/en/stable/ as a reference.

> Another issue with sphinx 4 I noticed is that it now always(?) renders the full signatures of overloaded functions, like `pytest.raise` and others. These were the reason we switched to `autodoc_typehints = "description"` in the first place (see #8405) so that's also something we need to fix. Maybe the right solution it to just not use autodoc for those.

I think this is an issue with Sphinx 3 as well, only that Sphinx 4 now also includes full qualnames for `TypeVar`s, making a `Type[_pytest.python_api.E]` out of `Type[_E]`. I will see what I can dig out, I think [aliasing types](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases) in Sphinx conf should mitigate that already.

bluetech:
Hmm Sphinx 3 no longer works in Python 3.10 (`from types import Union`) so this becomes more urgent -- I should get back to this.

The-Compiler:
@bluetech Anything I can help to get this in? Archlinux has [updated](https://github.com/archlinux/svntogit-packages/commit/daa634fe77bee66b804f3adc75a6672fbbf906d2) to Python 3.10 in December, and having to edit `tox.ini` every time I work on the docs is somewhat cumbersome...

doc/en/requirements.txt
```diff
@@ -2,6 +2,6 @@ pallets-sphinx-themes
 pluggy>=1.0
 pygments-pytest>=2.2.0
 sphinx-removed-in>=0.2.0
-sphinx>=3.1,<4
+sphinx>=3.1,<5
```
hugovk, line 5: Maybe bump the minimum as well? 4.2.0 is the first to support Python 3.10:

```suggestion
sphinx>=4.2.0,<5
```

And this can also be removed (need to rebase):

https://github.com/pytest-dev/pytest/blob/b4ab2f094234a0e5a1b14cf8e6d7cd9f48ecbadb/doc/en/requirements.txt#L9-L10

And should we bump Python 3.10 for `docs` and `doctesting` on the CI?

https://github.com/pytest-dev/pytest/blob/b4ab2f094234a0e5a1b14cf8e6d7cd9f48ecbadb/.github/workflows/test.yml#L145-L153

hugovk (approved):

hugovk (commented):
(I meant the previous review to be a "comment")

bluetech:
I updated the PR. Changes:

- Now requiring Sphinx 5 instead of 4. I set the bounds to >=5, <6. Lower bound because I probably relied on something new, and upper bound to avoid possible breaking changes in Sphinx.
- I fixed the private `__init__` problem discussed above by setting `autodoc_typehints_description_target = "documented"` and painfully documenting all params. I went over all of the API reference and think it's OK, probably an improvement actually.
- I fixed the overloads problem with some hack to fool Sphinx (see commit).

So I think this is good to merge.

src/_pytest/compat.py
```diff
@@ -20,6 +20,16 @@
 import attr
 import py
 
+# fmt: off
+# Workaround for https://github.com/sphinx-doc/sphinx/issues/10351.
+# If `overload` is imported from `compat` instead of from `typing`,
+# Sphinx doesn't recognize it as `overload` and the API docs for
+# overloaded functions look good again. But type checkers handle
```
nicoddemus, line 27: IIUC, should this read:

```suggestion
# Sphinx doesn't recognize it as `overload` and the API docs for
# overloaded functions look bad. But type checkers handle
```

?

src/_pytest/compat.py
```diff
@@ -20,6 +20,16 @@
...
+# it fine.
+# fmt: on
+if True:
+    from typing import overload as overload
```
nicoddemus, line 31: Curious: is the "import overload as overload" part of the workaround?

src/_pytest/logging.py
```diff
@@ -382,20 +384,19 @@ def _finalize(self) -> None:
...
-        """Get the logging handler used by the fixture.
-
-        :rtype: LogCaptureHandler
-        """
+        """Get the logging handler used by the fixture."""
         return self._item.stash[caplog_handler_key]
 
-    def get_records(self, when: str) -> List[logging.LogRecord]:
+    def get_records(
+        self, when: "Literal['setup', 'call', 'teardown']"
```
nicoddemus, line 391: No need to change, but we can use `from __future__ import annotations` and get away with writing the annotations directly (instead in string form).

nicoddemus (approved):
Looks great, looks like this was a lot of work. 😅 

Thanks a lot!

src/_pytest/compat.py
bluetech: The problem is that Sphinx started recognizing overloads and rendering them, but the way it does it has some issues (see picture in https://github.com/sphinx-doc/sphinx/issues/10351, pretty similar to the we cleaned up a while back). So this hack actually makes things look good by making Sphinx *not* recognize `overload` :)

nicoddemus: AHH I see, thanks!

src/_pytest/compat.py
bluetech: This is the way to explicitly re-export a value in type checking (see https://mypy.readthedocs.io/en/stable/config_file.html?highlight=export#confval-implicit_reexport).

src/_pytest/logging.py
bluetech: Hmm it might be nice, but probably should do it for the entire code base at once?

nicoddemus: I think doing piecewise is fine, and of course doing all at once is fine too. But of course, just want to comment as this is fine. 👍

bluetech:
Thanks for the review @nicoddemus.

PR #9252, opened 2021-10-30
9202 Update codecov uploader

### LINKED ISSUE #9202: Codecov bash uploader is deprecated

cristianMeli:

Closes #9202


cristianMeli:
This PR absolutely needs a squash and merge 😅.

bluetech (commented):
This seems to work, thanks @cristianMeli!

Two things:

- We can remove the `scripts/upload-coverage.sh` file now - can you do it?

- Any particular reason you added the explicit `use_coverage: false` lines? I think they were omitted intentionally so the jobs that are reporting coverage stand out.

cristianMeli:
Hi @bluetech 

- Yes I can remove it.

- First I added because I had coverage errors in some steps, but maybe it was that the `if` was wrong. Besides that, I think that is better to explicitly set `use_coverage: false` for those steps that the coverage report is not generated.

bluetech:
> What do you think?

I think, let's try to remove the explicit `use_coverage: false` and see if it works; if not, we can keep them.

cristianMeli:
@bluetech yeah, it works. Thanks!

bluetech (approved):

PR #9276, opened 2021-11-07
Replace hardcoded links to Python bugs with extlinks

hoefling:
This PR replaces referencing Python bugs via hardcoded URLs to https://bugs.python.org/ by using `sphinx.ext.extlinks` plugin.

Old usage example:
```
`bpo-123 <https://bugs.python.org/issue123>`__
```
New usage replacement:
```
:bpo:`123`
```
This PR partially addresses #9081.

Note that ATM there is only one reference to BPO in the docs. Introducing a new extlink just for one URL thus may be a complete overkill! Maybe it's wise to close this PR w/o merging.


bluetech (approved):
I think it's nice to have.

nicoddemus:
I like it as well. 👍

PR #9293, opened 2021-11-10
Replace hardcoded links to Github user accounts with extlinks

hoefling:
This PR replaces referencing Github users via hardcoded URLs to `https://github.com/{username}` by using `sphinx.ext.extlinks` plugin.

Old usage example:
```
`@hoefling <https://github.com/hoefling>`__
```
New usage replacement:
```
:user:`hoefling`
```
This PR partially addresses #9081.

### DIFF
--- a/doc/en/announce/release-2.9.0.rst
+++ b/doc/en/announce/release-2.9.0.rst
@@ -45,29 +45,29 @@ The py.test Development Team
 **New Features**
 
 * New ``pytest.mark.skip`` mark, which unconditionally skips marked tests.
-  Thanks `@MichaelAquilina`_ for the complete PR (:pull:`1040`).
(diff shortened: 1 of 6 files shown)


bluetech (approved):
Really nice, thanks!

I was thinking that maybe `user` is a bit too generic, and we better use `ghuser` instead. But then I figured that if we ever switch away from github, we will probably just carry over the usernames, like issues/pull numbers. So maybe better to keep it `user`.

nicoddemus (approved):

webknjaz:
> I was thinking that maybe `user` is a bit too generic,

FTR many projects use `:user:` as a convention (tox/setuptools/aiohttp/etc.)
Moreover, there's an extension `sphinx-issues` that also implements this with the same name.

PR #9299, opened 2021-11-11
Reemit deprecation warnings from `pytest.warns`

### LINKED ISSUE #9288: Document that `with pytest.warns()` captures all warnings, not just those it tests were emitted
I have a test that ensures a specific warning type is produced by a function, but since I switched to `pytest.warns` I no longer see Python 3.10 compatibility warnings in Pytest `warnings summary` the code was emitting before switching to `pytest.warns`. Is it expected behavior (in which case it contradicts https://docs.pytest.org/en/6.2.x/warnings.html#deprecationwarning-and-pendingdeprecationwarning) and how I can opt-out of hiding `DeprecationWarning`s by `pytest.warns` (there is nothing about that in the documentation https://docs.pytest.org/en/6.2.x/reference.html#pytest-warns)? 

MVCE:
```python
import pytest
import warnings

def test_warning():
    with pytest.warns(DeprecationWarning):
        with pytest.warns(UserWarning):  # catches UserWarning, but also every other type too
            warnings.warn("my warning", UserWarning)
            warnings.warn("some deprecation warning", DeprecationWarning)
```
```
________________________________ test_warning _________________________________

    def test_warning():
>       with pytest.warns(DeprecationWarning):
E       Failed: DID NOT WARN. No warnings of type (<class 'DeprecationWarning'>,) was emitted. The list of emitted warnings is: [].
```

Kojoley:
Only suppress deprecation types that are passed in `expected_warning` argument.

Fixes #9288

Looking for a feedback, are you ok with this change?

### DIFF
--- a/changelog/9288.improvement.rst
+++ b/changelog/9288.improvement.rst
@@ -1,0 +1,1 @@
+``pytest.warns`` no longer suppresses warnings derived from ``DeprecationWarning`` and ``PendingDeprecationWarning`` by default.
--- a/src/_pytest/recwarn.py
+++ b/src/_pytest/recwarn.py
@@ -292,6 +292,15 @@ class WarningsChecker(WarningsRecorder):
 
         # only check if we're not currently handling an exception
         if exc_type is None and exc_val is None and exc_tb is None:
+            for w in self:
+                reemit = {DeprecationWarning, PendingDeprecationWarning}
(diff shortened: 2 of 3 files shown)


Zac-HD:
Let's go with the more general solution in https://github.com/pytest-dev/pytest/pull/10937.

PR #9309, opened 2021-11-14
Change deprecation warnings into errors for 7.0 release, using a new mechanism

### LINKED ISSUE #9308: Change deprecation warnings into errors for 7.0
Per our [backward compatibility policy](https://docs.pytest.org/en/latest/backwards-compatibility.html), change deprecations for pytest 7 to errors, before complete removal in pytest 7.1.

Note: @nicoddemus already opened a similar issue in #7363, but it was changed to track the work for pytest 8.0. I believe we can still do some removals in 7.0 -- (TODO: link to PR). Specifically deprecations from 6.0 are eligible, but no later.

bluetech:
This fixes #9308. The first attempt #8837 ran into trouble because some deprecations weren't ready for removal. I make things work by adding `PytestRemovedInX` warnings. Please see the commit messages for more details.

The reason I worked on this is I'm eager to remove one deprecation - the `FILLFUNCARGS` - and don't want to wait until pytest 8.1 for that :)

### DIFF
--- a/src/pytest/__init__.py
+++ b/src/pytest/__init__.py
@@ -68,6 +68,8 @@ from _pytest.warning_types import PytestCollectionWarning
 from _pytest.warning_types import PytestConfigWarning
 from _pytest.warning_types import PytestDeprecationWarning
 from _pytest.warning_types import PytestExperimentalApiWarning
+from _pytest.warning_types import PytestRemovedIn7Warning
+from _pytest.warning_types import PytestRemovedIn8Warning
 from _pytest.warning_types import PytestUnhandledCoroutineWarning
 from _pytest.warning_types import PytestUnhandledThreadExceptionWarning
 from _pytest.warning_types import PytestUnknownMarkWarning
@@ -127,6 +129,8 @@ __all__ = [
     "PytestConfigWarning",
     "PytestDeprecationWarning",
     "PytestExperimentalApiWarning",
+    "PytestRemovedIn7Warning",
+    "PytestRemovedIn8Warning",
     "Pytester",
     "PytestPluginManager",
     "PytestUnhandledCoroutineWarning",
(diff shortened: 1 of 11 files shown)


src/pytest/__init__.py
```diff
@@ -127,6 +129,8 @@
     "PytestConfigWarning",
     "PytestDeprecationWarning",
     "PytestExperimentalApiWarning",
+    "PytestRemovedIn7Warning",
```
nicoddemus, line 132: When we will remove the `PytestRemovedIn7Warning` class itself? In pytest 8, just as we turn `PytestRemovedIn8Warning` into an error?

nicoddemus (approved):
Looks great, thanks for tackling this!

src/pytest/__init__.py
bluetech: I was thinking to remove it in 7.1:

- Users of pytest should be pinning their pytest version. So if they added an ignore for `PytestRemovedIn7Warning`, once they've upgraded to 7.1 it's no longer relevant and can be removed.
- Plugins which want to support multiple pytest versions really shouldn't do `PytestRemovedIn7Warning` ignores...

nicoddemus: > So if they added an ignore for PytestRemovedIn7Warning, once they've upgraded to 7.1 it's no longer relevant and can be removed.

Hmm OK, just to mention that it would immediately break the test suite (the warnings module raises an error if you try to ignore an exception that doesn't exist). But of course that's just removing one line.

Even so might make sense to remove it in 8 because of that? Either way I'm fine with what you decide, just wanted to mention that bit in case it was not clear.

bluetech: If someone needed to add this line, then their testsuite will fail on 7.1 anyway because the feature they added the ignore for will have been removed. So failing fast seems better to me actually.

nicoddemus: > If someone needed to add this line, then their testsuite will fail on 7.1 anyway because the feature they added the ignore for will have been removed. 

Good point. 👍

The-Compiler:
This breaks the `docs` build, follow up: #9336

PR #9310, opened 2021-11-15
testing/test_session: add a regression test for an old bug

bluetech:
Nothing tests this currently. Make sure it doesn't regress if/when the complex code in `Session.collect` is cleaned up.

### DIFF
--- a/testing/test_session.py
+++ b/testing/test_session.py
@@ -335,6 +335,54 @@ def test_sessionfinish_with_start(pytester: Pytester) -> None:
     assert res.ret == ExitCode.NO_TESTS_COLLECTED
 
 
+def test_collection_args_do_not_duplicate_modules(pytester: Pytester) -> None:
+    """Test that when multiple collection args are specified on the command line
+    for the same module, only a single Module collector is created.
+
+    Regression test for #723, #3358.
+    """
+    pytester.makepyfile(
+        **{
+            "d/test_it": """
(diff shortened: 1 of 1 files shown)


testing/test_session.py
```diff
@@ -335,6 +335,54 @@ def pytest_sessionfinish():
...
+    result.stdout.fnmatch_lines(
+        [
+            "<Module d/test_it.py>",
+            "  <Function test_1>",
+            "  <Function test_2>",
+        ],
+        consecutive=True,
+    )
+
+    # Different, but related case.
```
bluetech, line 367: I added this case as well, in relation to `node_cache2` in `Session.collect()`.

But I wonder if this is actually the behavior we want. If a user passes `d` twice with `--keep-duplicates`, do they expect e.g. module fixtures the be run once or twice?

nicoddemus: I would say run once makes more sense, I consider a "module" fixture being executed once to be an invariant; the keep duplicates only is there to "duplicate" the tests themselves, not fixture execution.

But I'm only really guessing.

nicoddemus (approved):

testing/test_session.py
bluetech: I feel similarly - it seems better but not quite sure. Anyway, I figure if we change something here it should be a conscious choice so the test is good to have.

PR #9336, opened 2021-11-24
Fix doc issues in deprecation change

The-Compiler:
Follow-up to #9309.

The issue in the changelog broke the docs build with:

doc/en/_changelog_towncrier_draft.rst:47: WARNING: Inline literal start-string
without end-string.

The second change isn't as critical, but caused the text to be rendered as
monospace including the tilde (i.e. `~pytest.PytestDeprecationwarning`).


bluetech (approved):
Oops, missed the real failure due to the other one. Thanks for fixing!

PR #9341, opened 2021-11-26
Improve reference and path/fspath docs

### LINKED ISSUE #9283: pytest 7: path vs. fspath, pathlib vs. py.path
I've just been looking through the changelog for pytest 7 and noticed something I'd like to discuss before the release:

In #8144, @bluetech added `pathlib` alternatives to various hooks, and called them `fspath`, with the old `py.path.local` alternatives keeping their `path` name.

However, in #8251, @RonnyPfannschmidt added a `pathlib` attribute to `Node`, and called it `path`, with the old `py.path` one being called `fspath`...

We then reverted the `Node.fspath` deprecation in #8903 after some discussion in #8821, but kept the new `path` attribute... 

However, the naming inconsistency seems like something that could confuse people. What to do? Keep it as is, or revert the new attribute entirely until we figured out the rest of the deprecation? Make it private perhaps?

The-Compiler:
Closes #9283

IMHO the last step towards a 7.0 rc (finally! :sweat_smile:), unless I missed something (if so, please yell now!).

I also fixed some other issues I noticed while looking through the reference docs.

### DIFF
--- a/changelog/7259.deprecation.rst
+++ b/changelog/7259.deprecation.rst
@@ -3,8 +3,10 @@
 ``py.path.local`` arguments to Node constructors have been deprecated. See :ref:`the deprecation note <node-ctor-fspath-deprecation>` for full details.
 
 .. note::
-    The name of the ``Node`` arguments and attributes (old ``fspath``, new ``path``) is **the opposite**
-    of the situation for hooks (old ``path``, new ``fspath``).
+    The name of the :class:`~_pytest.nodes.Node` arguments and attributes (the
+    new attribute being ``path``) is **the opposite** of the situation for hooks
+    (the old argument being ``path``).
+
     This is an unfortunate artifact due to historical reasons, which should be
     resolved in future versions as we slowly get rid of the :pypi:`py`
     dependency (see :issue:`9283` for a longer discussion).
--- a/changelog/7259.deprecation.rst
+++ b/changelog/7259.deprecation.rst
@@ -1,3 +1,7 @@
 ``py.path.local`` arguments for hooks have been deprecated. See :ref:`the deprecation note <legacy-path-hooks-deprecated>` for full details.
 
 ``py.path.local`` arguments to Node constructors have been deprecated. See :ref:`the deprecation note <node-ctor-fspath-deprecation>` for full details.
+
+.. note::
+    The name of the ``Node`` arguments and attributes (old ``fspath``, new ``path``) is **the opposite**
+    of the situation for hooks (old ``path``, new ``fspath``).
--- a/changelog/7259.deprecation.rst
+++ b/changelog/7259.deprecation.rst
@@ -5,3 +5,6 @@
 .. note::
     The name of the ``Node`` arguments and attributes (old ``fspath``, new ``path``) is **the opposite**
     of the situation for hooks (old ``path``, new ``fspath``).
+    This is an unfortunate artifact due to historical reasons, which should be
+    resolved in future versions as we slowly get rid of the :pypi:`py`
+    dependency (see :issue:`9283` for a longer discussion).
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -696,7 +696,7 @@ class Item(Node):
 
     def runtest(self) -> None:
         """Run the test case for this item.
-        
+
         Must be implemented by subclasses.
 
         .. seealso:: :ref:`non-python tests`
--- a/src/_pytest/nodes.py
+++ b/src/_pytest/nodes.py
@@ -159,6 +159,13 @@ class Node(metaclass=NodeMeta):
     Collector subclasses have children; Items are leaf nodes.
     """
 
+    # Implemented in the legacypath plugin.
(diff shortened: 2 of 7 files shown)


src/_pytest/nodes.py
```diff
@@ -159,6 +159,13 @@ class Node(metaclass=NodeMeta):
     Collector subclasses have children; Items are leaf nodes.
     """
 
+    # Implemented in the legacypath plugin.
+    #: A ``LEGACY_PATH`` copy of the :attr:`path` attribute. Intended for usage
+    #: for methods not migrated to ``pathlib.Path`` yet, such as
+    #: :meth:`Item.reportinfo`. Will be deprecated in a future release, prefer
+    #: using :attr:`path` instead.
+    fspath: LEGACY_PATH
+
```
The-Compiler, line 168: @bluetech What's your take on this? From what I understand, the aim of #9208 was to decouple legacy stuff like `fspath`, so this might seem out of place. However, I'd really prefer to keep this attribute documented (even more so because it's not deprecated yet!), and having this here is the only way I found to make Sphinx pick it up.

src/_pytest/nodes.py
```diff
@@ -188,26 +195,26 @@ def __init__(
         #: The parent collector node.
         self.parent = parent
 
-        #: The pytest config object.
```
The-Compiler, line 191: Looks like Sphinx silently ignores those comments (and hides the attribute as undocumented) if this comment isn't next to the line actually setting the attribute.

src/_pytest/nodes.py
bluetech: Two aspects to this - docs and typing. For the docs it might work to add it in reference.rst as regular sphinx attr instead of autodoc? As for typing, without this PR it would show up as a missing attribute error, but maybe it's a good thing?

Anyway, I'm fine with whatever you think is best (including this approach).

bluetech (approved):
LGTM!

changelog/7259.deprecation.rst
```diff
@@ -1,3 +1,7 @@
 ``py.path.local`` arguments for hooks have been deprecated. See :ref:`the deprecation note <legacy-path-hooks-deprecated>` for full details.
 
 ``py.path.local`` arguments to Node constructors have been deprecated. See :ref:`the deprecation note <node-ctor-fspath-deprecation>` for full details.
+
+.. note::
+    The name of the ``Node`` arguments and attributes (old ``fspath``, new ``path``) is **the opposite**
+    of the situation for hooks (old ``path``, new ``fspath``).
```
nicoddemus, line 7: I'm sure many people will ask the question "What, why??" when reading this... it might save us some explaining if we add a quick explanation, perhaps something along the lines: 

```suggestion
    of the situation for hooks (old ``path``, new ``fspath``). This is an unfortunate artifact due to historical reasons,
    which should be resolved in future versions as we slowly get rid of the :pypi:`py` dependency (see :issue:`9283` for a longer discussion).
```

nicoddemus (approved):

src/_pytest/nodes.py
The-Compiler: I feel like having this as a "missing" attribute would be wrong - especially given that we explicitly un-deprecated it. Until we deprecate (or, ideally, remove) it, I feel like typing should know that it's there too.

The-Compiler:
Great suggestion! I opted to add the additional infos to all locations having the original explanation for visibility.

The-Compiler:
Converting to draft given the ongoing discussion in #9283, only noticed it now.

The-Compiler:
Updated based on #9363. I decided to leave the notes in, as this is still a bit of a problem (albeit less confusing than before certainly). Let me know what you think!

changelog/7259.deprecation.rst
```diff
@@ -1,3 +1,12 @@
 ``py.path.local`` arguments for hooks have been deprecated. See :ref:`the deprecation note <legacy-path-hooks-deprecated>` for full details.
 
 ``py.path.local`` arguments to Node constructors have been deprecated. See :ref:`the deprecation note <node-ctor-fspath-deprecation>` for full details.
+
+.. note::
+    The name of the :class:`~_pytest.nodes.Node` arguments and attributes (the
+    new attribute being ``path``) is **the opposite** of the situation for hooks
```
nicoddemus, line 7: Not sure this still applies, as we don't have the same situation anymore.

nicoddemus (approved):

changelog/7259.deprecation.rst
The-Compiler: Well, kind of - we still have the situation that `Node.path` is new, but `path` in hooks is old. But if people prefer, I'm fine with removing those notes.

nicoddemus: Ahh you are correct. Nevermind, I misunderstood the text. 👍

The-Compiler: If you have a suggestion for a rewording to make things clearer, let me know! Probably going to be Monday or Tuesday until I can finish the release anyways.

nicoddemus (approved):

Issue #244, opened 2012-12-26
parametrize fails when values are unhashable


pytestbot:
Originally reported by: **Ronny Pfannschmidt (BitBucket: [RonnyPfannschmidt](http://bitbucket.org/RonnyPfannschmidt), GitHub: [RonnyPfannschmidt](http://github.com/RonnyPfannschmidt))**

---

seems like a recent change in parametrizing disallows hashable items
this broke the hgdistver testsuite

---
- Bitbucket: https://bitbucket.org/pytest-dev/pytest/issue/244

PR #9350, opened 2021-11-29
Fix test reordering for indirect parameterization

### LINKED ISSUE #8914: Indirect session fixture / item reordering issue
```python
import pytest

def prepare(request):
    print("\nprepare setup with {!r}".format(request.param))

@pytest.mark.parametrize('prepare', ["dina"], indirect=True, scope="session")
def test_1(prepare):
    print("Test1 done\n")

@pytest.mark.parametrize('prepare', ["mor"], indirect=True, scope="session")
def test_2(prepare):
    print("Test2 done\n")

@pytest.mark.parametrize('prepare', ["dina"], indirect=True, scope="session")
def test_3(prepare):
    print("Test3 done\n")
```

I get this output:

```
λ pytest .tmp\test_fix_session.py -sq

prepare setup with 'dina'
Test1 done

.
prepare setup with 'mor'
Test2 done

.
prepare setup with 'dina'
Test3 done

.
3 passed in 0.02s
```

We would expect it to reorder the tests to `[test_1, test_3, test_2]` so the fixture would execute only twice, instead of 3 times as it is doing now, so definitely looks like a bug (unless I'm missing something).

If this looks right, I suggest we open a new issue with this final example/description, so one doesn't need to read the whole thread to get to the central point of the problem. What do you folks think?

_Originally posted by @nicoddemus in https://github.com/pytest-dev/pytest/issues/8328#issuecomment-880683779_

haxtibal:
`reorder_items` groups tests so that each group can share indirectly parameterized fixture instances. This optimizes for fewer fixture setup/teardown cycles. Prior to this PR, grouping considered an parameter's index, not the parameter value itself, to determine whether a parameter is "the same" and therefore can be shared.

Relying on indexes is fast, however only works when there's exactly one parameter list for one fixture. If we provide multiple (possibly different) lists for the same fixture, e.g. by decorating test items, one index can no longer refer to "the same" parameter. In order to still be able to group for fixture reuse, we have to group by parameter value.

Caution: The value ends up inside the key of another dict, and therefore must be hashable. This was true for indexes, but no longer is guaranteed for user provided values. A user may e.g. provide `dict`s or numpy arrays. The `SafeHashWrapper` ensures a fallback to `id()` in such a case.

Fixes #8914.

### DIFF
--- a/src/_pytest/fixtures.py
+++ b/src/_pytest/fixtures.py
@@ -5,6 +5,7 @@ import sys
 import warnings
 from collections import defaultdict
 from collections import deque
+from collections.abc import Hashable
 from contextlib import suppress
 from pathlib import Path
 from types import TracebackType
@@ -238,6 +239,23 @@ def getfixturemarker(obj: object) -> Optional["FixtureFunctionMarker"]:
 _Key = Tuple[object, ...]
 
 
+@attr.s(auto_attribs=True, eq=False, slots=True)
+class SafeHashWrapper:
+    obj: Any
+
+    def __eq__(self, other: object) -> bool:
+        try:
+            res = self.obj == other
+            return bool(res)
+        except Exception:
+            return id(self.obj) == id(other)
+
+    def __hash__(self) -> int:
+        if isinstance(self.obj, Hashable):
+            return hash(self.obj)
+        return hash(id(self.obj))
+
+
 def get_parametrized_fixture_keys(item: nodes.Item, scope: Scope) -> Iterator[_Key]:
     """Return list of keys for all parametrized arguments which match
     the specified scope."""
@@ -254,15 +272,19 @@ def get_parametrized_fixture_keys(item: nodes.Item, scope: Scope) -> Iterator[_K
         for argname, param_index in sorted(cs.indices.items()):
             if cs._arg2scope[argname] != scope:
                 continue
+            # param ends up inside a dict key, and therefore must be hashable.
+            # Parameter values are possibly unhashable (dicts, numpy arrays, ...).
+            # SafeHashWrapper makes them hashable by a fallback to their identity.
+            param = SafeHashWrapper(cs.params[argname])
             if scope is Scope.Session:
-                key: _Key = (argname, param_index)
+                key: _Key = (argname, param)
             elif scope is Scope.Package:
-                key = (argname, param_index, item.path.parent)
+                key = (argname, param, item.path.parent)
             elif scope is Scope.Module:
-                key = (argname, param_index, item.path)
+                key = (argname, param, item.path)
             elif scope is Scope.Class:
                 item_cls = item.cls  # type: ignore[attr-defined]
-                key = (argname, param_index, item.path, item_cls)
+                key = (argname, param, item.path, item_cls)
             else:
                 assert_never(scope)
             yield key
--- a/testing/example_scripts/issue_519.py
+++ b/testing/example_scripts/issue_519.py
@@ -22,13 +22,13 @@ def checked_order():
     assert order == [
         ("issue_519.py", "fix1", "arg1v1"),
         ("test_one[arg1v1-arg2v1]", "fix2", "arg2v1"),
-        ("test_two[arg1v1-arg2v1]", "fix2", "arg2v1"),
         ("test_one[arg1v1-arg2v2]", "fix2", "arg2v2"),
+        ("test_two[arg1v1-arg2v1]", "fix2", "arg2v1"),
         ("test_two[arg1v1-arg2v2]", "fix2", "arg2v2"),
         ("issue_519.py", "fix1", "arg1v2"),
         ("test_one[arg1v2-arg2v1]", "fix2", "arg2v1"),
-        ("test_two[arg1v2-arg2v1]", "fix2", "arg2v1"),
         ("test_one[arg1v2-arg2v2]", "fix2", "arg2v2"),
+        ("test_two[arg1v2-arg2v1]", "fix2", "arg2v1"),
         ("test_two[arg1v2-arg2v2]", "fix2", "arg2v2"),
     ]
 
--- a/testing/python/fixtures.py
+++ b/testing/python/fixtures.py
@@ -1308,6 +1308,34 @@ class TestFixtureUsages:
         result = pytester.runpytest()
         result.stdout.fnmatch_lines(["*4 passed*"])
 
+    def test_optimize_by_reorder_indirect(self, pytester: Pytester) -> None:
+        """Test reordering for minimal setup/teardown with indirectly parametrized fixtures. See #8914, #9350."""
+        pytester.makepyfile(
+            """
+            import pytest
+
+            @pytest.fixture(scope="session")
+            def fix(request):
+                print(f'prepare foo-%s' % request.param)
+                yield request.param
+                print(f'teardown foo-%s' % request.param)
+
+            @pytest.mark.parametrize("fix", ["data1", "data2"], indirect=True)
+            def test1(fix):
+                pass
+
+            @pytest.mark.parametrize("fix", ["data2", "data1"], indirect=True)
+            def test2(fix):
+                pass
+        """
+        )
+        result = pytester.runpytest("-s")
+        output = result.stdout.str()
+        assert output.count("prepare foo-data1") == 1
+        assert output.count("prepare foo-data2") == 1
+        assert output.count("teardown foo-data1") == 1
+        assert output.count("teardown foo-data2") == 1
+
     def test_funcarg_parametrized_and_used_twice(self, pytester: Pytester) -> None:
         pytester.makepyfile(
             """
--- a/AUTHORS
+++ b/AUTHORS
@@ -321,6 +321,7 @@ Thomas Grainger
 Thomas Hisch
 Tim Hoffmann
 Tim Strazny
+Tobias Deiminger
 Tom Dalton
 Tom Viner
 Tomáš Gavenčiak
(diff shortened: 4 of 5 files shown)


testing/example_scripts/issue_519.py
```diff
@@ -22,13 +22,13 @@ def checked_order():
     assert order == [
         ("issue_519.py", "fix1", "arg1v1"),
         ("test_one[arg1v1-arg2v1]", "fix2", "arg2v1"),
-        ("test_two[arg1v1-arg2v1]", "fix2", "arg2v1"),
```
nicoddemus, line 25: Hmm, this is odd, because we have increased the number of setups/teardowns for fixture 2 with this change.

On `main`, we have 2 s/t for `fix1`, and 4 s/t for `fix2`.

Here we have the same 2 s/t for `fix1`, but now we have 8 s/t for `fix2`. Seems like a regression for this case. 🤔

nicoddemus:
Thanks @haxtibal!

The PR looks great, but I left a comment on one of the tests you changed, please take a look?

testing/example_scripts/issue_519.py
haxtibal: > now we have 8 s/t for fix2

Yes, but we also had 8 s/t for `fix2` even before this PR. `fix2` is always (and correctly) recreated, because it has function scope. It's just not obvious from the assert statement. Tested it by adding print output pre and post the yield in `foo{1,2}`. But please double check...

haxtibal: I think the same holds for fixture `arg2`, it also has function scope and therefore shall always be recreated.

`fix1` and `arg1` have module scope, and both are successfully reused even after the PR, are they?

nicoddemus: Ahh right my bad, missed the fact that `fix2` is function scoped, so its order there doesn't matter as it will be recreated each time. 👍

nicoddemus (approved):
LGTM thanks!

It would be nice if another maintainer also reviewed this (@RonnyPfannschmidt @bluetech)

src/_pytest/fixtures.py
```diff
@@ -240,6 +241,24 @@ def getfixturemarker(obj: object) -> Optional["FixtureFunctionMarker"]:
...
+            res = self.obj == other
+            bool(res)
+            return res
+        except Exception:
+            return id(self.obj) == id(other)
+
+    def __hash__(self) -> Any:
+        if isinstance(self.obj, Hashable):
+            return hash(self.obj)
+        return hash(id(self.obj))
```
bluetech, line 259: Some suggestions here (I haven't checked that they all work):

```suggestion
@attr.s(auto_attribs=True, eq=False, slots=True)
class SafeHashWrapper:
    obj: object

    def __eq__(self, other: object) -> bool:
        try:
            res = self.obj == other
            return bool(res)
        except Exception:
            return id(self.obj) == id(other)

    def __hash__(self) -> int:
        if isinstance(self.obj, Hashable):
            return hash(self.obj)
        return hash(id(self.obj))
```

src/_pytest/fixtures.py
```diff
@@ -256,15 +275,16 @@ def get_parametrized_fixture_keys(item: nodes.Item, scope: Scope) -> Iterator[_K
         for argname, param_index in sorted(cs.indices.items()):
             if cs._arg2scope[argname] != scope:
                 continue
+            param = SafeHashWrapper(cs.params.get(argname, param_index))
```
bluetech, line 278: Why is the `get` needed here?

testing/python/fixtures.py
```diff
@@ -1312,6 +1312,36 @@ def test2(no_eq):
         result = pytester.runpytest()
         result.stdout.fnmatch_lines(["*4 passed*"])
 
+    def test_reorder_indirect(self, pytester: Pytester) -> None:
+        pytester.makepyfile(
+            """
+            import pytest
```
bluetech, line 1318: Maybe a simpler test will also get the point across:

```py
import pytest


def fix(request):
    yield request.param


@pytest.mark.parametrize("fix", ["data1", "data2"], indirect=True)
def test1(fix):
    pass


@pytest.mark.parametrize("fix", ["data2", "data1"], indirect=True)
def test2(fix):
    pass
```

src/_pytest/fixtures.py
bluetech, line 278: A comment above this line explaining the purpose of `SafeHashWrapper` would be good.

testing/python/fixtures.py
```diff
@@ -1312,6 +1312,36 @@ def test2(no_eq):
         result = pytester.runpytest()
         result.stdout.fnmatch_lines(["*4 passed*"])
 
+    def test_reorder_indirect(self, pytester: Pytester) -> None:
```
bluetech, line 1315: A docstring with a reference to the PR (since there is no issue) would be good.

BTW, if you explicitly want to test the reordering here and not the runtime behavior, I suggest running the pytester with `--collect-only --quiet` which just spits the test order, and checking that, instead of doing prints and all that.

bluetech (commented):
Thanks @haxtibal, interesting. This does seem to give better behavior.

To give a proper review, I would need to dig into the details here (like understand the surrounding code, and figure out why the param *index* was even used in the first place). That might take some time, so I wouldn't want to hold the PR for too long.

One possible problem with using the param value is that some params might be large (like large numpy arrays and what not) and comparing them might be slow. See for example 80d4dd6f0bf289194f540f4ab6d6cf96776b8072. We will need to consider if this is a problem for this case or not.

haxtibal:
Thanks for the review @nicoddemus and @bluetech

> figure out why the param _index_ was even used in the first place

0cfd873 and #244 give some insight. Seems we switched from value to index in 2013 because unhashable params caused failure. This PR works around the unhashable problem by falling back to `id()` via the `SafeHashWrapper`.

> some params might be large (like large numpy arrays and what not) and comparing them might be slow.

Agreed. Comparing here means hashing, but that could be slow too. It's probably not too bad, but still... A way out could be to consider a user provided id, instead of calculating the hash (such a user id would also be better than the `id()` fallback for unhashable types). Maybe even `pytest.param(id="use_this_one_for_reorder")` could be good enough? Give me some time, I'll evaluate that.

> numpy arrays

They're special, as they're unhashable and won't compare to `bool`. It's a case where the `SafeHashWrapper` falls back to `id()`.

bluetech:
> 0cfd873 and #244 give some insight.

Indeed, knowing it was originally value based and only switched to indexes for the hashability reassures me that there's not something we're missing.

haxtibal:
> Maybe even pytest.param(id="use_this_one_for_reorder") could be good enough?

They could :) dfc773e demonstrates how. Not yet code complete, but all tests pass. Just wanted to hear your opinion about the approach early.

The benefit should be obvious from this example:
```python
@pytest.mark.parametrize("fix", [1, pytest.param({"data": 2}, id="2")], indirect=True)
def test1(fix):
    pass

@pytest.mark.parametrize("fix", [pytest.param({"data": 2}, id="2"), 1], indirect=True)
def test2(fix):
    pass
```

Instead of subtle surprises triggered by implicit fallback to identity compare, a user can now explicitly control which
parameters are "the same". The `id` arg seems like the natural choice for that purpose. Otoh, the changes are less
trivial compared to the first commit, so I can understand if you don't like it.

How shall we proceed from here?

testing/python/fixtures.py
haxtibal: > explicitly want to test the reordering here

Afaik pytest doesn't give guarantees about absolute test order, but only about certain constraints. So I intended to assert for the number of setups/teardowns to be minimal, not for a specific sequence. I think that's only possible if we actually let the fixture requests happen. They happen after collect.

But `--collect-only` could still be good for kind of unit testing the reorder function, would such an additional test be useful?

bluetech:
@haxtibal Using ids sounds interesting. I'm not sure if it can work, but if does, it would be an elegant solution to the problem.

Your commit is hard to understand directly (because the underlying code is not the prettiest). What would be great is if you could:

1. Open a separate PR for the id approach, based directly on main.
2. Add comments (can be code comments or github comments) explaining what each non-trivial change is doing.

haxtibal:
@bluetech

> 1. Open a separate PR for the id approach, based directly on main.

Done, see #9420. It's now basically a combination of using parameter ids from API plus a fallback to the behavior introduced in #9350.

> 2. Add comments (can be code comments or github comments) explaining what each non-trivial change is doing.

Did it as series of smaller commits, each with reasoning in commit message, hope that helps a bit.

> if it can work

All tests pass, "can work" is "yes" until proven otherwise😏

src/_pytest/fixtures.py
haxtibal: Thank's, applied all except `obj: Any`, because it makes mypy think `obj` is always hashable and therefore we have an unreachable statement in `SafeHashWrapper.__hash__`.

I think the type constraint I'd like to express for `obj` is "type supports identity" and "type is not NoneType". But I don't know a way how, do you?

A bit more OT, if `isinstance(object(), Hashable)` and `isinstance(dict(), object)` are `True`, doesn't `isinstance(dict(), Hashable) == False` break Liskov?

src/_pytest/fixtures.py
haxtibal: Wasn't sure if `argname` is guaranteed to be in `cs.params`. But now I think it is (and if not, it would be y pytest bug), so get is not required.

obestwalter:
@haxtibal it would be a pity to loose the effort that already went into this, so trying to get the ball rolling again here. #9420 is the preferred approach, so I think this should be closed and maybe we can bring that one home then, right?

haxtibal:
> #9420 is the preferred approach, so I think this should be closed and maybe we can bring that one home then, right?

Agreed, let's close this in favor of #9420.

The latter will bother maintainers. There's potential of breaking use cases or user expectations and I'm not in a position to foresee them all. I'd help where I can, but it's your turn to signal what's next.

PR #9352, opened 2021-11-30
Fix output of example test method in special.rst

### LINKED ISSUE #9351: Correct typo in doc/en/example/special.rst
Trivial typo in doc/en/example/special.rst, line 46:
* def test_method2() prints the same output as def test_method1().
* the test output requires an update.

See here for an example:
https://docs.pytest.org/en/latest/example/special.html?highlight=test_method1

truth-quark:
This PR closes #9351.

It's a very minor typo :-)

### DIFF
--- a/changelog/9351.trivial.rst
+++ b/changelog/9351.trivial.rst
@@ -1,0 +1,1 @@
+Correct minor typos in doc/en/example/special.rst.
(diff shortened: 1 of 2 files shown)


Zac-HD (approved):
Beautiful!

Thanks for reporting this @truth-quark, and congrats on your first PR to pytest 🎉

PR #9354, opened 2021-11-30
Use exact comparison for bool in approx()

### LINKED ISSUE #9353: pytest.approx considers boolean numeric types
From [the docs](https://docs.pytest.org/en/6.2.x/reference.html#pytest-approx):

> You can also use `approx` to compare nonnumeric types, or dicts and sequences containing nonnumeric types, in which case it falls back to strict equality. This can be useful for comparing dicts and sequences that can contain optional values:
>
> ```python
> >>> {"required": 1.0000005, "optional": None} == approx({"required": 1, "optional": None})
> True
> >>> [None, 1.0000005] == approx([None,1])
> True
> >>> ["foo", 1.0000005] == approx([None,1])
> False
> ```

However, that seems to treat booleans as numerical values too:

```python
import pytest

def test_approx():
    d = {"value": 0.3, "active": True}
    expected = {"value": 0.1 + 0.2, "active": False}
    assert d == pytest.approx(expected)
```

results in:

```python
AssertionError: assert {'active': True, 'value': 0.3} == approx({'value': 0.30000000000000004 ± 3.0e-07, 'active': False ± 1.0e-12})
```

(on Python 3.9.7 and pytest 6.2.5). Note the `False ± 1.0e-12`. While this is *technically* correct, it doesn't seem like a very reasonable way of comparing booleans.

cc @jvansanten who seems to have implemented this originally in #7710.

jvansanten:
Fixes #9353.


nicoddemus:
Hi @jvansanten,

Was taking a look at the list of PRs and this one seems to have gone unnoticed, sorry about that.

Overall all changes look good to me, I left a comment about the changelog only. Would you like to finish it up? If you don't have the time please let us know so we can do it. Thanks, and sorry for the delay again!

RonnyPfannschmidt (commented):
For approx of bool, doesn't it semantically make sense to check for truthy values when requested explicitly, but also ensure is checks for implicit matches in lists mappings

marcelotrevisani:
hey, I just got into this problems these days and found this PR. What is it missing to get this over the line please?

nicoddemus:
Seems this fell through the cracks, thanks for the ping. I will rebase.

@RonnyPfannschmidt anything else you would like changed here before we merge it?

nicoddemus (approved):

nicoddemus:
Will squash/merge in the next few days. Thanks @jvansanten for the PR and sorry that we missed this one. Also thanks @marcelotrevisani for the ping!

RonnyPfannschmidt (approved):
its good as is atm - the more tricky details are for when we implement matchers

Issue #9316, opened 2021-11-16
Remove gateway lines from xdist failure output


GergelyKalmar:
#### What's the problem this feature will solve?
Currently the failure section of the output of a pytest run with xdist enabled looks similar to this:
```
=== FAILURES ===
___ [pylint] file1.py ___
[gw1] linux -- Python 3.8.10 .../bin/python3.8
W: 31, 1: TODO: ... (fixme)
W: 32, 1: TODO: ... (fixme)
___ [pylint] file2.py ___
[gw4] linux -- Python 3.8.10...bin/python3.8
W: 47, 1: TODO: ... (fixme)
...
=== 3 failed, 69 passed, 131 skipped in 27.71s ===
```

Note that the lines starting with `[gwX]` do not provide any additional (useful) information in this case (the platform and Python version information is shown at the beginning of the session already).

#### Describe the solution you'd like
It would be great if these lines could be shown only when there is a benefit in doing so. Alternatively, making it possible to hide this via configuration or some pytest hook would be also acceptable (e.g. to make it possible to override it with a plugin).

#### Alternative Solutions
This behavior seems to be hardcoded at https://github.com/pytest-dev/pytest/blob/9d17cebe64878f695fbd7596fcfb643f53a56830/src/_pytest/terminal.py#L570 at the moment.

#### Additional context
The issue was originally raised in the pytest-xdist plugin (https://github.com/pytest-dev/pytest-xdist/issues/718), however, it seems that the issue is specific to pytest itself.

nicoddemus:
Hi @GergelyKalmar,

Thanks for writing! 

> Note that the lines starting with [gwX] do not provide any additional (useful) information in this case (the platform and Python version information is shown at the beginning of the session already).

Actually the `[gwX]` itself is useful, as often knowing which worker run a test can be useful. But I agree the platform and Python version are superfluous.

Can you elaborate why that line is such a problem however?

GergelyKalmar:
Well it is not a gigantic problem for sure, but if I have a few small errors but in many files, this can add quite a lot of lines to the overall failure output. These lines make the output less compact, so I see fewer errors on the screen than I could and I need to scroll more too. It is a bit of an inconvenience.

I see what you mean with the worker numbers, however, I don't remember any occasion where I actually needed that information, so I would definitely prefer the default output to be the same as when running pytest without xdist with the possibility to add this information when really needed (perhaps displaying it in verbose mode or when the environments are not the same would be the best).

[...]

PR #9357, opened 2021-11-30
Do not display empty worker info

GergelyKalmar:
This PR fixes an issue where empty workerinfo lines are displayed (resulting in an unnecessary newline in that case) when using xdist. Closes https://github.com/pytest-dev/pytest/issues/9316.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -129,6 +129,7 @@ Garvit Shubham
 Gene Wood
 George Kussumoto
 Georgy Dyuldin
+Gergely Kalmár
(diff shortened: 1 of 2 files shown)


GergelyKalmar:
The test error with `macos-py37` seems to be unrelated to this change (probably some flakiness is going on).

nicoddemus (approved):
Thanks @GergelyKalmar, I think the fix is OK to include.

GergelyKalmar:
Merge? :)

PR #9360, opened 2021-12-01
ApproxScalar: Reduce float rounding errors

henzef:
This commit reduces float rounding errors when comparing numbers with
a high (and non-float representable) mantissa and a low tolerance in
pytest.approx().

This came up when comparing geo coordinates with pytest.approx, e.g.:

    >>> tolerance = 5e-7
    >>> expected = 12.3456785
    >>> actual = 12.345678
    >>> abs(expected - actual) <= tolerance
    False
    >>> expected - tolerance <= actual <= expected + tolerance
    True

versus

    >>> expected = 51.500744
    >>> actual = 51.5007435
    >>> abs(expected - actual) <= tolerance
    True
    >>> expected - tolerance <= actual <= expected + tolerance
    True

I am not sure if this minor change warrants a changelog entry. Please let me know if I should add one.

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -3,6 +3,7 @@ import pprint
 from collections.abc import Sized
 from decimal import Decimal
 from numbers import Complex
+from numbers import Real
 from types import TracebackType
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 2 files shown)


bluetech:
This seems sensible to me although I'm not a numerical expert. Are the comparisons always more accurate, or are there cases where the abs might be?

For comparison, I checked a few other test libraries.

Python unittest `assertAlmostEquals` uses abs: https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/unittest/case.py#L872-L875

C++ gtest library `ASSERT_NEAR` uses abs: https://github.com/google/googletest/blob/054a986a8513149e8374fc669a5fe40117ca6b41/googletest/src/gtest.cc#L1593-L1594

JUnit `assertEquals(a, b, delta)` uses abs: https://github.com/junit-team/junit5/blob/4cd1fc064c6864d01ccacf08c4117c7c39d05e0a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java#L180-L183

nicoddemus:
Just a quick chime in:

> I am not sure if this minor change warrants a changelog entry.

It definitely does, as it changes how comparisons work.

rliebscher:
Actual this is more a problem of number representation in general, you can even find simpler examples which easily let both methods fail.
```
expected = 2.2
actual = 1.2
tolerance = 1
```
Depending where the numbers come from using Decimal from the decimal module might be the way to go.

Zac-HD:
Closing this PR as stale; we'd be happy to see a new one opened if the issues above can be addressed.

PR #9363, opened 2021-12-01
Rename pathlib hook parameters 

### LINKED ISSUE #9283: pytest 7: path vs. fspath, pathlib vs. py.path

nicoddemus:
Fixes #9283

### DIFF
--- a/testing/plugins_integration/requirements.txt
+++ b/testing/plugins_integration/requirements.txt
@@ -4,7 +4,9 @@ pytest-asyncio==0.16.0
 pytest-bdd==5.0.0
 pytest-cov==3.0.0
 pytest-django==4.4.0
-pytest-flakes==4.0.4
+# See https://github.com/asmeurer/pytest-flakes/pull/45
+# pytest-flakes==4.0.4
+git+https://github.com/nicoddemus/pytest-flakes@pytest-7.0-compat-part-2
 pytest-html==3.1.1
 pytest-mock==3.6.1
 pytest-rerunfailures==10.2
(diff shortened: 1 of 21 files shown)


testing/plugins_integration/requirements.txt
```diff
@@ -4,7 +4,9 @@ pytest-asyncio==0.16.0
 pytest-bdd==5.0.0
 pytest-cov==3.0.0
 pytest-django==4.4.0
-pytest-flakes==4.0.4
+# See https://github.com/asmeurer/pytest-flakes/pull/45
```
nicoddemus, line 7: Needed a new PR on `pytest-flakes`. 👍

bluetech (approved):

RonnyPfannschmidt (approved):
its a bit unfortunate the names got longer but it may serve users better after all :+1:

testing/plugins_integration/requirements.txt
The-Compiler: Given that @asmeurer plans a release according to https://github.com/asmeurer/pytest-flakes/pull/45#issuecomment-984171700, should we wait for that, so we can use `==4.0.5` or whatever here instead?

nicoddemus: Yeah that's my plan. 👍

nicoddemus: Done!

The-Compiler (approved):
Looking great, thanks!

PR #9369, opened 2021-12-06
Revert "Fix Dead link in README for 'assert statements' "

### LINKED ISSUE #8831: Update all links after doc restructuring
Similarly to what 93d2ccbfb7b99755e8f4a2a3947c466949ee3c46 did, we'll have to update all links in the source code to point to the correct `stable/` locations for when 7.0.0 is released (and/or add redirects in RTD's backend).

Otherwise, as I understand it, once the release is out, all the `stable/` links which currently work (due to `stable` pointing to `6.2.x`, before the restructuring) will break :cry:

The-Compiler:
This reverts commit 15989ddc8f20e49602858e6e4a2631b71e8b9ef1.

Reverts #8945, thus reintroducing #8926 temporarily, but we should be close to
the 7.0.0 release now, where this is the correct URL to use.

Closes #8831 as follow-up to #8858.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,7 +88,7 @@ Due to ``pytest``'s detailed assertion introspection, only plain ``assert`` stat
 Features
 --------
 
-- Detailed info on failing `assert statements <https://docs.pytest.org/en/stable/assert.html>`_ (no need to remember ``self.assert*`` names)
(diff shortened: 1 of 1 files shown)


nicoddemus (approved):

The-Compiler:
FYI I also added reverse redirects for everything now, so the links in the README already work again now!

PR #9373, opened 2021-12-06
Fix hook param name in nonpython example

The-Compiler:
Follow up to #9363

See https://github.com/pytest-dev/pytest/pull/9371/files#diff-523905715b1e18335a8ff84ab46e2c3a28f2c5fec4c63e001574286339fb8c74R30


nicoddemus (approved):
Thanks for the catch!

PR #9377, opened 2021-12-07
Cherry pick release version 7.0.0rc1 (#9375)

The-Compiler:
Co-authored-by: pytest bot <pytestbot@gmail.com>
(cherry picked from commit 85897eddc6d31c5825f5ca28f85e69b8d1506fc4)

Not sure if we should do this for the RC already or not? But I suppose if we have another PR for the final release and cherry pick that back too, I guess it would make sense?

Leaving this open for review despite it being a clean cherry pick, since I'd like a second opinion.


bluetech (approved):
Yes, since 7.0.x is now officially branched off, this is the right time to cherry-pick. Anything from now on should go through backports.

For the final 7 release notes we should add a sentence like "See release notes of 7.0.0rc1" if possible so people understand they need to look a bit further for the actual changes.

The-Compiler:
macOS failure seems unrelated, also turns up on main out of the sudden... will force merge and open a separate issue.

PR #9380, opened 2021-12-07
scripts: Use release branch for changelog URL

The-Compiler:
With a prerelease, /stable won't show the correct changelog.

Can't really test this until the next prerelease though... See #9379.

Should be backported just in case we need a rc2.


The-Compiler:
I was wondering whether to do this for all releases to keep the links more relevant in the future - but it looks like we can't link to old versions on RTD after disabling them, so I suppose keeping /stable for ordinary releases is fine.

nicoddemus (approved):

PR #9381, opened 2021-12-07
[7.0.x] Fixed error message prints function decorators when using assert in Python 3.9 and above

nicoddemus:
Backport of #9359

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -348,6 +348,7 @@ Xixi Zhao
 Xuan Luong
 Xuecong Liao
 Yoav Caspi
+Yuval Shimon
 Zac Hatfield-Dodds
 Zachary Kneupper
 Zoltán Máté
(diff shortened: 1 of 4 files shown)


The-Compiler (approved):

The-Compiler:
Would be good to at least wait for the CI before merging, just in case :wink:. FWIW it looks like that wasn't enforced for the 7.0.x branch (hence the button doing an immediate merge instead of enabling auto-merging).

I just adjusted the [branch protection rules](https://github.com/pytest-dev/pytest/settings/branches) so that the rules for `4.6.x` now apply to `*.x` instead.

nicoddemus:
You are right; I merged when there was only 1 job left to complete, but doesn't hurt to play safe and wait them all.

> I just adjusted the branch protection rules so that the rules for 4.6.x now apply to *.x instead.

Thanks!

PR #9394, opened 2021-12-07
Hide internal stack when using pytest.approx() in bool context

nicoddemus:
This makes the error traceback point directly to the offending usage, rather
than to the internal `Approx.__bool__` method.

Not sure if this warrants a test.

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -101,6 +101,7 @@ class ApproxBase:
         )
 
     def __bool__(self):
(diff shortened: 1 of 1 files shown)


The-Compiler:
Given that this is an usability improvement for a feature new in 7.0 with (hopefully) almost no risk associated, should we backport this?

The-Compiler (approved):

The-Compiler:
Thanks!

The-Compiler:
Backport: 9399

PR #9401, opened 2021-12-08
doc: add a `hook` crossref type

bluetech:
Allow writing

    :hook:`pytest_cmdline_main`

instead of

    :func:`pytest_cmdline_main <_pytest.hookspec.pytest_cmdline_main>`


bluetech:
Not sure if this is the best way to do it, I went by the example of `:fixture:`.

TODO: Go over all existing refs and change them to use `:hook:`.

nicoddemus (approved):
Awesome 😎

bluetech:
Changed all existing refs I could find to use `:hook:`.

PR #9416, opened 2021-12-16
doc: fix a reference in "Writing hook functions"

bluetech:


nicoddemus (approved):

bluetech:
So the backport workflow doesn't work - permission error: https://github.com/pytest-dev/pytest/runs/4546997247?check_suite_focus=true. It did work for me in a private test repo, so I'll try to figure out why it fails here. Will create a manual backport for now.

PR #9420, opened 2021-12-16
Improved fixture reuse by new param keys that can be derived from API ids

### LINKED ISSUE #8914: Indirect session fixture / item reordering issue

haxtibal:
This is a less trivial but (arguably) more powerful alternative to #9350. It fixes #8914. For a review, please follow the commit series and messages.

**Situation**: Both `reorder_items` and `FixtureDef.execute` must decide and agree about whether fixture instances can be reused. The obvious is to decide by comparing parameter values. Sadly none of `prm1 == prm2`, `prm1 is prm2` or `hash(prm1) == hash(prm2)` is generally suited to decide about "are parameters the same?":
- `dict`s are not hashable
- numpy arrays don't compare to bool, and are not hashable either
- `object`s compare and hash by identity, but a user may expect reuse only if it's equal by value

Using the parameter index instead of value doesn't work either, because, well, see #8914.

**Idea**: Give users the option to explicitly control equality of parameters by leveraging [parameter ids](https://docs.pytest.org/en/6.2.x/reference.html#pytest.python.Metafunc.parametrize). Under the hood we introduce a `param_key` to support fixture reordering and caching. The key is built according to the following rules
1. If parameter ids have been passed via API, use that parameter id as key.
1. Else, if a parameter value is hashable, use that parameter value as key.
2. Else, fallback to a parameter values identity.

**Example 1**: Use `id=` to control equality of the unhashable `dict` parameters (rule 1 applies).
```python
def foo(request.param):
    pass  # goal: minimize nrof setup/teardown
@pytest.mark.parametrize("foo", [pytest.param({"data": "a"}, id="0"), pytest.param({"data": "b"}, id="1")], indirect=True)
def test1(foo):
    pass
@pytest.mark.parametrize("foo", [pytest.param({"data": "b"}, id="1"), pytest.param({"data": "a"}, id="0")], indirect=True)
def test2(foo):
    pass
```

**Example 2**: No need to use `id=` or `ids=` if the parameter is hashable by value (rule 2 applies).
```python
def foo(request.param):
    pass  # goal: minimize nrof setup/teardown
@pytest.mark.parametrize("foo", ["a", "b"], indirect=True)
def test1(foo):
    pass
@pytest.mark.parametrize("foo", ["b", "a"], indirect=True)
def test2(foo):
    pass
```

Also related: #244, #5693, #6497, #6541.


src/_pytest/python.py
```diff
@@ -911,6 +912,165 @@ def hasnew(obj: object) -> bool:
...
+        ]
+        return "-".join(this_id)
+
+    def _param_keys_from_parameterset(
+        self, parameterset, valindex
+    ) -> Dict[str, Hashable]:
+        this_id: Dict[str, Hashable] = {}
+        for val, argname in zip(parameterset.values, self.argnames):
+            evaluated_id = self._evaluate_id_function(val, argname, valindex)
+            if isinstance(evaluated_id, str):
```
haxtibal, line 1011: TODO: Support cases where a user provided `id=` function doesn't directly return `str`, but a type that can reasonably converted to `str` (like `Enum`, see `_autoconvert_idval`).

testing/python/fixtures.py
```diff
@@ -1308,6 +1308,32 @@ def test2(no_eq):
         result = pytester.runpytest()
         result.stdout.fnmatch_lines(["*4 passed*"])
 
+    def test_optimize_by_reorder_indirect(self, pytester: Pytester) -> None:
```
haxtibal, line 1311: TODO: Test more cases of how users can provide ids, e.g. `ids=`_Sequence of (or generator for) ids_ (with various return types), and the [pytest_make_parametrize_id](https://docs.pytest.org/en/6.2.x/reference.html#pytest.hookspec.pytest_make_parametrize_id) hook.

src/_pytest/python.py
```diff
@@ -911,6 +912,172 @@ def hasnew(obj: object) -> bool:
     return False
 
 
+@attr.s(auto_attribs=True, eq=False)
```
haxtibal, line 915: TODO: Apply suggestions from https://github.com/pytest-dev/pytest/pull/9350#discussion_r764843625.

bluetech:
@haxtibal Thanks a lot for creating this separate PR with nice commits and all!

I'm looking forward to reviewing this PR, when I have the time available. Right now we are mostly focused on getting the 7.0 release out the door, just wanted to write this so you don't get discouraged :)

src/_pytest/fixtures.py
```diff
@@ -1015,7 +1020,7 @@ def execute(self, request: SubRequest) -> FixtureValue:
             # note: comparison with `==` can fail (or be expensive) for e.g.
```
haxtibal, line 1020: TODO: The comment doesn't hold anymore. Numpy arrays would now be compared by parameter id if provided, or else by identity. Both methods won't fail and ain't expensive.

src/_pytest/fixtures.py
```diff
@@ -251,18 +252,18 @@ def get_parametrized_fixture_keys(item: nodes.Item, scope: Scope) -> Iterator[_K
         # cs.indices.items() is random order of argnames.  Need to
```
haxtibal, line 252: TODO: Comment is now misleading, as we don't use cs.indices anymore.

src/_pytest/python.py
```diff
@@ -911,6 +911,124 @@ def hasnew(obj: object) -> bool:
     return False
 
 
+@final
+@attr.s(frozen=True, auto_attribs=True)
```
bluetech, line 933: Can add `slots=True` here as well, doesn't hurt.

src/_pytest/python.py
```diff
@@ -911,6 +911,124 @@ def hasnew(obj: object) -> bool:
...
+class IdMaker:
+    argnames: Iterable[str] = attr.Factory(list)
```
bluetech, line 935: When refactoring to a class, it's less natural to assume it will only be called once. So I think it will be better to avoid `Iterable` and use `Sequence` instead.

bluetech, line 935: Since only one place calls `IdMaker`, and it provides all parameters, I think we should get rid of the defaults of all fields.

src/_pytest/python.py
```diff
@@ -911,6 +911,124 @@ def hasnew(obj: object) -> bool:
...
+    parametersets: Iterable[ParameterSet] = attr.Factory(list)
+    idfn: Optional[Callable[[Any], Optional[object]]] = None
+    ids: Optional[List[Union[None, str]]] = None
```
bluetech, line 938: List -> Sequence, since we don't mutate it.

src/_pytest/python.py
```diff
@@ -911,6 +911,124 @@ def hasnew(obj: object) -> bool:
...
+        return None
+
+    def _test_id_from_parameterset(self, parameterset, valindex) -> str:
+        this_id = [
+            self._idval(val, argname, valindex)
+            for val, argname in zip(parameterset.values, self.argnames)
+        ]
+        return "-".join(this_id)
+
+    def _evaluate_idval_function(
```
bluetech, line 1018: Up to here the order of the functions in the class was `<function F>, <functions called by F>`. So I think it would be better to bring `def _idval` here before `_evaluate_idval_function` and the functions it calls.

src/_pytest/python.py
```diff
@@ -911,6 +911,124 @@ def hasnew(obj: object) -> bool:
...
+                    yield self._test_id_from_parameterset(parameterset, valindex)
+                else:
+                    yield _ascii_escaped_by_config(id, self.config)
+
+    def _id_at_index(self, valindex: int) -> Optional[str]:
+        if self.ids and valindex < len(self.ids):
+            return self.ids[valindex]
+        return None
+
+    def _test_id_from_parameterset(self, parameterset, valindex) -> str:
```
bluetech, line 998: Missing types here

src/_pytest/python.py
```diff
@@ -1357,61 +1476,8 @@ def _idval(
...
-            for val, argname in zip(parameterset.values, argnames)
-        ]
-        return "-".join(this_id)
-    else:
-        return _ascii_escaped_by_config(id, config)
+    id_maker = IdMaker(idfn=idfn, config=config, nodeid=nodeid)
+    return id_maker._idval(val, argname, idx)
 
 
 def idmaker(
```
bluetech, line 1544: Let's remove `_idval` and `idmaker`, as they're not used anymore.

Some tests in `testing/python/metafunc.py` use them, so we should either convert the tests to use `IdMaker`, or just move `_idval` and `idmaker` over to `testing/python/metafunc.py` for now.

src/_pytest/python.py
```diff
@@ -911,6 +911,124 @@ def hasnew(obj: object) -> bool:
...
+@attr.s(frozen=True, auto_attribs=True)
+class IdMaker:
+    argnames: Iterable[str] = attr.Factory(list)
+    parametersets: Iterable[ParameterSet] = attr.Factory(list)
+    idfn: Optional[Callable[[Any], Optional[object]]] = None
+    ids: Optional[List[Union[None, str]]] = None
+    config: Optional[Config] = None
+    nodeid: Optional[str] = None
+
+    def make_unique_test_ids(self) -> List[str]:
```
bluetech, line 942: If you can add some docstring for this function that would be nice (don't have to).

bluetech (commented):
I think I'm going to leave a separate review for each commit. This is for "Refactor idmaker functions into class IdMaker'.

The refactoring LGTM, with some comments.

src/_pytest/python.py
```diff
@@ -911,6 +911,124 @@ def hasnew(obj: object) -> bool:
...
+            # Suffix non-unique IDs to make them unique.
+            for index, test_id in enumerate(resolved_ids):
+                if test_id_counts[test_id] > 1:
+                    resolved_ids[index] = f"{test_id}{test_id_suffixes[test_id]}"
+                    test_id_suffixes[test_id] += 1
+
+        return resolved_ids
+
+    def _make_resolved_ids(self) -> Iterable[str]:
+        """Combine multiple parameter IDs to a single string for each parameter set."""
```
bluetech, line 982: Not entirely accurate. Suggestion: "Resolve IDs for all parameter sets (may contain duplicates)."

bluetech (commented):
Some more comments on the first commit. Sorry for all of this, I just hate to pass on an opportunity to make some pytest code clearer :)

If it's too much I can cherry pick the commit to a separate PR along with my suggestions.

src/_pytest/python.py
bluetech: Right, some of the suggestions there are still relevant.

src/_pytest/python.py
```diff
@@ -911,6 +912,23 @@ def hasnew(obj: object) -> bool:
...
+@attr.s(auto_attribs=True, eq=False)
+class SafeHashWrapper:
+    obj: Any
+
+    def __eq__(self, other) -> bool:
+        try:
+            res = self.obj == other
+            return bool(res)
+        except Exception:
+            return id(self.obj) == id(other)
```
bluetech, line 924: I guess this is equivalent to `return self.obj is other` so can maybe just do that

src/_pytest/python.py
```diff
@@ -942,6 +960,24 @@ def make_unique_test_ids(self) -> List[str]:
 
         return resolved_ids
 
+    def make_parameter_keys(self) -> Iterable[Dict[str, Hashable]]:
+        """Make hashable keys for parameters in a parametrized test.
+
```
bluetech, line 965: Will add a paragraph describing the return value:

For each parameter set, generates a dict mapping each parameter to its key.

src/_pytest/python.py
```diff
@@ -942,6 +960,24 @@ def make_unique_test_ids(self) -> List[str]:
...
+        This key will be considered to determine if parameters
```
bluetech, line 966: This key will be considered (along with the parameter name) to determine ...

src/_pytest/python.py
```diff
@@ -966,6 +1002,19 @@ def _test_id_from_parameterset(self, parameterset, valindex) -> str:
         ]
         return "-".join(this_id)
 
+    def _param_keys_from_parameterset(
+        self, parameterset, valindex
+    ) -> Dict[str, Hashable]:
+        this_id: Dict[str, Hashable] = {}
+        for val, argname in zip(parameterset.values, self.argnames):
+            evaluated_id = self._evaluate_idval_function(val, argname, valindex)
+            if isinstance(evaluated_id, str):
```
bluetech, line 1011: This test should be just `if evaluated_id is not None` no?

bluetech (commented):
Review for "Add IdMaker.make_parameter_keys".

The commit message note writes `pytest.param("apple", ids=["fruit"])` but meant to write `pytest.param("apple", id="fruit")`.

Regarding the the problem with the duplicates, when there are conflicts in `make_unique_test_ids` pytest disambiguates them with a suffix. This should probably also be done for parameter keys then, since duplicates probably don't make sense within the same group of parameter sets.

bluetech (commented):
Review of "Extend CallSpec2 with param_keys".

LGTM, though note #9531 causes some conflicts, sorry about that.

bluetech (commented):
"Extend SubRequest with param_key" LGTM.

src/_pytest/fixtures.py
bluetech: Right, so fix the comment.

bluetech (commented):
"Let reorder_items use our new parameter key" LGTM with comment

src/_pytest/fixtures.py
bluetech: Arr I definitely did not mean to write "Right, so fix the comment." - I meant "Right, need to fix the comment." (just to mark it)

src/_pytest/fixtures.py
bluetech: > TODO: The comment doesn't hold anymore. Numpy arrays would now be compared by parameter id if provided, or else by identity. Both methods won't fail and ain't expensive.

Are you sure about this? IIUC, numpy arrays wrapped in `SafeHashWrapper` would still compare by `==`, so the expensive bit still holds.

Further, that makes me think that some types (like numpy arrays but also others) may implement `__eq__` but not `__hash__`. The `SafeHashWrapper` makes them implement `__hash__`, but the hash value doesn't match `__eq__` which is not allowed and can cause trouble. So this is a problem... Maybe `SafeHashWrapper` should check `Hashable` also in `__eq__`?

src/_pytest/fixtures.py
```diff
@@ -1037,7 +1037,7 @@ def execute(self, request: SubRequest) -> FixtureValue:
         return result
 
     def cache_key(self, request: SubRequest) -> object:
-        return request.param_index if not hasattr(request, "param") else request.param
+        return request.param_key
```
bluetech, line 1040: (I need to remind myself a bit how CallSpec2 works with indirect parametrization...)

bluetech (commented):
Review for " Let FixtureDef.cache_key use our new parameter key "

src/_pytest/python.py
haxtibal: > doesn't hurt

IIUC [attrs docs](https://www.attrs.org/en/stable/how-does-it-work.html#summary) correctly, `@attr.s(frozen=True, slots=True, ...)` actually could hurt in some cases:
> You should avoid instantiating lots of frozen slotted classes (i.e. @frozen) in performance-critical code

I can't reason offhand about whether we'd be affected, can you?

src/_pytest/python.py
haxtibal: > less natural to assume it will only be called once

Yes... It shouldn't be a problem to iterate over an `Iterable` multiple times (`__iter__()` will provide a fresh iterator on each call). But users of `IdMaker` may expect index access to `argnames` and `parametersets`, so `Sequence` is the better choice anyway.

src/_pytest/python.py
haxtibal: > (along with the parameter name)

What is the "parameter name"? `reorder_items` uses e.g. `key = (argname, param_key, item.path, item_cls)`, and `FixtureDef.cache_key` uses `param_key`. Do you mean argument name?

haxtibal:
> Regarding the the problem with the duplicates, when there are conflicts in make_unique_test_ids pytest disambiguates them with a suffix. This should probably also be done for parameter keys then, since duplicates probably don't make sense within the same group of parameter sets.

For `make_parameter_keys`, leaving duplicates duplicated is at the heart of what makes the new feature work: Only duplicates hash to the same value, and same hash will in later stages be the criterion to skip all but one duplicate for the sake of fixture s/t optimization. Does this reasoning make sense?

src/_pytest/fixtures.py
haxtibal: >  numpy arrays wrapped in SafeHashWrapper would still compare by ==, so the expensive bit still holds

You're right, `numpy.ndarray.__eq__` is still executed and compares value by value. i.e. the expensive bit holds. Then it returns list (or exception) instead of bool which is catched by `SafeHashWrapper` and causes fallback to `id()`. I think the only reliable way to get rid of the expensive bit are user provided ids. If not given, we actually wish to prioritize fallback to deep value compare (possibly expensive) over fallback to id. That's the trade-off.

> the hash value doesn't match __eq__ which is not allowed and can cause trouble

Full ack. If `__eq__` and `__hash__` are not consistent yet, I'll have to fix it.

haxtibal: > Maybe SafeHashWrapper should check Hashable also in __eq__?

Quite good idea, thanks, did that. As a bonus that makes at least numpy arrays cheap, because we now fall back early.

One remaining issue: A user could mutate parameters after reorder items took place, which renders the reordering invalid. It becomes obvious when we talk about hashability here, but I think the problem was always there, and I have no solution.

src/_pytest/python.py
haxtibal: Tests now use IdMaker directly, and became a bit more verbose because we removed default values for IdMaker.

src/_pytest/python.py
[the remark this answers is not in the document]
haxtibal: Renamed to `make_unique_parameterset_ids`. But otoh, I also added a docstring that says `"Make a string for each ParameterSet to identify test invocations in pytest logs"`. In case the docstring is right, wouldn't it be more a "testinvocation_id" rather than a "parameterset_id"? Or in case the docstring is wrong, how is it wrong?

haxtibal:
@bluetech Thanks for taking your time! That was a really helpful review.

Most of the remarks are hopefully done, and the commit series is rebased on latest main. I left conversations open where response from you would be good.

src/_pytest/python.py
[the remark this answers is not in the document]
haxtibal: `make_unique_parameterset_ids` makes perfect sense, sorry for the noise. I'll change the docstring.

It's `CallSpec2.id` that (again) joins `parameterset_ids` in `CallSpec2._idlist` (e.g. `['indirectparamval1', 'directparamval1-directparamval2']`) to the string that appears for a test invocation (like in `test_ids.py::test_foo[indirectparamval1-directparamval1-directparamval2] PASSED`). And even then, a "test invocation id" as suggested in my earlier commit would additionally have to include the test function name.

src/_pytest/python.py
bluetech: Hmm interesting. We use this all over the place so we should probably check if it has any effect separately. But until then let's keep it in this PR.

src/_pytest/python.py
```diff
@@ -929,6 +929,130 @@ def hasnew(obj: object) -> bool:
...
+class IdMaker:
+    argnames: Sequence[str]
+    parametersets: Sequence[ParameterSet]
+    idfn: Optional[Callable[[Any], Optional[object]]]
+    ids: Optional[Sequence[Union[None, str]]]
+    config: Optional[Config]
+    nodeid: Optional[str]
+
+    def make_unique_parameterset_ids(self) -> List[str]:
+        """Make a string for each ParameterSet to identify test invocations in pytest logs.
```
bluetech, line 976: The nodeid (where the parameterset ID ends up) is somewhat more than just for the pytest logs. So maybe something like this:

Make a unique identifier for each ParameterSet, that may be used to identify
the parametrization in a node ID.

src/_pytest/python.py
```diff
@@ -929,6 +929,130 @@ def hasnew(obj: object) -> bool:
...
+
+        Format is <prm_1_token>-...-<prm_n_token>[counter],
+        where prmset_x_token is
```
bluetech, line 979: prm_x_token

src/_pytest/python.py
```diff
@@ -929,6 +929,130 @@ def hasnew(obj: object) -> bool:
...
+        - the parameter id, derived from value or user provided id, if applicable
```
bluetech, line 980: Would split this to two:

- user-provided id, if given
- id derived from the value, for certain types

src/_pytest/python.py
```diff
@@ -929,6 +929,130 @@ def hasnew(obj: object) -> bool:
...
+        - else <argument_name><test_invocation_nr>
```
bluetech, line 981: Would use names matching the field names for clarity:

- else `<argname><parameterset index>`

src/_pytest/python.py
```diff
@@ -929,6 +929,130 @@ def hasnew(obj: object) -> bool:
...
+        The counter suffix is appended only in case a string wouldn't be unique otherwise.
+        """
+        resolved_ids = list(self._make_resolved_ids())
+
+        # All IDs must be unique!
+        unique_ids = set(resolved_ids)
+        if len(unique_ids) != len(resolved_ids):
+
+            # Record the number of occurrences of each test ID.
```
bluetech, line 990: I think we shouldn't use the term "test ID" here because it is not an ID for the entire test (the nodeid) but a component of it. I suggest just dropping the "test_" prefix from here and below.

bluetech (commented):
Second review of the " Refactor idmaker functions into class IdMaker " commit.

I left a few comments, but it looks good to me now. It's nice regardless, so if you agree, I can cherry-pick it to main already, to reduce the size of the remaining PR.

I will try to review the rest again soon.

haxtibal:
> if you agree, I can cherry-pick it to main already, to reduce the size of the remaining PR.

Just applied your recent comments. Sure, cherry-pick it, or in case you find some more bits you'd like to change, it's fine for me if you amend the commit to spare another review iteration on "Refactor idmaker functions into class IdMaker".

bluetech:
(I rebased on top of the cherry-picked commit)

haxtibal:
@bluetech Sorry for delay in response.

> the example in the first commit message is not very fitting

Agreed, my example not only missed `scope="module"`, but also contained syntax errors.

> I suggest the following example [...]

Agreed. That's better suited.

> it does not make much sense for someone to duplicate an id within a single parametrization (basically, for a single test)?

Agreed. I'd suggest to print a warning or error message then, and maybe even bail out early, instead of deduplicating silently?

> The mechanism here is only useful for better managing the lifetime of higher-scoped indirect fixtures _across_ tests.

Agreed. Right now I'm slightly puzzled why direct parametrization (as in my commit example) is subject to fixture lookup at all - as opposed to unconditionally take values from `CallSpec2.funcargs`. Because, as you said, there's just no way to optimize something for direct parametrization. It's not new behavior from this PR (I hope), I just need to think twice to maybe understand the original intent.

> And that made me further think, if we shouldn't just completely do away with falling back on comparing values (with the entire `SafeHashWrapper` business it entails), and instead always generate string IDs (basically, call `_idval` in `_parameter_keys_from_parameterset`). This will mostly keep the existing behavior (since `_idval_from_argname` eventually falls back to the parameterset idx), but still allow the user to override with an explicit ID (which we could recommend they do).
> 
> This stuff is pretty entangled and complicated so I may be confused or missing something, but WDYT?

We can't simply rely on `_idval`, as the string it generates is not guaranteed to be unique within fixture scope. If we do what you suggested we can get incorrect parametrization:

```python
import pytest

@pytest.fixture(scope="module")
def fruitfix(request):
    print(f"setup fruitfix with {request.param}")
    return request.param

@pytest.mark.parametrize("fruitfix", [(1, 2), (3, 4)], indirect=True)
def test_fruits_1(fruitfix):
    print(f"test1 using {fruitfix}")

@pytest.mark.parametrize("fruitfix", [(3, 4), (5, 6)], indirect=True)
def test_fruits_2(fruitfix):
    print(f"test2 using {fruitfix}")
```

Output with `_evaluate_idval_function`, or `SafeHashWrapper` if the former is `None` (the current approach):
```
setup fruitfix with (1, 2)
test1 using (1, 2)
setup fruitfix with (3, 4)
test1 using (3, 4)
test2 using (3, 4)
setup fruitfix with (5, 6)
test2 using (5, 6)
```

Output with using `_idval` as suggested (note how tuple `(5, 6)` is not passed to `test2`):
```
setup fruitfix with (1, 2)
test1 using (1, 2)
test2 using (1, 2)
setup fruitfix with (3, 4)
test1 using (3, 4)
test2 using (3, 4)
```

This is because `_idval` only generates unique value representations for scalar parameters like `int`, `float`, `double`. But not for e.g. tuples, where it would fallback to argument names + counter. This is not unique in scope, and therefore suffers from a similar problem as using the index (the strategy before this PR).

haxtibal:
> but WDYT?

Sorry for the edits in yesterdays answer. They make decision even simpler: We can trade the current `SafeHashWrapper` approach for a documentation statement _"Provide ids, else your parametrized fixtures won't get optimized"_.

My personal opinion is to keep `SafeHashWrapper` and the implied compare-by-value. When I came to pytest I was under the impression that reuse of parametrized fixtures is an [advertised feature](https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes) and found it to be the most outstanding one compared to other testing frameworks (although it mostly didn't work because of a bug). As a user, I would like if it works without custom annotations where possible.

But I can understand if you want to leave the compare-by-value fallback out: It's simpler to document and understand, and it's probably simpler to maintain.

@bluetech It's your turn to decide, I will for sure be fine with either decision.

obestwalter:
@bluetech would you want to have another look at this, otherwise this might better be closed?

RonnyPfannschmidt:
as theres no feedback since the last pytest sprint im closing this for now
im happy to review a revival once its feasible

PR #9438, opened 2021-12-25
pytest: bring back direct imports of TempdirFactory, Testdir

### LINKED ISSUE #9432: [prerelease] pytest.TmpdirFactory is not importable anymore
Discovered in https://github.com/ClearcodeHQ/pytest-postgresql/issues/533#issuecomment-999567968 via #9415:

Doing e.g. `python3 -c "import pytest; pytest.TempdirFactory"` fails since  5e883f51959f900fcf985493ce2d09e07bad7e00. This is a bit similar to #9396, but #9400 won't help with this one.

Test modules often use `pytest.TempdirFactory` for type annotations - in fact, that's why [that was added](https://docs.pytest.org/en/stable/changelog.html#improvements) in 6.2.0.

However, it being set from a plugin means that:

a) Test modules aren't even importable anymore outside of pytest - in case of `pytest-postgresql`, this [breaks an importability check](https://github.com/ClearcodeHQ/pytest-postgresql/runs/4607221002?check_suite_focus=true), but I can imagine various other things breaking...
b) Perhaps more importantly, static analysis in IDEs or tools like mypy [aren't aware anymore](https://github.com/ClearcodeHQ/pytest-postgresql/runs/4607221090?check_suite_focus=true) that the attribute exists - however, after all, that's *the whole point* of it existing, and [lots of projects use it that way](https://sourcegraph.com/search?q=context:global+:%5C+*%28pytest%5C.%29%3FTempdirFactory+lang:python+-file:pytester.py&patternType=regexp). (Some import it from `_pytest.tmpdir`, probably because they did that before 6.2.0 - I'm not too worried about those, they knew what they're in for)

Reproducer:

```python
from pytest import TempdirFactory

def test_nothing(tmpdir: TempdirFactory):
    pass
```

Fails with mypy:

```
test_x.py:1: error: Module "pytest" has no attribute "TempdirFactory"; maybe "TempPathFactory"?
Found 1 error in 1 file (checked 1 source file)
```

What to do? I suppose as a stop-gap until we want to distribute the plugin independently from pytest, we could import the class from the plugin or so?

bluetech:
The monkeypatch approach doesn't work for `import pytest;
pytest.TempdirFactory`.

Fix #9432.

### DIFF
--- a/src/_pytest/legacypath.py
+++ b/src/_pytest/legacypath.py
@@ -10,13 +10,26 @@ from typing import Union
 import attr
 from iniconfig import SectionWrapper
 
-import pytest
+from _pytest.cacheprovider import Cache
 from _pytest.compat import final
 from _pytest.compat import LEGACY_PATH
 from _pytest.compat import legacy_path
+from _pytest.config import Config
+from _pytest.config import hookimpl
+from _pytest.config import PytestPluginManager
 from _pytest.deprecated import check_ispytest
+from _pytest.fixtures import fixture
+from _pytest.fixtures import FixtureRequest
+from _pytest.main import Session
+from _pytest.monkeypatch import MonkeyPatch
+from _pytest.nodes import Collector
+from _pytest.nodes import Item
 from _pytest.nodes import Node
+from _pytest.pytester import HookRecorder
+from _pytest.pytester import Pytester
+from _pytest.pytester import RunResult
 from _pytest.terminal import TerminalReporter
+from _pytest.tmpdir import TempPathFactory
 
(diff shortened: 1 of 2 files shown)


bluetech:
Adding the imports from `_pytest.legacypath` to `pytest` causes an import loop which necessitates all of the changes to the plugin to import from `_pytest` instead of `pytest`. This is unfortunate and will need to be undone when moving to a separate package but is required for now.

nicoddemus (approved):
Thanks for tackling this!

PR #9440, opened 2021-12-25
Restore {Code,TracebackEntry}.path to py.path and add alternative

### LINKED ISSUE #9423: [prerelease] `ExceptionInfo.traceback[...].path` changed from py.path to pathlib.Path
As for the `ExceptionInfo.traceback[...].path` one: The PR looks good, but it seems to me that we should retain and perhaps deprecate the `path` attribute properly, instead of (more or less silently) replacing it, despite it being public API)?

_Originally posted by @The-Compiler in https://github.com/pytest-dev/pytest/discussions/9415#discussioncomment-1835181_

bluetech:
In 92ba96b0612e6b06bb8f4ab05bd75481d2504806 we have changed the `path`
attribute to return a `pathlib.Path` instead of `py.path.local` without
a deprecation hoping it would be alright. But these types are somewhat
public, reachable through `ExceptionInfo.traceback`, and broke code in
practice. So restore them in the legacypath plugin and add `Path`
alternatives under a different name - `source_path`.

Fix #9423.

### DIFF
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -1,5 +1,6 @@
 import ast
 import inspect
+import os
 import re
 import sys
 import traceback
@@ -83,7 +84,7 @@ class Code:
         return self.raw.co_name
 
     @property
-    def path(self) -> Union[Path, str]:
+    def source_path(self) -> Union[Path, str]:
         """Return a path object pointing to source code, or an ``str`` in
         case of ``OSError`` / non-existing file."""
         if not self.raw.co_filename:
@@ -218,7 +219,7 @@ class TracebackEntry:
         return self.lineno - self.frame.code.firstlineno
 
     def __repr__(self) -> str:
-        return "<TracebackEntry %s:%d>" % (self.frame.code.path, self.lineno + 1)
+        return "<TracebackEntry %s:%d>" % (self.frame.code.source_path, self.lineno + 1)
 
     @property
     def statement(self) -> "Source":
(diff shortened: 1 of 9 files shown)


bluetech:
The alternative to this as discussed in the issue is to go through with the breaking change.

For this PR:

- Honor our promise to not break stuff without deprecation
- Will restore compat for the `promise` project and whatever other unknown projects that will be broken by this

Against this PR:

- `source_path` is uglier and less discoverable than `path`
- We estimate not many users are using this at all
- Some code might work fine with the silent switch from py.path -> pathlib and will end up better
- The `path` fix is temporary as we aim to deprecate and remove py.path stuff eventually

Overall it will be better to go with the breaking change in this case, i.e. close this PR and document the change as breaking rather than `trivial` as 92ba96b0612e6b06bb8f4ab05bd75481d2504806 did. WDYT?

src/_pytest/_code/code.py
```diff
@@ -343,10 +344,10 @@ def f(cur: TracebackType) -> Iterable[TracebackEntry]:
 
     def cut(
         self,
-        path: Optional[Union[Path, str]] = None,
+        path: Optional[Union["os.PathLike[str]", str]] = None,
```
bluetech, line 347: This part also fixes the `sybil` failure - it passes a string absolute path to `cut`, which used to work because `py.path.local('/foo') == '/foo'` but `Path('/foo') != '/foo'`. So the code now compares strings.

Even if we go with the breaking change option we should get this included in pytest 7.0 (this part is non breaking)

nicoddemus:
> Overall it will be better to go with the breaking change in this case, i.e. close this PR and document the change as breaking rather than trivial as 92ba96b did.

As I commented in the issue, I think we should just go for it, but I'm not against if others prefer to play safe here and go with the PR instead. 👍

bluetech:
Replaced by #9447.

PR #9451, opened 2021-12-27
doc: document {Code,TracebackEntry}.path changes as breaking

### LINKED ISSUE #9423: [prerelease] `ExceptionInfo.traceback[...].path` changed from py.path to pathlib.Path

bluetech:
Closes #9423.


nicoddemus (approved):

bluetech:
The CI failures are github issues, the job that matters (docs) passed.

PR #9454, opened 2021-12-27
Assert quote type in docs

### LINKED ISSUE #9431: The -k expression works with double quotes but not single quotes
When using single quotes and trying a conjunction of expressions:
` -k 'vet_a and vet_a'`
I get the error message:
`ERROR: file or directory not found: and`

I finally tried double quotes and it worked.

Note: the docs and -h help text using single quotes in the -k examples so at least update those as a quick workaround...

pytest 6.2.4
Windows 8.1

SauravMaheshkar:
This PR closes #9431 with minimal changes to the docs.

### DIFF
--- a/doc/en/how-to/usage.rst
+++ b/doc/en/how-to/usage.rst
@@ -35,11 +35,13 @@ Pytest supports several ways to run and select tests from the command-line.
 
 .. code-block:: bash
 
-    pytest -k "MyClass and not method"
+    pytest -k 'MyClass and not method'
 
 This will run tests which contain names that match the given *string expression* (case-insensitive),
 which can include Python operators that use filenames, class names and function names as variables.
 The example above will run ``TestMyClass.test_something``  but not ``TestMyClass.test_method_simple``.
+For the default windowns terminal ``cmd.exe`` you should use double quotes i.e. ``"``. For all other
+shells ``'`` is preferred.
 
 .. _nodeids:
 
--- a/doc/en/reference/reference.rst
+++ b/doc/en/reference/reference.rst
@@ -1825,7 +1825,8 @@ All the command-line flags can be obtained by running ``pytest --help``::
                             functions containing extra names in their
                             'extra_keyword_matches' set, as well as functions
                             which have names assigned directly to them. The
-                            matching is case-insensitive.
+                            matching is case-insensitive. For the default
+                            windows terminal double quotes are preferred.
       -m MARKEXPR           only run tests matching given mark expression.
                             For example: -m 'mark1 and not mark2'.
       --markers             show markers (builtin, plugin and per-project ones).


doc/en/reference/reference.rst
```diff
@@ -1815,17 +1815,18 @@ All the command-line flags can be obtained by running ``pytest --help``::
...
                             expression where all names are substring-matched
                             against test names and their parent classes.
-                            Example: -k 'test_method or test_other' matches all
+                            Example: -k "test_method or test_other" matches all
                             test functions and classes whose name contains
-                            'test_method' or 'test_other', while -k 'not
-                            test_method' matches those that don't contain
-                            'test_method' in their names. -k 'not test_method
-                            and not test_other' will eliminate the matches.
+                            "test_method" or "test_other", while -k "not
```
nicoddemus, line 1820: I thought we wanted to use `'` in general, with `"` (per `cmd.exe`) being the exception?

nicoddemus: Oh btw, you don't need to update this text, it is generated automatically when we build the documentation.

SauravMaheshkar: I think there might some confusion here but @taigit request changes in this [reply](https://github.com/pytest-dev/pytest/issues/9431#issuecomment-1001625011)

nicoddemus: Not sure, but regardless this file is updated automatically so any changes will get overwritten anyway.

doc/en/how-to/usage.rst
```diff
@@ -35,11 +35,13 @@ Pytest supports several ways to run and select tests from the command-line.
...
 .. code-block:: bash
 
-    pytest -k "MyClass and not method"
+    pytest -k 'MyClass and not method'
 
 This will run tests which contain names that match the given *string expression* (case-insensitive),
 which can include Python operators that use filenames, class names and function names as variables.
 The example above will run ``TestMyClass.test_something``  but not ``TestMyClass.test_method_simple``.
+For the default windowns terminal ``cmd.exe`` you should use double quotes i.e. ``"``. For all other
+shells ``'`` is preferred.
```
bluetech, line 44: I think it would be nice to have this comment as a Sphinx "note". Also some wording suggestions:

```suggestion

.. note::

    If you are using the Windows command interpreter ``cmd.exe`` to run pytest, you should use double quotes, as in ``"MyClass and not method"``.
    In most other shells, ``'`` is preferred.
```

doc/en/reference/reference.rst
bluetech: Right, changes to this file should be dropped. This means it wouldn't have helped @taigit but I don't think there is enough space in the pytest `--help` to explain the issue, and for most users `'` is the better quoting.

bluetech (changes requested):

nicoddemus:
@SauravMaheshkar gentle ping.

SauravMaheshkar:
Sorry folks got busy with work and school. Will close this for now and reopen at a later time.

PR #9456, opened 2021-12-28
ci: run plugins job with python 3.9 instead of 3.7

bluetech:
Latest Django release dropped support for 3.7.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -100,9 +100,9 @@ commands =
 [testenv:plugins]
 # use latest versions of all plugins, including pre-releases
 pip_pre=true
-# use latest pip and new dependency resolver (#7783)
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -100,9 +100,9 @@ commands =
 [testenv:plugins]
 # use latest versions of all plugins, including pre-releases
 pip_pre=true
-# use latest pip and new dependency resolver (#7783)
+# use latest pip to get new dependency resolver (#7783)
 download=true
-install_command=python -m pip --use-feature=2020-resolver install {opts} {packages}
```
bluetech, line 105: `--use-feature=2020-resolver` is the default now and the flag emits a warning

nicoddemus (approved):

PR #9477, opened 2022-01-04
- instead of throwing exception, return empty sequence

franzhaas:
When using pytest to test pyoxidizer projects, it was necessary to have the source code under test on the file system, instead of inside the single file executable. For me this meant that the actuall final product, could not be tested by pytest.

this was due to the fact that pytest needs a file path in order to rewrite assertions.

By returning a empty sequence instead, the tests can be executed. (without the rewriting).

An alternative would be to use  the --assert=plain switch.
however, i did not find a way how to set this flag from inside a repl...

Thanks for the great tool!

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1126,12 +1126,12 @@ class Config:
             return
 
         def _get_files(dist):
-            #dist.files does not make sense for dists
-            #who are not stored on a filesystem
-            #at least pyoxidizer does throw a
-            #not implemented assertion in this case
+            # dist.files does not make sense for dists
+            # who are not stored on a filesystem
+            # at least pyoxidizer does throw a
+            # not implemented assertion in this case
             try:
-                return dist.files 
+                return dist.files
             except NotImplementedError:
                 return []
 
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1115,35 +1115,31 @@ class Config:
                 self._mark_plugins_for_rewrite(hook)
         self._warn_about_missing_assertion(mode)
 
+
     def _mark_plugins_for_rewrite(self, hook) -> None:
         """Given an importhook, mark for rewrite any top-level
         modules or packages in the distribution package for
         all pytest plugins."""
-        self.pluginmanager.rewrite_hook = hook
+        try:
+            self.pluginmanager.rewrite_hook = hook
 
-        if os.environ.get("PYTEST_DISABLE_PLUGIN_AUTOLOAD"):
-            # We don't autoload from setuptools entry points, no need to continue.
-            return
+            if os.environ.get("PYTEST_DISABLE_PLUGIN_AUTOLOAD"):
+                # We don't autoload from setuptools entry points, no need to continue.
+                return
 
-        def _get_files(dist):
-            # dist.files does not make sense for dists
-            # who are not stored on a filesystem
-            # at least pyoxidizer does throw a
-            # not implemented assertion in this case
-            try:
-                return dist.files
-            except NotImplementedError:
-                return []
+            package_files = (
+                str(file)
+                for dist in importlib_metadata.distributions()
+                if any(ep.group == "pytest11" for ep in dist.entry_points)
+                for file in dist.files or []
+            )
 
-        package_files = (
-            str(file)
-            for dist in importlib_metadata.distributions()
-            if any(ep.group == "pytest11" for ep in dist.entry_points)
-            for file in _get_files(dist) or []
-        )
+            for name in _iter_rewritable_modules(package_files):
+                hook.mark_rewrite(name)
+        except Exception as e:
+            apropriate_output_function("unexpected exception %s while marking plugins for assertion rewrites, see LINK TO HELPFULL DOCUMENTATION", e)
+            raise e
 
-        for name in _iter_rewritable_modules(package_files):
-            hook.mark_rewrite(name)
 
     def _validate_args(self, args: List[str], via: str) -> List[str]:
         """Validate known args."""
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -1125,11 +1125,21 @@ class Config:
             # We don't autoload from setuptools entry points, no need to continue.
             return
 
+        def _get_files(dist):
+            #dist.files does not make sense for dists
+            #who are not stored on a filesystem
+            #at least pyoxidizer does throw a
+            #not implemented assertion in this case
+            try:
+                return dist.files 
+            except NotImplementedError:
+                return []
+
         package_files = (
             str(file)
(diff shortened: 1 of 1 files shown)


src/_pytest/config/__init__.py
```diff
@@ -1125,11 +1125,21 @@ def _mark_plugins_for_rewrite(self, hook) -> None:
             # We don't autoload from setuptools entry points, no need to continue.
             return
 
+        def _get_files(dist):
+            # dist.files does not make sense for dists
+            # who are not stored on a filesystem
+            # at least pyoxidizer does throw a
+            # not implemented assertion in this case
+            try:
+                return dist.files
```
nicoddemus, line 1134: Hi @franzhaas, thanks for the PR! 

However, according to [the api reference](https://importlib-metadata.readthedocs.io/en/latest/api.html#importlib_metadata.Distribution.files), `dist.files` should return `List of PackagePath for this distribution or None`, so seems like pyoxidizer is wrong in raising `NotImplementedError`.

I suggest reporting/opening a PR in `pyoxidizer` to fix that instead.

franzhaas: I understand what you are pointing to.

The dist is not in a file system, therefore the concept of path becomes questionable. Returning  None, should indicate that meta data is missing, and a empty list should indicate empty meta data.

I have sympathy for pyoxidizer to raise notimplentederror. At this point, i can not see an obvious value that is both, correct, and helpfull.

Is there a API to disable the assertion rewriting, from within python that i did overlook? There is already a command line argument. To have this accessible would solve my problem as well.

franzhaas:
thanks @nicoddemus, aproaching pyoxidizer is surely a good idea,

https://github.com/indygreg/PyOxidizer/issues/492

src/_pytest/config/__init__.py
nicoddemus: > The dist is not in a file system, therefore the concept of path becomes questionable. Returning None, should indicate that meta data is missing, and a empty list should indicate empty meta data.

I understand, but that point should be raised with the `imporlib-metadata` maintainers then, to extend the interface to allow for `NotImplementedError`.

> Is there a API to disable the assertion rewriting, from within python that i did overlook? There is already a command line argument. To have this accessible would solve my problem as well.

How are you executing pytest exactly? `pytest.main()`?

franzhaas: Thanks for the hint!

I use numpy.test(), which calls pytest.main...

The solution with disabeling assert rewrites has the disadvantage that _all_ assert rewrites do not happen, also those who would work without issue.

I still believe that this change makes sense.

Alternatively I would propose to presented the erroro in a more helpfull way. 

I post the working and the failing call here, hopefully it helps someone.

>>> numpy.test()
NumPy version 1.23.0.dev0+308.gf82b5dacc
NumPy relaxed strides checking option: True
NumPy CPU features:  SSE SSE2 SSE3 SSSE3* SSE41* POPCNT* SSE42* AVX* F16C* FMA3* AVX2* AVX512F? AVX512CD? AVX512_SKX? AVX512_CLX? AVX512_CNL?
  File "numpy._pytesttester", line 204, in __call__
  File "_pytest.config", line 143, in main
  File "_pytest.config", line 318, in _prepareconfig
  File "pluggy._hooks", line 265, in __call__
  File "pluggy._manager", line 80, in _hookexec
  File "pluggy._callers", line 55, in _multicall
  File "_pytest.helpconfig", line 100, in pytest_cmdline_parse
  File "pluggy._result", line 60, in get_result
  File "pluggy._callers", line 39, in _multicall
  File "_pytest.config", line 1003, in pytest_cmdline_parse
  File "_pytest.config", line 1283, in parse
  File "_pytest.config", line 1167, in _preparse
  File "_pytest.config", line 1114, in _consider_importhook
  File "_pytest.config", line 1134, in _mark_plugins_for_rewrite
  File "_pytest.config", line 785, in _iter_rewritable_modules
  File "_pytest.config", line 1131, in <genexpr>
NotImplementedError
>>> numpy.test(extra_argv=["--assert=plain"]) 
NumPy version 1.23.0.dev0+308.gf82b5dacc
NumPy relaxed strides checking option: True
NumPy CPU features:  SSE SSE2 SSE3 SSSE3* SSE41* POPCNT* SSE42* AVX* F16C* FMA3* AVX2* AVX512F? AVX512CD? AVX512_SKX? AVX512_CLX? AVX512_CNL?
....

nicoddemus: > The solution with disabeling assert rewrites has the disadvantage that all assert rewrites do not happen, also those who would work without issue.

Indeed. Perhaps you can set the `PYTEST_DISABLE_PLUGIN_AUTOLOAD` environment variable to `1`? This won't rewrite assertions in plugins, but IIUC in your case you don't have plugin files laying around on disk anyway?

> I still believe that this change makes sense.

Sorry I have to disagree.

`Distribution.files` documentation is clear that it should return either a `list[PackagePath]` or `None`. Raising `NotImplementedError` is not part of the interface, so it should be fixed in `pyoxidizer` (I think it should return `None` to mean "this distribution doesn't have files on the file system).

Why is that important? Imagine if packages decided to implement the `files` property slightly differently, for example project `X` decides to raise `ValueError` instead. Is pytest supposed to change to accommodate for that too, in addition to `NotImplementedError`? That's why I mention that it should be fixed in `pyoxidizer`, or brought up to `importlib-metadata` maintainers to change the interface to allow for `NotImplementedError` to be raised (I'm not advocating for the latter, I think `None` is sufficient, but that would be a valid route if you feel strongly about it).

Sorry if I'm seeming difficult, but I don't think pytest should accommodate for implementations that are broken just because it is easy, as this can have a rippling effect on the ecosystem or hide bugs.

franzhaas: Ok, i can understand your argument against the first sugestion. But I dont see why this is an argument against more helpful output.

Something along the idea of the latest commit of this PR (this comit just ilustrates the idea).

The-Compiler:
FWIW I still don't think having a "helpful output" is a good idea here, for reasons similar to what @nicoddemus has said. I think for common mistakes in test code it's good to have those - but if some very common assumptions are broken in the ecosystem somewhere around pytest, there are infinite places in pytest's codebase where it would have to account for that.

nicoddemus:
> but if some very common assumptions are broken in the ecosystem somewhere around pytest, there are infinite places in pytest's codebase where it would have to account for that.

Exactly. We would have to litter `try/except Exception` around pretty much everywhere pytest interacts with the ecosystem to account for broken assumptions, which is not really feasible.

So I think we can close this then. Once again thanks @franzhaas for the report and the discussion here: even though  we are closing the PR, we definitely appreciate the contribution nonetheless. 👍

franzhaas:
Thanks for considering,

having the error trace and the discussion findable here, is already an improvement.

All the best
Franz

PR #9495, opened 2022-01-09
Add docs on pytest.warns(None) deprecation

olgarithms:
Closes https://github.com/pytest-dev/pytest/issues/9404.

I added more context on the deprecation doc, but I have been wondering whether it would be useful to have these example on [the main doc of `pytest.warns()`](https://docs.pytest.org/en/latest/reference/reference.html?highlight=warns#pytest-warns), as people might not reach the first one.

Thoughts?

### DIFF
--- a/doc/en/deprecations.rst
+++ b/doc/en/deprecations.rst
@@ -221,11 +221,32 @@ Using ``pytest.warns(None)``
 
 .. deprecated:: 7.0
 
-:func:`pytest.warns(None) <pytest.warns>` is now deprecated because many people used
-it to mean "this code does not emit warnings", but it actually had the effect of
-checking that the code emits at least one warning of any type - like ``pytest.warns()``
+:func:`pytest.warns(None) <pytest.warns>` is now deprecated because it was frequently misused.
+Its correct usage was checking that the code emits at least one warning of any type - like ``pytest.warns()``
 or ``pytest.warns(Warning)``.
 
+If you are looking to:
+
+-  ensure that **no** warnings are emitted, consider using
+
+.. code-block:: python
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+
+- suppress warnings, you could use
+
+.. code-block:: python
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+
+- ensure that **any** warning is emitted, please use
+
+.. code-block:: python
+
+    with pytest.warns():
+        pass
 
 The ``--strict`` command-line option
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/en/deprecations.rst
+++ b/doc/en/deprecations.rst
@@ -225,28 +225,7 @@ Using ``pytest.warns(None)``
 Its correct usage was checking that the code emits at least one warning of any type - like ``pytest.warns()``
 or ``pytest.warns(Warning)``.
 
-If you are looking to:
-
--  ensure that **no** warnings are emitted, consider using
-
-.. code-block:: python
-
-    with warnings.catch_warnings():
-        warnings.simplefilter("error")
-
-- suppress warnings, you could use
-
-.. code-block:: python
-
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
-
-- ensure that **any** warning is emitted, please use
-
-.. code-block:: python
-
-    with pytest.warns():
-        pass
+See https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests for use cases.
 
 The ``--strict`` command-line option
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/en/deprecations.rst
+++ b/doc/en/deprecations.rst
@@ -225,7 +225,7 @@ Using ``pytest.warns(None)``
 Its correct usage was checking that the code emits at least one warning of any type - like ``pytest.warns()``
 or ``pytest.warns(Warning)``.
 
-See :ref:`warns-use-cases` for examples.
+See :ref:`warns use cases` for examples.
 
 The ``--strict`` command-line option
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


doc/en/deprecations.rst
```diff
@@ -221,11 +221,32 @@ Using ``pytest.warns(None)``
...
-:func:`pytest.warns(None) <pytest.warns>` is now deprecated because many people used
-it to mean "this code does not emit warnings", but it actually had the effect of
-checking that the code emits at least one warning of any type - like ``pytest.warns()``
+:func:`pytest.warns(None) <pytest.warns>` is now deprecated because it was frequently misused.
+Its correct usage was checking that the code emits at least one warning of any type - like ``pytest.warns()``
 or ``pytest.warns(Warning)``.
 
+If you are looking to:
+
+-  ensure that **no** warnings are emitted, consider using
```
nicoddemus, line 230:
```suggestion
-  Ensure that **no** warnings are emitted, use:
```

doc/en/deprecations.rst
```diff
@@ -221,11 +221,32 @@ Using ``pytest.warns(None)``
...
+
+.. code-block:: python
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+
+- suppress warnings, you could use
```
nicoddemus, line 237:
```suggestion
- Suppress warnings, use:
```

doc/en/deprecations.rst
```diff
@@ -221,11 +221,32 @@ Using ``pytest.warns(None)``
...
+
+.. code-block:: python
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+
+- ensure that **any** warning is emitted, please use
```
nicoddemus, line 244:
```suggestion
- Ensure that **any** warning is emitted, use:
```

doc/en/deprecations.rst
```diff
@@ -221,11 +221,32 @@ Using ``pytest.warns(None)``
...
 .. deprecated:: 7.0
 
-:func:`pytest.warns(None) <pytest.warns>` is now deprecated because many people used
-it to mean "this code does not emit warnings", but it actually had the effect of
-checking that the code emits at least one warning of any type - like ``pytest.warns()``
+:func:`pytest.warns(None) <pytest.warns>` is now deprecated because it was frequently misused.
+Its correct usage was checking that the code emits at least one warning of any type - like ``pytest.warns()``
 or ``pytest.warns(Warning)``.
 
+If you are looking to:
```
nicoddemus, line 228:
```suggestion
Here are some use cases involving warnings that often come up in tests, and suggestions on how to deal with them:
```

nicoddemus, line 228: I like this section and the examples a lot, so much I feel they should be part of the normal documentation, as opposed to just a note on the deprecations page.

How about a new section in `capture-warnings.rst`, pretty much what we have here? Then we can put a link to that section in here.

src/_pytest/deprecated.py
```diff
@@ -89,7 +89,10 @@
 
 WARNS_NONE_ARG = PytestRemovedIn8Warning(
     "Passing None to catch any warning has been deprecated, pass no arguments instead:\n"
-    " Replace pytest.warns(None) by simply pytest.warns()."
+    "Replace pytest.warns(None) by simply pytest.warns().\n"
```
The-Compiler, line 92: I feel like we should remove the "pass no arguments" suggestion entirely - clearly this solution is only correct in a fraction of cases, and with that, people will probably not look at the docs to find the proper fix for their situation...

doc/en/how-to/capture-warnings.rst
```diff
@@ -300,6 +300,32 @@ filter at the end of the test, so no global state is leaked.
 
 .. _recwarn:
```
nicoddemus, line 301: This is a [cross-reference label](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations) referring to the `Recording warnings` section, so it should be kept close to it.

We should also create a new cross reference label here that can be used instead of the full URL in the changelog:

```suggestion
.. _warns-use-cases:
```

doc/en/deprecations.rst
```diff
@@ -221,11 +221,11 @@ Using ``pytest.warns(None)``
...
 .. deprecated:: 7.0
 
-:func:`pytest.warns(None) <pytest.warns>` is now deprecated because many people used
-it to mean "this code does not emit warnings", but it actually had the effect of
-checking that the code emits at least one warning of any type - like ``pytest.warns()``
+:func:`pytest.warns(None) <pytest.warns>` is now deprecated because it was frequently misused.
+Its correct usage was checking that the code emits at least one warning of any type - like ``pytest.warns()``
 or ``pytest.warns(Warning)``.
 
+See https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests for use cases.
```
nicoddemus, line 228:
```suggestion
See :ref:`warns-use-cases` for examples.
```

src/_pytest/deprecated.py
```diff
@@ -88,8 +88,10 @@
 )
 
 WARNS_NONE_ARG = PytestRemovedIn8Warning(
-    "Passing None to catch any warning has been deprecated, pass no arguments instead:\n"
-    " Replace pytest.warns(None) by simply pytest.warns()."
+    "Passing None has been deprecated.\n"
+    "See https://docs.pytest.org/en/latest/how-to/capture-warnings.html"
+    "#additional-use-cases-of-warnings-in-tests"
+    " for alternativesin common use cases."
```
nicoddemus, line 94:
```suggestion
    " for alternatives in common use cases."
```

testing/deprecated_test.py
```diff
@@ -138,8 +138,10 @@ def test_warns_none_is_deprecated():
     with pytest.warns(
         PytestDeprecationWarning,
         match=re.escape(
-            "Passing None to catch any warning has been deprecated, pass no arguments instead:\n "
-            "Replace pytest.warns(None) by simply pytest.warns()."
+            "Passing None has been deprecated.\n"
+            "See https://docs.pytest.org/en/latest/how-to/capture-warnings.html"
+            "#additional-use-cases-of-warnings-in-tests"
+            " for alternativesin common use cases."
```
nicoddemus, line 144:
```suggestion
            " for alternatives in common use cases."
```

nicoddemus (approved):
Thanks @olgarithms!

Zac-HD:
Thanks @olgarithms!  This is a really nice solution for a very long discussion thread 💖

olgarithms:
Thank you all for your guidance and making this far less intimidating than it was!

PR #9496, opened 2022-01-09
hookspec: document conftest behavior for all hooks

bluetech:
Have each hook explain how implementing it in conftests works. This is part of the functional specification of a hook.

Also a few other related small related doc commits.

### DIFF
--- a/src/_pytest/hookspec.py
+++ b/src/_pytest/hookspec.py
@@ -62,7 +62,7 @@ def pytest_addhooks(pluginmanager: "PytestPluginManager") -> None:
     Use in conftest plugins
     =======================
 
-    If a conftest file implements this hook, it will be called immediately
+    If a conftest plugin implements this hook, it will be called immediately
     when the conftest is registered.
     """
 
@@ -85,7 +85,7 @@ def pytest_plugin_registered(
     Use in conftest plugins
     =======================
 
-    If a conftest file implements this hook, it will be called immediately
+    If a conftest plugin implements this hook, it will be called immediately
     when the conftest is registered, once for each plugin registered thus far
     (including itself!), and for all plugins thereafter when they are
     registered.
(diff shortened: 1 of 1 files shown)


bluetech:
Completing this is a bit of an effort, so this is still WIP, but you can get the idea!

src/_pytest/hookspec.py
```diff
@@ -56,7 +56,10 @@ def pytest_addhooks(pluginmanager: "PytestPluginManager") -> None:
     """Called at plugin registration time to allow adding new hooks via a call to
     ``pluginmanager.add_hookspecs(module_or_class, prefix)``.
 
-    :param pytest.PytestPluginManager pluginmanager: The pytest plugin manager.
+    :param pluginmanager: The pytest plugin manager.
```
diazona, line 59: This may be a silly question, but will the documentation still display the type of `pluginmanager` (based on the type annotation)? Personally, I've found that type information extremely helpful to know what kind of object each hook parameter is and what I can do with it.

bluetech: Yes, it is still displayed; sphinx knows how to get this information from the type annotation. You can see a preview of this PR rendering here to verify: https://pytest--9496.org.readthedocs.build/en/9496/reference/reference.html#bootstrapping-hooks

diazona: Perfect :+1: Thanks!

RonnyPfannschmidt (commented):
I like where this is going, just lovely

src/_pytest/hookspec.py
```diff
@@ -56,7 +56,10 @@ def pytest_addhooks(pluginmanager: "PytestPluginManager") -> None:
...
+
+    If a conftest file implements this hook, it will be called immediately
```
nicoddemus, line 61: Really like the new sections!

How about making them more prominent by adding a quick title, something like:

```suggestion
    **Note on conftests**
    
    If a conftest file implements this hook, it will be called immediately
```

This makes them standout and something to look for when browsing the hook documentation.

nicoddemus (approved):
This is a great addition!

Left minor suggestions about the term "initial conftests", not sure if there is a better way to clarify that.

src/_pytest/hookspec.py
```diff
@@ -174,12 +186,34 @@ def pytest_load_initial_conftests(
...
+    Stops at first non-None result, see :ref:`firstresult`.
+
+    :param config: The pytest config object.
+    :returns: The exit code.
+
+    Use in conftest plugins
+    =======================
+
+    This hook is only called for conftest files situated at the tests root
+    directory. It is not called for other conftest files.
```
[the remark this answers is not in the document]
nicoddemus: Oh right.

Sure thing, no problem.

> and their test* subdirs...

I always forget this one.

bluetech:
I finished annotating all of the hooks (separate commit to make reviewing the changes easier). I think it's good to merge now.

PR #9522, opened 2022-01-18
Add Test To Verify Rewritten *.pyc Files Are Readable

### LINKED ISSUE #9521: Add test coverage to pyc rewrite code
No current coverage for rewrite->write->read (that I can tell).

PR incoming (this bug is to reserve a changelog number).

holmanb:
Additional test verifies that rewrite->compile->write->read results in a valid pyc file.

closes #9521

### DIFF
--- a/changelog/9521.improvement.rst
+++ b/changelog/9521.improvement.rst
@@ -1,0 +1,1 @@
+Add test coverage to assertion rewrite path.
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -1123,6 +1123,28 @@ class TestAssertionRewriteHookDetails:
 
         assert _read_pyc(source, pyc) is None  # no error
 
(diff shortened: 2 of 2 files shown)


nicoddemus (approved):
Thanks!

Out of curiosity, why do you say it was not covered before?

holmanb:
> Out of curiosity, why do you say it was not covered before?

I suppose that was misleading wording, apologies.  I see that tests exist for `_read_pyc()`, `_write_pyc()`, and `_rewrite_test()` (which is called in `exec_module()`). What I don't see are any tests that verify that a pyc written with `_write_pyc()` is actually valid and readable with `_read_pyc()`. This "covers" that case. If this is already tested in some other way that I haven't considered I can drop this PR.

nicoddemus:
No that's fine, thanks for the clarification! I think such a test is valuable to have. 👍

nicoddemus:
Backport: https://github.com/pytest-dev/pytest/pull/9530

PR #9532, opened 2022-01-21
config: avoid stat storm in _getconftestmodules

### LINKED ISSUE #9478: [prerelease] Performance regression?
[According to nikolaizujev on Twitter](https://twitter.com/nikolaizujev/status/1468341977724997633):

> Tested on several internal projects, noticed almost 2x slowdown compared to 6.2.5 on py3.8

Unfortunately no reproducer, but I got `.prof` dumps from them: [Dropbox - pytest - Simplify your life](https://www.dropbox.com/sh/z6t42vphcbodl17/AAC6P05hj_6DW4NGKCsh3WJna?dl=0)

Opening this to remind myself to have a look at those before the final release - but ideally we'd have a proper reproducer so we can e.g. bisect this. Does anyone have any guess what could be the culprit? I only see a very small slowdown for my own project.

@hroncok Is there perhaps a way to find out which packages in Fedora took significantly longer to run tests with pytest 7 compared to 6? This would mainly be for runs which *still pass*, because if there are failures, all bets regarding the runtime are off anyways.

bluetech:
Fix #9478.

I'll also say Closes #9125 since it shouldn't be that slow anymore (also with next version of pluggy). I ran a profile on pandas collection with this PR and `gethookproxy` (which #9125 caches) comes out at ~1.5s out of ~30s.

I've looked at various ways to speed this up, but they were either too ugly/intrusive or not as effective. In the end, the conftest data structure `_dirpath2confmods`  is pretty efficient for what it does, and the API with the `is_file`/`parent` is mostly unavoidable given that. So a cache it is.

Note that this cache is safe (assuming a path does not change from file <-> dir during the run, which is already assumed), unlike the previous caches which worked on the entire `_getconftestmodules` function.

### DIFF
--- a/src/_pytest/config/__init__.py
+++ b/src/_pytest/config/__init__.py
@@ -330,6 +330,14 @@ def _prepareconfig(
         raise
 
 
+def _get_directory(path: Path) -> Path:
+    """Get the directory of a path - itself if already a directory."""
+    if path.is_file():
+        return path.parent
+    else:
+        return path
+
+
 @final
 class PytestPluginManager(PluginManager):
     """A :py:class:`pluggy.PluginManager <pluggy.PluginManager>` with
@@ -356,6 +364,11 @@ class PytestPluginManager(PluginManager):
         self._confcutdir: Optional[Path] = None
         # If set, conftest loading is skipped.
         self._noconftest = False
+
+        # _getconftestmodules()'s call to _get_directory() causes a stat
+        # storm when it's called potentially thousands of times in a test
+        # session (#9478), often with the same path, so cache it.
+        self._get_directory = lru_cache(256)(_get_directory)
 
         self._duplicatepaths: Set[Path] = set()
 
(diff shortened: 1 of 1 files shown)


src/_pytest/config/__init__.py
```diff
@@ -357,6 +365,11 @@ def __init__(self) -> None:
         # If set, conftest loading is skipped.
         self._noconftest = False
 
+        # _getconftestmodules()'s call to _get_directory() causes a stat
+        # storm when it's called potentially thousands of times in a test
+        # session (#9478), often with the same path, so cache it.
+        self._get_directory = lru_cache(256)(_get_directory)
```
nicoddemus, line 371: Does it make sense to remove the bounds here, using `None` instead of 256?

nicoddemus (approved):
Great work!

The-Compiler:
Any opinions whether this should be in 7.0.0 (due to #9478) or 7.0.1? I'd like to avoid adding more to the pile, but given that this is part of solving a regression and seems simple enough I guess we can take it in.

nicoddemus:
I think it should go into 7.0.0, as it is simple and fixes a regression, like you commented. 👍

The-Compiler:
Looks like the backport failed due to conflicts: https://github.com/pytest-dev/pytest/runs/4938147553?check_suite_focus=true

src/_pytest/config/__init__.py
bluetech: I just remembered I forgot to answer this.

I think it's always good to have limits on lru caches to avoid them growing too large. We can increase it though if it's too low for someone.

PR #9537, opened 2022-01-23
With -vv, display the full skip/xfail reason instead of "..."

### LINKED ISSUE #9536: Skip reasons get truncated, even with -vv
If I say `-vv`, I want to see the skip reason. Instead I get this:

```
tests/test_commands.py::test_kick SKIPPED (hircd doesn't support KICK)
tests/test_commands.py::test_op_deop SKIPPED (hircd doesn't support ...)
tests/test_commands.py::test_error_response SKIPPED (hircd doesn't s...)
tests/test_messaging.py::test_private_messages SKIPPED (hircd doesn'...)
```

pytest==6.2.5

- [ ] minimal example if possible

Akuli:
Closes #9536


bluetech (approved):
Makes sense to me

eamanu (commented):
IMO you need to add this option to see it in `pytest --help`

Akuli:
`  -v, --verbose         increase verbosity.`

What else do you want?

eamanu:
`-vv` what's the meaning?

Akuli:
`-v` is already in the help, and it can be specified multiple times, like `-vv`. This isn't anything that this PR adds, and pytest already has other features that do something different when `-v` or `-vv` is passed.

What exactly do you want me to add to the help message?

eamanu:
> What exactly do you want me to add to the help message?
I'm just giving an opinion about that I'd like to know in `pytest --help` that if I use `-vv`  I will have more details. No more than. I'm not core maintainer, so you can just ignore me.

nicoddemus (approved):

nicoddemus:
Thanks!

PR #9546, opened 2022-01-25
 fixtures: document FixtureDef's attributes

bluetech:
Also a refactor commit to make some code flow clearer.

This conflicts with #9420 so not meant for merging right away.


nicoddemus (approved):

bluetech:
@nicoddemus Thanks for the review.

This is just some internal docs so I'm going to skip the backport.

nicoddemus:
Wouldn't be better to do it anyway to avoid any potential conflicts?

bluetech:
I personally prefer to deal with conflicts if/when they happen.

PR #9547, opened 2022-01-26
Refactor idmaker functions into class IdMaker

bluetech:
ran: This is a commit cherry-picked from #9420 by @haxtibal with some tweaks by me. It is meant to accommodate the other changes in #9420 but I think it's nice regardless so picking it out already to reduce the size of #9420.

---

This commit only refactors, it does not change or add functionality yet. Public
API is retained. Reason or refactoring:

User provided parameter IDs (e.g. Metafunc.parametrize(ids=...)) had so far
only been used to calculate a unique test ID for each test invocation. That
test ID was a joined string where each parameter contributed some partial ID.

We're soon going to reuse functionality to generate parameter keys for
reorder_items and FixtureDef cache. We will be interested in the partial
IDs, and only if they originate from explicit user information. Refactoring
makes logic and data accessible for reuse, and increases cohesion in general.

### DIFF
--- a/src/_pytest/python.py
+++ b/src/_pytest/python.py
@@ -930,6 +930,139 @@ def hasnew(obj: object) -> bool:
 
 
 @final
+@attr.s(frozen=True, auto_attribs=True, slots=True)
+class IdMaker:
+    """Make IDs for a parametrization."""
+
+    # The argnames of the parametrization.
+    argnames: Sequence[str]
+    # The ParameterSets of the parametrization.
+    parametersets: Sequence[ParameterSet]
+    # Optionally, a user-provided callable to make IDs for parameters in a
+    # ParameterSet.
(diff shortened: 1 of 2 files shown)


nicoddemus (approved):

nicoddemus:
Should we backport this to `7.0.x` to avoid conflicts in the future?

bluetech:
IMO, no. If there's ever a conflict that's too hard to resolve, we can backport it then. But the stable branch should be as stable as possible - there's always a risk of backporting bugs...

The-Compiler:
Agreed. I'd really like to avoid anything rather risky at this point, we've had way too much delay already. Maybe for 7.0.1 if we decide it's stable enough and we need it to make backporting easier, but I'd like to keep it away from 7.0.0 (also to make the diff between rc1 and final smaller)

PR #9558, opened 2022-01-27
doc: Add ellipsis to warning usecase list

The-Compiler:
I feel like this makes it clearer where the code which was inside 'pytest.warns' is supposed to go.

Follow-up for #9495 - great work there, @olgarithms, this is just something minor that only came to mind when I re-read it now.


bluetech (approved):

Zac-HD:
Nice, thabks for spotting and fixing this!

Issue #9540, opened 2022-01-24
adding a link to the Python logging page


FIOpwK:
#### What's the problem this feature will solve?
Finding information on logging levels may prove simple if link to:
Python docs: [https://docs.python.org/3/library/logging.html#logging-levels] 

was found on page:

 Pytest docs: [https://docs.pytest.org/en/6.2.x/logging.html#live-logs]

Perhaps include the link in the  `reference external` class.

nicoddemus:
Good idea! Would you like to contribute that?

itsyourOOP:
Is this issue open? I'd like to try my hand at it.

The-Compiler:
@itsyourOOP it looks like @FIOpwK has opened a PR for this in the meantime (#9568), but thanks for your interest!

Achintya-Chaware:
Is this issue still open?

Issue #3013, opened 2017-12-08
Proposal: disable "live logging" by default and write to terminal writer instead


nicoddemus:
The [Live Logs](https://docs.pytest.org/en/latest/logging.html?highlight=logging#live-logs) feature currently is enabled by default and requires the user to pass `-s` to see the logs.

I see two problems:

1. If the user is interested in the feature, having to pass `-s` feels clumsy because now all sort of outputs will come through, reducing the actual usefulness of the feature because it will be harder to see the actual logging messages.
2. To solve 1), I believe we will need to write to the TerminalWriter, which by consequence will show up for all users from that point on even if they are not interested in the feature because the default is `WARNING`.

My proposal: 

1. Create a new property, `live_logs`, defaulting to `False`.
1. Write the "live logs" to the TerminalWriter, probably in color according to the level, yellow for WARNING and red for CRITICAL. This means users won't need to use `-s` to see live logs anymore.

This feature was originally implemented in eisensheng/pytest-catchlog#33 but was never released to PyPI, so I assume "live logs" has seen little use in the wild. 

cc @thisch

thisch:
I would introduce a logcapture (or log-capture) property instead, similar to the existing capture property. In the course of changing the property(ies) of the logging plugin I would also rename the `log-cli-*` properties (maybe to `log-capture-*`).

If we use the TerminalWriter for outputting the logs, we could hopefully write a newline character before the first log message s.t. loglines do not appear on the same lines as the testname. This is how it looks when capturing is turned off:

```
...
collected 1 item                                                                                                                                                                                                              

test_logging_in_pytest.py::test test_logging_in_pytest.py   27 WARNING  Some warning
PASSED                                                                                                                                                                                  [100%]
```

nicoddemus:
> I would introduce a logcapture (or log-capture) property instead, similar to the existing capture property.

You mean a command-line option? I ask because the "capture" configuration that exists today is `--capture=X` on the command-line. If so I agree, I think we should add an option to disable logging capture completely as well (similar to `-s` or `--capture=no`).

Having said that, I still would like a separate option for "live logs": seems useful to be able to be able to capture logs and only having them appear on the captured output section if a test fails. 

How about:

* `--capture-log`
  * `yes` (default): capture plugin is active; captured logs are displayed next to `stderr` and `stdout` sections if a test fails.
  * `no`: capture plugin is not active

* `--live_logs`: 
  * `no` (default): nothing special.
[...]

PR #9568, opened 2022-01-27
Added logging-level link to logging level table

FIOpwK:
This PR adds a link to the log level table for quick look up of all logging options.

### DIFF
--- a/doc/en/logging.rst
+++ b/doc/en/logging.rst
@@ -270,3 +270,5 @@ file:
 
 More details about the discussion that lead to this changes can be read in
 issue `#3013 <https://github.com/pytest-dev/pytest/issues/3013>`_.
+
+Log level table can be read at `logging-level <https://docs.python.org/3/library/logging.html#logging-levels>`_
--- a/doc/en/logging.rst
+++ b/doc/en/logging.rst
@@ -183,7 +183,7 @@ logging records as they are emitted directly into the console.
 
 You can specify the logging level for which log records with equal or higher
 level are printed to the console by passing ``--log-cli-level``. This setting
-accepts the logging level names as seen in python's documentation or an integer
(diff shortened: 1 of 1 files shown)


FIOpwK:
Addresses Issue [9540](https://github.com/pytest-dev/pytest/issues/9540)

The-Compiler (changes requested):
Thanks for the contribution! However, I don't think the location at the end of the file makes much sense: The preceding

> More details about the discussion that lead to this changes can be read in
￼issue #3013

Is referring to the "Incompatible changes in pytest 3.4" heading, while this is clearly unrelated to that. I doubt anyone reading the text and wondering about logging levels would find it there.

Further above under "Live Logs" (which is I think you were referring to in the issue) there are two places saying *This setting accepts the logging level names as seen in python's documentation*. What do you think about turning "in python's documentation" into a link instead?

Finally, it would be good if you could retarget this to the `main` branch instead - while docs are (currently) served from 6.2.x, we always fix things in `main` and then backport the changes to maintenance branches.

FIOpwK:
#What do you think about turning "in python's documentation" into a link instead?

I must've overlooked that line. It does make more sense in that location. Thanks for pointing that out. Further, I didn't know opening a PR on main was how things were done in this project. Good to know.

FIOpwK:
@The-Compiler, 'doubt' is measure there, sure. If  not the `python's documentation` then `logging level` or `logging level names` ? I moved before this discussion was truly resolved. Checking myself in for it.

FIOpwK:
I say logging level to keep the theme of what the text is about. Yes, the link is specific to python's documentation and deserves to be included if mentioned I guess. The logging level text *is* what that documentation references.

The-Compiler:
Closing this in favor of #9575 (which is against the correct branch) - let's continue this there!

PR #9575, opened 2022-01-29
Linking text statement to mentioned  docs

FIOpwK:
refactor PR 9568 for issue 9540


The-Compiler (changes requested):
There is another such reference further down (for `--log-file-level`), I think it would be great to update both. Note that you can update the file again in your `patch-2` branch, without having to open a new PR.

PR #9596, opened 2022-02-03
releasing: Add template for major releases

The-Compiler:
With pytest 6.0.0, we still used a manual releasing workflow (at least if I
remember correctly), and apparently we never wrote a release announcement
template for major releases. Instead, the minor release template claimed that
the release would contain "breaking changes", which doesn't seem reasonable.
Thus, this adds a new major template based on the former minor template, and
adjusts the latter to only mention fixes and new features instead.

### DIFF
--- a/scripts/release.minor.rst
+++ b/scripts/release.minor.rst
@@ -4,7 +4,7 @@ pytest-{version}
 The pytest team is proud to announce the {version} release!
 
 This release contains new features, improvements, and bug fixes,
-A full list of changes is available in the changelog:
+the full list of changes is available in the changelog:
 
     https://docs.pytest.org/en/stable/changelog.html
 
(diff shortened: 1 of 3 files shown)


scripts/release.minor.rst
```diff
@@ -3,8 +3,8 @@ pytest-{version}
 
 The pytest team is proud to announce the {version} release!
 
-This release contains new features, improvements, bug fixes, and breaking changes, so users
-are encouraged to take a look at the CHANGELOG carefully:
+This release contains new features, improvements, and bug fixes,
+A full list of changes is available in the changelog:
```
nicoddemus, line 7:
```suggestion
the full list of changes is available in the changelog:
```

The capitalize `A` here is wrong because the previous line ends in a comma, so this should be either `a` (or `the`?),  or a new phrase.

nicoddemus (approved):

scripts/release.minor.rst
The-Compiler: Good catch, thanks!

PR #9603, opened 2022-02-04
doc: Streamline remote handling in releasing docs

The-Compiler:
The docs already assume an 'upstream' remote, so we can only fetch from there instead of fetching all remotes. We also don't need to hardcode the remote URL.


nicoddemus (approved):
Good catch. 👍

PR #9607, opened 2022-02-04
Mention basetemp in docs as a way to only keep last temp dir

szabgab:
Help people who might want to reduce the number of temporary directories kept. See #9599

### DIFF
--- a/doc/en/how-to/tmp_path.rst
+++ b/doc/en/how-to/tmp_path.rst
@@ -122,7 +122,7 @@ the system temporary directory.  The base name will be ``pytest-NUM`` where
 than 3 temporary directories will be removed.
 
 The number of entries currently cannot be changed, but using the ``--basetemp``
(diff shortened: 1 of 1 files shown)


doc/en/how-to/tmp_path.rst
```diff
@@ -121,6 +121,10 @@ the system temporary directory.  The base name will be ``pytest-NUM`` where
 ``NUM`` will be incremented with each test run.  Moreover, entries older
 than 3 temporary directories will be removed.
 
+The number of entries currently cannot be changed, but using the ``--basetemp``
+option will remove the directory before every run. Effectively meaning the tempfiles
```
nicoddemus, line 125:
```suggestion
option will remove the directory before every run, effectively meaning the temporary directories
```

nicoddemus (approved):
Thanks!

Issue #9608, opened 2022-02-04
No module importlib.readers breaks pytest==7.0.0 with python 3.9


kdelee:
My project is using python 3.9 and we hit this today when 7.0.0 released:
https://github.com/pytest-dev/pytest/blob/5eb4d6977cf3604070172af018e84dc2d4cbcdee/src/_pytest/assertion/rewrite.py#L279-L283
Because this import does not seem to be working/reference existing code
```
/var/lib/awx/venv/awx/lib64/python3.9/site-packages/_pytest/assertion/rewrite.py:283: in get_resource_reader
    from importlib.readers import FileReader
E   ModuleNotFoundError: No module named 'importlib.readers'
```
If you'd like to see where this is happening on our CI, I'm pretty sure you can see the logs for the `api-test` stage here https://github.com/ansible/awx/runs/5072172116?check_suite_focus=true

Downgrading pytest fixes the issue, so it seems to a problem exclusively in 7.0.0 right now.

- [x] a detailed description of the bug or problem you are having
- [x] output of `pip list` from the virtual environment you are using [piplist.txt](https://github.com/pytest-dev/pytest/files/8006185/piplist.txt)
- [x] pytest==7.0.0 
- [x] CentOS Stream release 9 
- [x] minimal example if possible

This minimal reproducer re-creates the error:
say we have `broken.py` with the following contents
```
from pip._internal.req import parse_requirements


def test_broken():
    assert True
```

then running `pytest broken.py` results in
```
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.9.10, pytest-7.0.0, pluggy-1.0.0
django: settings: awx.settings.development (from ini)
rootdir: /awx_devel, configfile: pytest.ini
plugins: mock-1.11.1, forked-1.4.0, django-4.5.2, xdist-1.34.0, cov-3.0.0, timeout-2.1.0
collected 0 items / 1 error                                                                                                                                                                  

=========================================================================================== ERRORS ===========================================================================================
_________________________________________________________________________________ ERROR collecting broken.py _________________________________________________________________________________
ImportError while importing test module '/awx_devel/broken.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
broken.py:1: in <module>
    from pip._internal.req import parse_requirements
[...]

PR #9173, opened 2021-10-07
Support the importlib.resources files API in rewritten files

gaborbernat:
Resolves https://github.com/pytest-dev/pytest/issues/9169

PR #9609, opened 2022-02-04
importlib.readers not valid until python 3.10

kdelee:
This exists https://github.com/python/cpython/blob/3.10/Lib/importlib/readers.py and FileReader is in there
This is a 404 https://github.com/python/cpython/blob/3.9/Lib/importlib/readers.py

This change needs to get backported to the 7.0.z branch(s) too
closes https://github.com/pytest-dev/pytest/issues/9608

I'm going to go ahead and open this PR, if you think we need a changelog for this LMK

### DIFF
--- a/changelog/9608.bugfix.rst
+++ b/changelog/9608.bugfix.rst
@@ -1,0 +1,1 @@
+Fix invalid importing of ``importlib.reader`` in Python 3.9.
--- a/changelog/9608.bugfix.rst
+++ b/changelog/9608.bugfix.rst
@@ -1,1 +1,1 @@
-Fix invalid importing of ``importlib.reader`` in Python 3.9.
+Fix invalid importing of ``importlib.readers`` in Python 3.9.
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -43,7 +43,7 @@ from _pytest.stash import StashKey
 
 if TYPE_CHECKING:
     from _pytest.assertion import AssertionState
-    
+
 
 assertstate_key = StashKey["AssertionState"]()
 
@@ -280,7 +280,7 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder, importlib.abc.Loader)
                 from importlib.readers import FileReader
             else:
                 from importlib.resources.readers import FileReader
-                
+
             return FileReader(types.SimpleNamespace(path=self._rewritten_names[name]))
 
 
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -43,11 +43,7 @@ from _pytest.stash import StashKey
 
 if TYPE_CHECKING:
     from _pytest.assertion import AssertionState
-
-if sys.version_info >= (3, 11):
-    from importlib.resources.readers import FileReader
-elif sys.version_info >= (3, 10):
-    from importlib.readers import FileReader
+    
 
 assertstate_key = StashKey["AssertionState"]()
 
@@ -280,6 +276,11 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder, importlib.abc.Loader)
     if sys.version_info >= (3, 10):
 
         def get_resource_reader(self, name: str) -> importlib.abc.TraversableResources:  # type: ignore
+            if sys.version_info < (3, 11):
+                from importlib.readers import FileReader
+            else:
+                from importlib.resources.readers import FileReader
+                
             return FileReader(types.SimpleNamespace(path=self._rewritten_names[name]))
 
 
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -44,6 +44,8 @@ from _pytest.stash import StashKey
 if TYPE_CHECKING:
     from _pytest.assertion import AssertionState
 
+if sys.version_info >= (3, 10):
+    from importlib.readers import FileReader
 
 assertstate_key = StashKey["AssertionState"]()
 
@@ -276,10 +278,7 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder, importlib.abc.Loader)
     if sys.version_info >= (3, 10):
 
         def get_resource_reader(self, name: str) -> importlib.abc.TraversableResources:  # type: ignore
-            from types import SimpleNamespace
-            from importlib.readers import FileReader
-
-            return FileReader(SimpleNamespace(path=self._rewritten_names[name]))
+            return FileReader(types.SimpleNamespace(path=self._rewritten_names[name]))
 
 
 def _write_pyc_fp(
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -44,7 +44,9 @@ from _pytest.stash import StashKey
 if TYPE_CHECKING:
     from _pytest.assertion import AssertionState
 
-if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 11):
+    from importlib.resources.readers import FileReader
+elif sys.version_info >= (3, 10):
     from importlib.readers import FileReader
 
 assertstate_key = StashKey["AssertionState"]()
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -273,7 +273,7 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder, importlib.abc.Loader)
         with open(pathname, "rb") as f:
             return f.read()
 
-    if sys.version_info >= (3, 9):
+    if sys.version_info >= (3, 10):
 
         def get_resource_reader(self, name: str) -> importlib.abc.TraversableResources:  # type: ignore
             from types import SimpleNamespace


bluetech:
Thanks, this LGTM. Since I have no idea about the underlying importlib stuff, I'll leave it to someone else to actually approve, maybe @gaborbernat who added it (#9173), or @asottile who knows about this stuff.

I do notice that latest python says this about `importlib.readers`:

https://github.com/python/cpython/blob/bf95ff91f2c1fc5a57190491f9ccdc63458b089e/Lib/importlib/readers.py#L2-L5

Maybe we can be forward-compatible with Python 3.11 already by importing from `importlib.resources.readers` instead on python>=3.11.

bluetech:
And if someone knows how to test this that would be great as well (though not necessary for this PR).

kdelee:
Not surgical, but this test has my reproducer in it
```
diff --git a/testing/test_assertrewrite.py b/testing/test_assertrewrite.py
index ae18b4be3..f424dba98 100644
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -1836,3 +1836,16 @@ class TestReprSizeVerbosity:
         self.create_test_file(pytester, DEFAULT_REPR_MAX_SIZE * 10)
         result = pytester.runpytest("-vv")
         result.stdout.no_fnmatch_line("*xxx...xxx*")
+
+    def test_pytest_get_resource_reader_works_with_pip(pytester: Pytester) -> None:
+        """Assert we dont regress on https://github.com/pytest-dev/pytest/issues/9608"""
+        pytester.makepyfile(
+            """
+            from pip._internal.req import parse_requirements
+
+            def test_import_works():
+                assert True
+        """
+        )
+        result = pytester.runpytest()
+        result.stdout.fnmatch_lines(["*1 passed*"])
```

src/_pytest/assertion/rewrite.py
```diff
@@ -273,7 +273,7 @@ def get_data(self, pathname: Union[str, bytes]) -> bytes:
         with open(pathname, "rb") as f:
             return f.read()
 
-    if sys.version_info >= (3, 9):
+    if sys.version_info >= (3, 10):
 
         def get_resource_reader(self, name: str) -> importlib.abc.TraversableResources:  # type: ignore
             from types import SimpleNamespace
```
nicoddemus, line 279: How about we move these two imports outside the function? This would at least caught the error in the correct Python versions.

```python
    if sys.version_info >= (3, 10):
        from types import SimpleNamespace
        from importlib.readers import FileReader

        def get_resource_reader(self, name: str) -> importlib.abc.TraversableResources:  # type: ignore
            return FileReader(SimpleNamespace(path=self._rewritten_names[name]))
```

nicoddemus: Hmm nevermind that, that local import does not work.

nicoddemus:
Took the liberty of moving the import to the global level, and adding a CHANGELOG so this is "ready to go", but still would like to hear from someone who knows "importlib" stuff better.

changelog/9608.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix invalid importing of ``importlib.reader`` in Python 3.9.
```
kdelee, line 1:
```suggestion
Fix invalid importing of ``importlib.readers`` in Python 3.9.
```

nicoddemus: Thanks!

src/_pytest/assertion/rewrite.py
```diff
@@ -44,6 +44,8 @@
 if TYPE_CHECKING:
     from _pytest.assertion import AssertionState
 
+if sys.version_info >= (3, 10):
+    from importlib.readers import FileReader
```
bluetech, line 48:
```suggestion
if sys.version_info >= (3, 11):
    from importlib.resources.readers import FileReader
elif sys.version_info >= (3, 10):
    from importlib.readers import FileReader
```

nicoddemus: Thanks!

The-Compiler:
I'm a bit confused by two things here:

- Why didn't the test added in #9173 by @gaborbernat catch this?
- Why does the same test not fail now on Python 3.9, given that the code added there is now disabled there, but the test is unaltered?

src/_pytest/assertion/rewrite.py
```diff
@@ -44,6 +44,10 @@
 if TYPE_CHECKING:
     from _pytest.assertion import AssertionState
 
+if sys.version_info >= (3, 11):
+    from importlib.resources.readers import FileReader
+elif sys.version_info >= (3, 10):
+    from importlib.readers import FileReader
```
asottile, line 50: this change isn't needed -- and should remain inline in the function since it's verion-specific

src/_pytest/assertion/rewrite.py
```diff
@@ -273,13 +277,10 @@ def get_data(self, pathname: Union[str, bytes]) -> bytes:
         with open(pathname, "rb") as f:
             return f.read()
 
-    if sys.version_info >= (3, 9):
+    if sys.version_info >= (3, 10):
```
asottile, line 280: I agree that this should be changed to `(3, 10)` since `FileReader` isn't needed there

asottile (changes requested):
the test in the original review passes in 3.9 because the implementation of importlib.resources doesn't attempt `get_resource_reader` and just utilizes the `.submodule_search_locations` of the module's spec

so the `(3, 9)` vs. `(3, 10)` difference is ~mostly harmless -- is there a place where you're seeing this break?

kdelee:
Not sure about the others reporting https://github.com/pytest-dev/pytest/issues/9608, but this specifically occurs for us in a test file that imports `from pip._internal.req import parse_requirements`.

```
from pip._internal.req import parse_requirements

def test_broken():
    assert True
```
In the full traceback shown in  https://github.com/pytest-dev/pytest/issues/9608 we see it hitting this in https://github.com/pypa/pip/blob/44018de50cafba25445a225c1a1986d6312e1ef3/src/pip/_vendor/certifi/core.py#L50 which calls this `path` function in importlib.resources which ends up calling https://github.com/python/cpython/blob/459e26f0987a12a19238baba422e13a8f7fcfca3/Lib/importlib/resources.py#L161 which then checks if there is a attribute `get_resource_reader` on the "spec.loader" https://github.com/python/cpython/blob/459e26f0987a12a19238baba422e13a8f7fcfca3/Lib/importlib/resources.py#L74 -- which it looks like `AssertionRewritingHook` is a subclass of `importlib.abc.Loader`. This class are defines `get_resource_loader` on on the `importlib.abc.Loader` subclass and way down in importlib goes ahead and tries to call that function: https://github.com/python/cpython/blob/459e26f0987a12a19238baba422e13a8f7fcfca3/Lib/importlib/resources.py#L76.
This `get_resource_loader` defined on `AssertionRewritingHook` then tries to import code that doesn't exist in python 3.9 and everything breaks.

Should we import `from pip._internal.req import parse_requirements`? I'm not sure, but in the end, this code gets called, and it tries to import something that doesn't exist in python 3.9.

asottile:
> Not sure about the others reporting #9608, but this specifically occurs for us in a test file that imports `from pip._internal.req import parse_requirements`.
> 
> ```
> from pip._internal.req import parse_requirements
> 
> def test_broken():
>     assert True
> ```
> 
> In the full traceback shown in #9608 we see it hitting this in https://github.com/pypa/pip/blob/44018de50cafba25445a225c1a1986d6312e1ef3/src/pip/_vendor/certifi/core.py#L50 which calls this `path` function in importlib.resources which ends up calling https://github.com/python/cpython/blob/459e26f0987a12a19238baba422e13a8f7fcfca3/Lib/importlib/resources.py#L161 which then checks if there is a attribute `get_resource_reader` on the "spec.loader" https://github.com/python/cpython/blob/459e26f0987a12a19238baba422e13a8f7fcfca3/Lib/importlib/resources.py#L74 -- which it looks like `AssertionRewritingHook` is a subclass of `importlib.abc.Loader`. This class are defines `get_resource_loader` on on the `importlib.abc.Loader` subclass and way down in importlib goes ahead and tries to call that function: https://github.com/python/cpython/blob/459e26f0987a12a19238baba422e13a8f7fcfca3/Lib/importlib/resources.py#L76. This `get_resource_loader` defined on `AssertionRewritingHook` then tries to import code that doesn't exist in python 3.9 and everything breaks.

ah yeah this was the whole 3.9.3 fiasco where the resources api was partially rewritten and reverted in a patch version but some of the apis use the readers already

I think if you take the test that was in the previous PR and use `open_binary` or any of the other resource apis that are not `files()` then it would demonstrate the failure -- that said, they wouldn't be supported pre-3.9 anyway due the missing reader helpers

so I guess to move forward, I would inline the conditional import and personally I wouldn't make it version-specific but that's up to you

there isn't really a way to test this "regression" because the feature itself doesn't exist in previous versions

kdelee:
@nicoddemus @bluetech so what would ya'll like to do. It sounds like @asottile would like us to revert the move of the import outside the function.

bluetech:
@kdelee Can you move the imports back into the function? Then I think we can merge.

kdelee:
@bluetech done, lmk if anything else is needed

asottile (approved):

kdelee:
:grimacing: well not sure what those tests are blowing up about
Tried running them locally and I'm not seeing similar results

bluetech:
The failures are not due to this PR, it suddenly started to happen, we're not sure why yet. We will rerun the tests once this is resolved.

pombredanne:
Thanks... :bow:  Do you mind to push a dot release and may be yank 7.0.0 on PyPI as this is making CI fails all over? That would be awesome!

nicoddemus:
https://github.com/pytest-dev/pytest/pull/9654 is up, if no other maintainer would like to include another fix, we can release it. (Doesn't make sense to yank 7.0.0 if we are releasing 7.0.1 right away).

pombredanne:
You rock :guitar:

src/_pytest/assertion/rewrite.py
```diff
@@ -273,13 +273,15 @@ def get_data(self, pathname: Union[str, bytes]) -> bytes:
...
 
         def get_resource_reader(self, name: str) -> importlib.abc.TraversableResources:  # type: ignore
```
matthiasmoke, line 278: Hey, I'm currently writing a code analysis tool for Python projects. It pointed me to this code part, which I find quite interesting. Hence, I would have some questions about it:

1. Where exactly is the `get_resource_reader()` function called? I can't find any references to it. 
2. Why is the function defined by a condition? Would it make sense to move the version check inside the function with an early return? (As there is already a version check within it) As the function is public, it would keep the interface of the `AssertionRewritingHook` independent to the Python version.

I think these questions may be adressed to @kdelee or @nicoddemus? I would be really happy to hear from you as this would give me some qualitative insights on the reports of my tool.

Kind regards,
Matthias

asottile: it's called by the python import machinery itself. this function isn't valid until 3.10 (it breaks 3.9)

matthiasmoke: I saw the Github issue on the FileReader import causing the break for 3.9

Before, the import statement in the function was without any version check. But having the check `sys.version_info >= (3, 10)` within the function with an early return shouldn't break anything as the import statement would never be reached for 3.9. 
Or am I overlooking anything why the function would be invalid otherwise?

asottile: there's backports of parts of importlib which would reach into this and if the function is defined try and call it (and then error) -- by not defining the function, those backports won't even attempt to call it and therefore won't crash

matthiasmoke: Ahh ok I understand. Thanky you very much for your explanation, this helps me a lot!

PR #9611, opened 2022-02-04
DOC: pytest.warns() fails on zero warning

pllim:
Follow up of #9495 . Turns out `pytest.warns()` is not a drop-in replacement for the way we use `pytest.warns(None)` (astropy/astropy#12823). "Any" in the doc got me to believe that it will pass even with zero warning but the test fails when there is no warning with this, so this is an attempt to improve a doc a bit more.

p.s. This does not need a change log and I don't think a one word change qualify my name on authors list.


nicoddemus (approved):

nicoddemus:
Thanks!

PR #9628, opened 2022-02-07
Fix count of selected tests on terminal collection summary

### LINKED ISSUE #9626: terminal collection summary wrong count if there is an error
Lets say I have a test module with content:

```
def test_one():
    pass
def test_two():
    pass
def test_three():
    pass
```

`pytest -k test_` would should show:
`collected 3 items / 1 deselected / 2 selected`

But if there is another test module that raises an exception when imported: i.e.:

```
5/0
def test_foo():
    pass
```

The output for same command would be:
`collected 3 items / 1 error / 1 deselected / 1 selected`

Note how no new item was collected from the module raising an exception,
but the "selected" count substracted 1 due to the error on collecting the module.

The expected output would be:
`collected 3 items / 1 error / 1 deselected / 2 selected`

pytest version 7.0.0

schettino72:
If there were errors or skipped modules on collection,
pytest would mistakenly subtract those from the selected count.

Close #9626

### DIFF
--- a/changelog/9626.bugfix.rst
+++ b/changelog/9626.bugfix.rst
@@ -1,0 +1,3 @@
+Fix count of selected tests on terminal collection summary.
+
+If there were errors or skipped modules on collection, pytest would mistakenly subtract those from the selected count.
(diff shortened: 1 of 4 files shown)


nicoddemus (approved):
Thanks!

PR #9633, opened 2022-02-07
Fix 'warns' type. 'expected_warning' can accept None as the value

beliaev-maksim:
as per documentation https://docs.pytest.org/en/6.2.x/warnings.html#recwarn `recwarn` can accept None as the input

However, types are specified in a wrong way which leads to `mypy` complaints, example is this pipe: https://github.com/getsentry/responses/runs/5068593004?check_suite_focus=true

Current PR fixes the issue

Please let me know if you would like me to add/remove something else in the PR

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -201,6 +201,7 @@ Luke Murphy
 Maciek Fijalkowski
 Maho
 Maik Figura
+Maksim Beliaev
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
@The-Compiler 
you are right, sorry for that, thanks!

PR #9638, opened 2022-02-07
work around test pollution caused by new setuptools mutating global logger level

### LINKED ISSUE #9629: test_junitxml.py::TestPython::test_failure_function broke on CI
See e.g. https://github.com/pytest-dev/pytest/runs/5086149463?check_suite_focus=true - happens in multiple unrelated PRs. No idea why off-hand...

asottile:
resolves #9629


nicoddemus (approved):
Great find!

PR #9640, opened 2022-02-07
pythonpath: make it compatible with pytest-pythonpath until pytest 8

### LINKED ISSUE #9636: Pytest 7.0.0 appears to break pytest-pythonpath python_paths in pytest.ini
### Discussed in https://github.com/pytest-dev/pytest/discussions/9635

<div type='discussions-op-text'>

<sup>Originally posted by **enewton** February  7, 2022</sup>
We have some utility helper code in a separate folder to our tests. Up until now we have this configured via pytest.ini like so:

```
[pytest]
python_paths = ./utils/
```

This definition appears to be getting ignored after upgrading to pytest 7.0.0, which then gives us 'module not found' errors on tests trying to import utils. Reading around, it looks like this functionality is now built in to pytest and so is pytest-pythonpath obsolete?

I can get our tests passing again by updating to:

```
[pytest]
pythonpath = ./utils/
```

So the main question is, is this change in behaviour expected? Has the old method been deliberately disabled or should I raise a bug?
</div>

bluetech:
Fix #9636.

### DIFF
--- a/changelog/9636.bugfix.rst
+++ b/changelog/9636.bugfix.rst
@@ -1,0 +1,2 @@
+Restore compatiblity to users of the pytest-pythonpath plugins by adding two ini options ``python_paths`` and ``site_dirs``.
+These options are immediately marked as deprecated. See :ref:`python_paths-site_dir-ini` for details.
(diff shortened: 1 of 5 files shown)


bluetech:
I will add tests for these, too lazy to do it now...

bluetech:
Replaced by #9651.

PR #9661, opened 2022-02-10
Fix diff output for data types where `-v` would show less information

### LINKED ISSUE #5192: Failed test output is less verbose with -v than not using -v

nicoddemus:
`_compare_eq_verbose` doesn't seem to useful in the end, with the examples we
have in the test suite.

Close #5192

cc @mdmintz

### DIFF
--- a/changelog/5192.improvement.rst
+++ b/changelog/5192.improvement.rst
@@ -1,0 +1,3 @@
+Fixed test output for some data types where ``-v`` would show less information.
+
+Also, when showing diffs for sequences, ``-q`` would produce full diffs instead of the expected diff.
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -223,8 +223,6 @@ def _compare_eq_any(left: Any, right: Any, verbose: int = 0) -> List[str]:
             explanation = _compare_eq_set(left, right, verbose)
         elif isdict(left) and isdict(right):
             explanation = _compare_eq_dict(left, right, verbose)
-        elif verbose > 0:
-            explanation = _compare_eq_verbose(left, right)
 
         if isiterable(left) and isiterable(right):
             expl = _compare_eq_iterable(left, right, verbose)
@@ -281,18 +279,6 @@ def _diff_text(left: str, right: str, verbose: int = 0) -> List[str]:
     return explanation
 
 
-def _compare_eq_verbose(left: Any, right: Any) -> List[str]:
-    keepends = True
-    left_lines = repr(left).splitlines(keepends)
-    right_lines = repr(right).splitlines(keepends)
-
-    explanation: List[str] = []
-    explanation += ["+" + line for line in left_lines]
-    explanation += ["-" + line for line in right_lines]
-
-    return explanation
-
-
 def _surrounding_parens_on_own_lines(lines: List[str]) -> None:
     """Move opening/closing parenthesis/bracket to own lines."""
     opening = lines[0][:1]
@@ -308,8 +294,8 @@ def _surrounding_parens_on_own_lines(lines: List[str]) -> None:
 def _compare_eq_iterable(
     left: Iterable[Any], right: Iterable[Any], verbose: int = 0
 ) -> List[str]:
-    if not verbose and not running_on_ci():
-        return ["Use -v to get the full diff"]
+    if verbose <= 0 and not running_on_ci():
+        return ["Use -v to get more diff"]
     # dynamic import to speedup pytest
     import difflib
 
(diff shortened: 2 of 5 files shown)


mdmintz:
@nicoddemus Yes, this is much better. This resolves the issue seen earlier. I'm currently testing various assertion combinations, and so far this change seems to have the desired outcome.

bluetech (approved):
TL;DR - LGTM :)

IIUC the situation is this:

- Assertion rewrite calls the reprcompare mechanism, with custom "expls" for the assertion (the rewrite special sauce)
- But before using these special expls, it also calls the `pytest_assertrepr_compare` hook to see if it wants to offer its own formatting (which is preferred). The hook only gets the op and values (no special stringification sauce).
- The built-in `pytest_assertrepr_compare` hookimpl in this case eventually reaches `_compare_eq_any`.

Without verbose, `_compare_eq_any` returns None for this, and the custom expls are used.
With verbose, before this fix, `_compare_eq_verbose` would be called, which ends up with

1. Less nice formatting, because no special sauce.
2. Less verbose than the expls, because no special sauce, only shows the actual values.

This PR removes `_compare_eq_verbose` so it falls back to the rewrite expls.

So I looked into why `_compare_eq_any` was added in the first place. AFAIU, it used to be that the rewrite formatting didn't take verbosity in account when truncating - it always truncated. So `_compare_eq_any` was added as a workaround. But the underlying problem has since been fixed by @nicoddemus in be8d63e33b199cafa2656d49ad7128cc32214004. So this is probably no longer needed.

nicoddemus:
Thanks @bluetech for looking into why `_compare_eq_verbose` existed in the first place. 😁 

Do you agree this is an improvement, or should we treat it as bug fix and backport to 7.0.x?

bluetech:
I think the compare-any is an improvement, I wouldn't backport it since it's not risk-free either.

The `-q` is a bugfix, can backport if you feel like it :)

We probably shouldn't wait too long with pytest 7.1 though. I will open an issue for a release schedule.

PR #9668, opened 2022-02-11
CI: Test latest 3.10 available

hugovk:
Test against `3.10` instead of specific `3.10.1` so newer versions are used when available; 3.10.2 is already available for macOS and Ubuntu.


nicoddemus (approved):

nicoddemus:
Thanks @hugovk!

PR #9682, opened 2022-02-13
Expand warnings output for ResourceWarning

### LINKED ISSUE #9644: Expand warnings output for ResourceWarning
It is difficult to debug ``ResourceWarning`` from the existing output.

In [Python3.6 a ``source`` field was added](https://bugs.python.org/issue26567) to the ``warnings.WarningMessage`` objects.

Since pytest now requires Python 3.7 I propose the following addition to the existing ``warning_record_to_str``:

```python
def warning_record_to_str(warning_message: warnings.WarningMessage) -> str:
    """Convert a warnings.WarningMessage to a string."""
    warn_msg = warning_message.message
    msg = warnings.formatwarning(
        str(warn_msg),
        warning_message.category,
        warning_message.filename,
        warning_message.lineno,
        warning_message.line,
    )
    tb = tracemalloc.get_object_traceback(warning_message.source)
    if tb is not None:
        msg = "%sSource of unclosed object:\n%s" % (msg, '\n'.join(tb.format()))
    elif issubclass(warning_message.category, ResourceWarning):
        # tracemalloc is inactive
        # TODO needs to be adjusted, see below in ticket
        msg = "%sGet a traceback for where the unclosed object was allocated by ..." % msg
    return msg
```

The output looks like this as an example:
```
  /.../devpi/devel/lib/python3.8/enum.py:315: ResourceWarning: unclosed <socket.socket [closed] fd=53, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6>
    return cls.__new__(cls, value)
  Source of unclosed object:
    File "/.../devpi/devel/lib/python3.8/site-packages/urllib3/connection.py", line 205
      conn = self._new_conn()
    File "/.../devpi/devel/lib/python3.8/site-packages/urllib3/connection.py", line 174
      conn = connection.create_connection(
    File "/.../devpi/devel/lib/python3.8/site-packages/urllib3/util/connection.py", line 77
      sock = socket.socket(af, socktype, proto)
```

This requires that ``tracemalloc`` is active. So the open question is, when should ``tracemalloc.start()`` be called and how to pass the size of the stack trace to it, as the default of ``1`` is often too small (I needed 20 to find the source of an unclosed requests Response). I propose three possible solutions:

1. Document running pytest like this: ``python -X tracemalloc=20 -m pytest ...``
2. Add a command line option to enable ``tracemalloc`` with the limit as parameter.
3. Enable tracing when warnings for ``ResourceWarning`` are set to ``once`` or ``always`` and always use a stack trace limit of ``3``.

nicoddemus:
Fix #9644

cc @fschulze

### DIFF
--- a/src/_pytest/warnings.py
+++ b/src/_pytest/warnings.py
@@ -92,9 +92,7 @@ def warning_record_to_str(warning_message: warnings.WarningMessage) -> str:
                 formatted_tb = "\n".join(tb.format())
                 # Use a leading new line to better separate the (large) output
                 # from the traceback to the previous warning text.
-                msg += (
-                    f"\nObject allocated at (most recent call first):\n{formatted_tb}"
-                )
+                msg += f"\nObject allocated at:\n{formatted_tb}"
             else:
                 # No need for a leading new line.
                 url = "https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings"
--- a/src/_pytest/warnings.py
+++ b/src/_pytest/warnings.py
@@ -81,6 +81,25 @@ def warning_record_to_str(warning_message: warnings.WarningMessage) -> str:
         warning_message.lineno,
         warning_message.line,
     )
+    if warning_message.source is not None:
+        try:
+            import tracemalloc
+        except ImportError:
+            pass
+        else:
+            tb = tracemalloc.get_object_traceback(warning_message.source)
+            if tb is not None:
+                formatted_tb = "\n".join(tb.format())
+                # Use a leading new line to better separate the (large) output
+                # from the traceback to the previous warning text.
+                msg += (
+                    f"\nObject allocated at (most recent call first):\n{formatted_tb}"
+                )
+            else:
+                # No need for a leading new line.
+                url = "https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings"
+                msg += "Enable tracemalloc to get traceback where the object was allocated.\n"
+                msg += f"See {url} for more info."
     return msg
 
 
--- a/testing/test_warnings.py
+++ b/testing/test_warnings.py
@@ -789,6 +789,10 @@ def test_resource_warning(pytester: Pytester, monkeypatch: pytest.MonkeyPatch) -
     except ImportError:
         has_tracemalloc = False
 
+    # Explicitly disable PYTHONTRACEMALLOC in case pytest's test suite is running
+    # with it enabled.
+    monkeypatch.delenv("PYTHONTRACEMALLOC", raising=False)
+
     pytester.makepyfile(
         """
         def open_file(p):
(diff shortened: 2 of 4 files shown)


src/_pytest/warnings.py
```diff
@@ -81,6 +81,25 @@ def warning_record_to_str(warning_message: warnings.WarningMessage) -> str:
...
+        except ImportError:
+            pass
+        else:
+            tb = tracemalloc.get_object_traceback(warning_message.source)
+            if tb is not None:
+                formatted_tb = "\n".join(tb.format())
+                # Use a leading new line to better separate the (large) output
+                # from the traceback to the previous warning text.
+                msg += (
+                    f"\nObject allocated at (most recent call first):\n{formatted_tb}"
```
fschulze, line 96: I'm not sure ``most recent call first`` is correct here, maybe just remove it?

src/_pytest/warnings.py
```diff
@@ -81,6 +81,25 @@ def warning_record_to_str(warning_message: warnings.WarningMessage) -> str:
         warning_message.lineno,
         warning_message.line,
     )
+    if warning_message.source is not None:
+        try:
+            import tracemalloc
```
fschulze, line 86: Are there platforms which don't have ``tracemalloc`` available?

fschulze (approved):
Nice! Thanks for the PR, you beat me to it :wink:

src/_pytest/warnings.py
fschulze: It is likely that PyPy doesn't support it at all due to differing memory management.

testing/test_warnings.py
```diff
@@ -774,3 +775,53 @@ def test_it():
...
+    # available, using `importorskip("tracemalloc")` for example,
+    # because we want to ensure the same code path does not break in those platforms.
+    try:
+        import tracemalloc  # noqa
+
+        has_tracemalloc = True
+    except ImportError:
+        has_tracemalloc = False
+
+    pytester.makepyfile(
```
bluetech, line 792: I think this test (tracemalloc not enabled) will fail if the person running the *pytest* test suite enables tracemalloc.

bluetech (approved):
Looks great to me! I left one comment.

testing/test_warnings.py
nicoddemus: Hmm why do you say that? And do you mean running in CPython?

I tried it here:

```
python -X tracemalloc=20 -m pytest testing\test_warnings.py
```

And it passes.

nicoddemus: Note the try/except is checking if tracemalloc is available at all, which is not on PyPy for example, not if tracemalloc is enabled or not.

bluetech: I guess `python -X tracemalloc=20` doesn't transfer, but an envvar does, so try this:

```
PYTHONTRACEMALLOC=100 pytest testing/test_warnings.py -k resource
```

nicoddemus: Ahh good catch. 😁 

Updated the test, thanks.

PR #9714, opened 2022-02-25
pytester now requests monkeypatch instead of creating its own instance

nicoddemus:
It is tempting to use `monkeypatch` to replace the other mechanisms in pytester which change global 
state: `CwdSnapshot`, `SysModulesSnapshot`, `SysPathsSnapshot`, however those are more delicate 
than they look at first glance so leaving those alone for now.

Close #9708

PR #9719, opened 2022-03-01
prevent import-time side-effects from setuptools bundled importlib-metadata

asottile:
this works around the pollution identified here by avoiding the import of `setuptools` entirely: https://github.com/pytest-dev/pytest/pull/9714#issuecomment-1051068152


nicoddemus (approved):

nicoddemus:
Thanks for working on this!

PR #9740, opened 2022-03-07
Improve first paragraph of gettings-started.rst

soumyas567:
In the first example, the point that the test function and .py file should be prefixed or suffixed with 'test' is missing. It would be better if it is mentioned in the first example itself. So Added an extra phrase emphasizing to name the file as test_sample.py

### DIFF
--- a/doc/en/getting-started.rst
+++ b/doc/en/getting-started.rst
@@ -35,7 +35,7 @@ Install ``pytest``
 Create your first test
 ----------------------------------------------------------
 
-Create a simple test function with just four lines of code:
+Create a Python file test_sample.py and write a simple test function with just four lines of code:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


doc/en/getting-started.rst
```diff
@@ -35,7 +35,7 @@ Install ``pytest``
 Create your first test
 ----------------------------------------------------------
 
-Create a simple test function with just four lines of code:
+Create a Python file test_sample.py and write a simple test function with just four lines of code:
```
nicoddemus, line 38:
```suggestion
Create a Python file named ``test_sample.py`` and write a simple test function with just four lines of code:
```

nicoddemus (approved):
Great suggestion, thanks!

nicoddemus:
Hmm just realized your commit was based on an old version of the docs, the latest version reads:

```rst
Create a new file called ``test_sample.py``, containing a function, and a test:
```

So closing this then, but thanks anyway for the PR! Appreciate it.

PR #9744, opened 2022-03-08
Update upcoming trainings

The-Compiler:
Also see https://twitter.com/the_compiler/status/1501247315809640464

Submitted but not accepted yet trainings commented out for now.

### DIFF
--- a/doc/en/index.rst
+++ b/doc/en/index.rst
@@ -6,7 +6,7 @@
    - `PyConIT <https://pycon.it/en/talk/pytest-simple-rapid-and-fun-testing-with-python>`__, June 3rd 2022 (4h), Florence, Italy
    - `Professional Testing with Python <https://python-academy.com/courses/python_course_testing.html>`_, via `Python Academy <https://www.python-academy.com/>`_, March 7th to 9th 2023 (3 day in-depth training), Remote and Leipzig, Germany
 
-   Also see `previous talks and blogposts <talks.html>`_.
+   Also see :doc:`previous talks and blogposts <talks>`.
 
 ..
    -  `Europython <https://ep2022.europython.eu/>`__, July 11th to 17th (3h), Dublin, Ireland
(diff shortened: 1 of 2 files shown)


The-Compiler:
Preview: https://pytest--9744.org.readthedocs.build/en/9744/

nicoddemus (approved):

doc/en/index.rst
```diff
@@ -1,11 +1,16 @@
...
-..
-    .. sidebar:: Next Open Trainings
+.. sidebar:: Next Open Trainings
+
+   - `PyConDE <https://2022.pycon.de/program/W93DBJ/>`__, April 11th 2022 (3h), Berlin, Germany
+   - `PyConIT <https://pycon.it/en/talk/pytest-simple-rapid-and-fun-testing-with-python>`__, June 3rd 2022 (4h), Florence, Italy
+   - `Professional Testing with Python <https://python-academy.com/courses/python_course_testing.html>`_, via `Python Academy <https://www.python-academy.com/>`_, March 7th to 9th 2023 (3 day in-depth training), Remote and Leipzig, Germany
 
-       - `Professional Testing with Python <https://www.python-academy.com/courses/specialtopics/python_course_testing.html>`_, via `Python Academy <https://www.python-academy.com/>`_, February 1st to 3rd, 2022, Leipzig (Germany) and remote.
+   Also see `previous talks and blogposts <talks.html>`_.
```
hoefling, line 9: Just a small nit: a crossref instead of a hardcoded link could be used here.
```suggestion
   Also see :doc:`previous talks and blogposts <talks>`.
```

The-Compiler:
I'd actually propose to remove the link to previous talks/trainings completely - what do you think, @nicoddemus @hoefling? I feel like we're not doing a very good job at keeping it updated, so I don't think it really deserves to be linked that prominently.

hoefling:
I wouldn't miss it TBH, but right now this is the only spot that makes the reader aware of the "Talks and tutorials" chapter existence.

Just a side note, this wouldn't be an issue anymore when/If the theme is switched to `furo` (which I would love to see). I see the TOC properly listing all chapters ([preview](https://pytest--8420.org.readthedocs.build/en/8420/)), so the talks are easily accessible from any page.

nicoddemus:
>  I feel like we're not doing a very good job at keeping it updated, so I don't think it really deserves to be linked that prominently.

Not sure, feels like showing them in the front page gives them the importance they deserve. If we bury this down into the docs somewhere else, I feel they will go unnoticed.

The-Compiler:
Ok, let's keep it around then. I updated it to a crossref. I also added another talk I did in 2021 while I was at it.

PR #9766, opened 2022-03-14
Typo correction

Rowin:
A word was missing in the docstring of the warns function.

As the change is trivial, I didn't run any test or add anything to the changelog.


bluetech (approved):
Thanks!

PR #9768, opened 2022-03-14
testing: fix tests when run under `-v` or `-vv`

bluetech:
Regressed in fac8f284cd6ffd44c7d401093c67799766473be1, didn't notice
since we don't run tests in CI with `-v`.

### DIFF
--- a/extra/setup-py.test/setup.py
+++ b/extra/setup-py.test/setup.py
@@ -1,4 +1,5 @@
 import sys
+
 from distutils.core import setup
 
 if __name__ == "__main__":
--- a/testing/test_assertrewrite.py
+++ b/testing/test_assertrewrite.py
@@ -204,16 +204,8 @@ class TestAssertionRewrite:
         def f4() -> None:
             assert sys == 42  # type: ignore[comparison-overlap]
 
-        verbose = request.config.getoption("verbose")
         msg = getmsg(f4, {"sys": sys})
-        if verbose > 0:
-            assert msg == (
-                "assert <module 'sys' (built-in)> == 42\n"
-                "  +<module 'sys' (built-in)>\n"
-                "  -42"
-            )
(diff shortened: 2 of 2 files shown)


bluetech:
@nicoddemus the tests *do* give me some pause. One of the tests looks a bit like this (modified a bit):

```py
class X1: pass
class X2: pass

cls1 = X1
cls2 = X2

def test_it():
    assert cls1 == cls2
```

before fac8f284cd6ffd44c7d401093c67799766473be the output with `pytest -vv` iss

```
___________________________________ test_it ___________________________________

    def test_it():
>       assert cls1 == cls2
E       AssertionError: assert <class 'x.X1'> == <class 'x.X2'>
E         +<class 'x.X1'>
E         -<class 'x.X2'>

x.py:8: AssertionError
=========================== short test summary info ===========================
FAILED x.py::test_it - AssertionError: assert <class 'x.X1'> == <class 'x.X2'>
```

After (current main), it is

```
___________________________________ test_it ___________________________________

    def test_it():
>       assert cls1 == cls2
E       assert cls1 == cls2

x.py:8: AssertionError
=========================== short test summary info ===========================
FAILED x.py::test_it - assert cls1 == cls2
```

Now it's not printed anywhere what the `cls1`/`cls2` runtime values are. This only happens for types which `_compare_eq_any` doesn't know how to handle.

nicoddemus:
> Now it's not printed anywhere what the cls1/cls2 runtime values are. This only happens for types which _compare_eq_any doesn't know how to handle.

Yeah that seems bad. While the `+/-` diff doesn't seem that useful, at the very least the `AssertionError: assert <class 'x.X1'> == <class 'x.X2'>` should be displayed.

nicoddemus:
Thanks for tackling this as well. 👍

PR #9788, opened 2022-03-18
Switch documentation theme to Furo

pradyunsg:
Follow up to #8420


nicoddemus:
Thanks a lot @pradyunsg for following up with this!

I have a few comments, enumerating them for easy reference: 

1. The `docs` [build failed](https://github.com/pytest-dev/pytest/runs/5604701761?check_suite_focus=true) because of some warnings... do you know if the theme can affect those @pradyunsg ?

2. In https://pytest--9788.org.readthedocs.build/en/9788/reference/reference.html, I see this warning:

image

3. The topics on the left side contain too much information initially, for example:

image

I would expect less topics, because the API reference, changelog, etc are all buried all the way down on the scroll... is that something we can adjust by reorganizing the files a bit?

pradyunsg:
1. Yea, I'm not sure what the fix for that is.
2. Yup, and that's mostly a matter of removing the `..  contents::` from those page sources.
3. Yes, we can. We'd need to introduce a page for that, so I'd probably add an "How To Guides" page, that has all the relevant how to pages "under" it. (see the sidebar in https://pradyunsg.me/furo and the reference section or the development section)

pradyunsg:
I've excluded Sphinx 4.4, since it added an additional check for extlinks that... I don't want to deal with cleaning up to be honest.

And, yes, I'll split this into a bunch of sensible commits. I just wanted to whip something up quickly to make sure I don't drop the ball on this immediately after picking it back up. :)

pradyunsg:
Closing in favour of #12326.

PR #9793, opened 2022-03-19
ci: don't require `build` check again in `deploy` job

bluetech:
According to our release flow, a commit is only tagged after it had
already passed CI, so this additional check just adds one more annoying
wait.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,9 +6,6 @@ on:
       - main
       - "[0-9]+.[0-9]+.x"
       - "test-me-*"
-    tags:
-      - "[0-9]+.[0-9]+.[0-9]+"
-      - "[0-9]+.[0-9]+.[0-9]+rc[0-9]+"
 
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


bluetech:
In addition to this, I'd also like to enable [tag protection rule](https://github.blog/changelog/2022-03-09-tag-protection-rules/) in this repo (for all tags) - WDYT?

nicoddemus:
> In addition to this, I'd also like to enable [tag protection rule](https://github.blog/changelog/2022-03-09-tag-protection-rules/) in this repo (for all tags) - WDYT?

Definitely!


Regarding to this PR, I have changed the configuration in some other repos (https://github.com/pytest-dev/pytest-mock, https://github.com/pytest-dev/pytest-cpp) to have separated workflow for testing/deploy:

https://github.com/pytest-dev/pytest-mock/tree/main/.github/workflows

This way when tags are pushed, we don't even waste resources running the test suite again. With tag protection rules in place, this makes it even safer as only maintainers/admins can push tags (reducing the chance of a random contributor pushing a tag by mistake).

bluetech:
Makes sense - updated to move the deploy to a separate workflow, and removed running tests on tag.

nicoddemus:
Heh, should have noted that I was working on it (was about to push). 😁

bluetech:
:grin: Feel free to push if you did anything differently

.github/workflows/main.yml
```diff
@@ -6,9 +6,6 @@ on:
       - main
```
nicoddemus, line 6: Let's rename this file to `test.yml` for consistency, and rename the workflow name at the top to `name: test`.

Also, we refer to the workflow in README.rst, that needs to be updated as well:

```diff
diff --git a/README.rst b/README.rst
index f0fe35632..034034a40 100644
--- a/README.rst
+++ b/README.rst
@@ -20,8 +20,8 @@
     :target: https://codecov.io/gh/pytest-dev/pytest
     :alt: Code coverage Status

-.. image:: https://github.com/pytest-dev/pytest/workflows/main/badge.svg
-    :target: https://github.com/pytest-dev/pytest/actions?query=workflow%3Amain
+.. image:: https://github.com/pytest-dev/pytest/workflows/test/badge.svg
+    :target: https://github.com/pytest-dev/pytest/actions?query=workflow%3Atest
```

nicoddemus:
>  Feel free to push if you did anything differently

Yeah might as well, I was literally writing `git push` on the terminal when I noticed the comment popup in the browser window hehehe: https://github.com/pytest-dev/pytest/pull/9794

PR #9794, opened 2022-03-19
Split test/deploy workflows

nicoddemus:
This will trigger a deploy directly when a tag is pushed, avoiding
having to wait for another lenghty build.

This is safe since we only push release tags after the release PR has passed all its tests.

See comments in #9793.

Close #9793

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,56 @@
+name: deploy
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


bluetech (approved):

nicoddemus:
Went ahead and protected the tags we use here:

https://github.com/pytest-dev/pytest/settings/tag_protection

PR #9800, opened 2022-03-21
Fix CI for Python 3.11

Zac-HD:
Getting the current working directory has no side-effects, so these context managers were no-ops.
Discovered because `Path.__enter__` is deprecated in Python 3.11, for removal in 3.13.

See e.g. https://github.com/pytest-dev/pytest/runs/5621307685?check_suite_focus=true for previous failures.

### DIFF
--- a/testing/test_collection.py
+++ b/testing/test_collection.py
@@ -660,8 +660,7 @@ class Test_getinitialnodes:
         subdir = tmp_path.joinpath("subdir")
         x = ensure_file(subdir / "x.py")
         ensure_file(subdir / "__init__.py")
-        with subdir.cwd():
-            config = pytester.parseconfigure(x)
+        config = pytester.parseconfigure(x)
         col = pytester.getnode(config, x)
         assert col is not None
         assert col.name == "x.py"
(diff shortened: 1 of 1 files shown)


nicoddemus (approved):
Ahh right. That's a leftover when we ported over from `py.local` to `pathlib`.

Thanks!

nicoddemus:
Not sure, they don't seem to affect the test at all?

The-Compiler:
I haven't looked at the tests involved in detail - I'm wondering if they are really still testing what they should, or if they just happen to always pass now.

nicoddemus:
Hmm good point. But it is troubling that we remove the change to cwd and the test outcome is not affected...

One thing that might have happened is that at some point we changed `pytester` to always change the cwd to its `path`, which was not done automatically before. This might explain `test_collect_with_chdir_during_import` for example.

bluetech:
Oops, guess I missed this when reviewing 3bcd316f076b185bcc89c41a41345861f752aff7 (although I did catch this in one of the other PRs in that series. Anyway, good deprecation by Python there...).

The `pytester.path.cwd` are probably good to entirely remove since as @nicoddemus says they should be noops. The other two are presumably intentionally done as part of the test. We should change them to

```py
with monkeypatch.context() as mp:
    mp.chdir(...)
```

nicoddemus:
I will work on this then, will open a PR later. 👍

nicoddemus:
PR open: https://github.com/pytest-dev/pytest/pull/9804

PR #9823, opened 2022-03-27
Clarify error message in case no collectors are found for a file

tobiasdiez:
In case no collector is found for a given file (e.g. if you run `pytest something.unrecognized`), then the somewhat cryptic error message
```
ERROR: not found: something.unrecognized
(no name 'something.unrecognized' in any of [])
``` 
With this PR, a more meaningful error message is displayed.

Not sure if such an almost-trivial change warrants an a changelog and authors entry.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -325,6 +325,7 @@ Thomas Grainger
 Thomas Hisch
 Tim Hoffmann
 Tim Strazny
+Tobias Diez
 Tom Dalton
 Tom Viner
 Tomáš Gavenčiak
--- a/changelog/9823.improvement.rst
+++ b/changelog/9823.improvement.rst
@@ -1,0 +1,1 @@
+Improved error message that is shown when no collector is found for a given file.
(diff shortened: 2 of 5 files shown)


nicoddemus (changes requested):
Thanks for the PR!

> Not sure if such an almost-trivial change warrants an a changelog and authors entry.

I think so, improvements (however small) are worthy of note. 😁 

Plus we need a test for the change, specially to ensure it doesn't regress. 👍

tobiasdiez:
Thanks for the feedback. I've now added a changelog and authors entry, and fixed the tests.

nicoddemus (approved):
Looks great, thank you!

Zac-HD:
@tobiasdiez - you'll need to merge in `main` in order to fix the `jinja2` issue; then we should be ready to merge 🎉

tobiasdiez:
Thanks, that worked like a charm!

Zac-HD:
Thank *you* for the patch! Good error messages are enormously important to users, but so often neglected - it's a tricky form of "expert blindness" which makes contributions especially valuable 😍

Issue #8730, opened 2021-06-04
6.2.4: sphinx warnings


kloczek:
There are some sphinx documentation minor issues:
```console
+ /usr/bin/python3 setup.py build_sphinx -b man
running build_sphinx
Running Sphinx v4.0.2
making output directory... done
[autosummary] generating autosummary for: adopt.rst, announce/index.rst, announce/release-2.0.0.rst, announce/release-2.0.1.rst, announce/release-2.0.2.rst, announce/release-2.0.3.rst, announce/release-2.1.0.rst, announce/release-2.1.1.rst, announce/release-2.1.2.rst, announce/release-2.1.3.rst, ..., sponsor.rst, talks.rst, tidelift.rst, tmpdir.rst, unittest.rst, usage.rst, warnings.rst, writing_plugins.rst, xunit_setup.rst, yieldfixture.rst
loading intersphinx inventory from https://pluggy.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
building [mo]: all of 0 po files
building [man]: all source files
updating environment: [new config] 198 added, 0 changed, 0 removed
reading sources... [100%] yieldfixture
/usr/lib/python3.8/site-packages/_pytest/pytester.py:docstring of _pytest.pytester.Pytester.TimeoutExpired:1: WARNING: duplicate object description of _pytest.pytester.Pytester.TimeoutExpired, other instance in reference, use :noindex: for one of them
/usr/lib/python3.8/site-packages/_pytest/main.py:docstring of _pytest.main.Session:1: WARNING: duplicate object description of _pytest.main.Session, other instance in reference, use :noindex: for one of them
/usr/lib/python3.8/site-packages/_pytest/main.py:docstring of _pytest.main.Failed:1: WARNING: duplicate object description of _pytest.main.Failed, other instance in reference, use :noindex: for one of them
docstring of builtins.Interrupted:1: WARNING: duplicate object description of builtins.Interrupted, other instance in reference, use :noindex: for one of them
/usr/lib/python3.8/site-packages/_pytest/main.py:docstring of _pytest.main.Session:1: WARNING: duplicate object description of _pytest.main.Session, other instance in reference, use :noindex: for one of them
docstring of builtins.Interrupted:1: WARNING: duplicate object description of builtins.Interrupted, other instance in reference, use :noindex: for one of them
[...]

PR #9836, opened 2022-04-04
Unpin `sphinx` upper bound

mxr:
CI fails with `ImportError: cannot import name 'environmentfilter' from 'jinja2'` ([example](https://github.com/pytest-dev/pytest/runs/5726444207)).

Bumping Sphinx as described in https://github.com/sphinx-doc/sphinx/issues/10291

### DIFF
--- a/doc/en/requirements.txt
+++ b/doc/en/requirements.txt
@@ -2,6 +2,6 @@ pallets-sphinx-themes
 pluggy>=1.0
 pygments-pytest>=2.2.0
 sphinx-removed-in>=0.2.0
-sphinx>=3.1,<4
+sphinx>=3.1
 sphinxcontrib-trio
 sphinxcontrib-svg2pdfconverter


mxr:
The tox docs step fails with 

```
ERROR: InvocationError for command /home/runner/work/pytest/pytest/.tox/docs/bin/sphinx-build -W --keep-going -b html doc/en doc/en/_build/html -t changelog_towncrier_draft (exited with code 1)
```

I think this is from the below warning (since `-W` makes warnings into errors)

```
/home/runner/work/pytest/pytest/src/_pytest/pytester.py:docstring of _pytest.pytester.Pytester.TimeoutExpired:1: WARNING: duplicate object description of _pytest.pytester.Pytester.TimeoutExpired, other instance in reference/reference, use :noindex: for one of them
```

It seems there was some work to fix that in https://github.com/pytest-dev/pytest/issues/8730#issuecomment-860059789 but no change was merged

Beyond that, CI will still be red because of https://github.com/psf/black/issues/2964

The-Compiler:
This seems like a duplicate of #9248.

kshahaji166:
zxzxz

bluetech:
Indeed, we're handling this in #9248.

PR #9842, opened 2022-04-08
fix comparison of dataclasses with `InitVar`

### LINKED ISSUE #9820: Assertion diffs print-outs break on dataclasses with InitVar fields
When an assertion fails comparing dataclasses which has [`InitVar` fields](https://docs.python.org/3/library/dataclasses.html#init-only-variables), the diff printing function fails with an `AttributeError` as it attempts to access the `InitVar` field. `InitVar` fields are not actually attributes stored in the dataclass, and I believe pytest needs to inspect the type here as `dataclass` does to ensure it's not treated as a real attribute.

Complete example:

```python
from dataclasses import InitVar, dataclass

@dataclass
class Foo:
    init_only: InitVar[int]
    real_attr: int


def test_demonstrate():
    assert Foo(1, 2) == Foo(1, 3)
```

Gives this error:

```
___________________________________________ test_demonstrate ___________________________________________
def test_demonstrate():
>       assert Foo(1, 2) == Foo(1, 3)
E       AssertionError: assert Foo(real_attr=2) == Foo(real_attr=3)
E         (pytest_assertion plugin: representation of details failed: /home/ayoon/.cache/pypoetry/virtualenvs/neoscore-ptOHLkol-py3.10/lib/python3.10/site-packages/_pytest/assertion/util.py:454: AttributeError: 'Foo' object has no attribute 'init_only'.
```

I'm running on Python 3.10.2, and my reproduction environment's `pip list` gives:
```
(.venv) pytest_bug $ pip list
Package    Version
---------- -------
attrs      21.4.0
iniconfig  1.1.1
packaging  21.3
pip        21.3.1
pluggy     1.0.0
py         1.11.0
pyparsing  3.0.7
pytest     7.1.1
setuptools 60.9.3
tomli      2.0.1
wheel      0.37.1
```

asottile:
resolves #9820

### DIFF
--- a/src/_pytest/assertion/util.py
+++ b/src/_pytest/assertion/util.py
@@ -437,8 +437,10 @@ def _compare_eq_cls(left: Any, right: Any, verbose: int) -> List[str]:
     if not has_default_eq(left):
         return []
     if isdatacls(left):
-        all_fields = left.__dataclass_fields__
-        fields_to_check = [field for field, info in all_fields.items() if info.compare]
+        import dataclasses
+
+        all_fields = dataclasses.fields(left)
+        fields_to_check = [info.name for info in all_fields if info.compare]
     elif isattrs(left):
         all_fields = left.__attrs_attrs__
         fields_to_check = [field.name for field in all_fields if getattr(field, "eq")]
--- a/changelog/9820.bugfix.rst
+++ b/changelog/9820.bugfix.rst
@@ -1,0 +1,1 @@
+Fix comparison of  ``dataclasses`` with ``InitVar``.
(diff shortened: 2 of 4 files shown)


nicoddemus (approved):
Awesome, thanks!

src/_pytest/assertion/util.py
```diff
@@ -437,8 +437,10 @@ def _compare_eq_cls(left: Any, right: Any, verbose: int) -> List[str]:
     if not has_default_eq(left):
         return []
     if isdatacls(left):
-        all_fields = left.__dataclass_fields__
-        fields_to_check = [field for field, info in all_fields.items() if info.compare]
+        import dataclasses
+
+        all_fields = dataclasses.fields(left)
```
nicoddemus, line 442: 👍

PR #9845, opened 2022-04-09
[7.1.x] update pre-commit hooks

asottile:
should make things pass


asottile:
actually -- I'll just tack this on to the jina2 fix

PR #9854, opened 2022-04-12
Docs: link to easy issues in contributing guide

hugovk:
The [status: easy](https://github.com/pytest-dev/pytest/labels/status%3A%20easy) issues are a good place for beginners to start, let's include them in the contrib guide.

Those could fall under either "Fix bugs" or "Implement features", but they tend to be bugs so I put it in that section.


nicoddemus (approved):
Thanks!

PR #9863, opened 2022-04-16
Update goodpractices.rst

gresm:
Fix example package tree.


asottile (approved):
either of these directory names work -- since we standardized on `tests` above I think this is good

thanks!

PR #9869, opened 2022-04-20
Increase stacklevel to point at user's code

hugovk:
The deprecation is raised fairly deep in pytest's code, so let's bump `stacklevel` so it points to the user's code and not somewhere inside pytest.

# Before

Points to pytest:

```consoe
=============================================================================== warnings summary ================================================================================
../../../Users/huvankem/github/pytest/src/_pytest/nodes.py:264
  /Users/huvankem/github/pytest/src/_pytest/nodes.py:264: PytestRemovedIn8Warning: The (fspath: py.path.local) argument to SanitizerFile is deprecated. Please use the (path: pathlib.Path) argument instead.
  See https://docs.pytest.org/en/latest/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path
    return cls._create(parent=parent, **kw)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

# After

Points to user code, in this case, `html5lib-python` where it calls `return SanitizerFile.from_parent(parent, fspath=path)`:

```
=============================================================================== warnings summary ================================================================================
html5lib/tests/conftest.py:108
  /private/tmp/html5lib-python/html5lib/tests/conftest.py:108: PytestRemovedIn8Warning: The (fspath: py.path.local) argument to SanitizerFile is deprecated. Please use the (path: pathlib.Path) argument instead.
  See https://docs.pytest.org/en/latest/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path
    return SanitizerFile.from_parent(parent, fspath=path)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```


nicoddemus (approved):
Thanks!

PR #9875, opened 2022-04-21
Add colors to summary

### LINKED ISSUE #9873: Short test summary info should use colors
While https://github.com/pytest-dev/pytest/pull/5061 added nice solors for the single line end-of-summary report, it did not include colors for the summary of the run, which makes harder to read the lines and to distinguish between skips xfail and failed statuses.

Some coloring on the summary entries is needed, either one of:
1. first word colored only, such SKIPPED, XFAIL, FAILED
2. entire line being colored

#### Alternative Solutions

Optionally we could try to color the filename using dimmed text, to make it easier to spot the test name that failed. The idea is that we current display 6 different texts on a single line without any visual help to make it easier to read
```
<MARKER> <file-path>::<test_function[parameter]><optional message/error>
```

#### Additional context

Attention should be taken to not over-do-it, too much coloring would also make reading harder.

PS. Please do not propose use of emoji ;)

nicoddemus:
## [0] Original output, for comparison:

image

## [1] Coloring status words:

image

Note that here I also moved the **XFAIL reason** to the same line, similar to the summary we have for failures. I think this is more pleasing, and was implemented when we didn't have the failure summary on each line, and not updated as an oversight.

## [2] Test names in bold:

image

Fix #9873

---

I personally like [2], but posting [1] in case people prefer it.

---
The fact that SKIPS are formatted differently and also folded together bothers me a bit, opened https://github.com/pytest-dev/pytest/issues/9876 for discussion.

### DIFF
--- a/src/_pytest/terminal.py
+++ b/src/_pytest/terminal.py
@@ -35,6 +35,7 @@ from _pytest import nodes
 from _pytest import timing
 from _pytest._code import ExceptionInfo
 from _pytest._code.code import ExceptionRepr
+from _pytest._io import TerminalWriter
 from _pytest._io.wcwidth import wcswidth
 from _pytest.compat import final
 from _pytest.config import _PluggyPlugin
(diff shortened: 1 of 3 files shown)


nicoddemus:
cc @ssbarnea

ssbarnea:
@nicoddemus LOL! I did not expect anyone to react so quickly on my suggestion. I was waiting for some feedback, imagining that I will have to make it myself. Not a complaint.

I love it! And is it obvious from the photo how much easier is to read.

My vote is for first option, I personally dislike bold on terminal because it makes the shape of the letters harder to read, some terminals do not even support bold. Using default text color is good here.

If you use vscode, you might also be interested about https://github.com/microsoft/vscode/issues/147769 -- i was bit engaged with microsoft on making the terminal links easier to click as sometimes it fails to open the files.

ssbarnea (approved):

nicoddemus:
>  LOL! I did not expect anyone to react so quickly on my suggestion. I was waiting for some feedback, imagining that I will have to make it myself. Not a complaint.

Heh got some time to work on pytest and seemed like a nice improvement to tackle in a couple of hours.

> My vote is for first option, I personally dislike bold on terminal because it makes the shape of the letters harder to read

We can use another color for the test names too.

bluetech:
I think this is nice, :+1:.

BTW, here is the output of a JavaScript test runner called jest:

image

jest1

Two relevant things:

1. In some cases it uses green/yellow/red background instead of letter color.
2. It hyperlinks the test names to the file using `file://` URLs. This is a [feature](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) that a growing number of terminal emulators support. It adds the thin dashed underline.

src/_pytest/terminal.py
```diff
@@ -1249,9 +1263,14 @@ def _build_collect_only_summary_stats_line(
         return parts, main_color
 
 
-def _get_pos(config: Config, rep: BaseReport):
+def _get_node_id_with_markup(tw: TerminalWriter, config: Config, rep: BaseReport):
     nodeid = config.cwd_relative_nodeid(rep.nodeid)
-    return nodeid
+    path, *parts = nodeid.split("::")
+    if parts:
+        parts_markup = tw.markup("::".join(parts), white=True, bold=True)
```
bluetech, line 1270: Hmm I'm not sure but wouldn't `white=True` be a problem on white terminal backgrounds? Why is it needed?

bluetech (approved):

src/_pytest/terminal.py
nicoddemus: Right, this was probably left-over from testing, removed. 👍

nicoddemus:
> In some cases it uses green/yellow/red background instead of letter color.

Yeah we can probably use that more. 👍

nicoddemus:
@bluetech how is your opinion on the use of bold for test names? I kind like it, but @ssbarnea said it makes the text harder to read in some terminal configurations.

bluetech:
I somewhat like the bold, and jest uses it too. I'll leave it to you to decide :)

ssbarnea:
Some things need to hit the market. If it proves less than ideal, we will see others complaining. Anyway thanks for the improvements!

PR #9878, opened 2022-04-21
Fix wrong `log_file` docs

### LINKED ISSUE #7336: inaccurate documentation

Zac-HD:
If a relative path is passed for the `--log-file` CLI argument, *or* the `log_file` config option is set, it is always resolved relative to the current working directory - pytest doesn't have any special logic beyond that in the standard library `logging.FileHandler`.

Unfortunately this was incorrectly documented as being relative to the `pytest.ini` file (*not* rootdir), which kicked off a considerable chain of confusion.  This PR therefore fixes #7336, closes #7350, and closes #9700.

Any proposal to change the current internally-consistent behaviour should start from a fresh issue, and include a description of why the alternative semantics are more desirable, and how to manage the backwards-compatibility implications of moving so many users' log files.

### DIFF
--- a/doc/en/how-to/logging.rst
+++ b/doc/en/how-to/logging.rst
@@ -198,6 +198,9 @@ option names are:
 If you need to record the whole test suite logging calls to a file, you can pass
 ``--log-file=/path/to/log/file``. This log file is opened in write mode which
 means that it will be overwritten at each run tests session.
+Note that relative paths for the log-file location are always resolved relative
+to the current working directory, whether passed on the CLI or declared in a
+config file.
 
 You can also specify the logging level for the log file by passing
 ``--log-file-level``. This setting accepts the logging level names as seen in
--- a/doc/en/how-to/logging.rst
+++ b/doc/en/how-to/logging.rst
@@ -198,9 +198,8 @@ option names are:
 If you need to record the whole test suite logging calls to a file, you can pass
 ``--log-file=/path/to/log/file``. This log file is opened in write mode which
 means that it will be overwritten at each run tests session.
-Note that relative paths for the log-file location are always resolved relative
-to the current working directory, whether passed on the CLI or declared in a
-config file.
+Note that relative paths for the log-file location, whether passed on the CLI or declared in a
+config file, are always resolved relative to the current working directory.
 
 You can also specify the logging level for the log file by passing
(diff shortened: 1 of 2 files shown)


doc/en/how-to/logging.rst
```diff
@@ -198,6 +198,9 @@ option names are:
 If you need to record the whole test suite logging calls to a file, you can pass
 ``--log-file=/path/to/log/file``. This log file is opened in write mode which
 means that it will be overwritten at each run tests session.
+Note that relative paths for the log-file location are always resolved relative
+to the current working directory, whether passed on the CLI or declared in a
+config file.
```
nicoddemus, line 203:
```suggestion
Note that relative paths for the log-file location, whether passed on the CLI or declared in a
config file, are always resolved relative to the current working directory.
```

Reads better to me (but I'm not a native speaker so take the suggestion with a grain of salt).

nicoddemus (changes requested):
Thanks @Zac-HD! 

We should also update the description of `log_file` in `reference.rst`, it is incorrect in there as well.

Zac-HD:
> We should also update the description of `log_file` in `reference.rst`, it is incorrect in there as well.

This PR *does* update `reference.rst`?  If you mean the CLI help in `logging.py`, I thought that it was unnecessary because "relative to cwd" is the obvious interpretation for a CLI path argument, and the existing text doesn't mention anything else.  It's also described in the reference as identical to the `log_file` config, and *that* is explicitly relative to cwd.

nicoddemus:
Oh I could sworn it was updating only `logging.rst`, sorry my bad!

nicoddemus (approved):

PR #9883, opened 2022-04-23
Document the --code-highlight default

hugovk:
Recently had to look this up in the source code.

So include in help text it shows during `pytest --help` and on https://docs.pytest.org/en/7.1.x/reference/reference.html#command-line-flags

---

Whilst looking at the the help texts at https://docs.pytest.org/en/7.1.x/reference/reference.html#command-line-flags, there's quite a bit of variation in case and whether to use a full stop after a single or multiple sentence:

* Sentence case
* Sentence case.
* Sentence case. Sentence case.
* Sentence case. Sentence case
* lower case
* lower case.
* lower case. Sentence case
* lower case. Sentence case.

I can update them if you have a preference? 

I suggest: use sentence case, no full stop for single sentence. If there's more than one sentence, end each with a full stop.

### DIFF
--- a/src/_pytest/helpconfig.py
+++ b/src/_pytest/helpconfig.py
@@ -49,7 +49,7 @@ def pytest_addoption(parser: Parser) -> None:
         action="count",
         default=0,
         dest="version",
-        help="display pytest version and information about plugins. "
+        help="Display pytest version and information about plugins. "
         "When given twice, also display information about plugins.",
     )
     group._addoption(
@@ -57,7 +57,7 @@ def pytest_addoption(parser: Parser) -> None:
         "--help",
         action=HelpAction,
         dest="help",
-        help="show help message and configuration info",
+        help="Show help message and configuration info",
     )
     group._addoption(
         "-p",
@@ -65,7 +65,7 @@ def pytest_addoption(parser: Parser) -> None:
         dest="plugins",
         default=[],
         metavar="name",
-        help="early-load given plugin module name or entry point (multi-allowed).\n"
+        help="Early-load given plugin module name or entry point (multi-allowed).\n"
         "To avoid loading of plugins, use the `no:` prefix, e.g. "
         "`no:doctest`.",
     )
@@ -74,7 +74,7 @@ def pytest_addoption(parser: Parser) -> None:
         "--trace-config",
         action="store_true",
         default=False,
-        help="trace considerations of conftest.py files.",
+        help="Trace considerations of conftest.py files",
     )
     group.addoption(
         "--debug",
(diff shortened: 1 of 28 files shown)


nicoddemus:
> I suggest: use sentence case, no full stop for single sentence. If there's more than one sentence, end each with a full stop.

Just make sure I got it right, do you mean?

* `one sentence help text`
* `first line of help.\nsecond line of help.`

hugovk:
I was suggesting:

* `One sentence help text`
* `First sentence of help. Second sentence of help.`

And not inserting any `\n`, let the terminal and argument parser wrap as usual.

nicoddemus:
👍 from me 😁

hugovk:
Updated!

src/_pytest/helpconfig.py
```diff
@@ -83,16 +83,17 @@ def pytest_addoption(parser: Parser) -> None:
...
-        "Defaults to 'pytestdebug.log'.",
+        "Default: pytestdebug.log.",
     )
     group._addoption(
         "-o",
         "--override-ini",
         dest="override_ini",
         action="append",
-        help='override ini option with "option=value" style, e.g. `-o xfail_strict=True -o cache_dir=cache`.',
+        help='override ini option with "option=value" style, '
```
nicoddemus, line 95:
```suggestion
        help='Override ini option with "option=value" style, '
```

src/_pytest/logging.py
```diff
@@ -228,7 +228,7 @@ def add_option_ini(option, dest, default=None, type=None, **kwargs):
         default=None,
         metavar="LEVEL",
         help=(
-            "level of messages to catch/display.\n"
+            "Level of messages to catch/display.\n"
```
nicoddemus, line 231:
```suggestion
            "Level of messages to catch/display."
```

src/_pytest/mark/__init__.py
```diff
@@ -96,7 +96,7 @@ def pytest_addoption(parser: Parser) -> None:
         dest="markexpr",
         default="",
         metavar="MARKEXPR",
-        help="only run tests matching given mark expression.\n"
+        help="Only run tests matching given mark expression.\n"
```
nicoddemus, line 99:
```suggestion
        help="Only run tests matching given mark expression. "
```

nicoddemus: I understand we want to remove the `\n` and let the terminal/formatter handle new lines, correct? If not please disregard.

nicoddemus (changes requested):
Looks great, I found a few places that seem to have been missed!

Besides that, I think we should add a CHANGELOG entry, `9883.improvement.rst`:

```rst
Normalize the help description of all command-line options.
```

I figure it is possible that someone is parsing/searching that text using some tool, so it is good to just mention it in the changelog as a heads-up. 👍

src/_pytest/logging.py
```diff
@@ -228,7 +228,7 @@ def add_option_ini(option, dest, default=None, type=None, **kwargs):
...
             "Not set by default, so it depends on the root/parent log handler's"
```
hugovk, line 232:
```suggestion
            " Not set by default, so it depends on the root/parent log handler's"
```

src/_pytest/mark/__init__.py
hugovk: Yes. I had decided to _not_ add any new ones, and leave the existing ones in, but it's a good idea to remove those too, there's not many.

nicoddemus: I see, thanks!

nicoddemus (approved):
Thanks @hugovk!

PR #9891, opened 2022-04-25
Add remove_finalizer function

RoniKish:
This PR adds remove_finalizer function to pytest.

This will enable removing a finalizer once it was added.
This is usefull in sitautions where you want to add finalizers at start of fixture, but in some conditions remove that finalizer.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -285,6 +285,7 @@ Roberto Polli
 Roland Puntaier
 Romain Dorgueil
 Roman Bolshakov
+Roni Kishner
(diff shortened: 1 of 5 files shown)


RonnyPfannschmidt (changes requested):
Strictly no to this one, even the test examples demonstrate bad behaviour


If you need a disengaging finalizer, pass a better finalizer

nicoddemus:
Thanks @RoniKish for the implementation anyway, we appreciate the well written PR, even if in the end it was not a good fit. 👍

PR #9893, opened 2022-04-26
Let mkdir() and mkpydir() receive PathLike names

jparise:
These pytester utility methods were annotated to only receive `str`
names, but they naturally support os.PathLike values, as well.

This makes writing some pytester calls a little nicer, such as when
creating a directory based on a `.joinpath()` call. We previously needed
to cast that intermediate value to a `str`.


Zac-HD (approved):
Looks great - thanks @jparise, and sorry it took so long for someone to review!

PR #9901, opened 2022-04-28
RFC: Add Downstream Plugin Testing

sommersoft:
Addresses #7342

I dabbled on a solution for this, taking a few different approaches and settling on this one. This is a bit of a compromise with all the concerns brought up in the issue's discussion.

This utilizes Docker (compose) inside a GitHub Actions workflow to establish the necessary environment(s) to run plugin tests. Current environment has Python 3.7, 3.8, 3.9, 3.10, and pypy3.7 (all via pyenv); also includes image layers for PostgreSQL and mySQL. The base image can be found here: https://github.com/sommersoft/docker-pyenv-pytest

Workflow runs can be viewed here: https://github.com/sommersoft/pytest/actions?query=branch%3Adownstream_testing_2

There are some (current) constraints:
- Limited CI Collection:
  - Will only run plugin tests that are defined in a GitHub Actions workflow and utilize ``tox``.
  - Adding non-GHA/tox _shouldn't_ be too difficult. Just would need to add parsing of Travis/etc config, and establish a modular ``toxenv`` to use.
- Failing/Flaky Plugin Tests:
  - If a plugin's current tip is failing it's own tests, those failures will occur here as well.
  - One way to workaround this is to query a plugin's last CI run and compare the status/results. Easier with GHA, obviously; Travis/etc may prove more difficult.

I'll be adding in some review comments to cover some questions to be answered, if this approach is acceptable.


.github/workflows/downstream_testing.yml
```diff
@@ -0,0 +1,110 @@
...
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      # Only run downstream tests on PRs for releases
+      - "[0-9]+.[0-9]+.x"
+  push:
+    branches:
+      - downstream_testing_2
```
sommersoft, line 11: I took these guesses at when to run. The `on.push.branches` will of course have my branch removed.

testing/downstream_testing/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM sommersoft/pyenv-pytest:latest
```
sommersoft, line 1: Ownership of this should likely be transferred. If you'd rather not have to maintain Docker Hub, GitHub's container registry is probably an option. Again, base image is here: https://github.com/sommersoft/docker-pyenv-pytest

nicoddemus:
Hi @sommersoft,

Just passing by to let you know we appreciate this effort a lot, however it is a lot to take in so that's why I believe people are taking their time to review it (at least that's the case for me).

Appreciate the effort and patience!

Zac-HD:
We've added a simple version of this via tox: https://github.com/pytest-dev/pytest/blob/e7355d70771b5f34f78d7012f73bdc8eb79574d6/tox.ini#L178-L200

and given the complexity and maintainence burden of this more complex approach I'm going to close the PR.  We really do appreciate the offered contribution though, and encourage you to take a look at another open issue if you'd like!

PR #9904, opened 2022-04-30
Mention use of re_assert for strings in approx docs

### LINKED ISSUE #9861: Feature request: pytest.match for regex-base equality-checking a la pytest.approx
One thing I really love about `pytest.approx` is that it means that I can do all equality checking the same way, regardless of whether it is approximate or exact. That is, I can do stuff like this:
```
@pytest.mark.parametrize(
    ("attribute", "expected_value"),
    [
        ("foo", 1),
        ("bah", pytest.approx(3.0)),
    ]
)
def test_attributes(object_under_test):
    assert getattr(object_under_test, attribute) == expected_value
```
I don't need separate tests for strict equality checking and approximate equality checking.

By the same token it would be great to have a `pytest.match` method that similarly reduces string matching to an equality check. This would let me
```
@pytest.mark.parametrize(
    ("attribute", "expected_value"),
    [
        ("foo", 1),
        ("bah", pytest.approx(3.0)),
        ("url", pytest.match(url_regex))
    ]
)
def test_attributes(object_under_test):
    assert getattr(object_under_test, attribute) == expected_value
```

johnkangw:
Closes #9861

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -666,6 +666,11 @@ def approx(expected, rel=None, abs=None, nan_ok: bool = False) -> ApproxBase:
         specialised test helpers in :std:doc:`numpy:reference/routines.testing`
         if you need support for comparisons, NaNs, or ULP-based tolerances.
 
+        To match strings using regex, you can use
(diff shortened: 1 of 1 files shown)


nicoddemus (approved):

nicoddemus:
Thanks @johnkangw!

PR #9906, opened 2022-05-01
Re-export importlib.metadata from _pytest.compat

hauntsaninja:
I'm fixing a bug in mypy's `--no-implicit-reexport` logic in
https://github.com/python/mypy/pull/12704 and mypy-primer flagged this


bluetech (approved):
Thanks!

PR #9907, opened 2022-05-01
Add deprecations for tests written for `nose`

### LINKED ISSUE #9886: Deprecate and remove nose plugin
`nose` has long been in maintenance only mode, and we support it via our `nose.py` plugin.

The `nose.py` plugin however is not trivial to maintain, and it dips its toes in a few places through the code:

https://github.com/pytest-dev/pytest/blob/d949b3f7d98fa7641055251765d3d6d0b1ef719c/src/_pytest/python.py#L382-L389

https://github.com/pytest-dev/pytest/blob/d949b3f7d98fa7641055251765d3d6d0b1ef719c/src/_pytest/python.py#L534

https://github.com/pytest-dev/pytest/blob/d949b3f7d98fa7641055251765d3d6d0b1ef719c/src/_pytest/python.py#L869

https://github.com/pytest-dev/pytest/blob/d949b3f7d98fa7641055251765d3d6d0b1ef719c/src/_pytest/unittest.py#L353-L355

Also it can cause confusion such as in https://github.com/pytest-dev/pytest/issues/9549.

Perhaps it is time to deprecate `nose`, and eventually removing it? If there is a large test suite which still requires this, they can just pin pytest until the last version that still supports it.

It could become a third party plugin. but unless someone volunteers to do it, I don't think it should be pytest's core responsibility to maintain such a plugin.

---

How to deprecate it?

`nose.py` actually just handles calling the nose-specific `setup` and `teardown` methods:

https://github.com/pytest-dev/pytest/blob/d949b3f7d98fa7641055251765d3d6d0b1ef719c/src/_pytest/nose.py#L9-L27

We could issue a deprecation warning when one of those functions are called, which AFAIK is basically what our nose support entails.

symonk:
Adds deprecation warnings for tests written in `nose`, namely around `setup` and `teardown` aspects; as we support limited functionality.

I went with a `pytest 8.x` deprecation as I think that is the norm here.  I also notice a few things when reviewing docs locally:

~my version was showing `6.3.x-dev*`~ was my fault;
`tox -e docs` does not work in python `3.10+` due to an upstream issue in `sphinx/util/typing.py` we may need a version bump there? I will chase that up in a second

closes #9886 

Todo:
 - Implement a solution for `call_optional(...)`
 - Investigate the mention of `@with_setup` etc.


doc/en/deprecations.rst
```diff
@@ -18,6 +18,14 @@ Deprecated Features
 Below is a complete list of all pytest features which are considered deprecated. Using those features will issue
 :class:`~pytest.PytestWarning` or subclasses, which can be filtered using :ref:`standard warning filters <warnings>`.
 
+Support for tests written for nose
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 7.1.3
```
symonk, line 24: `7.2.0` here?

src/_pytest/nose.py
```diff
@@ -38,5 +41,6 @@ def call_optional(obj: object, name: str) -> bool:
         return False
     # If there are any problems allow the exception to raise rather than
     # silently ignoring it.
+    warnings.warn(NOSE_SUPPORT, stacklevel=2)
```
symonk, line 44: not sure if we can unwind this into a user-space to be more helpful

doc/en/deprecations.rst
```diff
@@ -18,6 +18,14 @@ Deprecated Features
 Below is a complete list of all pytest features which are considered deprecated. Using those features will issue
 :class:`~pytest.PytestWarning` or subclasses, which can be filtered using :ref:`standard warning filters <warnings>`.
 
+Support for tests written for nose
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 7.2.0
+
+Support for running tests written for nose is now deprecated.
```
nicoddemus, line 26:
```suggestion
Support for running tests written for `nose <https://nose.readthedocs.io/en/latest/>`__ is now deprecated.

`nose` has been in maintenance mode-only for years, and maintaining the plugin is not trivial as it spills through the code base (see :issue:`9886` for more details).
```

doc/en/how-to/nose.rst
```diff
@@ -3,7 +3,8 @@
 How to run tests written for nose
 =======================================
 
-``pytest`` has basic support for running tests written for nose_.
+``pytest`` has basic support for running tests written for nose_. This functionality has been
```
nicoddemus, line 6: Probable better to state the deprecation as a warning:

```suggestion
``pytest`` has basic support for running tests written for nose_. 

.. warning:: 
    This functionality has been deprecated and is likely to be removed in ``pytest 8.x``.
```

src/_pytest/nose.py
nicoddemus: Indeed, but I think we should at least display the method name (and module), as well as the test that triggered this, if possible.

nicoddemus (changes requested):
Thanks @symonk!

Left a few suggestions, please take a look. 👍

The-Compiler:
Does this only affect `@with_setup`, or also undecorated plain `setup` and `teardown` methods on test classes? The latter is technically part of our nose support, but as far as I'm aware, there are many projects using them with pytest-style tests without being aware that they are supposed to be nose specific...

nicoddemus:
> there are many projects using them with pytest-style tests without being aware that they are supposed to be nose specific...

Good point. In that case they will get a warning, and we should mention in the deprecation docs how to port it to use something pytest-native.

symonk:
Sorry folks, I will get this one wrapped up shortly

nicoddemus:
@symonk gentle ping. 😁

symonk:
> @symonk gentle ping. grin

will get a look, thanks!

hwaking:
邮件已收到！

src/_pytest/nose.py
symonk: any preferred way to handle this one @nicoddemus; we have the `item` in scope when called, could just pass the `nodeid` into a formattable warning? however some unit tests don't pass that obviously; I'm not sure how many edge cases there can be for this `nose` stuff;  I've everything else fixed up I think but as-is its something like:

```python
def call_optional(obj: object, name: str, nodeid: str) -> bool:
    method = getattr(obj, name, None)
    if method is None:
        return False
    is_fixture = getfixturemarker(method) is not None
    if is_fixture:
        return False
    if not callable(method):
        return False
    # If there are any problems allow the exception to raise rather than
    # silently ignoring it.
    warnings.warn(NOSE_SUPPORT.format(nodeid=nodeid, nose_method=name), stacklevel=2)
    method()
    return True
```
which I don't really like; ok to use `inspect` here to derive some of that info? or perhaps `__qualname__ / `__module__` would suffice on the existing function obj passed?

nicoddemus (approved):
Thanks @symonk!

Went ahead and fixed the tests and added information on how to port code in the deprecation docs. 👍

nicoddemus:
@The-Compiler would appreciate a review if possible. 👍

doc/en/deprecations.rst
```diff
@@ -18,6 +18,109 @@ Deprecated Features
...
 :class:`~pytest.PytestWarning` or subclasses, which can be filtered using :ref:`standard warning filters <warnings>`.
 
+Support for tests written for nose
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 7.2.0
+
+Support for running tests written for `nose <https://nose.readthedocs.io/en/latest/>`__ is now deprecated.
+
+`nose` has been in maintenance mode-only for years, and maintaining the plugin is not trivial as it spills
```
The-Compiler, line 28:
```suggestion
``nose`` has been in maintenance mode-only for years, and maintaining the plugin is not trivial as it spills
```

for consistency with the rest of the file?

doc/en/deprecations.rst
```diff
@@ -18,6 +18,109 @@ Deprecated Features
...
+
+        def test_bar(self):
+            ...
+
+
+@with_setup
+^^^^^^^^^^^
+
+Code using `@with_setup <with-setup-nose>`_ will also need to be ported to a supported
+pytest style:
```
The-Compiler, line 77: This makes it sound like the following snippet is the supported pytest style, maybe just use "such as this:" and move the rest of the sentence down to "One way to do it"?

testing/deprecated_test.py
```diff
@@ -231,3 +231,31 @@ def test_importing_instance_is_deprecated(pytester: Pytester) -> None:
         match=re.escape("The pytest.Instance collector type is deprecated"),
     ):
         from _pytest.python import Instance  # noqa: F401
+
+
+def test_nose_deprecated(pytester: Pytester) -> None:
```
The-Compiler, line 236: Should also have a test using bare `setup` and `teardown` methods?

The-Compiler (changes requested):
Seems fine implementation-wise other than a few doc nitpicks (feel free to ignore if you disagree), and perhaps a missing test.

I'm still somewhat lukewarm about dropping bare `setup` and `teardown` methods... Unfortunately, I see no easy way to find out how much usage of those there is, as [searching for it](https://sourcegraph.com/search?q=context:global+%22def+setup%28self%29:%22+path:test_.*%5C.py+repohasfile:conftest.py&patternType=standard&case=yes) will of course also pick up fixtures (only limited to repos containing `conftest.py` to try and avoid unittest.py testsuites).

Maybe we'll just need to do it and see how many people complain...

edit: The more I look at it, the more I seem to see genuine cases of this being used in numpy, celery, ceph, and similar high-profile projects...

nicoddemus:
> edit: The more I look at it, the more I seem to see genuine cases of this being used in numpy, celery, ceph, and similar high-profile projects...

I see, thanks for looking into it. 

I strongly believe we should deprecate `nose`, but given that people have been using `setup` and `teardown` thinking this was part of the pytest core feature set, we have a few options:

1. Go ahead with the deprecation: to be fair it is a simple find/replace, and is backward-compatible with very old pytest versions (ancient versions even).
2. Make support for `setup` and `teardown` pytest native, outside `nose`.

2) should be trivial to implement, we already support `setup_method` and `teardown_method` named-methods, so it should be just a matter of extending that to `setup` and `teardown` names.

Exirel:
> Implementing 2) should be trivial, we already support `setup_method` and `teardown_method` named-methods, so it should be just a matter of extending that to `setup` and `teardown` names.

Hi! Sopel dev here, and here is our story. For the next major release of our IRC Bot, we try to support Python 3.10 and the latest version of `pytest`. When working on that, we discovered an issue with our test suite: for some unknown reason at the time, `pytest` would suddenly think that our `setup` function was now a test hook function. That was very unfortunate for a moment.

We were very happy when we understood the problem, and the solution: we deactivated the nose plugin, and all our problems went away.

So, from our perspective, we are fine with the following options:

* remove the nose plugin from pytest
* keep the plugin, but make sure we can deactivate it
* officially support `setup`/`teardown` function, but making sure it is configurable (opt-out, opt-in, regex pattern, etc.)

Basically: we really need to make sure we can still run pytest without having to change our codebase, or our user's codebase, to accommodate for something that was not, as far as I know, officially documented (at least until now).

nicoddemus:
> officially support setup/teardown function, but making sure it is configurable (opt-out, opt-in, regex pattern, etc.)

Just to make it clear, my proposal is to support `setup`/`teardown` *methods*, not functions. IIUC that would not affect your test suite, right?

Exirel:
> to support `setup`/`teardown` _methods_, not functions.

Oh! You're right, I was worried about functions, not methods. In that case, and since pytest has a configuration for class naming convention, we would not have any problem with that. Thanks for the clarification. 👍

nicoddemus:
Thanks @Exirel for the feedback! 👍

nicoddemus:
@The-Compiler what do you think regarding https://github.com/pytest-dev/pytest/pull/9907#issuecomment-1248388970?

I'm leaning towards 1), mostly because we are showing a deprecation, and the fix is trivial by doing a simple find/replace.

testing/deprecated_test.py
nicoddemus: Glad you asked this, it was not emitting a warning for `setup` and `teardown` 😬

nicoddemus:
The warning message now includes the text `To remove this warning, rename it to `{method}_method(self)` for plain `setup`/`teardown` methods.

The-Compiler:
> @The-Compiler what do you think regarding [#9907 (comment)](https://github.com/pytest-dev/pytest/pull/9907#issuecomment-1248388970)?
> 
> I'm leaning towards 1), mostly because we are showing a deprecation, and the fix is trivial by doing a simple find/replace.

Seems fine to me, though I see one problem: Once the deprecation period is over, it turns into a silent failure, depending on what the user does in those methods.

nicoddemus:
> Seems fine to me, though I see one problem: Once the deprecation period is over, it turns into a silent failure, depending on what the user does in those methods.

Yeah I agree it is not perfect, unfortunately.

We should make a point to add a big note to the release notes when we drop the functionality, I think is the best we can do.

nicoddemus:
Btw @The-Compiler could you give this another pass over? Thanks!

.gitignore
```diff
@@ -56,3 +56,4 @@ pip-wheel-metadata/
 
 # pytest debug logs generated via --debug
 pytestdebug.log
+__pycache__/
```
The-Compiler, line 59: nitpick (feel free to ignore): This should maybe be in the big block above, not under "# pytest debug logs generated via --debug"

changelog/9886.deprecation.rst
```diff
@@ -0,0 +1 @@
+The functionality for running tests written for ``nose`` has been officially deprecated.
```
The-Compiler, line 1: This should be a bit more verbose: It should point out that this also includes `setup` and `teardown` methods (since users don't seem to be aware of that being a nose thing), and it should have a link to the detailed deprecation docs.

doc/en/deprecations.rst
```diff
@@ -18,6 +18,110 @@ Deprecated Features
 Below is a complete list of all pytest features which are considered deprecated. Using those features will issue
 :class:`~pytest.PytestWarning` or subclasses, which can be filtered using :ref:`standard warning filters <warnings>`.
 
+Support for tests written for nose
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
The-Compiler, line 22: nitpick:

```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The-Compiler (changes requested):
Looks great! Just a few small doc comments. Don't feel strongly about them, except the changelog one.

changelog/9886.deprecation.rst
nicoddemus: Indeed, done. 👍

nicoddemus:
Thanks @symonk and @The-Compiler!

Issue #9892, opened 2022-04-25
doctest-modules with NUMBER option ignores zero decimal precision


MatthewFlamm:
- [x] a detailed description of the bug or problem you are having
- [x] output of `pip list` from the virtual environment you are using
- [x] pytest and operating system versions
- [x] minimal example if possible

# Description
This was found in https://github.com/pyvista/pyvista/pull/2511.  Using NUMBER option causes zero-precision numbers to pass when the numbers are off by 1 by not by 2.  Also using NUMBER option in a module causes this behavior to leak out into other docstring tests that do not use NUMBER.  Not using NUMBER correctly fails the tests.

Thanks to @adeak who identified the off by 1 but not by 2 aspect of this.

Using commands `pytest --doctest-modules test_1.py` and `pytest --doctest-modules test_2.py`.

# MRE

```py
# test_1.py
def list_1():
    """
    Passes

    Examples
    --------
    >>> [0.]  # doctest: +NUMBER
    [1.]
    """

def list_2():
    """
    Passes

    Examples
    --------
    >>> [1.]  # doctest: +NUMBER
    [0.]
    """

def list_3():
    """
    Passes

    Examples
    --------
    >>> [0.0]  # doctest: +NUMBER
    [1.]
    """

def list_4():
    """
    Passes

    Examples
    --------
    >>> [1.0]  # doctest: +NUMBER
    [0.]
    """

def list_5():
    """
    Passes

    Examples
    --------
    >>> [1.]  # doctest: +NUMBER
    [2.]
    """

def list_6():
    """
    Fails

    Examples
    --------
    >>> [1.]
    [3.]
    """


def list_7():
    """
    Fails
    
    Examples
    --------
    >>> [1.1]  # doctest: +NUMBER
    [0.]
    """

def list_8():
    """
    Passes in this module

    Examples
    --------
    >>> [1.]  
    [0.]
    """
```

```py
# test_2.py
def list_8():
    """
    Fails in this module

    Examples
    --------
    >>> [1.]  
    [0.]
    """
```

# pip list
```txt
Package    Version
---------- -------
attrs      21.4.0
iniconfig  1.1.1
numpy      1.22.3
packaging  21.3
pip        22.0.4
pluggy     1.0.0
py         1.11.0
pyparsing  3.0.8
pytest     7.1.2
setuptools 41.2.0
tomli      2.0.1
```

# system version
CentOS 7.9

Edit: Added description of how I ran the tests.

[...]

PR #9914, opened 2022-05-03
Clarify precision when using NUMBER option in --doctest-modules

MatthewFlamm:
This PR addresses one issue in #9892: the original wording suggests that it compares to the level of precision.  The new text explicitly lays out how the numbers are compared using `pytest.approx`.


nicoddemus:
Thanks @MatthewFlamm!

nicoddemus (approved):

PR #5284, opened 2019-05-17
[WIP] Display full crash information in short test summaries on non-tty

blueyed:

PR #9933, opened 2022-05-09
Do Not Truncate Crash Messages In Short Test Summary on CI

### LINKED ISSUE #9920: Do no truncate summary messages on CI
Currently short test summary gets truncated to the terminal width, but when running on CI we can probably ditch the limit so they appear on the logs.

Follow up to https://github.com/pytest-dev/pytest/issues/5283.

sommersoft:
Fixes #9920

This skips crash message truncation, if a `CI` environment variable is set and truthy.

Should this be mentioned in the docs, maybe as a note? https://github.com/pytest-dev/pytest/blob/main/doc/en/how-to/output.rst#producing-a-detailed-summary-report

*Thanks to blueyed for some hints in #5284.*

### DIFF
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -1140,15 +1140,17 @@ class TestTerminalFunctional:
 
 
 @pytest.mark.parametrize(
-    "use_CI",
+    ("use_ci", "expected_message"),
     (
         (True, f"- AssertionError: {'this_failed'*100}"),
         (False, "- AssertionError: this_failedt..."),
     ),
     ids=("on CI", "not on CI"),
 )
-def test_fail_extra_reporting(pytester: Pytester, monkeypatch, use_CI) -> None:
-    if use_CI[0]:
+def test_fail_extra_reporting(
+    pytester: Pytester, monkeypatch, use_ci: bool, expected_message: str
+) -> None:
+    if use_ci:
         monkeypatch.setenv("CI", "true")
     else:
         monkeypatch.delenv("CI", raising=False)
@@ -1160,7 +1162,7 @@ def test_fail_extra_reporting(pytester: Pytester, monkeypatch, use_CI) -> None:
     result.stdout.fnmatch_lines(
         [
             "*test summary*",
-            f"FAILED test_fail_extra_reporting.py::test_this {use_CI[1]}",
+            f"FAILED test_fail_extra_reporting.py::test_this {expected_message}",
         ]
     )
 
--- a/testing/test_terminal.py
+++ b/testing/test_terminal.py
@@ -1139,7 +1139,19 @@ class TestTerminalFunctional:
         assert result.stdout.lines.count(expected) == 1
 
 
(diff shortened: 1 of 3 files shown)


testing/test_terminal.py
```diff
@@ -1139,7 +1139,19 @@ def test():
         assert result.stdout.lines.count(expected) == 1
 
 
-def test_fail_extra_reporting(pytester: Pytester, monkeypatch) -> None:
+@pytest.mark.parametrize(
+    "use_CI",
```
nicoddemus, line 1143: Let's use two variables instead:

```python
@pytest.mark.parametrize(
    "use_ci, expected_message",
    (
        (True, f"- AssertionError: {'this_failed'*100}"),
        (False, "- AssertionError: this_failedt..."),
    ),
    ids=("on CI", "not on CI"),
)
def test_fail_extra_reporting(pytester: Pytester, monkeypatch, use_ci: bool, expected_messsage: str) -> None:
    if use_ci:
        monkeypatch.setenv("CI", "true")
    else:
        monkeypatch.delenv("CI", raising=False)
    ...

            f"FAILED test_fail_extra_reporting.py::test_this {expected_message}",
```

nicoddemus (changes requested):
Thanks a lot @sommersoft!

Left a few suggestions, please take a look.

> Should this be mentioned in the docs, maybe as a note?

Not sure it is needed, but feel free to write one `.. note::` at the end of that section if you like. 👍

sommersoft:
Fixed the merge conflicts (hopefully).

> Feel free to add yourself to the `AUTHORS` file btw.

Maybe at some point. This was a tiny change, but I appreciate the sentiment.

nicoddemus:
Thanks again @sommersoft !

The-Compiler:
> Which version of pytest will this be in?

The next feature release, i.e., 7.2.

jainshilpi:
Which python version does it need? It seems it can work with 3.9. If I use python version 3.9 and install pytest version 7.2:
`python3.9 -m pip install --user pytest==7.2` , it says `this version does not exist`. What are the requirements to use this pytest version?

The-Compiler:
@jainshilpi There [has not been](https://pypi.org/project/pytest/#history) a 7.2 release yet, so that's correct - but not related to your Python version. See #10045.

(For the record, the supported Python versions are documented here: [Python version support](https://docs.pytest.org/en/7.1.x/backwards-compatibility.html#python-version-support). At the time of writing, 3.7+ is supported.)

nicoddemus:
> I'm new to pytest, but should ["CI" and "BUILD_NUMBER"](https://github.com/pytest-dev/pytest/blob/main/src/_pytest/assertion/util.py#L519-L522) be added to the ["Environment Variables" section ](https://docs.pytest.org/en/7.2.x/reference/reference.html#environment-variables)of the reference?

Good idea, created #10690 to track that. PRs are welcome! :grin:

PR #9938, opened 2022-05-10
Consistently add **Tutorial**: in front of how-to links in reference

timhoffm:
Up to now, some had the prefix, some didn't. I think it's good to have
this prefix to give more context, otherwise the links sometimes blend
into the description.

In one case, the link goes to an example, so I used `**Example**:` there.

Also, there was a mixture of having a period at the end or not. I normalized to not have a period because the displayed how to names are usually not full sentences.


nicoddemus:
This is great, thanks a lot @timhoffm!

PR #9940, opened 2022-05-10
Move documentation contents from reference.rst to docstrings

timhoffm:
It's better to have the documentation in one place, instead
of having some in the docstring and some additional
information added to the reference documentation in
`reference.rst`.


nicoddemus:
Hi @timhoffm,

I like the idea, however it has broken the docs and pre-commit jobs, could you please take a look? Thanks!

nicoddemus (approved):

PR #9951, opened 2022-05-12
doc: link to pytest-subtests

rcomer:
Today I learned of the existence of `pytest-subtests` by rummaging in your issue tracker.  Is it worth a link from the main docs?

Huge thanks for everything you do!


nicoddemus (approved):
Definitely, thanks a lot!

PR #9957, opened 2022-05-14
Remove docs on Python 2.7 and 3.4 Support

timhoffm:
This extensive explanation is not need anymore because they are now unsupported for a long time.

Instead, add a short section on Python version compatibility in the backward compatibility docs.


timhoffm:
I just saw, that the removed page was linked from some locations:

```
/home/runner/work/pytest/pytest/doc/en/changelog.rst:2621: WARNING: unknown document: py27-py34-deprecation
/home/runner/work/pytest/pytest/doc/en/changelog.rst:2845: WARNING: undefined label: what goes into 4.6.x releases
/home/runner/work/pytest/pytest/doc/en/changelog.rst:3017: WARNING: unknown document: py27-py34-deprecation
```

How do you want to handle this?
- Remove the links
- Change the links to a fixed url, e.g. https://docs.pytest.org/en/7.0.x/py27-py34-deprecation.html
- We want to keep the page for historical record in current docs. Only remove it from the top-level toc in the side bar.

doc/en/backwards-compatibility.rst
```diff
@@ -77,3 +77,18 @@ Deprecation Roadmap
...
+Python version support
+======================
+
+Released pytest versions support all Python versions that are actively maintained at the time of the release:
+
+==============  ==============
+pytest version  Python version
+==============  ==============
+7.1             3.7+
+6.2             3.6+
```
The-Compiler, line 91: 7.0 also still supports Python 3.6 - but I think it would make sens to list both 6.2 and 7.0 here, because the latter was a major release with a lot of pytest changes.

doc/en/backwards-compatibility.rst
```diff
@@ -77,3 +77,18 @@ Deprecation Roadmap
...
+5.0             3.5+
+3.3             2.7, 3.4+
```
The-Compiler, line 93: The [changelog](https://docs.pytest.org/en/7.1.x/changelog.html#pytest-4-6-0-2019-05-31) for 4.6 says:

> The 4.6.X series will be the last series to support Python 2 and Python 3.4.

so I think this should be 4.6?

The-Compiler (changes requested):
Thanks! Every time I saw that page I thought "oh, we should get rid of that", and then I forgot about it again... I left a couple of comments.

As for the changelog, I think it would be best to link to the historical version of the page - that way, the links still work, and the changelog is a historical document anyways.

doc/en/backwards-compatibility.rst
timhoffm: What I was trying to say is that the given pytest versions changed the dependencies. Implicitly everything in has the same dependencies as the next lower pytest number. But maybe that's too confusiong. I changed to ranges now , e.g. "6.2 - 7.0".

The-Compiler (approved):
Great work, thanks @timhoffm! :+1:

PR #9967, opened 2022-05-16
Remove "Python 2.7 and 3.4 Support" from globaltoc

timhoffm:
Follow up to #9957 - I missed that the globaltoc is hard-coded
and the page needs to be removed there as well.


nicoddemus (approved):
Thanks!

PR #9968, opened 2022-05-16
Reorder the reference guides in the docs

timhoffm:
This changes the order to:

- API Reference
- Fixtures reference
- Configuration
- Exit codes
- Plugin List

which is approximately sorted from general to specific and often used to less used.
Plugin List ist at the end because it points to further external resources.


nicoddemus (approved):

nicoddemus:
Thanks!

PR #9969, opened 2022-05-16
Declutter doc entry page

timhoffm:
- Remove the index page https://docs.pytest.org/en/7.1.x/genindex.html
  Such an index is reasonable in books but not so much in HTML pages.
  IMHO the integrated search is much more user-friendly and sufficient.

- Remove the PDF link from the entry page. The PDF is already linked
  in the sidebar. I anticipate that getting the PDF documentation is
  not a major concern for users.

  This is removed:  grafik
  
  because we still have
  grafik

- Remove the sections "Changelog" and "License" from the entry page.
  These section only contain links to the respective pages.
  The topics are purely technical. They should be easily accessible
  (which they still are through sidebar section "About the project"),
  but they don't need additional advertisement through a dedicated
  section on the entry page.

  Note: The entry page sections

  - Bugs/Requests
  - Support pytest
  - pytest for enterprise / Security

  are also duplicate or similar to the "About the project" sidebar.
  However, I left them in place for now since you might want to draw
  special attention to them.

before / after:

grafik grafik

### DIFF
--- a/doc/en/conf.py
+++ b/doc/en/conf.py
@@ -247,7 +247,7 @@ html_sidebars = {
 html_domain_indices = True
 
 # If false, no index is generated.
-html_use_index = True
+html_use_index = False
 
 # If true, the index is split into individual pages for each letter.
 # html_split_index = False
--- a/doc/en/_templates/globaltoc.html
+++ b/doc/en/_templates/globaltoc.html
@@ -30,5 +30,3 @@
 {%- endif %}
 
 <hr>
-<a href="{{ pathto('genindex') }}">Index</a>
-<hr>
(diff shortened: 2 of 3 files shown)


nicoddemus (approved):
Good changes, thanks!

doc/en/conf.py
```diff
@@ -247,7 +247,7 @@
 html_domain_indices = True
 
 # If false, no index is generated.
-html_use_index = True
+html_use_index = False
```
The-Compiler, line 250: Do we really want to disable it entirely? It seems useful to keep even if we don't link to it from the sidebar, no?

timhoffm: When would one want to use such an index? IMHO there are better options for all usecases:

- If you know the term you are looking for, it's simpler to type it into the search field.
- If you don't know the term and rather want to see what is there, topic-oriented pages like the references and how-tos are a better place to browse compared to a long alphabetical list.

Therefore I suggest to remove it. But if you really want to keep it, we can add that page to the "full table of contents".

The-Compiler:
Thanks!

PR #9970, opened 2022-05-17
Fix rst markup in TempdirFactory's docstring.

ezio-melotti:
The docstring of `TempdirFactory` has two markup errors that are visible at https://docs.pytest.org/en/6.2.x/reference.html#pytest.TempdirFactory (since the autodoc directive fetches the docstring):

image

This PR fixes the markup.


nicoddemus (approved):
Thanks, appreciate it!

While at it I changed the references from the `LEGACY_PATH` constant to `py.path.local`, which is what users understand/know about. 👍

PR #9988, opened 2022-05-23
Allow configuration file to be hidden

### LINKED ISSUE #9987: Offer alternative config filename ".pytest.ini"
pytest is the only config file in my repo that is not, and cannot be hidden. (hidden with linux dot notation).

Other examples of config files include:
.coveragerc
.editorconfig
.eslintrc.json
.flake8
.gitattributes
.gitignore
.pre-commit-config.yaml
.prettierrc.yaml
.pylintrc
.stylelintrc.json

Meanwhile, pytest uses "pytest.ini"

I would like to propose that pytest load ".pytest.ini" in situations where "pytest.ini" does not exist.

grintor:
Adds support for `.pytest.ini` as an alternative to `pytest.ini`. Closes #9987

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -66,6 +66,7 @@ Charnjit SiNGH (CCSJ)
 Chris Lamb
 Chris NeJame
 Chris Rose
+Chris Wheeler
 Christian Boelsen
 Christian Fetzer
 Christian Neumüller
--- a/changelog/9987.improvement.rst
+++ b/changelog/9987.improvement.rst
@@ -1,0 +1,1 @@
+Added support for hidden configuration file by allowing ``.pytest.ini`` as an alternative to ``pytest.ini``.
--- a/doc/en/reference/customize.rst
+++ b/doc/en/reference/customize.rst
@@ -29,9 +29,11 @@ pytest.ini
 
 ``pytest.ini`` files take precedence over other files, even when empty.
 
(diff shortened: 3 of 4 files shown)


nicoddemus (changes requested):
LGTM, but we need an explicit test that loads from a `.pytest.ini` file to avoid regressions.

For an example see `test_pyproject_toml` in `test_config.py`, I think we can create a new test which follows the same approach taken there.

ameenkhan003 (approved):

ameenkhan003 (approved):

nicoddemus:
Took the liberty of adding the test myself (we just needed to add a new parameter to a parametrized test actually).

Thanks @grintor!

testing/test_config.py
```diff
@@ -112,21 +112,26 @@ def test_tox_ini_wrong_version(self, pytester: Pytester) -> None:
...
+            ("pytest", ".pytest.ini"),
+        ],
     )
     def test_ini_names(self, pytester: Pytester, name, section) -> None:
         pytester.path.joinpath(name).write_text(
             textwrap.dedent(
                 """
             [{section}]
-            minversion = 1.0
+            minversion = 3.36
```
nicoddemus, line 127: Used a more unusual version number than 1.0 just to be more clear that we are getting the actual value (1.0 seems a bit like it could be a default value).

PR #10007, opened 2022-05-31
Do not advertise that importlib will be default import mode

### LINKED ISSUE #10003: Add Location for Where to Provide Feedback for `--import-mode=importlib`

Reading [here](https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html?highlight=import%20mode#import-modes), it says:

> We intend to make importlib the default in future releases, depending on feedback.

However, I don't know where to provide feedback.

Please add to the docs a location where users should provide feedback.

#### Alternative Solutions

N/A
#### Additional context

I've used `--import-mode=importlib` successfully and I think it should become the new default!

nicoddemus:
It is clear at this point that changing the default would break a lot of suites, and is not a clear win in all cases anyway.

Close #10003

### DIFF
--- a/doc/en/explanation/goodpractices.rst
+++ b/doc/en/explanation/goodpractices.rst
@@ -173,10 +173,9 @@ This layout prevents a lot of common pitfalls and has many benefits, which are b
 `blog post by Ionel Cristian Mărieș <https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure>`_.
 
 .. note::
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):

adam-grant-hendry:
@nicoddemus ...How is it clear? You give no proof.

nicoddemus:
I commented in your original issue: https://github.com/pytest-dev/pytest/issues/10003#issuecomment-1142099508

PR #10012, opened 2022-06-01
 Update :class:`pytest.PytestUnhandledCoroutineWarning` to a deprecation; it will raise an error in pytest 8. [SQUASH]

graingert:
currently the PytestUnhandledCoroutineWarning are a regular PytestWarning. However really they are just a more important special case of PytestReturnNotNoneWarning

If the plan is to upbraid PytestReturnNotNoneWarning into a PytestReturnNotNoneError in pytest 8, we should turn both of these into an error at the same time.

See https://github.com/pytest-dev/pytest/pull/9956#issuecomment-1143853734
And https://github.com/pytest-dev/pytest/issues/7363

### DIFF
--- a/changelog/10012.deprecation.rst
+++ b/changelog/10012.deprecation.rst
@@ -1,0 +1,1 @@
+ upgrade PytestUnhandledCoroutineWarning to PytestDeprecationWarning
--- a/changelog/10012.deprecation.rst
+++ b/changelog/10012.deprecation.rst
@@ -1,1 +1,1 @@
- upgrade PytestUnhandledCoroutineWarning to PytestDeprecationWarning
+upgrade :class:`pytest.PytestUnhandledCoroutineWarning` it will raise an error in pytest 8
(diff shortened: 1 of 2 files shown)


src/_pytest/warning_types.py
```diff
@@ -55,7 +55,6 @@ class PytestRemovedIn8Warning(PytestDeprecationWarning):
     __module__ = "pytest"
 
 
-@final
 class PytestReturnNotNoneWarning(PytestDeprecationWarning):
```
graingert, line 58: @RonnyPfannschmidt requested that PytestReturnNotNoneWarning be a RemovedIn8Warning and I agree

nicoddemus (changes requested):
Looks good, however I think this warrants a changelog entry?

changelog/10012.deprecation.rst
```diff
@@ -0,0 +1 @@
+upgrade :class:`pytest.PytestUnhandledCoroutineWarning` to a deprecation it will raise an error in pytest 8
```
nicoddemus, line 1:
```suggestion
Change :class:`pytest.PytestUnhandledCoroutineWarning` to a deprecation; it will raise an error in pytest 8.
```

graingert:
@nicoddemus what do you think about making RemovedIn8Warning not-final?

nicoddemus:
I'm OK with making it not-final 👍

nicoddemus (approved):

RonnyPfannschmidt (approved):

PR #10026, opened 2022-06-07
Update location of `usage.rst` to fix manpage compilation

sandrotosi:
`usage.rst` has been moved from `doc/en` to `doc/en/how-to`, so the `man_pages` configuration
value needs to be updated to the new location, so that we dont get this warning:

    writing... WARNING: "man_pages" config value references unknown document usage


nicoddemus (approved):
Thanks, appreciate it!

Issue #10042, opened 2022-06-14
pdb + expect tests failing in Python 3.7


nicoddemus:
Link to a failing build:

https://github.com/pytest-dev/pytest/runs/6860982658

```
FAILED testing/test_debugging.py::TestPDB::test_pdb_interaction_capturing_simple - pexpect.exceptions.TIMEOUT: Timeout exceeded.
FAILED testing/test_debugging.py::TestPDB::test_pdb_interaction_capturing_twice - pexpect.exceptions.TIMEOUT: Timeout exceeded.
FAILED testing/test_debugging.py::TestPDB::test_pdb_with_injected_do_debug - assert False
FAILED testing/test_debugging.py::TestDebuggingBreakpoints::test_pdb_not_altered - pexpect.exceptions.TIMEOUT: Timeout exceeded.
FAILED testing/test_debugging.py::test_pdb_suspends_fixture_capturing[capfd] - assert 'out_inner' not in ') c\r\n\r\n...========\r\n'
FAILED testing/test_debugging.py::test_pdb_suspends_fixture_capturing[capsys] - assert 'out_inner' not in ') c\r\n\r\n...========\r\n'
```

Marking those tests as `xfail` so they get out of the way for now, but the cause of the errors should be investigated eventually.

thmo:
The tests do not fail for me on Python 3.11.0:
```
=================================== FAILURES ===================================
________________ TestDebuggingBreakpoints.test_pdb_not_altered _________________
[gw4] linux -- Python 3.11.0 /usr/bin/python3
[XPASS(strict)] #10042
_________________ TestPDB.test_pdb_interaction_capturing_twice _________________
[gw5] linux -- Python 3.11.0 /usr/bin/python3
[XPASS(strict)] #10042
________________ TestPDB.test_pdb_interaction_capturing_simple _________________
[gw2] linux -- Python 3.11.0 /usr/bin/python3
[XPASS(strict)] #10042
___________________ TestPDB.test_pdb_with_injected_do_debug ____________________
[gw5] linux -- Python 3.11.0 /usr/bin/python3
[XPASS(strict)] #10042
__________________ test_pdb_suspends_fixture_capturing[capfd] __________________
[gw0] linux -- Python 3.11.0 /usr/bin/python3
[XPASS(strict)] #10042
_________________ test_pdb_suspends_fixture_capturing[capsys] __________________
[gw0] linux -- Python 3.11.0 /usr/bin/python3
[...]

PR #10043, opened 2022-06-14
Mark pdb+expect tests as xfail for now

nicoddemus:
Related to #10042


nicoddemus:
> Yikes, those will be a pain

😬 

I would try to take a look, but those only run on Linux.

PR #10054, opened 2022-06-16
Indicate support for a tuple of exceptions in xfail raises=

SnoopJeDi:
This PR adds a note in the reference documentation for `pytest.mark.xfail` that indicates that a tuple of exceptions is allowed. Currently, a user (:wave:) can only discover this functionality in the supplementary remarks about `xfail` in the [`skipping`](https://pytest.org/en/stable/how-to/skipping.html#raises-parameter) documentation.


nicoddemus (approved):
Thanks a lot, appreciate it, and sorry for the delay!

PR #10100, opened 2022-07-04
Make ApproxScalar hashable

kalekundert:
This PR allows the following:
```
assert {1, 2} == {approx(1), approx(2)}
```
Currently it is not possible to construct the set on the right-hand side of this expression, because `ApproxScalar` is not hashable.

To me this seems like a bug, since `ApproxScalar` instances are meant to behave like floats, which are hashable.  That said, I was surprised to find that the lack of hashing is not an oversight: hashing is explicitly disabled with `ApproxScalar.__hash__ = None`.  I don't understand why the code would be written like this, so perhaps there's some larger issue I'm overlooking.  Implementing a hash function doesn't seem to break any tests, though.

### DIFF
--- a/src/_pytest/python_api.py
+++ b/src/_pytest/python_api.py
@@ -459,8 +459,8 @@ class ApproxScalar(ApproxBase):
         result: bool = abs(self.expected - actual) <= self.tolerance
         return result
 
-    # Ignore type because of https://github.com/python/mypy/issues/4266.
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (commented):
Please try out what happens when the sets have integers/floats that differ from the expected value, but match the absolute /relative range

PR #10123, opened 2022-07-11
Fix mypy pre-commit run

nicoddemus:
This started to fail recently with:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/_pytest/assertion/rewrite.py:284: error: Returning Any from function declared to return "TraversableResources"  [no-any-return]
Found 1 error in 1 file (checked 219 source files)
```

Not sure why that started failing, but seems like ignoring that error specifically is OK.

### DIFF
--- a/src/_pytest/assertion/rewrite.py
+++ b/src/_pytest/assertion/rewrite.py
@@ -281,7 +281,9 @@ class AssertionRewritingHook(importlib.abc.MetaPathFinder, importlib.abc.Loader)
             else:
                 from importlib.resources.readers import FileReader
 
-            return FileReader(types.SimpleNamespace(path=self._rewritten_names[name]))
+            return FileReader(  # type:ignore[no-any-return]
+                types.SimpleNamespace(path=self._rewritten_names[name])
+            )
 
 
 def _write_pyc_fp(


The-Compiler (commented):
FWIW this happens because `reveal_type(FileReader)` is `Any`, because typeshed [doesn't know anything](https://github.com/python/typeshed/tree/master/stdlib/importlib) about an `importlib.readers`. Indeed it seems to be [completely undocmented](https://docs.python.org/3/library/importlib.html) too.

But clearly it's public to some degree, we need it, and there doesn't seem to be a better place to import it from. Not sure why this didn't error earlier, but I can't think of a better fix other than disabling it either. 

cc @gaborbernat @bluetech @kdelee @asottile who seem to have been involved in changes to this code (e.g. #9609) and might be able to say more.

asottile:
yeah probably just a deficiency in typeshed -- I'd recommend reporting there:

- 3.10: https://github.com/python/cpython/blob/3.10/Lib/importlib/readers.py
- 3.11: https://github.com/python/cpython/blob/3.11/Lib/importlib/resources/readers.py

asottile (approved):

Issue #10131, opened 2022-07-13
Test failures with Python 3.11.0b4


hroncok:
- [x] a detailed description of the bug or problem you are having
- [x] output of `pip list` from the virtual environment you are using
- [x] pytest and operating system versions
- [x] minimal example if possible

Hello. We have updated Python from 3.11.0b3 to 3.11.0b4 in Fedora Linux 37 and we see some test failures of pytest 7.1.2. I was able to reproduce them from git HEAD @ c2b1d5b368a1fc829c6c06c700ae2ba9cda55794.

```
# on pytest git clone
$ git rev-parse HEAD
c2b1d5b368a1fc829c6c06c700ae2ba9cda55794
$ tox -e py311 -- -k 'test_wrap_session_notify_exception or test_doctest_unexpected_exception'
py311 inst-nodeps: .../pytest/.tox/.tmp/package/1/pytest-7.2.0.dev210+gc2b1d5b36.tar.gz
py311 installed: argcomplete==2.0.0,attrs==21.4.0,certifi==2022.6.15,charset-normalizer==2.1.0,elementpath==2.5.3,hypothesis==6.50.1,idna==3.3,iniconfig==1.1.1,mock==4.0.3,nose==1.3.7,packaging==21.3,pluggy==1.0.0,py==1.11.0,Pygments==2.12.0,pyparsing==3.0.9,pytest @ file://.../pytest/.tox/.tmp/package/1/pytest-7.2.0.dev210%2Bgc2b1d5b36.tar.gz,requests==2.28.1,sortedcontainers==2.4.0,urllib3==1.26.10,xmlschema==1.11.3
py311 run-test-pre: PYTHONHASHSEED='1407110200'
py311 run-test: commands[0] | pytest -k 'test_wrap_session_notify_exception or test_doctest_unexpected_exception'
============================= test session starts ==============================
platform linux -- Python 3.11.0b4, pytest-7.2.0.dev210+gc2b1d5b36, pluggy-1.0.0
cachedir: .tox/py311/.pytest_cache
rootdir: .../pytest, configfile: pyproject.toml, testpaths: testing
plugins: hypothesis-6.50.1
collected 3164 items / 3160 deselected / 4 selected

testing/test_doctest.py F                                                [ 25%]
testing/test_main.py FFF                                                 [100%]

=================================== FAILURES ===================================
________________ TestDoctests.test_doctest_unexpected_exception ________________

[...]

PR #10132, opened 2022-07-13
Adjust tests for a last minute Python 3.11 change in the traceback format

hroncok:
See https://github.com/python/cpython/issues/93883
and https://github.com/python/cpython/pull/93994

Fixes https://github.com/pytest-dev/pytest/issues/10131


nicoddemus (approved):
Interesting, thanks @hroncok!

PR #10143, opened 2022-07-15
Add monkeypatch.context() to how-to doc intro

sirosen:
This is just a small doc addition.

I consider myself a moderately experienced pytest user, but I had no idea about `monkeypatch.context()` until today. (And I am very happy to start using it where I still had `mock.patch` calls!) I think making it more prominent will help more users like me.


nicoddemus (approved):
Great, appreciate it!

PR #10145, opened 2022-07-16
issue-10113/Change error message

### LINKED ISSUE #10113: Misleading error message when parameters are missing in the parametrize but presented in the test function

I'm using pytest version 6.2.4.
Say we have the following function and a corresponding test function:
```python
# ./add.py
def add(n1, n2):
  return n1 + n2

# ./test_add.py
@pytest.mark.parametrize(
  "arg1, expected_answer",
  [pytest.param(2, 3, 5)]
)
def test_add(arg1, arg2, expected_answer):
  assert add(arg1, arg2) == expected_answer
```

Running `pytest ./test_add.py::test_add` logs the following error message:
```
ERROR: not found: ./test_add.py::test_add
(no name './test_add.py::test_add' in any of [<Module test_add.py>])
```

But actually it fails because `arg2` is not presented in the parametrize decorator.
Similar issue when the argument name is wrong.

jekwatt:
Fixes #10113.

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -650,7 +650,7 @@ class Session(nodes.FSCollector):
                 for arg, collectors in self._notfound:
                     if collectors:
                         errors.append(
-                            f"coundn't collect: {arg}\n(no name {arg!r} in any of {collectors!r})"
+                            f"couldn't collect: {arg}\n(no name {arg!r} in any of {collectors!r})"
                         )
                     else:
                         errors.append(f"found no collectors for {arg}")
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -650,7 +650,7 @@ class Session(nodes.FSCollector):
                 for arg, collectors in self._notfound:
                     if collectors:
                         errors.append(
-                            f"not found: {arg}\n(no name {arg!r} in any of {collectors!r})"
+                            f"coundn't collect: {arg}\n(no name {arg!r} in any of {collectors!r})"
                         )
                     else:
                         errors.append(f"found no collectors for {arg}")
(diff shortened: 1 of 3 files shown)


src/_pytest/main.py
```diff
@@ -650,7 +650,7 @@ def perform_collect(
                 for arg, collectors in self._notfound:
                     if collectors:
                         errors.append(
-                            f"not found: {arg}\n(no name {arg!r} in any of {collectors!r})"
+                            f"coundn't collect: {arg}\n(no name {arg!r} in any of {collectors!r})"
```
Zac-HD, line 653:
```suggestion
                            f"couldn't collect: {arg}\n(no name {arg!r} in any of {collectors!r})"
```

Zac-HD (approved):
Looks great - thanks Jennifer!

The-Compiler:
That seems like it would now make the error message confusing for the far more common case, when there indeed is a test which wasn't found...

nicoddemus:
@The-Compiler would you like to decline the PR then?

Zac-HD:
@The-Compiler - it's been a year, let's either merge this or close it and close the issue as wontfix.

The-Compiler:
Sorry for the radio silence, this got lost in my inbox with all the Europython stuff :sweat_smile: 

I still think that in the common case of:

```
$ pytest tests/unit/test_app.py::test_bla
[...]
collected 0 items                                                              

============================ no tests ran in 0.02s =============================
ERROR: not found: .../tests/unit/test_app.py::test_bla
(no name '.../tests/unit/test_app.py::test_bla' in any of [<Module tests/unit/test_app.py>])
```

"not found" makes things clearer vs. "couldn't collect" (after all, the collection did work!). As shown in https://github.com/pytest-dev/pytest/issues/10113#issuecomment-1178868063 I believe the existing error message is sufficient for this far more exotic case.

Thus I'm closing this - thanks for the contribution anyways though, @jekwatt, and sorry for the loooong delay!

PR #10171, opened 2022-07-26
Update fixtures.rst w/ finalizer order

### LINKED ISSUE #10023: `request.addfinalizer` order of execution is undocumented
This is very important if you are creating fixtures which interact with others which were not written with `yield`, and is extra confusing when writing tests which themselves *do* use `yield`.

aizpurua23a:
Fixes #10023 
Add note to clarify the difference in execution order between after-yield finalizers and `request.addfinalizer` finalizers.

### DIFF
--- a/doc/en/how-to/fixtures.rst
+++ b/doc/en/how-to/fixtures.rst
@@ -768,6 +768,32 @@ Consider the differences in the following examples:
         print("after_yield_2")
 
 
+.. code-block:: pytest
+
+    $ pytest test_module.py
+    =========================== test session starts ============================
+    platform linux -- Python 3.x.y, pytest-7.x.y, pluggy-1.x.y
+    collected 1 item
+
+    test_module.py test_bar
+    .after_yield_2
+    after_yield_1
+
+
+.. code-block:: python
+
+    import pytest
+
+
+    @pytest.fixture
+    def fix_w_finalizers(request):
+        request.addfinalizer(partial(print, "finalizer_2"))
+        request.addfinalizer(partial(print, "finalizer_1"))
+
+
+    @pytest.fixture
+    def test_bar(fix_w_finalizers):
+        print("test_bar")
 
 
 .. code-block:: pytest
@@ -775,17 +801,11 @@ Consider the differences in the following examples:
     $ pytest test_module.py
     =========================== test session starts ============================
     platform linux -- Python 3.x.y, pytest-7.x.y, pluggy-1.x.y
-    collected 2 items
+    collected 1 item
 
-    main.py
-    test_foo
+    test_module.py test_bar
     .finalizer_1
     finalizer_2
-
-    test_bar
-    .after_yield_1
-    after_yield_2
-
 
 
 .. _`safe teardowns`:
--- a/doc/en/how-to/fixtures.rst
+++ b/doc/en/how-to/fixtures.rst
@@ -738,6 +738,8 @@ Note on finalizer order
 Finalizers are executed in a first-in-last-out order, while operations after yield are executed sequentially.
 Consider the differences in the following examples:
 
+.. regendoc:wipe
+
 .. code-block:: python
 
     import pytest
--- a/doc/en/how-to/fixtures.rst
+++ b/doc/en/how-to/fixtures.rst
@@ -742,7 +742,7 @@ Note on finalizer order
 """"""""""""""""""""""""
 
 Finalizers are executed in a first-in-last-out order.
-For yield fixtures, the first fixture to run is the right-most one, i.e. the last test parameter.
+For yield fixtures, the first teardown code to run is from the right-most fixture, i.e. the last test parameter.
 
 .. regendoc:wipe
 
--- a/doc/en/how-to/fixtures.rst
+++ b/doc/en/how-to/fixtures.rst
@@ -759,7 +759,6 @@ Consider the differences in the following examples:
 
     def test_foo(fix_w_finalizers):
         print("test_foo")
-        pass
 
 
     def test_bar(fix_w_yield):
--- a/doc/en/how-to/fixtures.rst
+++ b/doc/en/how-to/fixtures.rst
@@ -729,15 +729,66 @@ Here's how the previous example would look using the ``addfinalizer`` method:
         assert email in receiving_user.inbox
 
 
-It's a bit longer than yield fixtures and a bit more complex, but it
-does offer some nuances for when you're in a pinch.
+Here's how the previous example would look using the ``addfinalizer`` method:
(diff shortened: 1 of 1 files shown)


doc/en/how-to/fixtures.rst
```diff
@@ -732,12 +732,64 @@ Here's how the previous example would look using the ``addfinalizer`` method:
...
+    @pytest.fixture
+    def fix_w_finalizers(request):
+        request.addfinalizer(partial(print, "finalizer_2"))
+        request.addfinalizer(partial(print, "finalizer_1"))
+
+
+    @pytest.fixture
+    def fix_w_yield():
+        yield
+        print("after_yield_1")
```
nicoddemus, line 756: This is not a good example I think, because of course both prints will be executed sequentially (pytest cannot mess with how Python executes code within a function).

I suggest we use two fixtures:

```python
    @pytest.fixture
    def fix_w_yield1():
        yield
        print("after_yield_1")

    @pytest.fixture
    def fix_w_yield2():
        yield
        print("after_yield_2")
```

And then make `test_bar` request both:

```python
    def test_bar(fix_w_yield1, fix_w_yield2):
        print("test_bar")
```

doc/en/how-to/fixtures.rst
```diff
@@ -732,12 +732,64 @@ Here's how the previous example would look using the ``addfinalizer`` method:
...
+        print("after_yield_2")
+
+
+    def test_foo(fix_w_finalizers):
+        print("test_foo")
+        pass
```
nicoddemus, line 762: We don't need the `pass` statement.

```suggestion
```

doc/en/how-to/fixtures.rst
```diff
@@ -732,12 +732,64 @@ Here's how the previous example would look using the ``addfinalizer`` method:
...
+
+
+    def test_bar(fix_w_yield):
+        print("test_bar")
+        pass
```
nicoddemus, line 767:
```suggestion
```

doc/en/how-to/fixtures.rst
```diff
@@ -732,12 +732,64 @@ Here's how the previous example would look using the ``addfinalizer`` method:
 It's a bit longer than yield fixtures and a bit more complex, but it
 does offer some nuances for when you're in a pinch.
 
+Note on finalizer order
+""""""""""""""""""""""""
+
+Finalizers are executed in a first-in-last-out order, while operations after yield are executed sequentially.
+Consider the differences in the following examples:
+
```
nicoddemus, line 740: This will cleanup the files from the previous example.

```suggestion

.. regendoc:wipe

```

doc/en/how-to/fixtures.rst
```diff
@@ -732,12 +732,64 @@ Here's how the previous example would look using the ``addfinalizer`` method:
...
+    test_foo
+    .finalizer_1
+    finalizer_2
+
+    test_bar
+    .after_yield_1
+    after_yield_2
 
+
+.. code-block:: pytest
```
nicoddemus, line 788: This block should be moved above the `Note on finalizer order` I believe?

nicoddemus (changes requested):
Thanks a lot @aizpurua23a! Left a few comments, please take a look.

doc/en/how-to/fixtures.rst
aizpurua23a: Ah, yes. Moved!

doc/en/how-to/fixtures.rst
aizpurua23a: Ah, I see.
The execution order with the addfinalizer method is evident, so the execution order only needs clarification in this case.
Lovely! How about now?

nicoddemus: I think we can keep the examples using `addfinalizer` as they are useful for comparision, so could you please bring them back? Thanks

nicoddemus (changes requested):

doc/en/how-to/fixtures.rst
aizpurua23a: Sure thing, re-added.

doc/en/how-to/fixtures.rst
```diff
@@ -735,8 +735,78 @@ does offer some nuances for when you're in a pinch.
...
    $ pytest -q test_emaillib.py
-   .                                                                    [100%]
-   1 passed in 0.12s
+  .                                                                    [100%]
+  1 passed in 0.12s
+
+Note on finalizer order
+""""""""""""""""""""""""
+
+Finalizers are executed in a first-in-last-out order, while operations after yield are executed sequentially.
```
nicoddemus, line 744: >  while operations after yield are executed sequentially.

This sentence needs tweaking if we look at the output. 👍 

Finalizers are executed in FILO order, that is correct, because we can see that adding `fin2` and `fin1` we get back `fn1`, `fn2` in the output.

But the same can be said for the yield fixtures: we request them in `fix_w_yield1`, `fix_w_yield2` order, and we get `fix_w_yield2` and `fix_w_yield1` in the output.

That's because `yield` fixtures use `addfinalizer` behind the scenes: when the fixture executes, we register a function with `addfinalizer` that resumes the generator (which in turn calls the teardown code).

Not sure how to "fix" this sentence, or perhaps we can explain something along the lines of what I wrote above, if that makes things clearer.

aizpurua23a: Yeah, that makes sense.
I tried to explain how both examples show the similar FILO behavior, plus added the hidden usage of addfinalizer after the examples.
What do you think?

doc/en/how-to/fixtures.rst
```diff
@@ -735,8 +735,81 @@ does offer some nuances for when you're in a pinch.
...
-   .                                                                    [100%]
-   1 passed in 0.12s
+  .                                                                    [100%]
+  1 passed in 0.12s
+
+Note on finalizer order
+""""""""""""""""""""""""
+
+Finalizers are executed in a first-in-last-out order.
+For yield fixtures, the first fixture to run is the right-most one, i.e. the last test parameter.
```
nicoddemus, line 745:
```suggestion
For yield fixtures, the first teardown code to run is from the right-most fixture, i.e. the last test parameter.
```

The fixture setup code is executed left-to-right, while the teardown code is right-to-left.

nicoddemus (changes requested):

doc/en/how-to/fixtures.rst
aizpurua23a: That's right, thanks for clarifying :)

nicoddemus:
Thanks @aizpurua23a!

cc @fish-face

nicoddemus (approved):

PR #10189, opened 2022-08-05
Change example to reference from within same page

brylie:
The text references a "previous example" that doesn't exist in the document. Readers coming directly to this document via a search engine may not have seen the previous example. It is possible that the example has been removed from the documentation.

This change makes a reference to a test case within the same document, so the reader is able to find the example.


nicoddemus:
Thanks @brylie!

Merging as the failures are unrelated. 👍

PR #10190, opened 2022-08-05
JUnit XML: Escape error messages in setup/teardown

holesch:


RonnyPfannschmidt (approved):
Lovely catch, thanks!

The-Compiler:
I know nothing about JUnit XML to be honest, but this is already approved and seems reasonable, so I'm merging and backporting it.

PR #10191, opened 2022-08-05
Make sure `monkeypatch.undo()` is always called

webknjaz:
While reading the docs, I noticed that in one place (in the fixture), `monkeypatch.undo()` is not wrapped with try/finally, but the CM method in the class, it is. It seems this could lead to weird behaviors so I figured I'd offer this patch seeing that it should improve things. Not sure if a change note is needed, feel free to request it :)

- [ ] Allow maintainers to push and squash when merging my commits. Please uncheck this if you prefer to squash the commits yourself.


The-Compiler:
The teardown part of fixtures is always called, even if there was an exception in the test - so I'm afraid I don't really see how this changes anything over what the current behavior already is.

The-Compiler:
Taking the freedom to close this, as I believe the current code already works correctly, and I don't see any benefit in changing it.

PR #10196, opened 2022-08-08
PytestReturnNotNoneWarning now subclasses PytestRemovedIn8Warning

graingert:
…@final` decorator

### DIFF
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -342,6 +342,12 @@ else:
         return f
 
 
+# TODO: sealed classes are not supported yet
+# https://github.com/mypyc/mypyc/issues/846
+def sealed(f: _T) -> _T:
+    return f
+
+
 if sys.version_info >= (3, 8):
     from functools import cached_property as cached_property
 else:
--- a/src/_pytest/compat.py
+++ b/src/_pytest/compat.py
@@ -342,12 +342,6 @@ else:
         return f
 
 
-# TODO: sealed classes are not supported yet
-# https://github.com/mypyc/mypyc/issues/846
-def sealed(f: _T) -> _T:
-    return f
-
-
 if sys.version_info >= (3, 8):
     from functools import cached_property as cached_property
 else:
--- a/src/_pytest/warning_types.py
+++ b/src/_pytest/warning_types.py
@@ -54,7 +54,6 @@ class PytestRemovedIn8Warning(PytestDeprecationWarning):
     __module__ = "pytest"
 
 
-@sealed
 class PytestReturnNotNoneWarning(PytestRemovedIn8Warning):
     """Warning emitted when a test function is returning value other than None."""
 
--- a/src/_pytest/warning_types.py
+++ b/src/_pytest/warning_types.py
@@ -6,7 +6,6 @@ from typing import TypeVar
 import attr
 
 from _pytest.compat import final
-from _pytest.compat import sealed
 
 
 class PytestWarning(UserWarning):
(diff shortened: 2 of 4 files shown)


src/_pytest/compat.py
```diff
@@ -342,6 +342,12 @@ def final(f):
         return f
 
 
+# TODO: sealed classes are not supported yet
```
graingert, line 345: is it worth trying to pre-empt `@sealed` existing? or just move the TODO: into warning_types.py?

bluetech:
The `@final` annotations are really just a way to say "we did not design this class for inheritance, if you inherit from it you're on your own". But it's a very weak thing, if we actually want to subclass some class then we should just remove the `@final` from it, i.e. I don't think the `@sealed` placeholder is necessary.

src/_pytest/compat.py
nicoddemus: As @bluetech commented, it is fine to just go ahead and remove it. 👍

src/_pytest/warning_types.py
```diff
@@ -48,14 +49,15 @@ class PytestDeprecationWarning(PytestWarning, DeprecationWarning):
...
-@final
+@sealed
 class PytestRemovedIn8Warning(PytestDeprecationWarning):
     """Warning class for features that will be removed in pytest 8."""
 
     __module__ = "pytest"
 
 
-class PytestReturnNotNoneWarning(PytestDeprecationWarning):
+@sealed
```
graingert, line 59:
```suggestion
```

src/_pytest/warning_types.py
```diff
@@ -48,14 +49,15 @@ class PytestDeprecationWarning(PytestWarning, DeprecationWarning):
     __module__ = "pytest"
 
 
-@final
+@sealed
```
graingert, line 52:
```suggestion
```

src/_pytest/warning_types.py
```diff
@@ -6,6 +6,7 @@
 import attr
 
 from _pytest.compat import final
+from _pytest.compat import sealed
```
graingert, line 9:
```suggestion
```

nicoddemus (changes requested):
Let's just remove the `@final` decorators then.

src/_pytest/compat.py
```diff
@@ -342,6 +342,12 @@ def final(f):
...
+# https://github.com/mypyc/mypyc/issues/846
+def sealed(f: _T) -> _T:
```
nicoddemus, line 347: This is no longer used and can be removed?

graingert: Yep

nicoddemus:
Other than the existing comments we also need a new changelog entry. 👍

src/_pytest/warning_types.py
```diff
@@ -48,14 +48,13 @@ class PytestDeprecationWarning(PytestWarning, DeprecationWarning):
...
 
-@final
 class PytestRemovedIn8Warning(PytestDeprecationWarning):
     """Warning class for features that will be removed in pytest 8."""
 
     __module__ = "pytest"
 
 
-class PytestReturnNotNoneWarning(PytestDeprecationWarning):
+class PytestReturnNotNoneWarning(PytestRemovedIn8Warning):
```
nicoddemus, line 57: Sorry, I don't recall the original discussion, do we ever plan to make returning non-None an error? Do you have a link to the original discussion.

graingert: https://github.com/pytest-dev/pytest/pull/9956#issuecomment-1132710934

nicoddemus: Ahh thanks!

nicoddemus:
Thanks @graingert!

PR #10202, opened 2022-08-10
Updated fixtures.rst

piermarcob:
Added fixtures decorators, otherwise it seems possible to use standard Python functions as fixtures.


The-Compiler:
The whole point of that example is to demonstrate how fixtures would work in pure Python outside of pytest:

> Here's how this example would work if we did it by hand:

See the final part of the example:

```python
    entry = first_entry()
    the_list = order(first_entry=entry)
    test_string(order=the_list)
```

Thus, I'm closing this - thanks for the contribution, though!

PR #10204, opened 2022-08-10
Fix partially unknown type

brandonchinn178:
Related to https://github.com/pytest-dev/pytest/issues/7469


The-Compiler (approved):
Explicit is better than implicit, I guess - thanks!

brandonchinn178:
ah sorry I guess I didnt note the motivation: this fixes pyright in strict mode, which errored with "Partially unknown type"

PR #10206, opened 2022-08-12
Revamp good practices

flying-sheep:
The importlib import mode has no* drawbacks while the default mode `prepend` has a major footgun. That one has one popular workaround which in turn has further drawbacks.

Also one of the sections correctly says “subfolder”, but others misleadingly say “subpackage”. Should I get rid of that too?

*I consider the fact that you can’t import from test files a feature, as it cleans up the distinction between test files and python packages. It encourages to keep your test utils in one location and reduces dependencies between test files.

### DIFF
--- a/doc/en/explanation/goodpractices.rst
+++ b/doc/en/explanation/goodpractices.rst
@@ -117,7 +117,6 @@ Here, your application root package resides in a sub-directory of your root:
 .. code-block:: text
 
     pyproject.toml
-    setup.cfg
     src/
         mypkg/
             __init__.py
@@ -139,7 +138,6 @@ want to distribute them along with your application:
 .. code-block:: text
 
     pyproject.toml
-    setup.cfg
     mypkg/
         __init__.py
         app.py
--- a/doc/en/explanation/goodpractices.rst
+++ b/doc/en/explanation/goodpractices.rst
@@ -74,7 +74,6 @@ to keep tests separate from actual application code (often a good idea):
 .. code-block:: text
 
     pyproject.toml
-    setup.cfg
     mypkg/
         __init__.py
         app.py
@@ -245,7 +244,7 @@ setuptools intends to
     The test files in the first example would be imported as
     ``test_app`` and ``test_view`` top-level modules by adding ``tests/`` to ``sys.path``.
 
-    This results in the following drawback compared to the import mode ``import-lib``:
+    This results in a drawback compared to the import mode ``import-lib``:
     your test files must have **unique names**.
 
     If you need to have test modules with the same name,
@@ -255,7 +254,6 @@ setuptools intends to
     .. code-block:: text
 
         pyproject.toml
-        setup.cfg
         mypkg/
             ...
         tests/
--- a/doc/en/explanation/goodpractices.rst
+++ b/doc/en/explanation/goodpractices.rst
@@ -96,8 +96,7 @@ This has the following benefits:
     See :ref:`pytest vs python -m pytest` for more information about the difference between calling ``pytest`` and
     ``python -m pytest``.
 
-For the most convenient experience,
-choose the ``import-lib`` :ref:`import mode <import-modes>`.
+For new projects, we recommend to use ``importlib`` :ref:`import mode <import-modes>` (see [1]_ for a detailed explanation).
 To this end, add the following to your ``pyproject.toml``:
 
 .. code-block:: toml
@@ -111,7 +110,7 @@ The default :ref:`import mode <import-modes>` ``prepend`` has several drawbacks 
 
 .. _`src-layout`:
 
-Generally, but especially if you use the default import mode,
+Generally, but especially if you use the default import mode ``prepend``,
 it is **strongly** suggested to use a ``src`` layout.
 Here, your application root package resides in a sub-directory of your root:
 
@@ -244,7 +243,7 @@ setuptools intends to
     The test files in the first example would be imported as
     ``test_app`` and ``test_view`` top-level modules by adding ``tests/`` to ``sys.path``.
 
-    This results in a drawback compared to the import mode ``import-lib``:
+    This results in a drawback compared to the import mode ``importlib``:
     your test files must have **unique names**.
 
     If you need to have test modules with the same name,
--- a/doc/en/explanation/goodpractices.rst
+++ b/doc/en/explanation/goodpractices.rst
@@ -12,41 +12,26 @@ For development, we recommend you use :mod:`venv` for virtual environments and
 as well as the ``pytest`` package itself.
 This ensures your code and dependencies are isolated from your system Python installation.
 
-Next, place a ``pyproject.toml`` file in the root of your package:
+Create a ``pyproject.toml`` file in the root of your repository as described in
+:doc:`packaging:tutorials/packaging-projects`.
+The first few lines should look like this:
 
 .. code-block:: toml
 
     [build-system]
-    requires = ["setuptools>=42", "wheel"]
-    build-backend = "setuptools.build_meta"
-
-and a ``setup.cfg`` file containing your package's metadata with the following minimum content:
-
-.. code-block:: ini
+    requires = ["hatchling"]
+    build-backend = "hatchling.build"
 
     [metadata]
-    name = PACKAGENAME
-
-    [options]
-    packages = find:
+    name = "PACKAGENAME"
 
 where ``PACKAGENAME`` is the name of your package.
-
-.. note::
-
-    If your pip version is older than ``21.3``, you'll also need a ``setup.py`` file:
-
-    .. code-block:: python
-
-        from setuptools import setup
-
-        setup()
 
 You can then install your package in "editable" mode by running from the same directory:
 
 .. code-block:: bash
 
-     pip install -e .
+    pip install -e .
 
 which lets you change your source code (both tests and application) and rerun tests at will.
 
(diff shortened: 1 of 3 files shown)


flying-sheep:
No idea why there are test failures, but I’m going to assume you sort that out

The-Compiler:
> [flying-sheep](https://github.com/flying-sheep) wants to merge 1 commit into [pytest-dev:6.2.x](https://github.com/pytest-dev/pytest/tree/6.2.x) from [flying-sheep:patch-1](https://github.com/flying-sheep/pytest/tree/patch-1)

Note the PR should be against the `main` branch (and then will be backported to `7.1.x`). The `6.2.x` branch is long outdated.

> It’s misleading to have them there, as the `tests` directory isn’t (and shouldn’t be a package)

I don't think I agree with that. See [Good Integration Practices — pytest documentation](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#choosing-a-test-layout-import-rules) - however, they are still not really needed for the example, so I'd be fine with removing them.

> No idea why there are test failures, but I’m going to assume you sort that out

From a quick look, they indeed don't look like they are related to your changes.

flying-sheep:
> See [Good Integration Practices — pytest documentation](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#choosing-a-test-layout-import-rules)

I’m intimately familiar with that document:

> If you need to have test modules with the same name, you might add __init__.py files to your tests folder and subfolders, changing them to packages:

and

> Note
> 
> The --import-mode=importlib option (see [Import modes](https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html#import-modes)) does not have any of the drawbacks above because sys.path is not changed when importing test modules, so users that run into this issue are strongly encouraged to try it.

RonnyPfannschmidt (changes requested):
Unless the default import mode of pytest gets pep420 aware these are absolutely needed

flying-sheep:
@RonnyPfannschmidt please read the comment I made a few seconds before yours

RonnyPfannschmidt:
The files missing is a footgun, the failure modes i keep observing when people expand testsuites are a pain, it's simply horrendous to debug for less experienced developers

So unless we fix the default import mode up im opposed to leaving them out

flying-sheep:
Wait, so pytest doesn’t throw a simple clear error like

> you appear to have multiple test files with the same name. Please try `--import-mode=importlib` or make your tests folder into a package by adding empty `__init__.py` files to each folder with files containing tests.

RonnyPfannschmidt:
It can't know if it's a mistake or a fatal breakage in multiple ways

RonnyPfannschmidt:
So neither making it a package nor impormode are a viable fix in multiple cases of import missmatch errors

nicoddemus:
> The files missing is a footgun, the failure modes i keep observing when people expand testsuites are a pain, it's simply horrendous to debug for less experienced developers

So perhaps we should update the docs to at least add a note to that end? Because if we advertise to users that having or not having them is fine depending on the case, we should not be reluctant to remove the `__init__.py` files from that section on the docs.

flying-sheep:
My personal preferences would be

1. Ideally make `--import-mode=importlib` the default
2. If that’s not possible yet (what’s keeping us?) promote it more heavily in `explanation/goodpractices.html`. Basically start with “add it to `addopts`”, and then towards the end “if you can’t use it, make `tests` a package as a workaround”

nicoddemus:
Unfortunately we found that `importlib` mode is not perfect either, and would break a ton of test suites if we ever changed the defaults. The main issue is that test modules cannot import from other modules in `tests`, which is a very common idiom.

I'm OK with changing the good practices docs to mention it as an alternative approach (listing the benefits and drawbacks of each).

flying-sheep:
> The main issue is that test modules cannot import from other modules in tests, which is a very common idiom.

Yeah, test suites relying on this misfeature would break. If this was my personal project, I’d release a new major version with the change and tell people to do things this way, but I understand that your stability guarantees are stronger.

```
|- src/packagename/
|  |- test_utils  # public test utils to be used in packages depending on ours
|  '- _test_utils  # private test utils that somehow can’t be fixtures
'- tests/
   |- conftest.py  # private test fixtures used in most places
   |- test_stuff.py
   '- test_feature/
      |- conftest.py  # private test fixtures/overrides to test that feature
      '- test_stuff.py
```

> I'm OK with changing the good practices docs to mention it as an alternative approach (listing the benefits and drawbacks of each).

Great, I’ll see if I can come up with something :D

flying-sheep:
OK, I took a shot at updating [`goodpractices.rst`](https://github.com/pytest-dev/pytest/pull/10206/files#diff-2bd72f21d977750bb53cb22d58c1b0d4646d7d72f55674b7c0386f4b6516718b). Maybe someone has opinions.

I think it can be further slimmed by moving the huge `.. note::` block labelled `test package name` to what is now a footnote, and promote that footnote to its own page or header.

Then goodpractices.rst will only contain good practices and pointers to the default.

I also didn’t mention the `testing.mypkg`/`mypkg.testing` approach to “what if I want to import test utils”, that should probably also go in.

doc/en/explanation/goodpractices.rst
```diff
@@ -112,43 +96,24 @@ This has the following benefits:
...
-``pytest`` will import them as *top-level* modules since there are no packages
-to derive a full package name from. In other words, the test files in the example above will
-be imported as ``test_app`` and ``test_view`` top-level modules by adding ``tests/`` to
-``sys.path``.
-
-If you need to have test modules with the same name, you might add ``__init__.py`` files to your
-``tests`` folder and subfolders, changing them to packages:
+For the most convenient experience,
+choose the ``import-lib`` :ref:`import mode <import-modes>`.
+To this end, add the following to your ``pyproject.toml``:
```
nicoddemus, line 101:
```suggestion
For new projects, we recommend to use ``importlib`` :ref:`import mode <import-modes>` (see [1]_ for a detailed explanation).
To this end, add the following to your ``pyproject.toml``:
```

doc/en/explanation/goodpractices.rst
```diff
@@ -112,43 +96,24 @@ This has the following benefits:
...
 
-This is problematic if you are using a tool like `tox`_ to test your package in a virtual environment,
-because you want to test the *installed* version of your package, not the local code from the repository.
+The default :ref:`import mode <import-modes>` ``prepend`` has several drawbacks [1]_.
 
 .. _`src-layout`:
 
-In this situation, it is **strongly** suggested to use a ``src`` layout where application root package resides in a
-sub-directory of your root:
+Generally, but especially if you use the default import mode,
```
nicoddemus, line 114:
```suggestion
Generally, but especially if you use the default import mode ``prepend``,
```

doc/en/explanation/goodpractices.rst
```diff
@@ -112,43 +96,24 @@ This has the following benefits:
...
+it is **strongly** suggested to use a ``src`` layout.
+Here, your application root package resides in a sub-directory of your root:
 
 .. code-block:: text
```
nicoddemus, line 119: (I can't comment on it because it is not part of the diff)

We should remove `setup.cfg` from the listing below given we did the same with the first listing.

doc/en/explanation/goodpractices.rst
nicoddemus, line 114: I agree we should primarily recommend the src code layout, so it makes sense to move this layout to the top of the session.

doc/en/explanation/goodpractices.rst
```diff
@@ -159,26 +124,11 @@ sub-directory of your root:
...
 
-.. note::
-    The ``--import-mode=importlib`` option (see :ref:`import-modes`) does not have
-    any of the drawbacks above because ``sys.path`` is not changed when importing
-    test modules, so users that run into this issue are strongly encouraged to try it.
-
-    The ``src`` directory layout is still strongly recommended however.
-
 
 Tests as part of application code
```
nicoddemus, line 133: We should remove `setup.cfg` from the listing below too (I can't comment directly on it because it is not part of the diff).

doc/en/explanation/goodpractices.rst
```diff
@@ -285,3 +235,46 @@ See also `pypa/setuptools#1684 <https://github.com/pypa/setuptools/issues/1684>`
...
+
+
+.. [1] The default ``prepend`` :ref:`import mode <import-modes>` works as follows:
+
+    Since there are no packages to derive a full package name from,
+    ``pytest`` will import your test files as *top-level* modules.
+    The test files in the first example would be imported as
+    ``test_app`` and ``test_view`` top-level modules by adding ``tests/`` to ``sys.path``.
+
+    This results in a drawback compared to the import mode ``import-lib``:
```
nicoddemus, line 247:
```suggestion
    This results in a drawback compared to the import mode ``importlib``:
```

nicoddemus (changes requested):
Thanks @flying-sheep for the follow up.

Left a few comments, please take a look.

doc/en/explanation/goodpractices.rst
```diff
@@ -12,41 +12,26 @@ For development, we recommend you use :mod:`venv` for virtual environments and
...
 .. code-block:: toml
 
     [build-system]
-    requires = ["setuptools>=42", "wheel"]
-    build-backend = "setuptools.build_meta"
-
-and a ``setup.cfg`` file containing your package's metadata with the following minimum content:
-
-.. code-block:: ini
+    requires = ["hatchling"]
```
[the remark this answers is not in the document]
flying-sheep: hmm, maybe we can also use tabs like in [the packaging tutorial][]?

I don’t recommend setuptools as it’s:

- slow
- messy (leaves a .egg-info folder lying around in your workspace, emits a bunch of unnecessary log messages during build)
- hard to configure (a bunch of legacy config files like MANIFEST.in, and misleadingly named config keys)
- no longer the first choice on [the packaging tutorial][]

but of course I can leave it if you prefer. I’d still get rid of `setup.cfg` since setuptools these days [supports PEP 621](https://setuptools.pypa.io/en/latest/userguide/quickstart.html). Unfortunately configuring it with more than just basic metadata is labelled as “beta”, e.g. this:

```toml
[tool.setuptools.packages.find]
where = ["src"]  # ["."] by default
```

[the packaging tutorial]: https://packaging.python.org/en/latest/tutorials/packaging-projects/

doc/en/explanation/goodpractices.rst
```diff
@@ -112,43 +96,24 @@ This has the following benefits:
...
-the test modules from the ``tests`` directory, pytest prepends the root of the repository to
-``sys.path``, which adds the side-effect that now ``mypkg`` is also importable.
+    [tool.pytest.ini_options]
+    addopts = [
+        "--import-mode=importlib",
+    ]
 
-This is problematic if you are using a tool like `tox`_ to test your package in a virtual environment,
-because you want to test the *installed* version of your package, not the local code from the repository.
+The default :ref:`import mode <import-modes>` ``prepend`` has several drawbacks [1]_.
```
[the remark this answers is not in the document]
flying-sheep: Hmm, my understanding is that there‘s only the property that you can’t import from test files, which is an advantage in my book. It’s only a blocker for adoption by existing projects because it’s not backwards compatible, which I should mention.

Are there any disadvantages apart from “isn’t backwards compatible in this regard”?

doc/en/explanation/goodpractices.rst
flying-sheep: will do!

doc/en/explanation/goodpractices.rst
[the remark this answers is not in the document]
nicoddemus: > no longer the first choice on [the packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/)

Ahh thanks for pointing that out. If the packaging tutorial recommends `hatchling` as first choice, I think we should follow suit then.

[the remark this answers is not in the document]
flying-sheep: In case you would like some context, I followed that discussion, and the reasoning went something like that:

1. we don’t want to promote any standards following build backend
2. so we use tabs
3. but since the first (auto-selected) tab is kind of an endorsement, we need something there that is good for the vast majority of people, i.e. pure python packages
4. because of the above mentioned drawbacks of setuptools, the first tab will be hatchling, because it provides a better experience for people creating a pure python package than setuptools

Hatchling doesn’t have a sufficient amount of plugins or example code yet to allow a highly custom build like numpy or numba or so, but for anything with purely static build configuration, it’s a very good choice.

doc/en/explanation/goodpractices.rst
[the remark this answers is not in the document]
nicoddemus: @flying-sheep gentle ping here just in case you missed this comment. 👍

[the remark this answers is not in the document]
flying-sheep: Sure, I’ll spell out the differences there! Due to the amount of misunderstandings of the good practices document in the past, I think there should be several references, so I think referring to that footnote/section here in some form isn’t the worst idea.

nicoddemus (approved):

nicoddemus:
Thanks @flying-sheep!

nicoddemus:
Anybody else wants to review this?

Otherwise I plan to merge this tomorrow.

nicoddemus:
Thanks again @flying-sheep!

PR #10209, opened 2022-08-12
fallback to native traceback when handling ExceptionGroup (take 2) [SQUASH]

jakkdl:
Picking up work on and therefore closes #10071 to temporarily address #9159

Improvements over the previous PR:
* Handles `BaseExceptionGroup`
* Handles exception chains properly, using `excinfo_` instead of `excinfo`, and sets `reprcrash`
* fixed tox.ini
* rewrote tests extending them, previous ones also had regex bugs

I'm not entirely sure how to handle the envlist in tox.ini, since hypothesis currently depends on `exceptiongroup`, making an extra environment for it redundant. I also wanted a test for py311 without `exceptiongroup` installed, but it got messy setting up a separate testenv without `extras=testing`. An alternative is a 3.11-only test that modifies `_pytest._code.code.ExceptionGroups` removing `exceptiongroup.BaseExceptionGroup` from it.

The test is still somewhat WIP, I want some extra matching lines to differentiate the parametrized variables, but thought I'd push my current work for initial review.

@Zac-HD

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,6 +46,7 @@ jobs:
           "ubuntu-py39",
           "ubuntu-py310",
           "ubuntu-py311",
+          "ubuntu-py311-exceptiongroup",
           "ubuntu-pypy3",
 
           "macos-py37",
@@ -114,6 +115,10 @@ jobs:
             python: "3.11-dev"
             os: ubuntu-latest
             tox_env: "py311"
+          - name: "ubuntu-py311-exceptiongroup"
+            python: "3.11-dev"
+            os: ubuntu-latest
+            tox_env: "py311-exceptiongroup"
           - name: "ubuntu-pypy3"
             python: "pypy-3.7"
             os: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,6 +115,7 @@ jobs:
             python: "3.11-dev"
             os: ubuntu-latest
             tox_env: "py311"
+            use_coverage: true
           - name: "ubuntu-py311-exceptiongroup"
             python: "3.11-dev"
             os: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,7 +46,6 @@ jobs:
           "ubuntu-py39",
           "ubuntu-py310",
           "ubuntu-py311",
-          "ubuntu-py311-exceptiongroup",
           "ubuntu-pypy3",
 
           "macos-py37",
@@ -116,10 +115,6 @@ jobs:
             os: ubuntu-latest
             tox_env: "py311"
             use_coverage: true
-          - name: "ubuntu-py311-exceptiongroup"
-            python: "3.11-dev"
-            os: ubuntu-latest
-            tox_env: "py311-exceptiongroup"
           - name: "ubuntu-pypy3"
             python: "pypy-3.7"
             os: ubuntu-latest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,6 +68,7 @@ repos:
           - packaging
           - tomli
           - types-pkg_resources
+            # for python>=3.11
           - exceptiongroup>=1.0.0rc8
 -   repo: local
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,6 +68,7 @@ repos:
           - packaging
           - tomli
           - types-pkg_resources
+          - exceptiongroup>=1.0.0rc8
 -   repo: local
     hooks:
     -   id: rst
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,7 +68,8 @@ repos:
           - packaging
           - tomli
           - types-pkg_resources
-            # for python>=3.11
+            # for mypy running on python>=3.11 since exceptiongroup is only a dependency
+            # on <3.11
           - exceptiongroup>=1.0.0rc8
 -   repo: local
     hooks:
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -56,6 +56,18 @@ if TYPE_CHECKING:
 
     _TracebackStyle = Literal["long", "short", "line", "no", "native", "value", "auto"]
 
+ExceptionGroupTypes: tuple = ()  # type: ignore
+try:
+    ExceptionGroupTypes += (ExceptionGroup,)  # type: ignore
+except NameError:
+    pass  # Is missing for `python<3.10`
+try:
+    import exceptiongroup
+
+    ExceptionGroupTypes += (exceptiongroup.ExceptionGroup,)
+except ModuleNotFoundError:
+    pass  # No backport is installed
+
 
 class Code:
     """Wrapper around Python code objects."""
@@ -923,11 +935,20 @@ class FormattedExcinfo:
         seen: Set[int] = set()
         while e is not None and id(e) not in seen:
             seen.add(id(e))
-            if excinfo_:
-                reprtraceback = self.repr_traceback(excinfo_)
-                reprcrash: Optional[ReprFileLocation] = (
-                    excinfo_._getreprcrash() if self.style != "value" else None
+            if isinstance(e, ExceptionGroupTypes):
+                reprtraceback: Union[
+                    ReprTracebackNative, ReprTraceback
+                ] = ReprTracebackNative(
+                    traceback.format_exception(
+                        type(excinfo.value),
+                        excinfo.value,
+                        excinfo.traceback[0]._rawentry,
+                    )
                 )
+                reprcrash: Optional[ReprFileLocation] = None
+            elif excinfo_:
+                reprtraceback = self.repr_traceback(excinfo_)
+                reprcrash = excinfo_._getreprcrash() if self.style != "value" else None
             else:
                 # Fallback to native repr if the exception doesn't have a traceback:
                 # ExceptionInfo objects require a full traceback to work.
--- a/src/_pytest/_code/code.py
+++ b/src/_pytest/_code/code.py
@@ -56,17 +56,18 @@ if TYPE_CHECKING:
 
     _TracebackStyle = Literal["long", "short", "line", "no", "native", "value", "auto"]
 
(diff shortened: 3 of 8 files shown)


jakkdl:
failing code coverage is partly for the 3.11-only test, and partly for the code written for when `exceptiongroup` is not installed.

jakkdl:
also shoutout to @Cheukting for giving a good base to work off of. I'll add you to `AUTHORS` if you want :)

Cheukting:
Thank you @jakkdl 😊

Zac-HD (commented):
Two comments below; otherwise both tests and implementation look good to me.  I've also confirmed that it works locally 👍 

Definitely add Cheuk to the authors list though, it's important to give OSS credit (and copyright acknowledgement!) where it's due.

Zac-HD:
Tagging @RonnyPfannschmidt for review too because I'm pretty close to the whole ExceptionGroup project, between PEP-678 and implementing support in Hypothesis.

src/_pytest/_code/code.py
```diff
@@ -56,6 +56,19 @@
 
     _TracebackStyle = Literal["long", "short", "line", "no", "native", "value", "auto"]
 
+ExceptionGroupTypes: Tuple[Type[BaseException], ...] = ()
+
+if sys.version_info >= (3, 11):
+    ExceptionGroupTypes = (BaseExceptionGroup,)  # type: ignore # noqa: F821
```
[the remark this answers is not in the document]
jakkdl: Needed some wrangling to get mypy to accept a version of this (esp as I went all the way of getting mypy to pass it both on 3.10 and 3.11), but good call - no need for a tuple.

jakkdl:
Fixed comments, although Cheukting is already in `AUTHORS` :sweat_smile: Also changed the tests a little bit.
Adding `exceptiongroup` to the pre-commit-config is just for line 64, `from exceptiongroup import BaseExceptionGroup`, so maybe simpler to just `#type: ignore` that line instead.

I managed to figure out how to add a tox testenv to test py37 without exceptiongroup installed, but it's quite a hassle and doesn't seem worth it just to test ~2 lines of code.

Zac-HD (approved):
Looks great to me - thanks @jakkdl!

If I don't hear any objections (@nicoddemus @RonnyPfannschmidt), I'll merge this on Friday.

tox.ini
```diff
@@ -11,6 +11,7 @@ envlist =
     py311
     pypy3
     py37-{pexpect,xdist,unittestextras,numpy,pluggymain}
+    py311-exceptiongroup
```
nicoddemus, line 14: This new tox environment needs to be added to `.github/workflows/test.yml`, otherwise it won't execute in CI.

src/_pytest/_code/code.py
```diff
@@ -56,6 +56,15 @@
 
     _TracebackStyle = Literal["long", "short", "line", "no", "native", "value", "auto"]
 
+BaseExceptionGroup: Optional[Type[BaseException]]
```
nicoddemus, line 59: Perhaps we should make `exceptiongroup` a proper dependency in Python<3.11?

This way we can simplify this code quite a bit:

```python
if sys.version_info[:2] < (3, 11):
    from exceptiongroup import BaseExceptionGroup
```

And `BaseExceptionGroup` would always be available.

nicoddemus (changes requested):
Left a few comments, please take a look.

src/_pytest/_code/code.py
Zac-HD: Actually yeah, a hard dependency would also be useful for https://github.com/pytest-dev/pytest/issues/8217

tox.ini
jakkdl: added on ubuntu, do you think it's worth testing on windows too?
I'm also curious why py311 is not run with `xdist`, I defaulted to follow suit and not do as well.

src/_pytest/_code/code.py
jakkdl: Makes sense that it'll be used in more places going forward, fixed :+1:
Also means there's no need for any tests in the vein of `py37-noexceptgroup`

tox.ini
jakkdl: I also took the liberty of adding `use_coverage` to ubuntu-py311 to get 100% code coverage

.pre-commit-config.yaml
```diff
@@ -68,6 +68,8 @@ repos:
           - packaging
           - tomli
           - types-pkg_resources
+            # for python>=3.11
```
Zac-HD, line 71: `python<3.11`?

jakkdl: on python<3.11 it's required by `install_depends` and will be installed, but on python>=3.11 it won't be, and so mypy will fail typechecking exceptiongroup if the linting is run in a default environment of 3.11

src/_pytest/_code/code.py
```diff
@@ -56,6 +56,12 @@
 
     _TracebackStyle = Literal["long", "short", "line", "no", "native", "value", "auto"]
 
+BaseExceptionGroup: Type[BaseException]
```
nicoddemus, line 59: I suggest we use an explicit version check here:

```python
if sys.version_info[:2] < (3, 11):
    from exceptiongroup import BaseExceptionGroup
```

This makes it easy to remove this code once it is no longer needed, also it won't given a confusing error message in case we are running in corrupted environment.

.github/workflows/test.yml
```diff
@@ -114,6 +115,11 @@ jobs:
             python: "3.11-dev"
             os: ubuntu-latest
             tox_env: "py311"
+            use_coverage: true
+          - name: "ubuntu-py311-exceptiongroup"
+            python: "3.11-dev"
+            os: ubuntu-latest
+            tox_env: "py311-exceptiongroup"
```
nicoddemus, line 122: Don't we want to test this in Python < 3.11, so as to test our `exceptiongroup` external package support? IIUC this is testing in Python 3.11 using the native `ExceptionGroup`.

jakkdl: This environment is for testing that 3.11 works when tested code raises `exceptiongroup.ExceptionGroup` and it's caught as a native `ExceptionGroup`. The standard `py37` checks that `exceptiongroup.ExceptionGroup` works, and `py311` that native `ExceptionGroup` works when `exceptiongroup` is not installed.

src/_pytest/_code/code.py
jakkdl: yeah that's much cleaner

bluetech:
Since `exceptiongroup` hasn't had a stable release yet and has has some breaking changes in the recent releases, let's ask `exceptiongroup`'s developer:

@agronholm - pytest is considering adding a dependency on `exceptiongroup` for Python versions before 3.11. pytest is a popular project so the dependency will add a lot of installs of `exceptiongroup`. Is this OK with you? If so, do you think it's OK to depend on version `exceptiongroup>=1.0.0rc8` or should we wait for a final release? Thanks.

agronholm:
The only reason I've held back the final release is to ensure that it matches the behavior of Python 3.11 in its final release.
That said, I would like to resolve the last two open issues (https://github.com/agronholm/exceptiongroup/issues/18 and https://github.com/agronholm/exceptiongroup/issues/14) before making a final release.

And yes, I'm okay with pytest depending on `exceptiongroup`. But the latter of the open issues would need to be resolved for pytest which, IIRC, installs its own exception hook, yes?

.github/workflows/test.yml
nicoddemus: Ahh right, thanks for the clarification!

jakkdl: though since the implementation has ended up this simple, running CI for a single test that's pretty much only gonna break if `exceptiongroup` breaks is maybe a bit silly

nicoddemus: Indeed, agree.

Feel free to drop this then. 👍

agronholm:
If it overrides those hooks, then it needs some way to manually render exception groups – something the `exceptiongroup` package doesn't provide yet. I'm not sure how to even provide such functionality in a way that doesn't involve patching `TracebackException`. It's something that needs to be researched. I would welcome any help with this.

Zac-HD:
> If it overrides those hooks, then it needs some way to manually render exception groups – something the `exceptiongroup` package doesn't provide yet. I'm not sure how to even provide such functionality in a way that doesn't involve patching `TracebackException`.

I'd have thought that calling the `traceback.format_exception()` in (without loss of generality) pytest's hook function would work, since that ultimately delegates to the patched-by-`exceptiongroup` `TracebackException` methods?

And that's what Pytest's `sys.unraisablehook` and `threading.excepthook` currently do, _without_ going through the fancier formatting that this PR works around, so I think it already works.

agronholm:
> I'd have thought that calling the traceback.format_exception() in (without loss of generality) pytest's hook function would work, since that ultimately delegates to the patched-by-exceptiongroup TracebackException methods?

You're probably right – I am no so knowledgeable about the inner workings of the `traceback` module. I just adapted trio's `MultiError` patching to `exceptiongroup`. To be honest, I don't even know why pytest patches those two specific hooks but NOT `sys.excepthook`.

Zac-HD:
> I don't even know why pytest patches those two specific hooks but NOT `sys.excepthook`.

(I'm reconstructing rather than recalling, but) we don't need to register a `sys.excepthook` because we'll always catch the exception coming out of the user's test function.  We *do* want to register a simple hook for the others; they just convert from the default "print to stderr" to "emit a warning containing the same text" - I guess someone had a reason to prefer that at some point, maybe warnings capture is more reliable around threads? 🤷‍♂️

Zac-HD:
As a high-level comment, I'd also observe that this PR seems like a strict improvement on the status quo, so I propose that we merge it (without the `3.11-exceptiongroup` CI task) and move any remaining discussion to a new issue :slightly_smiling_face:

jakkdl:
Yeah not much meaningful has changed after reviews, though the code is cleaner now :sparkles: 
Removed the added toxenv from CI, and added comments to tox and pre-commit configs to avoid future confusion and make continued work with exceptiongroups easier.

nicoddemus (approved):
Looks great, thanks everyone for chiming in.

Let's please use squash when merging this. 👍  (I've added `[SQUASH]` to the title as a reminder).

Zac-HD:
🎉🎉

Thanks to everyone involved! I'm looking forward to using this 😁

PR #10217, opened 2022-08-15
Add reference to the Where to patch docs in monkeypatch.setattr

nicoddemus:
This should help users with the common issue of patching the wrong place.

Also took the opportunity of using proper links in the monkeypatch introduction.

Related to #10216


src/_pytest/monkeypatch.py
```diff
@@ -186,16 +187,39 @@ def setattr(
...
+        Raises :class:`AttributeError` if the attribute does not exist, unless
         ``raising`` is set to False.
+
+        **Where to patch**
+
+        ``monkeypatch.setattr`` works by (temporarily) changing the object that a name points to with another one.
+        There can be many names pointing to any individual object, so for patching to work you must ensure
+        that you patch the name used by the system under test.
+
+        See the section *Where to patch* in the :mod:`unittest.mock` docs for a complete explanation, which
```
bluetech, line 221:
```suggestion
        See the section :ref:`Where to patch <python:where-to-patch>` in the :mod:`unittest.mock` docs for a complete explanation, which
```

bluetech (approved):
Nice improvement. I left a small suggestion.

src/_pytest/monkeypatch.py
nicoddemus: TY, TIL. 😁

PR #10219, opened 2022-08-15
mark: allow any Sequence[str] for parametrize(argnames), not just list/tuple

bluetech:
The main motivation for this change is to simplify the type shown in code editors -- `Sequence[str]` is easier to follow than `Union[list[str], tuple[str, ...]]`.

It also permits using other types if desired. It might lead to problems if someone uses some oddball sequence type, but hopefully they won't do that.

### DIFF
--- a/changelog/10218.improvement.rst
+++ b/changelog/10218.improvement.rst
@@ -1,0 +1,5 @@
+``@pytest.mark.parametrize()`` (and similar functions) now accepts any ``Sequence[str]`` for the argument names,
+instead of just ``list[str]`` and ``tuple[str, ...]``.
+
+(Note that ``str``, which is itself a ``Sequence[str]``, is still treated as a
+comma-delimited name list, as before).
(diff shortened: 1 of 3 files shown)


changelog/10218.improvement.rst
```diff
@@ -0,0 +1,5 @@
+``@pytest.mark.parametrize()`` (and similar functions) now accepts any ``Sequence[str]`` for the argument names,
```
nicoddemus, line 1: Perhaps we should categorize this as `trivial`? Up to you.

nicoddemus (approved):
Nice!

changelog/10218.improvement.rst
bluetech: It is pretty trivial, but we define `trivial` as "fixing a small typo or internal change that might be noteworthy.", while this *is* user-visible, so I'll keep it as `improvement`.

PR #10229, opened 2022-08-19
Drop `wheel` from `pyproject.toml` example

webknjaz:
It is unnecessary and has been deleted from the setuptools' docs too.
The setuptools' PEP 517 build backend implementation has been
auto-adding the `wheel` dependency since it's first been implemented.

Looks like it's been dropped from the main packaging config too, recently: https://github.com/pytest-dev/pytest/commit/56862c03cbb3f9cae3464c33390fbbaf8a91823a.


nicoddemus (approved):

nicoddemus:
Thanks @webknjaz!

Merging as the failures were due to codecov.

Issue #10247, opened 2022-08-26
When the test item prints log, the print format looks strange


TonyMistark:
#### For now
```
========test session starts ==========
platform linux -- Python 3.10.6, pytest-7.1.2, pluggy-1.0.0
django: settings: gin2.config.settings (from ini)
rootdir: /home/ice/projects/gino/doxmaster2/src, configfile: pytest.ini
plugins: mock-3.8.2, anyio-3.6.1, django-4.5.2
collected 19 items                                                                                                                                                                                                                                               

pytests/common/utils/test_str_utils.py ..........
pytests/common/utils/test_time_utils.py this is my log info
.this is my log info
.this is my log info
.this is my log info
```

#### Expected
```
========test session starts ==========
platform linux -- Python 3.10.6, pytest-7.1.2, pluggy-1.0.0
django: settings: gin2.config.settings (from ini)
rootdir: /home/ice/projects/gino/doxmaster2/src, configfile: pytest.ini
plugins: mock-3.8.2, anyio-3.6.1, django-4.5.2
collected 19 items                                                                                                                                                                                                                                               

pytests/common/utils/test_str_utils.py
..........
pytests/common/utils/test_time_utils.py
this is my log info
.this is my log info
.this is my log info
```

nicoddemus:
The current behavior is correct, although I can see it being surprising. You get the same output if you were calling `print` for example.

PR #10248, opened 2022-08-26
When the test item prints log, the print format looks strange #10247

TonyMistark:


nicoddemus:
Closed as in #10266 10247

PR #10252, opened 2022-08-26
Fix regendoc

nicoddemus:
The finalizers example was wiping out code needed for the examples below.

Noticed this in #10250

Closes #10252

PR #10257, opened 2022-08-30
doc: reproducible finalizer order examples

eagleoflqj:
These 2 tests pass so stdout won't be printed by default.
Fix an import missing.
Rename tests as `test_module` is meaningless here.


nicoddemus:
This should be merged/adapted only after https://github.com/pytest-dev/pytest/pull/10252 is merged, because this section in the docs as is, is not generating the expected output.

nicoddemus:
While at it I applied the same fix (passing `-s`) to #10252. Thanks again!

nicoddemus:
Fixed in #10252

PR #10274, opened 2022-09-05
Fix nodeid parser edge case in test selection

### LINKED ISSUE #10273: pytest main does not detect some invalid nodeid names
### Test scenario

`test_file.py`
```python
import pytest

@pytest.mark.parametrize("bar", [1,2,3])
def test_foo(bar):
    print(bar)
```

The following invalid node id is accepted by `pytest` instead of printing an error
```bash
$ pytest "test_hello.py[foobar]" -q

...        [100%]
3 passed in 0.04s
```

### Expected

Similar to specifying wrong path/dir/parameterized node_id an error should be printed instead of running all the tests in the given file

```bash
$ pytest "test_hello.py[foobar]" -q

no tests ran in 0.01s
ERROR: file or directory not found: test_hello.py[foobar]
```

elbehery95:
Fixes #10273 and also a small change in `resolve_collection_argument`

Before this commit `resolve_collection_argument` returned a list with empty string if the `::` selection operator was appended to the given file name E.g.

```python
assert resolve_collection_argument(Path(), "test_foo.py::")[1] == ['']  # why empty string ?
assert resolve_collection_argument(Path(), "test_foo.py")[1] == []

# for dir selection
assert resolve_collection_argument(Path(), "test_bar/")[1] == []
assert resolve_collection_argument(Path(), "test_bar")[1] == []
```

I think if the `/` operator *selects* parts (which file in a dir) and with/out adding this operator `resolve_collection_argument` returns `[]` for the `parts` value, then same thing can be applied to the `::` selection operator (which function/class in a file)

Not really sure if code assumes somewhere that when `::` is appended to the file name a list with empty string must be returned or not.

Appreciate your feedback

### DIFF
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -886,7 +886,8 @@ def resolve_collection_argument(
         parts[-1] = f"{parts[-1]}{squacket}{rest}"
     if as_pypath:
         strpath = search_pypath(strpath)
-    fspath = absolutepath(invocation_path / strpath)
+    fspath = invocation_path / strpath
+    fspath = absolutepath(fspath)
     if not fspath.exists():
         msg = (
             "module or package not found: {arg} (missing __init__.py?)"
--- a/src/_pytest/main.py
+++ b/src/_pytest/main.py
@@ -877,14 +877,16 @@ def resolve_collection_argument(
     If the path doesn't exist, raise UsageError.
     If the path is a directory and selection parts are present, raise UsageError.
     """
-    base, squacket, rest = str(arg).partition("[")
-    strpath, *parts = base.split("::")
-    if parts:
+    strpath, selection, rest = arg.partition("::")
+    test_name, squacket, rest = rest.partition("[")
+    parts = []
+
+    if test_name:
+        parts = test_name.split("::")
         parts[-1] = f"{parts[-1]}{squacket}{rest}"
     if as_pypath:
         strpath = search_pypath(strpath)
-    fspath = invocation_path / strpath
-    fspath = absolutepath(fspath)
+    fspath = absolutepath(invocation_path / strpath)
     if not fspath.exists():
         msg = (
             "module or package not found: {arg} (missing __init__.py?)"
@@ -892,7 +894,7 @@ def resolve_collection_argument(
             else "file or directory not found: {arg}"
         )
         raise UsageError(msg.format(arg=arg))
-    if parts and fspath.is_dir():
+    if selection and fspath.is_dir():
         msg = (
             "package argument cannot contain :: selection parts: {arg}"
             if as_pypath
--- a/testing/test_main.py
+++ b/testing/test_main.py
@@ -180,6 +180,10 @@ class TestResolveCollectionArgument:
             invocation_path / "src/pkg",
             [],
         )
+        assert resolve_collection_argument(invocation_path, "src/pkg/test.py::") == (
+            invocation_path / "src/pkg/test.py",
+            [],
+        )
 
         with pytest.raises(
             UsageError, match=r"package argument cannot contain :: selection parts"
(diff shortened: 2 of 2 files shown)


src/_pytest/main.py
```diff
@@ -877,22 +877,24 @@ def resolve_collection_argument(
     If the path doesn't exist, raise UsageError.
     If the path is a directory and selection parts are present, raise UsageError.
     """
-    base, squacket, rest = str(arg).partition("[")
```
bluetech, line 880: In the interest of making the change minimal, I would prefer if we keep the same flow as before, i.e. first the `partition("[")` then the `split("::")`. The fix to the issue can be done as an additional step, which is removing the last component of `parts` of it is `""` (this is a known "fixup" to `split()` but which should be explained with a comment).

src/_pytest/main.py
```diff
@@ -877,22 +877,24 @@ def resolve_collection_argument(
...
+    parts = []
+
+    if test_name:
+        parts = test_name.split("::")
         parts[-1] = f"{parts[-1]}{squacket}{rest}"
     if as_pypath:
         strpath = search_pypath(strpath)
-    fspath = invocation_path / strpath
-    fspath = absolutepath(fspath)
+    fspath = absolutepath(invocation_path / strpath)
```
bluetech, line 889: Again in the interest of keeping the change minimal (and keeping the "git blame" in tact), would be better to drop this cosmetic change.

testing/test_main.py
```diff
@@ -201,6 +201,12 @@ def test_does_not_exist(self, invocation_path: Path) -> None:
         ):
             resolve_collection_argument(invocation_path, "foobar")
 
+        with pytest.raises(
```
bluetech, line 204: This test doesn't look right to me -- the path here should be `src/pkg/test.py` not `src/pkg/test.py[a::b]`.

bluetech (changes requested):
Thanks @elbehery95, to me the change to avoid `""` makes sense. I left some comments on the code.

I would also like to see a new test which checks the scenario you described in the issue #10273 (the `test_file.py[foobar]` situation). This is the more important test than the "low level" `resolve_collection_argument` test. I recommend writing the test *before* the code change, making sure that it fails, then adding the change and checking that it fixes the test.

testing/test_main.py
elbehery95: but considering that pytest supports non python tests. In such case someone could have a file with virtually any name E.g. `test.py[a::b]` 

Does this mean that `[` and `::` should be avoided in test file names in such case ?

elbehery95: In such case having `rest` and `not parts` should be invalid right ? I mean should `resolve_argument` raise an exception at such condition ?
E.g.

```py
# invalid usage for the split :: operator
resolve_collection_argument(Path(), "test_foo.py[a::b]")

# rest = 'a::b]'
# parts = []
```

bluetech: Right now, `test_foo.py[a::b]` ends up just ignoring the `[a::b]` part. I agree it's not so great.

Note that 2442034a1eca53dae4b3141732660f3384b6d159 explicitly changed the split order of `::` and `[`.

IMO, the should be the behavior (I'm always assuming the returned `Path` actually exists in the filesystem):

1. `test_foo.py[a::b]` -> `(Path('test_foo.py[a::b]'), [])`.
2. `test_foo.py::` -> `(Path('test_foo.py'), [])` (special case for empty string)
3. `test_foo.py::test_it[a::b]` -> `(Path('test_foo.py'), 'test_it[a::b]')`

This behavior requires `resolve_collection_argument` parsing to be a bit more elaborate, just splitting by `::` or `[` wouldn't work. If you're interested in working on that, I think that'd be a good improvement. We should probably get buy in from the other maintainers first though.

In the meantime, I suggest handling the `test_foo.py::` case minimally in this PR, to fix the original issue, and open a new issue for the other suggested changes, as they go beyond #10273.

elbehery95:
I have updated the PR to include the following tests

- Scenario for test file name with `::` inside between braces as file name E.g. `test_foo.py[a::b]` 
- Scenario for test file name described in #10273 E.g. `test_foo.py[foobar]`
- Scenario for test file name ending with `::` E.g. `test_foo.py::`

As per the comments in the code review, the function now achieves the following

> test_foo.py[a::b] -> (Path('test_foo.py[a::b]'), []).
test_foo.py:: -> (Path('test_foo.py'), []) (special case for empty string)
test_foo.py::test_it[a::b] -> (Path('test_foo.py'), 'test_it[a::b]')

Appreciate your thoughts and feedback

src/_pytest/main.py
elbehery95: I think this will only solve the case for `test_foo.py::` but not https://github.com/pytest-dev/pytest/issues/10273 E.g. `test_foo.py[foobar]` right?

PR #10303, opened 2022-09-16
Automatically attach to non-propagating loggers

### LINKED ISSUE #3697: Handling of loggers with propagate=False
Hi,

There is a 2-year old issue on the old repo and inspired by #3003 I thought it might be easier if I at least add some reference here:

https://github.com/eisensheng/pytest-catchlog/issues/44

haxtibal:
Currently, pytest attaches its log handlers only to the root logger, and relies on log propagation to catch logs from loggers deeper in the hierarchy. In consequence we miss logs if one of those child loggers has `logger.propagate` set to `False` by user code. This affects caplog, report logging and live logging.

This commit showcases a possible fix, where we attach not only to the root logger, but also to non-propagating loggers as they occur. To do so we patch the standard libraries logging module with our own descriptor that intercepts assignments to `logger.propagate`. This enables us to inject and remove our pytest handlers right on time, even during executing test calls, with negligible performance impact.

Attempts to fix #3697.

PS: I'm aware the PR needs extra caution regarding thread safety and compatibility. Before considering such safe guards, I'd like to hear opinions on the approach in general. Feel free to close the PR right away if you think patching the standard library is against all rules.

### DIFF
--- a/src/_pytest/logging.py
+++ b/src/_pytest/logging.py
@@ -8,11 +8,13 @@ from contextlib import nullcontext
 from io import StringIO
 from pathlib import Path
 from typing import AbstractSet
+from typing import cast
 from typing import Dict
 from typing import Generator
 from typing import List
 from typing import Mapping
 from typing import Optional
+from typing import Set
 from typing import Tuple
 from typing import TYPE_CHECKING
 from typing import TypeVar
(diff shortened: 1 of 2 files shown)


bluetech:
Thanks for the PR. I think the monkeypatching interceptor approach is a bit too complex and hacky for pytest. It has advantage in low overhead and being able to handle test-time changes to `propagate`, but I don't think we can go with it, so I'll close this PR.

I adopted your test case and propose a simpler approach (#14375), which is less efficient and doesn't handle the dynamic change case, but I think is acceptable.

Issue #10326, opened 2022-09-30
Cannot ignore internal `pytest` `EncodingWarning` warnings


AA-Turner:
`pytest` has several internal EncodingWarnings when running tests with `python -X warn_default_encoding -m pytest ...`:

```
=============================== warnings summary ===============================
.tox/du19/lib/python3.10/site-packages/_pytest/capture.py:357
  /home/runner/work/sphinx/sphinx/.tox/du19/lib/python3.10/site-packages/_pytest/capture.py:357: EncodingWarning: 'encoding' argument not specified
    self.tmpfile = open(os.devnull)

.tox/du19/lib/python3.10/site-packages/_pytest/cacheprovider.py:199
  /home/runner/work/sphinx/sphinx/.tox/du19/lib/python3.10/site-packages/_pytest/cacheprovider.py:199: EncodingWarning: 'encoding' argument not specified
    readme_path.write_text(README_CONTENT)

.tox/du19/lib/python3.10/site-packages/_pytest/cacheprovider.py:189
.tox/du19/lib/python3.10/site-packages/_pytest/cacheprovider.py:189
  /home/runner/work/sphinx/sphinx/.tox/du19/lib/python3.10/site-packages/_pytest/cacheprovider.py:189: EncodingWarning: 'encoding' argument not specified
    f = path.open("w")

.tox/du19/lib/python3.10/site-packages/_pytest/cacheprovider.py:160
  /home/runner/work/sphinx/sphinx/.tox/du19/lib/python3.10/site-packages/_pytest/cacheprovider.py:160: EncodingWarning: 'encoding' argument not specified
    with path.open("r") as f:

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

The last three have fixes (see #9910 and #9915, thanks to @bluetech and @peymanslh), though as a general case I tried ignoring the warnings with the below in my `pyproject.toml`:

```toml
[tool.pytest.ini_options]
filterwarnings = [
    "all",
    "ignore::EncodingWarning:_pytest/cacheprovider",
    "ignore::EncodingWarning:_pytest/capture",
]
```

This failed with the below on Python versions prior to 3.10, as `EncodingWarning` doesn't exist:

```pytb
Traceback (most recent call last):
  File "/home/runner/work/sphinx/sphinx/.tox/du14/lib/python3.6/site-packages/_pytest/config/__init__.py", line 1642, in parse_warning_filter
[...]

PR #10327, opened 2022-09-30
Use an explicit encoding when opening `os.devnul`

AA-Turner:
Although slightly pointless, this avoids an `EncodingWarning` in Python 3.10 or greater when running under the `-X warn_default_encoding` flag.

----

I think this meets the bar for triviality so I haven't included a changelog entry. PR created per @nicoddemus's request in #10326, though this is not sufficient to close that issue.

A

### DIFF
--- a/src/_pytest/capture.py
+++ b/src/_pytest/capture.py
@@ -381,7 +381,7 @@ class FDCaptureBinary:
         self.targetfd_save = os.dup(targetfd)
 
         if targetfd == 0:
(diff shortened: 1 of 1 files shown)


bluetech (approved):
Strictly speaking the encoding of `stdin` is [more complicated](https://docs.python.org/3/library/sys.html#sys.stdin) than `utf-8`, however I believe the encoding is in no way visible/effectual in this specific case, so `utf-8` works. Thanks.

PR #10328, opened 2022-09-30
Enable Python 3.10's ``EncodingWarning``

AA-Turner:
This warns on the default encoding being used in ``open`` or ``io.open``, which can lead to cross-platform bugs (for example on Windows where the default codepage is cp1252), or with the upcoming change to the default Python encoding (PEP 686, Make UTF-8 mode default https://peps.python.org/pep-0686/)

-----

I think this meets the bar for triviality so I haven't included a changelog entry. PR created per @nicoddemus's request in #10326, though this is not sufficient to close that issue.

A

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -38,6 +38,10 @@ passenv =
 setenv =
     _PYTEST_TOX_DEFAULT_POSARGS={env:_PYTEST_TOX_POSARGS_DOCTESTING:} {env:_PYTEST_TOX_POSARGS_LSOF:} {env:_PYTEST_TOX_POSARGS_XDIST:}
 
+    # Warn where default encoding is used
+    # https://docs.python.org/3/library/io.html#io-encoding-warning
+    PYTHONWARNDEFAULTENCODING=1
+
     # Configuration to run with coverage similar to CI, e.g.
     # "tox -e py37-coverage".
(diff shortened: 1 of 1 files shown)


bluetech:
Needs a rebase now that #10328 is merged, although I highly doubt that was the only case missing an encoding -- probably just the first to error. I suspect that are many more places that will need to be fixed.

inwaves:
I'm currently going through the codebase and updating calls to `open` without an encoding to include an encoding of UTF-8. There are around 500 in total, luckily most are trivial. 

Some tests are failing because they use `iniconfig`, which does not specify an encoding when opening an `.ini` file. I've opened a matching issue on the repo.

bluetech:
iniconfig had its unicode maniac release :) so let's try again

bluetech:
Looking at the result I think it's undesirable to fix all of the EncodingWarnings in the test suite, even just the adding of `'utf-8'` to all `write_text` calls will uglify the code considerably, and they are 99% (maybe even 100%) ascii anyway. I think it's better to just wait until it becomes the default in a few years. So my recommendation is to close this PR - if anyone disagrees, let me know.

I did find two places to fix in pytest itself (not its test suite), which will affect *users* who want to be clean. Will send a PR.

bluetech:
Closing per the above, thanks for bringing this up @AA-Turner.

PR #10331, opened 2022-10-01
Add note in documentation to help modernize code bases to use pathlib

### LINKED ISSUE #10199: Allow users to find usages of py.local
As far as I understand, `py` (especially `py.local`) and related fixtures like `tempdir` should be replace with `pathlib` and their corresponding fixtures like `tmp_path`. In large code bases switching is a longer task and requires multiple commits and work of multiple people

It would be great if there would be an option to block the usage of these "old" fixtures, e.g. by a "future mode" or a warning which can be changed to an error with warning filters. 

With this, 
1) it is easy to find, where a code change is needed
2) after switching to `pathlib` only, future usage is blocked

#### Alternative Solutions
* Using a linter, which is quite hard tue the dynamic calculation of fixtures

sgaist:
The legacypath plugin handles the tmpdir and tmpdir_factory which should be update to their modern counterparts that uses pathlib.

Disabling this plugin allows developer to easily find and thus update the use of these fixtures.

Fixes #10199

### DIFF
--- a/doc/en/how-to/tmp_path.rst
+++ b/doc/en/how-to/tmp_path.rst
@@ -109,6 +109,17 @@ rather than standard :class:`pathlib.Path` objects.
 .. note::
     These days, it is preferred to use ``tmp_path`` and ``tmp_path_factory``.
 
+    In order to help modernize old code bases, one can run pytest with the legacypath
+    plugin disabled:
+
+    .. code-block:: bash
+
+        pytest -p no:legacypath
+
+    This will trigger errors on tests using the legacy paths.
+    It can also be permanently set as part of the :confval:`addopts` parameter in the
+    config file.
+
 See :fixture:`tmpdir <tmpdir>` :fixture:`tmpdir_factory <tmpdir_factory>`
 API for details.
 


doc/en/how-to/tmp_path.rst
```diff
@@ -109,6 +109,17 @@ rather than standard :class:`pathlib.Path` objects.
 .. note::
     These days, it is preferred to use ``tmp_path`` and ``tmp_path_factory``.
 
+    In order to help modernize old code bases, one can run pytest with the legacypath
+    plugin disabled:
+
+    .. code-block:: bash
+
+        pytest -p no:legacypath-tmpdir
```
nicoddemus, line 117: Shouldn't this be just `-p no:legacypath`?

sgaist: Good question

I used the registered name defined in `legacypath.py`.

From a quick test:

```python
def test_tmpdir(tmpdir):
      assert True
def test_tmpdir_factory(tmpdir_factory):
      assert True
```

either `no:legacypath` or `no:legacypath-tmpdir` triggers the expected error.

However, using `pytest --trace-config`, `legacypath` is listed there, so you are correct.

I will correct that.

nicoddemus (approved):
Thanks!

nicoddemus:
@bluetech the [backport workflow](https://github.com/pytest-dev/pytest/actions/runs/3210484141/jobs/5247958919) failed with:

```
HTTP 401: Bad credentials (https://api.github.com/graphql)
```

This is not supercritical to port right now, just mentioned in case you were not aware.

PR #10346, opened 2022-10-06
ci: fix PR not triggering workflows by switching back to pytestbot access token

### LINKED ISSUE #9554: Backport workflow: Use access token to run CI on PRs?
Regarding CI not running, right - CI doesn't run when triggered this way - this also happens for the create-pull-request actions and similar. Workaround is to close & reopen, then it triggers. It's also possible to use a user token (pytestbot), but that has its own security considerations.

_Originally posted by @bluetech in https://github.com/pytest-dev/pytest/issues/9430#issuecomment-1004842182_

bluetech:
The builtin actions token is not allowed to trigger further workflows. So when we create a PR using that token, the PR checks don't run, and a maintainer must close & open the PR to make them run. This is quite annoying, so switch back to the "machine user" approach.

Fixes #9554.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -48,4 +48,4 @@ jobs:
           --title "${subject}" \
           --body "Backport of PR #${{ github.event.number }} to $target_branch branch. PR created by backport workflow."
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.CHATOPS }}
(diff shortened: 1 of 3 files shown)


bluetech:
I remember the CHATOPS token has the required permissions, but not 100% sure, so once this is merged I'll make sure to test that it actually works (must be merged to the main branch before I can test it).

nicoddemus (approved):

nicoddemus:
Thanks @bluetech  for tackling this!

bluetech:
Marking this for backport just to test it out, but not going to actually backport it.

PR #10348, opened 2022-10-07
doc: Add second day for enterPy training

The-Compiler:
First day has sold out :sparkles:

### DIFF
--- a/doc/en/index.rst
+++ b/doc/en/index.rst
@@ -2,7 +2,7 @@
 
 .. sidebar:: Next Open Trainings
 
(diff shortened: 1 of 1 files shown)


The-Compiler:
Preview: https://pytest--10348.org.readthedocs.build/en/10348/

@RonnyPfannschmidt I think I'd rather keep them as-is, to prevent the box from getting too big. Unfortunately, our template has some kind of bug which will cause the "a quick example" heading to always be *below* the box, so it looks rather ugly when the text is too long.

nicoddemus (approved):

PR #10407, opened 2022-10-21
Add note on tags to CONTRIBUTING.rst

NateMeyvis:
Tags from the main repository are required for testing to work properly, but the current documentation does not reflect this, and at least one new contributor has been tripped up by this. Update the documentation with a short explanation and directions.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -243,6 +243,11 @@ Here is a simple overview, with pytest-specific bits:
    Given we have "major.minor.micro" version numbers, bug fixes will usually
    be released in micro releases whereas features will be released in
    minor releases and incompatible changes in major releases.
+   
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -244,6 +244,11 @@ Here is a simple overview, with pytest-specific bits:
    be released in micro releases whereas features will be released in
    minor releases and incompatible changes in major releases.
 
+   You will need the tags to test locally, so be sure you have the tags from the main repository. If you suspect you don't, set the main repository as upstream and fetch the tags::
```
nicoddemus, line 247: To confirm: is that because of `setuptools-scm`, or something else?

hugovk: Yes, without the tags, the version was showing like 0.1.dev123+xyz. See the pytest Discord where we debugged this: https://discord.com/channels/810210251398840371/810245251502964736/1033019899670437998

nicoddemus: Thanks!

PR #10417, opened 2022-10-24
Use specific tag in the gh-action-pypi-publish action

nicoddemus:
Otherwise we see this warning:

```
Warning:  You are using "pypa/gh-action-pypi-publish@master". The "master" branch of this project has been sunset and will not receive any updates, not even security bug fixes. Please, make sure to use a supported version. If you want to pin to v1 major version, use "pypa/gh-action-pypi-publish@release/v1". If you feel adventurous, you may opt to use use "pypa/gh-action-pypi-publish@unstable/v1" instead. A more general recommendation is to pin to exact tags or commit shas.
```

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -43,9 +43,8 @@ jobs:
         python -m build
 
     - name: Publish package to PyPI
-      uses: pypa/gh-action-pypi-publish@master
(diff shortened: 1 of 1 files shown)


asottile (approved):

The-Compiler:
Should we backport this to 7.2.x?

nicoddemus:
Yes, thanks for the reminder. :+1:

PR #10471, opened 2022-11-04
Remove done trainings

The-Compiler:
Another 20 people who know more about pytest now! :sparkles:


nicoddemus (approved):

The-Compiler:
Not sure WTF happened to the CI there, but it's the same on `main`, so force-merging this for now.

PR #10480, opened 2022-11-07
scripts/update-plugin-list: Be liberal in accepted summaries

joukewitteveen:
The update of the plugin list has been [failing for two months now](https://github.com/pytest-dev/pytest/actions/workflows/update-plugin-list.yml) due to a plugin that managed to put `null` in its summary (instead of the empty string). I have proposed [a fix to that package](https://github.com/thewtex/pytest-web3-data/pull/6), but we might as well make the update script a bit more robust.

I don't know how it is possible that a null summary was produced, potentially this is a bug in hatchling.

### DIFF
--- a/scripts/update-plugin-list.py
+++ b/scripts/update-plugin-list.py
@@ -90,7 +90,9 @@ def iter_plugins():
                 last_release = release_date.strftime("%b %d, %Y")
                 break
         name = f':pypi:`{info["name"]}`'
-        summary = escape_rst(info["summary"].replace("\n", ""))
+        summary = ""
+        if info["summary"]:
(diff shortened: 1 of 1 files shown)


nicoddemus:
Thanks, appreciate it!

nicoddemus:
Merging as the failures are unrelated.

ofek:
> potentially this is a bug in hatchling

Nope, no summary in that case https://github.com/pypa/hatch/blob/c879e33400fd25ce259d7bfde00fbf7c1c2ee8bb/backend/src/hatchling/metadata/spec.py#L83-L84

PR #10484, opened 2022-11-08
scripts/update-plugin-list: Improve requirement detection

joukewitteveen:
PEP 566 does not require a space after the dependency name.

I tested this change and found that it detected all dependencies previously detected, plus some more. There were no unwanted detections. Note that the old test was technically wrong: `"pytest " in requirement` may trigger on a `pytest` suffix. The new test may be slower, but this was not noticeable (the script is not fast anyway).

### DIFF
--- a/scripts/update-plugin-list.py
+++ b/scripts/update-plugin-list.py
@@ -78,7 +78,7 @@ def iter_plugins():
         requires = "N/A"
         if info["requires_dist"]:
             for requirement in info["requires_dist"]:
-                if requirement == "pytest" or "pytest " in requirement:
+                if re.match(r"pytest(?![-.\w])", requirement):
                     requires = requirement
                     break
         releases = response.json()["releases"]


scripts/update-plugin-list.py
```diff
@@ -78,7 +78,7 @@ def iter_plugins():
         requires = "N/A"
         if info["requires_dist"]:
             for requirement in info["requires_dist"]:
-                if requirement == "pytest" or "pytest " in requirement:
+                if re.match(r"pytest(?![-.\w])", requirement):
```
nicoddemus, line 81: Just to be clear, the idea here is to match only `pytest`, but not `pytest-`, `pytest.` or `pytest `, correct?

joukewitteveen: Pretty much. The idea is to match `pytest` as a package name. The package name ends either
- when the string ends, or
- when the next character cannot be part of the package name.

Package names can include any word character (`\w`, which includes digits and the underscore) as well as hyphens and periods ([PEP 508](https://peps.python.org/pep-0508/)). The big difference from `requirement.startswith("pytest ")` is that the character following the package name may also be a version specifier (`=`, `~`, `<`, or `>`). Indeed, the change in the plugin list after this MR will be precisely the addition of such requirement specifications.

nicoddemus: Great, thanks for the clarification! :+1:

Issue #10491, opened 2022-11-10
the report of the failed subtest is not generated if when='call'


anb76ru:
Hello!
I use the _**"pytest_runtest_logreport"**_ method to get subtest reports. 
A report is generated for **when='call'.**
 If the test failed, the _report_ is not generated. I think, that the reason is the **True** value of the **_"log_passing_tests"_**. 
This attribute is used in the block **"elif report.failed:"**. If the test feiled, then when = 'call' we don't step into the code block:
**if not self.log_passing_tests:        
reporter.write_captured_output(report)**
and the report is not generated. I don't understand why  _**self.log_passing_test using**_ in **report.failed.**
Maybe this is a mistake. if you remove the 'if not self.log_passing_tests: ', then the report for the failed tests for when='call' is generated without problems

OS: win10
pytest_ver: 7.1.1
![log_passing_tests](https://user-images.githubusercontent.com/65860480/201186826-45eaba3a-2d7c-4363-bfad-a894d72b47b8.jpg)

I suggest the following edits (screenshot diff), but I don't have the rights to create my own branch
![diff](https://user-images.githubusercontent.com/65860480/201187596-4c14e495-e3ad-4e9b-b2c2-b2ae35065713.jpg)

if these changes are unacceptable, please explain the logic of the condition:
                if not self.log_passing_tests:
                    reporter.write_captured_output(report)

Zac-HD:
Hi @anb76ru - I've edited your issue a little to clarify the report and use `code formatting` rather than mixed bold and italics.

 I can't comment directly on this code, but you don't need permissions to create a branch in *this* repo in order to contribute: the standard open-source workflow is to [create your own copy ("fork")](https://docs.github.com/en/get-started/quickstart/fork-a-repo), make a branch in that, and then open an pull request ("PR").  [This quickstart guide is pretty good](https://docs.github.com/en/get-started/quickstart/contributing-to-projects), and Pytest maintainers are generally happy to have further discussion about proposed changes on the pull request.

anb76ru:
> Hi @anb76ru - I've edited your issue a little to clarify the report and use `code formatting` rather than mixed bold and italics. For future reference, we strongly prefer [markdown code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) over screenshots of code too.
> 
[...]

PR #10496, opened 2022-11-14
fix logreport

anb76ru:
Hello!  I use the `pytest_runtest_logreport` method to get subtest reports.  A report is generated for `when='call'`, but if the test failed, the `report` is not generated. I think this is because the `log_passing_tests=True` attribute is used in the `elif report.failed:` block. If the test failed, then if `when == 'call'` we don't step into the code block:
```python
if not self.log_passing_tests:        
    reporter.write_captured_output(report)
```
and the report is not generated. I don't understand why  `self.log_passing_test using` in `report.failed`.  Maybe this is a mistake.  If you remove the '```if not self.log_passing_tests: ```', then the report for the failed tests for `when='call'` is generated without problems.

I suggest making `reporter.write_captured_output(report)` unconditional

#https://github.com/pytest-dev/pytest/issues/10491

OS: win10
pytest_ver:  #7.2.0
log_passing_tests

### DIFF
--- a/doc/en/changelog.rst
+++ b/doc/en/changelog.rst
@@ -137,6 +137,8 @@ Bug Fixes
 - `#9877 <https://github.com/pytest-dev/pytest/issues/9877>`_: Ensure ``caplog.get_records(when)`` returns current/correct data after invoking ``caplog.clear()``.
 
 
+- `#10491 <https://github.com/pytest-dev/pytest/issues/10491>`_: Method ``pytest_runtest_logreport`` generated report in case test passed or test failed and ``when=='call'``.
+
 
 Improved Documentation
 ----------------------
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -561,32 +561,6 @@ class LogXML:
                 reporter = self._opentestcase(report)
                 reporter.append_pass(report)
                 reporter.write_captured_output(report)
-
-            if report.when == "teardown":
-                reporter = self._opentestcase(report)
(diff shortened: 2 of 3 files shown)


nicoddemus (changes requested):
Thanks! The fix does make sense.

I also noticed that `write_captured_output` should be called in the start of the function, in the `if report.passed`/`if report.when == "call"`, otherwise passing tests won't output the captured output even if `log_passing_tests` is True.

We also need tests and a changelog entry. 👍

anb76ru:
@nicoddemus , if ```write_captured_output``` call without condition ```if not self.log_passing_tests```, than
get failed test ```test_avoid_double_stdout```. 
result_test_avoid_double_stdout

Therefore, I moved the condition ```if not self.log_passing_tests``` in ```if report.passed``` / ```if report.when == "call"```. But I am not sure about the correctness of these changes
Now, all test passed
Also, I added information about changes in file changelog.

Please, check PR

anb76ru:
@nicoddemus, can you review, please ?

anb76ru:
@nicoddemus I remove ```condition if not self.log_passing_tests``` in ```if report.passed / if report.when == "call"```
add test, and change target branch to main. Review please

anb76ru:
@nicoddemus, Hello. Please check my changes

nicoddemus:
Closing as per https://github.com/pytest-dev/pytest/issues/10491#issuecomment-1340914543.

Sorry again for the misunderstanding of the use case, and thanks for the effort @anb76ru!

PR #10517, opened 2022-11-19
Fix `tmp_path_retention_policy` crash when skipping from fixture

### LINKED ISSUE #10502: `tmp_path_retention_policy` crash when skipping from fixture
With #10442, this:

```python
import pytest


@pytest.fixture
def fixt(tmp_path):
    pytest.skip()


def test_fixt(fixt):
    pass
```

results in:

```
________________________ ERROR at teardown of test_fixt ________________________

cls = <class '_pytest.runner.CallInfo'>
func = <function call_runtest_hook.<locals>.<lambda> at 0x7f0e90dab760>
when = 'teardown'
reraise = (<class '_pytest.outcomes.Exit'>, <class 'KeyboardInterrupt'>)

    @classmethod
    def from_call(
        cls,
        func: "Callable[[], TResult]",
        when: "Literal['collect', 'setup', 'call', 'teardown']",
        reraise: Optional[
            Union[Type[BaseException], Tuple[Type[BaseException], ...]]
        ] = None,
    ) -> "CallInfo[TResult]":
        """Call func, wrapping the result in a CallInfo.
    
        :param func:
            The function to call. Called without arguments.
        :param when:
            The phase in which the function is called.
        :param reraise:
            Exception or exceptions that shall propagate if raised by the
            function, instead of being wrapped in the CallInfo.
        """
        excinfo = None
        start = timing.time()
        precise_start = timing.perf_counter()
        try:
>           result: Optional[TResult] = func()

src/_pytest/runner.py:342: 
src/_pytest/runner.py:263: in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
.venv/lib/python3.10/site-packages/pluggy/_hooks.py:265: in __call__
    return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
.venv/lib/python3.10/site-packages/pluggy/_manager.py:80: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
src/_pytest/runner.py:183: in pytest_runtest_teardown
    item.session._setupstate.teardown_exact(nextitem)
src/_pytest/runner.py:538: in teardown_exact
    raise exceptions[0]
src/_pytest/runner.py:527: in teardown_exact
    fin()
src/_pytest/fixtures.py:684: in <lambda>
    subrequest.node.addfinalizer(lambda: fixturedef.finish(request=subrequest))
src/_pytest/fixtures.py:1036: in finish
    raise exc
src/_pytest/fixtures.py:1029: in finish
    func()
src/_pytest/fixtures.py:916: in _teardown_yield_fixture
    next(it)

request = <SubRequest 'tmp_path' for <Function test_fixt>>
tmp_path_factory = TempPathFactory(_given_basetemp=None, _trace=<pluggy._tracing.TagTracerSub object at 0x7f0e91345780>, _basetemp=PosixPath('/tmp/pytest-of-florian/pytest-29'), _retention_count=3, _retention_policy='failed')

    @fixture
    def tmp_path(
        request: FixtureRequest, tmp_path_factory: TempPathFactory
    ) -> Generator[Path, None, None]:
        """Return a temporary directory path object which is unique to each test
        function invocation, created as a sub directory of the base temporary
        directory.
    
        By default, a new base temporary directory is created each test session,
        and old bases are removed after 3 sessions, to aid in debugging. If
        ``--basetemp`` is used then it is cleared each session. See :ref:`base
        temporary directory`.
    
        The returned object is a :class:`pathlib.Path` object.
        """
    
        path = _mk_tmp(request, tmp_path_factory)
        yield path
    
        # Remove the tmpdir if the policy is "failed" and the test passed.
        tmp_path_factory: TempPathFactory = request.session.config._tmp_path_factory  # type: ignore
        policy = tmp_path_factory._retention_policy
>       if policy == "failed" and request.node._tmp_path_result_call.passed:
E       AttributeError: 'Function' object has no attribute '_tmp_path_result_call'. Did you mean: '_tmp_path_result_setup'?

src/_pytest/tmpdir.py:276: AttributeError
```

cc @yusuke-kadowaki

yusuke-kadowaki:
closes #10502

### DIFF
--- a/changelog/10502.bugfix.rst
+++ b/changelog/10502.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed tmp_path_retention_policy crash when skipping from fixture.
--- a/changelog/10502.bugfix.rst
+++ b/changelog/10502.bugfix.rst
@@ -1,1 +1,0 @@
-Fixed tmp_path_retention_policy crash when skipping from fixture.
--- a/src/_pytest/tmpdir.py
+++ b/src/_pytest/tmpdir.py
@@ -281,8 +281,7 @@ def tmp_path(
     policy = tmp_path_factory._retention_policy
     result_dict = request.node.stash[tmppath_result_key]
 
-    # "call" might be skipped so check if it exists first
-    if "call" not in result_dict or (policy == "failed" and result_dict["call"]):
+    if policy == "failed" and result_dict.get("call", True):
         # We do a "best effort" to remove files, but it might not be possible due to some leaked resource,
         # permissions, etc, in which case we ignore it.
         rmtree(path, ignore_errors=True)
--- a/src/_pytest/tmpdir.py
+++ b/src/_pytest/tmpdir.py
@@ -5,10 +5,14 @@ import sys
 import tempfile
 from pathlib import Path
 from shutil import rmtree
+from typing import Dict
 from typing import Generator
 from typing import Optional
 from typing import TYPE_CHECKING
 from typing import Union
+
+from _pytest.nodes import Item
+from _pytest.stash import StashKey
 
 if TYPE_CHECKING:
     from typing_extensions import Literal
@@ -32,6 +36,8 @@ from _pytest.deprecated import check_ispytest
 from _pytest.fixtures import fixture
 from _pytest.fixtures import FixtureRequest
 from _pytest.monkeypatch import MonkeyPatch
+
+tmppath_result_key = StashKey[Dict[str, bool]]()
 
 
 @final
@@ -273,10 +279,15 @@ def tmp_path(
     # Remove the tmpdir if the policy is "failed" and the test passed.
     tmp_path_factory: TempPathFactory = request.session.config._tmp_path_factory  # type: ignore
     policy = tmp_path_factory._retention_policy
-    if policy == "failed" and request.node._tmp_path_result_call.passed:
+    result_dict = request.node.stash[tmppath_result_key]
+
+    # "call" might be skipped so check if it exists first
+    if "call" not in result_dict or (policy == "failed" and result_dict["call"]):
         # We do a "best effort" to remove files, but it might not be possible due to some leaked resource,
         # permissions, etc, in which case we ignore it.
         rmtree(path, ignore_errors=True)
+
+    del request.node.stash[tmppath_result_key]
 
     # remove dead symlink
     basetemp = tmp_path_factory._basetemp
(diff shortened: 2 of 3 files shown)


src/_pytest/tmpdir.py
```diff
@@ -273,11 +279,16 @@ def tmp_path(
     # Remove the tmpdir if the policy is "failed" and the test passed.
     tmp_path_factory: TempPathFactory = request.session.config._tmp_path_factory  # type: ignore
     policy = tmp_path_factory._retention_policy
-    if policy == "failed" and request.node._tmp_path_result_call.passed:
+    result_dict = request.node.stash[tmppath_result_key]
+
+    # "call" might be skipped so check if it exists first
+    if "call" not in result_dict or (policy == "failed" and result_dict["call"]):
```
The-Compiler, line 285: This now always deletes the tmpdir if there was no `"call"`, no matter what the policy is. That seems odd to me. If I set the policy to `all`, I'd expect those to be kept, even if the test wasn't executed, no?

What about something like:

```suggestion
    if policy == "failed" and result_dict.get("call", True):
```

instead? This would delete them if the test wasn't run, but only if the policy is set to failed.

testing/test_tmpdir.py
```diff
@@ -139,6 +140,24 @@ def test_1(tmp_path):
...
+            def fixt(tmp_path):
+                pytest.skip()
+
+            def test_fixt(fixt):
+                pass
+        """
+        )
+
+        reprec = pytester.inline_run(p)
+        assert reprec.ret == ExitCode.OK
```
The-Compiler, line 159: There should probably also be an assertion here which makes sure the files are still around (or not around, depending on policy) - see the earlier comment above.

changelog/10502.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed tmp_path_retention_policy crash when skipping from fixture.
```
The-Compiler, line 1: No changelog needed I think, as the original change hasn't been released yet.

The-Compiler (changes requested):

testing/test_tmpdir.py
yusuke-kadowaki: Added some tests.

[e4cf09d](https://github.com/pytest-dev/pytest/pull/10517/commits/e4cf09d8cb03775a1b8598bc3e15898ab2d18b89)

changelog/10502.bugfix.rst
yusuke-kadowaki: Removed.

The-Compiler (approved):
Looks great to me, thanks! Let's see if @nicoddemus has anything to add, and if not, I think this can be merged.

PR #10526, opened 2022-11-22
Issue #10506

### LINKED ISSUE #10506: `pytest` uses filesystem root (`C:\`) as root dir on Windows
My setup is a bit elaborate since I am using `pyinstaller` and running `pytest` from the bundled executable. Anyway, I can tell from the header that `pytest` is using `C:\` as a rootdir and generates `C:\.pytest_cache`, which I would like to avoid.

By debugging `determine_setup`
https://github.com/pytest-dev/pytest/blob/bc4e70e0485a15654aa72658668056d091a4b5dc/src/_pytest/config/findpaths.py#L175-L218

I have found that it is being called with the following parameters:
```python
inifile = None
args = ['C:\\Users\\bers\\AppData\\Local\\Temp\\_MEI138802\\project']
rootdir_cmd_arg = None
rootdir = None
config = <_pytest.config.Config object at 0x0000021A33D60490>
```

I have no relevant config in the app bundle, so the function passes via

https://github.com/pytest-dev/pytest/blob/bc4e70e0485a15654aa72658668056d091a4b5dc/src/_pytest/config/findpaths.py#L205-L208

with
```python
cwd = WindowsPath('C:/Code/project')
ancestor = WindowsPath('C:/Users/bers/AppData/Local/Temp/_MEI138802/project')
```

`rootdir = get_common_ancestor([cwd, ancestor])` returns `WindowsPath('C:/')`, but the following check returns `false` as `os.path.splitdrive(str(rootdir))[1]` is `"\\"`, not `"/"`. As a consequence, `rootdir` is *not* set to `ancestor`, but remains at `WindowsPath('C:/')`, which I believe is not intended.

TLDR: In https://github.com/pytest-dev/pytest/blob/bc4e70e0485a15654aa72658668056d091a4b5dc/src/_pytest/config/findpaths.py#L206 change `"/"` to `os.sep`.

PrerakPatelCS:
Closes #10506 
Helps run pytest on machines where the root directory being "C:\" 
is an issue some people have "C://" as their root directory so os.sep fixes this issue, os.path.sep can also work.

### DIFF
--- a/changelog/10506.bugfix.rst
+++ b/changelog/10506.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bug where sometimes pytest would use the file system root directory as :ref:`rootdir <rootdir>` on Windows.
--- a/src/_pytest/config/findpaths.py
+++ b/src/_pytest/config/findpaths.py
@@ -203,8 +203,7 @@ def determine_setup(
                     else:
                         cwd = Path.cwd()
                     rootdir = get_common_ancestor([cwd, ancestor])
-                    is_fs_root = os.path.splitdrive(str(rootdir))[1] == os.sep
-                    if is_fs_root:
+                    if is_fs_root(rootdir):
                         rootdir = ancestor
     if rootdir_cmd_arg:
         rootdir = absolutepath(os.path.expandvars(rootdir_cmd_arg))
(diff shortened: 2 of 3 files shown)


nicoddemus (approved):

nicoddemus:
Thanks @PrerakPatelCS!

PR #10535, opened 2022-11-26
Modify documentation to use `.stash` when storing test results.

yusuke-kadowaki:
To store test results, it's better to use `.stash` instead of adding attributes on items. However, the documentation currently advises to add attributes. So it's probably better to update the documentation.

Some contexts: 
https://github.com/pytest-dev/pytest/pull/10442#discussion_r1023643840
https://github.com/pytest-dev/pytest/pull/10517

### DIFF
--- a/doc/en/example/simple.rst
+++ b/doc/en/example/simple.rst
@@ -906,10 +906,7 @@ here is a little example implemented via a local plugin:
 
         # store test results for each phase of a call, which can
         # be "setup", "call", "teardown"
-        if your_unique_key not in item.stash:
-            item.stash[your_unique_key] = {rep.when: rep.passed}
-        else:
-            item.stash[your_unique_key][rep.when] = rep.passed
+        item.stash.setdefault(your_unique_key, {})[rep.when] = rep
 
 
     @pytest.fixture
--- a/doc/en/example/simple.rst
+++ b/doc/en/example/simple.rst
@@ -895,6 +895,8 @@ here is a little example implemented via a local plugin:
 
     import pytest
 
+    your_unique_key = StashKey[Dict[str, bool]]()
+
 
     @pytest.hookimpl(tryfirst=True, hookwrapper=True)
     def pytest_runtest_makereport(item, call):
@@ -902,10 +904,12 @@ here is a little example implemented via a local plugin:
         outcome = yield
         rep = outcome.get_result()
 
-        # set a report attribute for each phase of a call, which can
+        # store test results for each phase of a call, which can
         # be "setup", "call", "teardown"
-
-        setattr(item, "rep_" + rep.when, rep)
+        if your_unique_key not in item.stash:
+            item.stash[your_unique_key] = {rep.when: rep.passed}
+        else:
+            item.stash[your_unique_key][rep.when] = rep.passed
 
 
     @pytest.fixture
@@ -913,11 +917,13 @@ here is a little example implemented via a local plugin:
         yield
         # request.node is an "item" because we use the default
         # "function" scope
-        if request.node.rep_setup.failed:
+        result = request.node.stash[your_unique_key]
+        if not result.get("setup", False):
             print("setting up a test failed!", request.node.nodeid)
-        elif request.node.rep_setup.passed:
-            if request.node.rep_call.failed:
-                print("executing test failed", request.node.nodeid)
+        elif not result.get("call", False):
+            print("executing test failed", request.node.nodeid)
+        elif not result("teardown", False):
+            print("tear down test failed", request.node.nodeid)
 
 
 if you then have failing tests:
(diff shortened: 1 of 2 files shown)


yusuke-kadowaki:
@The-Compiler @nicoddemus 
What do you think?

doc/en/example/simple.rst
```diff
@@ -895,29 +895,35 @@ here is a little example implemented via a local plugin:
...
 
-        # set a report attribute for each phase of a call, which can
+        # store test results for each phase of a call, which can
         # be "setup", "call", "teardown"
-
-        setattr(item, "rep_" + rep.when, rep)
+        if your_unique_key not in item.stash:
+            item.stash[your_unique_key] = {rep.when: rep.passed}
+        else:
+            item.stash[your_unique_key][rep.when] = rep.passed
```
bluetech, line 912: Can use `setdefault` to make this more succinct:

```suggestion
        item.stash.setdefault(your_unique_key, {})[rep.when] = rep.passed
```

doc/en/example/simple.rst
```diff
@@ -895,29 +895,35 @@ here is a little example implemented via a local plugin:
...
         outcome = yield
         rep = outcome.get_result()
 
-        # set a report attribute for each phase of a call, which can
+        # store test results for each phase of a call, which can
         # be "setup", "call", "teardown"
-
-        setattr(item, "rep_" + rep.when, rep)
+        if your_unique_key not in item.stash:
+            item.stash[your_unique_key] = {rep.when: rep.passed}
```
bluetech, line 910: The previous code stored the entire report, I think that's a bit more generally useful, so would keep that instead of storing just `passed`.

doc/en/example/simple.rst
```diff
@@ -895,29 +895,35 @@ here is a little example implemented via a local plugin:
 
     import pytest
 
+    your_unique_key = StashKey[Dict[str, bool]]()
```
bluetech, line 898: `your_unique_key` is not a very good name. How about `phase_report_key`, if you choose to store the full report, or `phase_result_key` if not?

doc/en/example/simple.rst
yusuke-kadowaki: Didn't know this. Thanks.
[ca2cc3d](https://github.com/pytest-dev/pytest/pull/10535/commits/ca2cc3da2a0b218a8ada409c5b4c24baa716ee20)

Also refactored the `tmpdir.py` implementation this way. I can make it to another PR if that's better.
[5959e97](https://github.com/pytest-dev/pytest/pull/10535/commits/5959e979d52e754558af03eb2516a4b43ab5a4e2)

doc/en/example/simple.rst
yusuke-kadowaki: Agreed.
e20dfb6

doc/en/example/simple.rst
yusuke-kadowaki: e20dfb6

Yeah but isn't `stash` shared with all the other plugins and stuff? If that's the case, it should be remarked that the key name should be unique to avoid collision.

doc/en/example/simple.rst
bluetech: > Also refactored the tmpdir.py implementation this way. I can make it to another PR if that's better.

Sure!

doc/en/example/simple.rst
bluetech: > Yeah but isn't stash shared with all the other plugins and stuff?

The name doesn't matter, every stash and stash key (i.e. `StashKey()`) pair are unique.

doc/en/example/simple.rst
```diff
@@ -895,29 +895,32 @@ here is a little example implemented via a local plugin:
...
         yield
         # request.node is an "item" because we use the default
         # "function" scope
-        if request.node.rep_setup.failed:
-            print("setting up a test failed!", request.node.nodeid)
-        elif request.node.rep_setup.passed:
-            if request.node.rep_call.failed:
-                print("executing test failed", request.node.nodeid)
+        report = request.node.stash[phase_report_key]
+        if ("setup" not in report) or report["setup"].failed:
```
bluetech, line 918: Generally I don't think it's possible to reach here when `"setup" not in report` is true (but didn't test it), so I'd remove it. Same for `"call" not in report` (as long as it's in an `elif`). Regarding `"teardown" not in report` -- it's either never true or always true, but I'm not sure which one it is off hand.

doc/en/example/simple.rst
yusuke-kadowaki: Understood. I was confused about it.

doc/en/example/simple.rst
yusuke-kadowaki: This is probably correct on `setup` and `teardown` (tested that `teadown` doesn't exist here).

But `call` may not exist like this https://github.com/pytest-dev/pytest/pull/10442#discussion_r1023643840 no?

bluetech (approved):
LGTM now, thanks!

Issue #10547, opened 2022-11-30
AttributeError: module 'posixpath' has no attribute 'sep'


danschef:
Within a Python 3.10 environment and with pytest 7.2.0 installed, it seems like I cannot start the debugger in PyCharm. Instead PyCharm raises the following error:

```python
============================= test session starts ==============================
platform linux -- Python 3.10.8, pytest-7.2.0, pluggy-1.0.0 -- /home/me/mambaforge/envs/myenv/bin/python
cachedir: .pytest_cache
rootdir: /home/me/EnPT
plugins: reporter-html1-0.8.2, cov-4.0.0, reporter-0.5.2
collecting ... 
tests/test_1.py:None (tests/test_1.py)
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/runner.py:339: in from_call
    result: Optional[TResult] = func()
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/runner.py:370: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/python.py:532: in collect
    self._inject_setup_module_fixture()
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/python.py:546: in _inject_setup_module_fixture
    self.obj, ("setUpModule", "setup_module")
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/python.py:311: in obj
    self._obj = obj = self._getobj()
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/python.py:529: in _getobj
    return self._importtestmodule()
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/python.py:618: in _importtestmodule
    mod = import_path(self.path, mode=importmode, root=self.config.rootpath)
../../../../mambaforge/envs/myenv/lib/python3.10/site-packages/_pytest/pathlib.py:547: in import_path
    if module_file.endswith(os.path.sep + "__init__.py"):
E   AttributeError: module 'posixpath' has no attribute 'sep'

```
The environment is running on Ubuntu 20.04 and the error did not appear with Python 3.9. 

When replacing `os.path.sep` with `os.sep` in .../site-packages/_pytest/pathlib.py, the error goes away.

nicoddemus:
[...]

PR #10550, opened 2022-12-01
Use os.sep instead of os.path.sep.

danschef:
Closes https://github.com/pytest-dev/pytest/issues/10547.


Zac-HD (approved):
Thanks @danschef!

nicoddemus:
We probably should have a "trivial" changelog entry for that, in case anyone encounter a problem with *that* change.

Issue #6723, opened 2020-02-13
Separate terminal verbosity from diff verbosity


nicoddemus:
Currently `-v` controls both terminal verbosity (how to display test session progress in the terminal) and diff verbosity (how verbose is our diffs when assertions fail).

I've often wanted to increase diff verbosity while also keeping terminal verbosity at the normal level, but this is not possible today.

## Proposal

Introduce a *new* option that controls only *diff* verbosity (for discussion here `-dv`, but please let's bikeshed on this).

For backward compatibility reasons, `-v` keeps controlling both terminal verbosity *and* diff verbosity, but `-dv` can be used to increase diff verbosity independently. 

* `-v`: when used, increase both terminal verbosity and diff verbosity.
* `-dv`: when used, increase diff verbosity only (we also need the equivalent of `-q` for diff verbosity).

With `-dv` in place, we might even consider changing to a higher level of diff verbosity in the future by default, but that remains to be seen.

Related: #6292, #5192

blueyed:
I have `assert_truncate_level` already (https://github.com/blueyed/pytest/pull/63), but it only affects the truncation of lines (which is related to diffs / what you want, but yet another knob in this regard).

I'm planning to have something like `assert_verbosity_level` that would be added/substracted from `verbosity` for all of the assertion rewriting (including truncating lines - I will likely remove `assert_truncate_level` then, or keep it (based on `assert_verbosity_level` then)).

This could also default to "auto", which means that it would be 2 for running on CI.


> Related: #6292, #5192

https://github.com/pytest-dev/pytest/issues/5192 is a bug, not really related to having the option / control for it.
https://github.com/pytest-dev/pytest/issues/3962 is related however.

bluetech:
I almost always use `-v`/`-vv` just the get the full diff, and in these cases I always prefer not to increase the terminal verbosity. So :+1: from me.

---

Will the `diff is truncated, to see full diff pass -v` message be changed to the new flag?

---

How does the parsing of `-dv` work? Is `-d` a flag that just modifies what `-v` does? I think this doesn't compose very well, e.g. if I have `pytest -dv long_test_name`, and I want terminal verbosity, I expect to be able to just quickly jump to the end of the line and append `-v`, but it wouldn't work. So I prefer to just have `-d` and `-dd` as standalone flags.

> With -dv in place, we might even consider changing to a higher level of diff verbosity in the future by default, but that remains to be seen.

[...]

Issue #6292, opened 2019-11-30
Add option to force "short summary" despite high verbosity


Delgan:
Hi!

By "short summary", I mean the following format with dots:

```
========================= test session starts ==========================
platform linux -- Python 3.7.4, pytest-4.6.2, py-1.8.0, pluggy-0.12.0
rootdir: /home/delgan/Programmation/loguru, inifile: tox.ini
plugins: xdist-1.29.0, forked-1.0.2, cov-2.7.1
collected 1154 items                                                   

tests/test_activation.py ..............................          [  2%]
tests/test_add_option_backtrace.py .                             [  2%]
tests/test_add_option_catch.py ...........                       [  3%]
tests/test_add_option_colorize.py .............................. [  6%]
tests/test_add_option_diagnose.py .                              [  7%]
tests/test_add_option_enqueue.py .........                       [  7%]
tests/test_add_option_filter.py ................................ [ 10%]
tests/test_add_option_format.py ..................               [ 14%]
tests/test_add_option_kwargs.py .......                          [ 14%]
tests/test_add_option_level.py ..........                        [ 15%]
tests/test_add_option_serialize.py ...                           [ 15%]
tests/test_add_sinks.py ...........................FFF.......... [ 19%]
tests/test_ansimarkup_basic.py ................................. [ 23%]
[...]
```

Versus the verbose one:
```
============================= test session starts ==============================

platform linux -- Python 3.5.6, pytest-5.0.1, py-1.8.0, pluggy-0.12.0
cachedir: .tox/py35/.pytest_cache
rootdir: /home/travis/build/Delgan/loguru, inifile: tox.ini
plugins: cov-2.7.1
collected 945 items                
                                            
tests/test_activation.py::test_disable[-False] PASSED                      [  0%]
tests/test_activation.py::test_disable[tests-False] PASSED                 [  0%]
tests/test_activation.py::test_disable[test-True] PASSED                   [  0%]
tests/test_activation.py::test_disable[testss-True] PASSED                 [  0%]
tests/test_activation.py::test_disable[tests.-True] PASSED                 [  0%]
tests/test_activation.py::test_disable[tests.test_activation-False] PASSED [  0%]
tests/test_activation.py::test_disable[tests.test_activation.-True] PASSED [  0%]
tests/test_activation.py::test_disable[test_activation-True] PASSED        [  0%]
tests/test_activation.py::test_disable[.-True] PASSED                      [  0%]
[...]

Issue #5285, opened 2019-05-17
Suggestion: Ability to truncate long parameters in output in `-v` mode, still display full value in `-vv`


nixjdm:
AFAICT this ability doesn't exist.

I have a parameterized test that sometimes receives params that are long strings. I would generally prefer verbose output, but the long strings really clutter up the output pretty badly.

For example:

```python
strings_list = [
    # Imagine several of these
    (
        """Imagine a whole paragraph here.""",
        """Imagine a whole paragraph here.""",
    ),
]

pytest.mark.parametrize(["content", "expected"], string_list)
def test_sample(content, expected):
    ....
```

Testing the above with `pytest -v` Will print the entirety of the long strings used in `string_list`. Normally this is just what I want, and the output for each line fits nicely on a single line, or maybe a little spill over to two. With long params though, this creates a mess that makes it more difficult to understand the total results of the test at a glance, and needs a lot more scrolling to get through.

Nonetheless, a hint to the params would be great. I suggest, to not change current behavior, the ability to decorate or otherwise mark specific tests to affect this behavior for just that test.

Alternatives:
- A global setting to change this behavior (a good supplement)
- Actually changing the default behavior of `-v`. I'd guess this isn't a welcome option.

Yay295:
I found this code in pytest's tests that I think could be modified to do what you want.

https://github.com/pytest-dev/pytest/blob/326ae0cd88f5e954c8effc2b0c986832e9caff11/testing/python/approx.py#L48-L65

Something like this:

```python
strings_list = [
    # Imagine several of these
    (
        """Imagine a whole paragraph here.""",
        """Imagine a whole paragraph here.""",
    ),
]

@pytest.mark.parametrize(["content", "expected"], strings_list)
def test_sample(content, expected, pytestconfig):
    original_verbosity = pytestconfig.getoption("verbose")
    pytestconfig.option.verbose = 0
    try:
        ....
    finally:
        pytestconfig.option.verbose = original_verbosity
```

PR #10588, opened 2022-12-14
Add 'assert_truncate_level' to show the full diff without ``-vv``

### LINKED ISSUE #3962: Add distinct option to show full (assertion) diffs always

Pierre-Sassoulas:
Add an option to get the full diff regardless of the verbosity. 

This would Closes #3962

Co-authored-by: Daniel Hahler <github@thequod.de> (expanded from it and adapted from the fork. I kept the option name but changing it wouldn't be that hard).

### DIFF
--- a/changelog/3962.feature.rst
+++ b/changelog/3962.feature.rst
@@ -1,0 +1,1 @@
+``assert_truncate_level`` option added to be able to get the full diff without using ``-vv``.
--- a/doc/en/how-to/output.rst
+++ b/doc/en/how-to/output.rst
@@ -270,6 +270,10 @@ situations, for example you are shown even fixtures that start with ``_`` if you
 Using higher verbosity levels (``-vvv``, ``-vvvv``, ...) is supported, but has no effect in pytest itself at the moment,
 however some plugins might make use of higher verbosity.
 
+By default for verbosity inferior to ``-vv`` really long output are truncated. It's also possible to
+control the truncation directly using the ``assert_truncate_level`` option. You can use ``assert_truncate_level=0``
+to get the full diff regardless of the verbosity level.
+
 .. _`pytest.detailed_failed_tests_usage`:
 
 Producing a detailed summary report
--- a/testing/test_assertion.py
+++ b/testing/test_assertion.py
@@ -1290,6 +1290,16 @@ class TestTruncateExplanation:
             ]
         )
 
+        # test assert_truncate_level ini option.
+        result = pytester.runpytest("-o", "assert_truncate_level=1")
+        result.stdout.fnmatch_lines(
+            ["E         ...Full output truncated (2 lines hidden), use '-vv' to show"]
+        )
+        result = pytester.runpytest("-o", "assert_truncate_level=0")
+        result.stdout.fnmatch_lines(["* 6*"])
+        result = pytester.runpytest("-v", "-o", "assert_truncate_level=0")
+        result.stdout.fnmatch_lines(["* 6*"])
+
         result = pytester.runpytest("-vv")
         result.stdout.fnmatch_lines(["* 6*"])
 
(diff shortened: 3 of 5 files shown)


doc/en/how-to/output.rst
```diff
@@ -270,6 +270,10 @@ situations, for example you are shown even fixtures that start with ``_`` if you
 Using higher verbosity levels (``-vvv``, ``-vvvv``, ...) is supported, but has no effect in pytest itself at the moment,
 however some plugins might make use of higher verbosity.
 
+By default for verbosity inferior to ``-vv`` really long output are truncated. It's also possible to
```
nicoddemus, line 273: This is great, but in addition to this, we should also include this option in the `Configuration Options` section in `reference.rst`, making use of the `confval` directive, for example:

https://github.com/pytest-dev/pytest/blob/f6adebb99064ad2a894ea7c04a1257aba5876e21/doc/en/reference/reference.rst?plain=1#L1180-L1184

(Make sure to include the new option in alphabetical order)

doc/en/how-to/output.rst
```diff
@@ -270,6 +270,10 @@ situations, for example you are shown even fixtures that start with ``_`` if you
...
+control the truncation directly using the ``assert_truncate_level`` option. You can use ``assert_truncate_level=0``
```
nicoddemus, line 274: By using the `confval` as I mentioned before, we can create a link to it automatically:

```suggestion
control the truncation directly using the :confval:`assert_truncate_level` option. You can use ``assert_truncate_level=0``
```

changelog/3962.feature.rst
```diff
@@ -0,0 +1 @@
+``assert_truncate_level`` option added to be able to get the full diff without using ``-vv``.
```
nicoddemus, line 1: By using `confval` as I mentioned before, we can create a link to it automatically:

```suggestion
:confval:`assert_truncate_level` option added to be able to get the full diff without using ``-vv``.
```

nicoddemus (changes requested):
Thanks a lot @Pierre-Sassoulas, we really appreciate it!

As for the option name, `assert_truncate_level` reads strange to me, given `assert` is a verb. Perhaps we should use `assertion_truncate_level`? I'm not a native English speaker, so would appreciate if others share their opinions.

testing/test_assertion.py
```diff
@@ -1664,7 +1674,9 @@ def test_raising_repr():
     """
     )
     result = pytester.runpytest()
-    result.stdout.fnmatch_lines(["E       AssertionError: <exception str() failed>"])
+    result.stdout.fnmatch_lines(
+        ["E       AssertionError: <unprintable AssertionError object>"]
```
nicoddemus, line 1678: This change should be reverted it seems.

Pierre-Sassoulas:
> As for the option name, assert_truncate_level reads strange to me, given assert is a verb. Perhaps we should use assertion_truncate_level?

I agree. ``assertion_truncation_level`` or ``truncation_level`` maybe ?

testing/test_assertion.py
Pierre-Sassoulas: Right, it's strange I had an issue locally.

Pierre-Sassoulas:
I need to take into account https://github.com/pytest-dev/pytest/issues/6292, #5285 and #6723 on top of #3962, I'm going to rethink this a little.

nicoddemus:
> I agree. assertion_truncation_level or truncation_level maybe ?

The second makes sense if we will be using it for other things than just the `assert` truncation.

Pierre-Sassoulas:
Closing as it was done in https://github.com/pytest-dev/pytest/pull/11473

PR #10589, opened 2022-12-15
Allow users to override the hardcoded maxsize value

ssbarnea:

### DIFF
--- a/changelog/10589.feature.rst
+++ b/changelog/10589.feature.rst
@@ -1,0 +1,2 @@
+Allow users to override repr maxsize value (240) by defining
+:envvar:`PYTEST_DEFAULT_REPR_MAX_SIZE` environment variable.
--- a/doc/en/reference/reference.rst
+++ b/doc/en/reference/reference.rst
@@ -1061,6 +1061,11 @@ processes can inspect it, see :ref:`pytest current test env` for more informatio
 
 When set, pytest will print tracing and debug information.
 
+.. envvar:: PYTEST_DEFAULT_REPR_MAX_SIZE
+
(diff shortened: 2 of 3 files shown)


changelog/10589.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Allow users to override repr maxsize value (240) by defining
```
nicoddemus, line 1: This should be considered a new feature, so please rename this file to `10589.feature.rst`.

changelog/10589.bugfix.rst
```diff
@@ -0,0 +1,2 @@
...
+``PYTEST_DEFAULT_REPR_MAX_SIZE`` environment variable.
```
nicoddemus, line 2: We should document this in the reference together with other environment variables:

https://github.com/pytest-dev/pytest/blob/f6adebb99064ad2a894ea7c04a1257aba5876e21/doc/en/reference/reference.rst?plain=1#L1050-L1053

It would be nice to include in the description an example of a use case (I'm guessing this is interesting to configure while constructing/testing external libraries in a CI environment?).

nicoddemus, line 2:
```suggestion
:envvar:`PYTEST_DEFAULT_REPR_MAX_SIZE` environment variable.
```

nicoddemus (changes requested):
Thanks @ssbarnea!

Besides my comments, we should include some test to avoid a future regression.

nicoddemus (approved):

Zac-HD:
Since implemented, see https://docs.pytest.org/en/stable/reference/reference.html#confval-truncation_limit_chars

thanks for offering this patch!

PR #10599, opened 2022-12-21
Fix update-plugin-list workflow due to new 'packaging'

nicoddemus:
The latest 'packaging' release has dropped support for `LegacyVersion`, meaning that version strings which do not conform to the standard now raise an error.


nicoddemus:
PR created by running the workflow from this branch: https://github.com/pytest-dev/pytest/pull/10600

bluetech (approved):

PR #10607, opened 2022-12-23
Mitigate directory creation race condition

Kadino:
Closes https://github.com/pytest-dev/pytest/issues/10604 which could intermittently display unexpected behavior between checking if the path exists and requesting creation. This was relatively common when pytest was being invoked in parallel by another test runner (CTest) and trying to create the same parent-folder for multiple XMLs. A modest amount of testing did not reproduce additional filesystem race conditions.

This notably does not work around another edge case where the parent path of the XML could be created as a file instead of a folder or link. That vanishingly rare case should cause file creation to fail on the next line, with a fairly obvious exception message.

### DIFF
--- a/src/_pytest/junitxml.py
+++ b/src/_pytest/junitxml.py
@@ -645,8 +645,8 @@ class LogXML:
 
     def pytest_sessionfinish(self) -> None:
         dirname = os.path.dirname(os.path.abspath(self.logfile))
-        if not os.path.isdir(dirname):
-            os.makedirs(dirname, exist_ok=True)
+        # exist_ok avoids filesystem race conditions between checking path existence and requesting creation
+        os.makedirs(dirname, exist_ok=True)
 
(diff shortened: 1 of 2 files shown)


src/_pytest/junitxml.py
```diff
@@ -646,7 +646,7 @@ def pytest_sessionstart(self) -> None:
     def pytest_sessionfinish(self) -> None:
         dirname = os.path.dirname(os.path.abspath(self.logfile))
         if not os.path.isdir(dirname):
-            os.makedirs(dirname)
+            os.makedirs(dirname, exist_ok=True)
```
nicoddemus, line 649: Probably makes sense to also remove the `if` condition. Also probably nice to leave a comment in the code explaining the rationale around that.

Other than that, can you please also add a changelog entry as "bugfix"?

nicoddemus (approved):

nicoddemus:
Thanks!

PR #10614, opened 2022-12-26
Defer removing tmpdir

ykadowak:
Partially addresses https://github.com/pytest-dev/pytest/issues/10564.

In the discussion on the issue, I believe majority agreed to defer the removal of temporary directories. Therefor, this PR has been implemented that way. If there needs to be more discussions, we can do it in the thread of this PR I think.

Also, I am planning to make another PR after this based on https://github.com/pytest-dev/pytest/pull/10546 for the following reasons.

1. As I mentioned [here](https://github.com/pytest-dev/pytest/issues/10564#issuecomment-1341090010), a lot of test cases need to be removed because of this PR's change, which is not ideal for the maintainability.
2. By separating https://github.com/pytest-dev/pytest/pull/10546, we can easily revert the commit and rollback to this PR if some regression happens.

CC: @The-Compiler @RonnyPfannschmidt @nicoddemus


yusuke-kadowaki:
Can this PR be reviewed? 
The situation of the issue is being incomplete right now.

Please leave any questions if something is not clear.

Zac-HD:
Hi @ykadowak,

Thanks for your work on this, and apologies it sat so long.

Closing because the mechanism here has been overtaken: #10546 ("Stop using atexit") merged and replaced `atexit.register` in this module with an `ExitStack` closed at session finish (tmpdir.py:79-81, :219-225, :339-340). Rebasing this would mean reintroducing the pattern we just retired. The branch also deletes four tests that #10546 deliberately kept and extended, and patches `cleanup_dead_symlink`, which no longer exists — it moved to `cleanup_dead_symlinks` in `pytest_sessionfinish`.

Your diagnosis was right, though, and the bug is still live — I reproduced the duplicate `pytest-0/test_..._0` path on main under `tmp_path_retention_policy = "failed"`. Notably you predicted this outcome yourself in #10564: with #10546 landed, the `rmtree` can just be registered on the ExitStack. That's now a small change on top of merged infrastructure and would preserve the existing tests.

This would be very welcome as a fresh PR if you're still interested, and if you ping me specifically I'll make reviewing it a priority.

Issue #10630, opened 2023-01-03
testing/io/test_terminalwriter.py is broken with the latest pygments 2.14.0 release


yilei:
After the latest pygments https://github.com/pygments/pygments/releases/tag/2.14.0 release (2 days ago), `testing/io/test_terminalwriter.py` is broken.

Reproducible steps:

```
$ git clone https://github.com/pytest-dev/pytest
$ cd pytest
$ tox -e py -- testing/io/test_terminalwriter.py
```

```
===================================================================================================================== FAILURES ======================================================================================================================
________________________________________________________________________________________________ test_code_highlight[with markup and code_highlight] ________________________________________________________________________________________________

has_markup = True, code_highlight = True, expected = '{kw}assert{hl-reset} {number}0{hl-reset}\n', color_mapping = <class 'conftest.color_mapping.<locals>.ColorMapping'>

    @pytest.mark.parametrize(
        ("has_markup", "code_highlight", "expected"),
        [
            pytest.param(
                True,
                True,
                "{kw}assert{hl-reset} {number}0{hl-reset}\n",
                id="with markup and code_highlight",
            ),
            pytest.param(
                True,
                False,
                "assert 0\n",
                id="with markup but no code_highlight",
            ),
            pytest.param(
                False,
                True,
                "assert 0\n",
                id="without markup but with code_highlight",
            ),
            pytest.param(
                False,
                False,
                "assert 0\n",
                id="neither markup nor code_highlight",
            ),
        ],
    )
    def test_code_highlight(has_markup, code_highlight, expected, color_mapping):
        f = io.StringIO()
        tw = terminalwriter.TerminalWriter(f)
        tw.hasmarkup = has_markup
        tw.code_highlight = code_highlight
        tw._write_source(["assert 0"])

>       assert f.getvalue().splitlines(keepends=True) == color_mapping.format([expected])
E       AssertionError: assert ['\x1b[94mass...[39;49;00m\n'] == ['\x1b[94mass...[39;49;00m\n']
E         At index 0 diff: '\x1b[94massert\x1b[39;49;00m \x1b[94m0\x1b[39;49;00m\x1b[90m\x1b[39;49;00m\n' != '\x1b[94massert\x1b[39;49;00m \x1b[94m0\x1b[39;49;00m\n'
E         Use -v to get more diff
```

When I forced with `pygments==2.13.0`, this test passed.

I'm not sure if this is a real issue in pytest itself, or just the tests need to be updated.

[...]

PR #10632, opened 2023-01-04
Fix tests pygments 2.14.0

danigm:
Fix https://github.com/pytest-dev/pytest/issues/10630

This PR adapt the color tests with the latest pygments 2.14.0 release, adding the some new ending sequence (`\x1b[90m\x1b[39;49;00m`) when it's needed.


nicoddemus (approved):
Thanks, appreciate it!

PR #10641, opened 2023-01-08
Dont update cache from xdist worker

s-padmanaban:
head-fork: s-padmanaban/pytest
compare: stepwise_cache_condition

base-fork: pytest-dev/pytest
base: main

This PR disables the update of the cache/stepwise data from xdist worker nodes (Prevents race condition and performance issue on NFS when involving large number of worker nodes trying to write to the same location)


src/_pytest/stepwise.py
```diff
@@ -48,6 +48,8 @@ def pytest_configure(config: Config) -> None:
 def pytest_sessionfinish(session: Session) -> None:
     if not session.config.getoption("stepwise"):
         assert session.config.cache is not None
+        if hasattr(session.config, "workerinput"):
```
nicoddemus, line 51:
```suggestion
        # Do not update cache if this process is a xdist worker to prevent race conditions (#10641).
        if hasattr(session.config, "workerinput"):
```

src/_pytest/stepwise.py
```diff
@@ -119,4 +121,6 @@ def pytest_report_collectionfinish(self) -> Optional[str]:
         return None
 
     def pytest_sessionfinish(self) -> None:
+        if hasattr(self.config, "workerinput"):
```
nicoddemus, line 124:
```suggestion
        # Do not update cache if this process is a xdist worker to prevent race conditions (#10641).
        if hasattr(self.config, "workerinput"):
```

nicoddemus (changes requested):
Thanks @s-padmanaban!

Besides the comments I left, could you also create `changelog/10641.bugfix.rst` describing the change from the user's point of view?

src/_pytest/stepwise.py
s-padmanaban: Your feedback has been incorporated into the code

src/_pytest/stepwise.py
s-padmanaban: Your feedback has been incorporated into the code

s-padmanaban:
> Thanks @s-padmanaban!
> 
> Besides the comments I left, could you also create `changelog/10641.bugfix.rst` describing the change from the user's point of view?

I've incorporated your feedback into the code and changelog accordingly.

nicoddemus (approved):
Thanks!

